PR #90, opened 2015-03-16
Refactor -set.xml file of the version with the alternative FDM

onox:
Refactors c172p-detailed2-set.xml to be a bit shorter. When testing it, the aircraft felt quite different, so it should indeed be using the alternative FDM.


gilbertohasnofb:
> When testing it, the aircraft felt quite different, so it should indeed be using the alternative FDM.

So you mean that, in your opinion, this different FDM is superior to the default one?

onox:
No. Just saying that the model is indeed using a different FDM.

gilbertohasnofb:
Understood.

PR #98, opened 2015-03-21
Fix MP lighting

onox:
* Make beacon and strobe lights of MP aircraft independent from local aircraft. If you turn on your beacon light, then it is no longer automatically the case that the beacon light of other aircraft are turned on as well.
* Transmit state of navigation lights over MP

~~I cannot really get the taxi and landing lights to work (not even locally).~~ ALS landing light is visible, but it seems displayed behind the aircraft in external view.

Currently the electrical output of the navigation, taxi, and landing lights are transmitted. Someone (a 3D modeller / animation expert) should fix the XML conditions so that the lights (taxi and landing) are visible depending on the voltage (scaling the brightness instead of on/off).

Navigation lights seem to work, but should use modified XML conditions (<greater-than> or <expression><difference> elements) as well.

I have tested this with:

`fgfs --aircraft=c172p-detailed --airport=KNID --multiplay=out,10,localhost,5701 --multiplay=in,10,localhost,5702 --callsign=A --runway=14`

and

`fgfs --aircraft=c172p-detailed --airport=KNID --multiplay=out,10,localhost,5702 --multiplay=in,10,localhost,5701 --callsign=B --runway=14`


Juanvvc:
Wow, this is an impressive amount of work!

- Taxi ligths: I think they are only visible in Rembrandt, not in ALS.
- Landing lights in ALS shouldn't be visible in external views. Lights are fake in ALS, and they only *seem* real in the cockpit view. There is no way to fix landing lights in ALS in external view.
- In the current state of your branch, landing lights are not working because you changed the property they should listen for an alias in your second commit, and the alias is not working after removing light.nas in your third commit.
- The idea of making the intensity of the light dependant of the voltage is interesting, but I'm afraid it is not in my priority list. Currently, the state on/off of a light depends on several switches and breakers. It is not difficult to check the voltage as well and use a float number as the output. Check electrical.nas for the details, ligths control is in electrical_bus_1()

onox:
The second commit does not really deal with the landing lights yet. Only beacon, strobes, and navigation. And those work.

Juanvvc:
Yes, sorry, you are right, landing lights are not modified in your second commit. It is late here (2:30AM) I better sleep and check again tomorrow :)

onox:
Me too :smile: 

The problem with the landing lights is that even if I would use something like <true/> in the condition, I don't see the 3D objects (landing light) that I would like FG to show me.

gilbertohasnofb:
Is there a reason this pull request hasn't been merged yet?

onox:
The reason is that I could not get the landing light to work.

onox:
Fixed the landing light. Branch can be merged now.

PR #113, opened 2015-03-30
Fix "/Aircraft/..." for "Aircraft/..." in Nasal

IAHM-COL:
please revise :calling: 

Thanks
Israel

### DIFF
--- a/Nasal/liveries.nas
+++ b/Nasal/liveries.nas
@@ -1,4 +1,4 @@
 
 # liveries =========================================================
-aircraft.livery.init("Aircraft/c172p-detailed/Models/Liveries", "sim/model/livery/name", "sim/model/liverytail/name", "sim/model/liverywing/name", "sim/model/livery/index");
+aircraft.livery.init("Models/Liveries", "sim/model/livery/name", "sim/model/liverytail/name", "sim/model/liverywing/name", "sim/model/livery/index");
 
--- a/Nasal/liveries.nas
+++ b/Nasal/liveries.nas
@@ -1,4 +1,4 @@
 
 # liveries =========================================================
-aircraft.livery.init("/Aircraft/c172p-detailed/Models/Liveries", "sim/model/livery/name", "sim/model/liverytail/name", "sim/model/liverywing/name", "sim/model/livery/index");
+aircraft.livery.init("Aircraft/c172p-detailed/Models/Liveries", "sim/model/livery/name", "sim/model/liverytail/name", "sim/model/liverywing/name", "sim/model/livery/index");
 


onox:
Could you explain the purpose of this change?

IAHM-COL:
Not really on detail.

But I seem to understand that the trailing slash on the /Aircraft path is a
deprecated code and it is to be replaced for the slash-less version
Aircraft, on every iteration of the code.

This merge replaces
"/Aircraft for "Aircraft in one line of the nasal file, affecting the
liveries.

Beyond this simple explanation, I am not very certain, but I believe this
change is relevant to when the Aircrafts are not hosted within either
$FG_ROOT or $FG_HOME.

Israel

Juanvvc:
Is there any documentation about this?

To me, "Aircraft/..." is a relative path from the current context, and "/Aircraft" is an absolute path from the root of the filesystem. In FlightGear, the root of the "filesystem" is a merge of many different paths on the disk: FG_ROOT, FG_AIRCRAFT...

In any case, the management of these "absolute" and "relative" paths is not consistent anywhere, both in FlightGear (sometimes you need relative paths, other times absolute paths) and in our code.

IAHM-COL:
oh ok. That's what I meant with "I am not sure"

Your position seems reasonable Juan.

So in that case, I believe the correct version is as master is currently
and this case can be closed.

Thanks for the revision.

Best,
Israel

onox:
Most paths in FlightGear models don't start with a /. Just prepending Aircraft/ to the path is sufficient for FlightGear to find it in either the FG_AIRCRAFT folder or in FG_DATA/Aircraft. So I don't mind if this gets merged or not. It does not seem to break changing liveries locally, but I'm not sure how it will work over MP.

IAHM-COL:
I think it does not break the livery on multiplayer either.
I checked the livery locally too before submitting the merge

I still believe that although this is not currently documented, the
behavior of the path
"/Aircraft/

Is probably more erratic than the one of
"Aircraft/

Likewise with <tag>/Aircraft vs <tag>Aircraft iterations.

I know the flightgear aircrafts are not very standardized on this respect,
but ideally work may be done to standardize the "Aircraft" (no slash) form.

I believe this is correct on the fact that not every aircraft (now, most no
aircraft) resides in $FG_ROOT anymore, and they should be compatible to
reside outside.
Meaning changes for relative paths such as

../../Instruments

or to

/Aircraft might be corrected instead.

I had sent a devel-list email a few days ago asking about this, but no-one
replies to me :(

(I am kinda banned there or something)

http://sourceforge.net/p/flightgear/mailman/message/33665462/

Juanvvc:
My position:

1) The project still have more than 50 open issues. Fixing something that doesn't seem to be broken shouldn't use any of our time constraints. This is very simple modification and very fast to merge, but I'm concerned accepting this modification now could make us looking at other "minor" enhancements in the future instead of the really important issues.
2) Path management is not inside FlightGear but in other external project: OpenSceneGraph. Specifically, inside file OpenSceneGraph/src/osgDB/FileUtils.cpp This file seems to use the well-known, long-accepted tradition of "if a path starts with a slash, it is absolute. If it doesn't, it is relative to the working directory" This obviously refers to the current namespace. In FlightGear, an absolute path refers to the merge of many directories: FG_ROOT, FG_AIRCRAFT, FG_SCENERY, FG_TERRASYNC and who knows what else. A relative path, as in any other namespace in the last 30 years, refer to the current working directory. "Aircraft/c172p-detailed/foo.txt" (without a slash) is ambiguous and depends on the context. It can to "/opt/fgdata/Aircraft/c172p-detailed/foo.txt" or "/opt/fgdata/Aircraft/c172p-detailed/Aircraft/c172p-detailed/foo.txt" depending of where the XML code defines its parent. This ambiguity is fixed using absolute paths (with slashes) and the complexity of multiple directories FG_AIRCRAFT, FG_ROOT... must be hidden to the XML programmer. If it is not, I will considere that a bug in FlightGear.

For these reasons, unless there are some new documentation or policy in the forum/wiki/source-code related to absolute/relative paths, I vote for not merging this commit.

Israel, in any case, thank you vey much for your efforts! Please, consider participating in the project and proposing many other enhancements. We have a list of 50 issues to fix! Fix any of them!

Juanvvc:
My position:

1) The project still have more than 50 open issues. Fixing something that doesn't seem to be broken shouldn't use any of our time constraints. This is very simple modification and very fast to merge, but I'm concerned accepting this modification now could make us looking at other "minor" enhancements in the future instead of the really important issues.
2) Path management is not inside FlightGear but in other external project: OpenSceneGraph. Specifically, inside file OpenSceneGraph/src/osgDB/FileUtils.cpp This file seems to use the well-known, long-accepted tradition of "if a path starts with a slash, it is absolute. If it doesn't, it is relative to the working directory" This obviously refers to the current namespace. In FlightGear, an absolute path refers to the merge of many directories: FG_ROOT, FG_AIRCRAFT, FG_SCENERY, FG_TERRASYNC and who knows what else. A relative path, as in any other namespace in the last 30 years, refer to the current working directory. "Aircraft/c172p-detailed/foo.txt" (without a slash) is ambiguous and depends on the context. It can to "/opt/fgdata/Aircraft/c172p-detailed/foo.txt" or "/opt/fgdata/Aircraft/c172p-detailed/Aircraft/c172p-detailed/foo.txt" depending of where the XML code defines its parent. This ambiguity is fixed using absolute paths (with slashes) and the complexity of multiple directories FG_AIRCRAFT, FG_ROOT... must be hidden to the XML programmer. If it is not, I will considere that a bug in FlightGear.

For these reasons, unless there are some new documentation or policy in the forum/wiki/source-code related to absolute/relative paths, I vote for not merging this commit.

Israel, in any case, thank you vey much for your efforts! Please, consider participating in the project and proposing many other enhancements. We have a list of 50 issues to fix! Fix any of them!

IAHM-COL:
Thanks for Considering it Juan ;)
As I said earlier, this case can be closed.

Best,
Israel

IAHM-COL:
-----------------------------------------------
>In fact, if a line like this works:
>
>aircraft.livery.init("Models/Liveries", "sim/model/livery/name", "sim/model/liverytail/name", "sim/model/liverywing/name", "sim/model/livery/index");
>
>I will happily merge it
-------------------------------------------------------

Dear @Juanvvc 
I followed your recomm.

The line as "Models/Liveries"
Still works on my side, and I agree with you is all-together a cleaner coding alternative that is resilient to path updates

best,
IHCOL

wlbragg:
The line as "Models/Liveries"

This is something I have been concerned with for awhile now. I am changing to more relevant paths because I am afraid at what will happen when aircraft are not stored in any defined %root%/Aircraft path
Will it be able to find its own parts? 
I have several aircraft directories that are standalone with no root or Aircraft designation and they have not failed yet so maybe it is an invalid concern, but I don't see how that can work when calling either relative or not using /Aircraft or Aircraft/ and there is no Aircraft in the path at all. Yet it still works.

At any rate, Models/ is cleaner in my opinion.

IAHM-COL:
Dear Wlbragg

Recent discussions on the devel.list upon this sort of problem.
Allow me to reffer you to @sanhozay, and this discussion over the same decision for the 727-230:

https://github.com/IAHM-COL/727-230/pull/15

Upon revision on the behavior of the several options, apparently the form 
"Models/Liveries" or "Models/" will have conflicts on multiplaying. 

Apparently the best FG behavior is upon the form

"Aircraft/path/subpath" 
in most situations.

See also work by L. Brenta on this sense 
here:https://github.com/FGMEMBERS/DC-10-30/commit/ec4ae9ccde772a2992ed6902b0abd46b6ab2e1b9
and here: https://github.com/FGMEMBERS/Lockheed1049h/commit/d50b0f2944e870e59e50899c2ebc430ea499b4b5

Best,
@IAHM-COL

PR #121, opened 2015-04-18
Update start-up tutorial.

algefaen:
To start the C172 you now have to have the correct mixture, set the master
switch and you may even have to prime the engine. This commit adds these points
to the engine start-up tutorial.


onox:
Nice. Thanks for the contribution! :+1:

Issue #92, opened 2015-03-16
Turn coordinator's moving ball texture


gilbertohasnofb:
The moving ball in the turn coordinator is slightly cut on each side (see picture below, note how both the left and right sides of the ball are totally straight, while the texture file is really round).

https://dl.dropboxusercontent.com/u/2730444/flightgear/devel/tc.png

Juanvvc:
I tried to fix this and a new bug appeared.

It seems the algorithm flightgear uses to move textures aims to fast instead of accurate transformations. As a result, the ball is perfect when it is on the right side, but the antialising of the border desapears when the ball is on the left side. I'm not sure how to fix this.

gilbertohasnofb:
Ok, so I exchanged the labels enhancement to bug, and let's leave this issue opened. Maybe in the future someone will find out how to solve this.

gilbertohasnofb:
Hasn't this been fixed by @dany93 ?

dany93:
I would say this is fixed (but I'm far from being a champion in Blender and Gimp).
This image http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=780#p237373

gilbertohasnofb:
I tested it and it seems to be fixed. The size and colour are now much better and the problem mentioned by Juan before in which the texture gets aliased on one side of the TC but not on the other is not present any longer. I'd say we close this issue now.

PR #126, opened 2015-04-23
Various modifications

dany93:
* The "Propwash Effect" I had added in the past is useless or overabundant, now that FlightGear has been fixed for that.

* The "Autostart" sets the elevator trim at -0.19
Result : the rotation and takeoff start by themselves around 45 - 50 kt, which is close to stall. After that and still with no action on the elevator, climbing is done at this same velocity (instead of about 80 kt, unless you push the yoke or correct the trim). This value of -0.19 is also far from the trim for cruise. Consequently, you have to permanently push the yoke unless you correct he trim.
Usually, the pilot has to pull the yoke to provoke the rotation (unless you have a long enough runway, waiting for a higher velocity)
I think that merely trim = 0 is a better value to start.

* The height of the pilot's eyes are a bit too high. This prevents from seeing the horizon under the wings at left and right. Moreover, I would add a default-field-of-view (change zoom) setting. But this can depend on everyone's screen.

* Fixed size of ball of turn coordinator and fixed the animation curvature and the min / max angles are not well matched to the geometry. I've also increased the sensitivity.


dany93:
I have NOT "Merged this pull request" because I don't understand what this will do.
I let it as a request.

onox:
If you close the the PR, it's not going to be merged, which is what you want I think. Thus I'm reopening this PR :smile:

onox:
@dany93 I have deleted the dany-suggest-20150422 branch for you. Could you edit the description of this PR to explain the changes to the TC?

dany93:
Thank you for having deleted the dany-suggest-20150422 branch.
onox wrote:
"However, you should always let someone else do the merge so that at least a second pair of eyes have looked at the changes."
That was my aim, as you suggested in this post http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=885#p239753
That is my first contribution on this aircraft (not on git). When you have enough confidence in me, I can merge it directly.

Changes to the TC are explained in this post (# 4) http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=870#p239270
Turn coordinator
see this post for the image http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=780#p237373
On your current turn coordinator, the ball is too small, the animation curvature and the min / max angles are not well matched to the geometry.
I've also increased the sensitivity.

onox:
@dany93 You have changed 2 .png files with .rgb files. Shouldn't everything be a .png file? You also deleted a .blend file. Was that intentional?

dany93:
You are right, .png seems to be the standard now in FG.

Rgb files were not intentional. That's only because I copied the 
_Generic_ TC in fgdata/Aircraft/Instruments-3d/tc before modifying it. 
And this folder, probably old, was with .rgb (this doesn't prevent it 
from being currently in the generic files FG/data....)

I hope I can convert the .rgb to .png, but it also needs to change it in 
the .ac files. Can it be done later  and pushed?
I there is a doubt, if you can, refuse my tc replacement, let the old 
one and I will see later, then I will propose only if I can do it well.

I haven't explicitly deleted the .blend file, I have merely replaced the 
initial TC folder of the c172p-detailed by "mine", which didn't contain 
this .blend. In fact, the .blend file has nothing to do here, it's only 
for Blender, FG cannot read it. Maybe the author has put it here to make 
it easier in case someone would like modifying it with Blender.

onox:
You can just add more commits to your dany93/c172p-detailed branch. They should automatically  show up on this page.

dany93:
I'll do this.
But I see that most instrument textures are still in .rgb.

wlbragg:
I need to read this a little closer to digest  exactly what all was done. I'll try to get to it today.

dany93:
Done for the turn coordinator: .png textures instead of .rgb.
Do I have to ask for another pull request? Or will you include it directly updated ?

Juanvvc:
- The .blend files are there because we are modifying many instruments. We also provide the .blend files of the new instruments/models we modify, in case someone in the future wants to improve them. These .blend files are not read by flightgear, but this is a development repository and (imho) it must also include the .blend files we are using.
- In the case of the turn coordinator, the previous .blend/.ac files not only changed the ball but also included a new white background. If this new version does not include a background (and it they are copied from the original tc, they don't include a background), let me consider this a regression and then, not acceptable. Our version of the tc included a small version of the ball on porpuse for testing issue #92 

The other modifications are ok for me.

wlbragg:
I think the .png textures will go in here.
I studied a little closer the changes. I have only one comment and that is on the pilots pov. The change is so minor I have no problem with it, but, if it is set the same as the copilot I have a question. The copilot is farther away from the left wingtip and closer to the right wingtip, shouldn't their pov's be slightly different? Everything else, I saw nothing of concern and must defer to your expertise. Looks good thanks, Go ahead and merged, we can always revert if something was missed.

dany93:
Turn coordinator:
I added a white background, that was my first aim (instead of that horrible purple of the generic tc).
Ball diameter: if we look at real photos, we can see that the ball always seem rather large in the tube (lens effect?).
Animation: The curvature radius was nor well fitted with that of the 3D tube. That is an xml modif, not Blender
Also, the ball went left and right farther than the 3D model limits. That is in the xml file too.
I have corrected these animations (easy).

I understand for the inclusion of the .blend file, I had guessed and written this.
I can add or upload my Blender file. The problem is that I'm not good at all at Blender, I'm not sure it's a good example, I'd prefer it be corrected by someone.
Blender can import files starting from the .ac and textures.
Juanvvc, is there a solution with your tc model if you prefer, which could take my remarks into account? Of course, I can upload my .blend file if you can look at it.
I do not wish to impose "my" model, but I think the current one can be improved.
I can do it easily for the animations not so for Blender.

dany93:
@wlbragg: I haven't touched (not looked at) the pov's copilot.
Anyway, I don't understand what you mean, the pov is not dependent on the view direction

Juanvvc:
Yes, the ball is much larger in the real plane than the size our model shows. But the small size for the ball was intended to help testing other issues :) I guess it is not important any more and we can use the normal (large) ball size. Of course, the final version must have a larger size.

Our tc was just the original with a white background. If your new model also includes the white background and enhances the animation, then it is better than the old version. I'm sorry for not testing your model, testing different branches is not easy for me right now. I prefer if you include the .blend file, but that is your call.

dany93:
OK, I will include my Blender file. But no miracle, I had it from the .ac original file of the generic tc. There's nothing more in it and it is a very simple model.
If you want to see it quickly, here's the image http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=780#p237373
As it was my first intention, I will include the previous tc folder (yours, just renamed) to make a return more easy just in case.

dany93:
Done.
I have:
- added "my" corresponding tc.blend (as Juan wished),
- re-included the previous tc folder (Juan-vvc's), renamed "tc-previous". This way, anyone can easily come back to the previous tc.

Juanvvc:
I think there is no need to include the folder tc-previous. We can always revert to the previous version of any file using git commands. This is why they call git a "version control system" :)

dany93:
Done. tc-previous deleted.

Juanvvc:
Merged. Thanks!

dany93:
Thanks too. And sorry for the bother with the tc...

PR #141, opened 2015-05-02
Propeller: such that idle RPM increase at diving

dany93:
Correction in the propeller file, such that the RPM increase at pitch down and diving, when the airspeed increases.
Initially, the propeller RPM increased when diving with enough throttle but it remained at idle (about 800 RPM) with throttle at zero.


onox:
Let me test this.

dany93:
Thanks.

Issue #128, opened 2015-04-27
some instruments seem to be in a bit wrong position


gilbertohasnofb:
The user dg-505 has reported in the forum thread: "[...] some instruments seem to be in a bit wrong position: http://abload.de/img/fgfs-screen-0032jxic.png "

I tested on my system and indeed both the vertical speed indicator and the tachometer are above their correct positions.

onox:
@dg-505 Are all instruments in the correct position now? Then this bug can be closed.

gilbertohasnofb:
I can confirm this is fixed.

PR #152, opened 2015-05-07
dialogs: Clean up "Aircraft Options" and "Registration"

onox:
Cleaned up the two XML files, formatting, and tabs+spaces mixture mess.

Aircraft Options:
* Reduced padding and code size
* Replaced the "----" lines with orange horizontal lines

Immatriculation / Registration:
* Renamed "Immatriculation" (is this French?) to "Registration"
* Replaced "OK" button with "Apply" and "Cancel" buttons

PR #154, opened 2015-05-08
Set instruments to correct position

dg-505:
As mentioned in issue #128, I corrected the positions of the VSI, the RPM, and also the Battery Gauge. To do this I had to made some changes in the .ac-files, so I created three new folders (BatteryGauge, RPM, and VSI in Models/Interior/Panel/Instruments.
When committing the changes I messed up something, but finally it should be right.
I'm a newcomer to Git and so I apologize for the chaos in these 6 commits. :confused:


gilbertohasnofb:
Why does it say that 2 of those commits were made by me?? Anyway, let's merge and see how it looks, thanks for the work!

gilbertohasnofb:
Tested, all seems to be fine, much better than before!

dg-505:
I don't know. Something went wrong when I comitted/added/pushed, or whatelse...
The main point is, that it works. Also after the chaos made by me.

onox:
@gilbertohasnofb If you want to see how something looks before merging it, click on "command line" in the "This pull request can be automatically merged." box, then you execute the instructions of "Step 1" to test the PR.

gilbertohasnofb:
@onox thanks for the info, onox; I will do that next time.

onox:
@gilbertohasnofb Could you also take a look at my PR #152 ? :smile:

gilbertohasnofb:
@onox just did it, easy peasy to test it before merging :)

PR #161, opened 2015-05-10
Mixture and Throttle levers control by scroll wheel, inversion

dany93:
Were counter-intuitive.
issue https://github.com/Juanvvc/c172p-detailed/issues/156#event-300990125


onox:
Tested and works! :+1:

Issue #103, opened 2015-03-24
instrument lights shouldn't be functioning when master is off


gilbertohasnofb:
Currently, ~~all lights (strobe, beacon, nav, landing and instrument lights)~~ ~~some lights (landing and instrument lights)~~ the instrument lights work even when the master is off.

onox:
I cannot reproduce this. If I turn off the battery master, then the lights turn off.

gilbertohasnofb:
@onox Hmm, I tested it again and I don't know how I managed to make this bug appear the first time! Now the strobe, beacon and nav lights are working properly, so either I got confused or this is a bug that needs more steps to reproduce.

But the instrument lights are still working when master switch is on, and the landing lights as well (I use ALS). Can you try to reproduce these two?

onox:
Yes, instrument lights still works if master switch is off and ALS landing light too.

gilbertohasnofb:
thanks!

gilbertohasnofb:
UPDATE: I found how to reproduce the strobe, beacon and nav lights being on while the master is off. Simply turn the master on, turn the lights on, reset the simulator. The aircraft will be reloaded with the master off (good decision), light switches (good) on and all lights will be visible (bug!), as if the battery would be on!

wlbragg:
This is probably a different problem but may be related. I don't know when this crept in or for sure if it is wrong, but I think it is.
When I start the sim the all the lights are on including the strobes, except the strobes are on solid, not blinking.

gilbertohasnofb:
@wlbragg I am away from my computer now (which has FG installed on it), but I guess what you are describing is exactly identical to what I was experiencing. So there are basically two issues here:

- initial state of nav, strobes and beacon lights is on even though master is off
- functioning  landing lights and instrument lights when master is off

onox:
The problem appears to be that `/sim/model/c172p/lighting/strobes/state` is somehow `true` after initialization. I could not find anywhere in the code where this could be happening. The only thing that controls this boolean is the `aircraft.light` class. When I added `debug.dump()` calls to the `light` class in `Nasal/aircraft.nas` I noticed that `state` was first `0`, and then later after JSBSim had printed some statistics, it had somehow become `1`, even though I did not see any of my debug statements being printed on lines where it might possibly set `state` to `1`.

[...]

PR #162, opened 2015-05-11
MP model fix

wlbragg:
Hopefully this is all done and I can move on. Someone,  please take a look and merge when satisfied.

### DIFF
--- a/Models/c172p-detailed.xml
+++ b/Models/c172p-detailed.xml
@@ -3294,36 +3294,17 @@
    <object-name>landinglightcover</object-name>
    <object-name>landinglight2</object-name>
    <condition>
-	<or>
-		<and>
-		 <equals>
-		  <property alias="../../params/wing_left_broken/property"/>
-		  <value>0</value>
-		 </equals>
-		 <equals>
-		  <property alias="../../params/wing_right_broken/property"/>
-		  <value>0</value>
-		 </equals>
-		 <equals>
-		   <property alias="../../params/crash/property"/>
-		  <value>0</value>
-		 </equals>
-		</and>
-		<and>
-		 <equals>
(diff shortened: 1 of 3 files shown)


wlbragg:
Oh, may want to HOLD ON!. I forgot, I still haven't found the fdm/jsbsim/crash getting switched on bug. We don't want that in master as it breaks starting the aircraft unless you click on "repair" first.
Let me find this bug before we merge.

onox:
Yes, I noticed it too, my plane is broken when I start FG. Other than that, the lightning properties still work :+1: 

Btw, if the wing is broken, shouldn't the landing light (for the left wing) and the navigation lights and strobes be turned off?

onox:
I have fixed the issue. It reminded me of an old issue I encountered in [#103](https://github.com/Juanvvc/c172p-detailed/issues/103#issuecomment-87404364). You can merge the branch :shipit: :smile:

wlbragg:
That is really funny. I was just getting ready to commit the same fix. Mine wasn't as elegant (I just setprop). Nor was I going to explained the situation as I figured it was just a hack fix. I figured it had to be the link between alias and the original props. But I don't understand why others don't pick up the bad values. Whatever, it's done and works.

onox:
It is indeed a strange issue :confused:

For future readers:

`/params` is an absolute path to the `<params>` element, while `../../../../params` is a relative path. And the latter is quite error-prone because it requires the correct number of `../`.

Juanvvc:
Keep in mind there is an animation that unselects (i.e., do not render) some parts in MP. This is mainly for the internal objects and instruments, but we can also add to this animation any other object. Also, I'm not sure if any Nasal code is executed locally for the remote aircraft.

PR #163, opened 2015-05-11
Added exhaust smoke when operating the engine starter.

dg-505:
When starting the engines, there come a little bit fumes out of the exhaust.


Juanvvc:
This is a great idea! I'm merging the changes because at first glance there is nothing disruptive with them, but I'm not sure if controls/engines/engine/starter is the property to check. Will the fumes appear when the starter is pressed, even if the master switch is off or the carburettor is not primed?

I'm merging but not closing the issue to run some tests.

dg-505:
/controls/engines/engine/starter only switches to "true", when the Master Switch is on, so with Master Switch off, the fumes don't appear.

PR #171, opened 2015-05-12
Fix #132 by using mixture instead of mixture-lever

### LINKED ISSUE #132: Map the mixture lever back to /controls/engines/engine/mixture
Currently, the mixture lever is controlled by /controls/engines/engine/mixture-lever as a proxy property, and  /controls/engines/engine/mixture is written by Nasal/engine.nas taking into account this value and (if the engine is off) the number of times the pilot primed the engine.

This was a temporal solution to implement the primer. I tried many times to create a jsbsim module to control this, but I was unable to due my lack of knowledge of jsbsim. Please, check this post: http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=525#p234882

The main problem with our approach is that input devices with a mixture lever (for example, the Saitek Quadrant) must be mapped specifically to /controls/engines/engine/mixture-lever instead of /controls/engines/engine/mixture as in any other aircraft. Forcing a joystick configuration specific to our aircraft is not acceptable, in my opinion.

This bug is my main concern for the c172p-detailed to be accepted as the default aircraft: all input devices must work out-of-the-box as expected. I prefer removing the primer support better than forcing a specific joystick configuration.

Keep in mind everything works ok if you are using a joystick without a mixture lever or the keyboard. If this is your case, you probably didn't notice this bug until now :)

onox:
This PR fixes #132 by calculating the actual mixture in JSBSim.


Juanvvc:
onox, I have tested this and it worked like a charm! Thank you very much! Please, change the reference to mixture-lever in Models/Effects/exhaust/exahust.xml and we can merge this to master.

Thank you very much!

PR #173, opened 2015-05-12
New FDM damage effects caused by overspeed and g-forces.

wlbragg:
Dany, please confirm I got this correct. I added the all changes by hand and refactored physics.nas a bit.

### DIFF
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -4,14 +4,20 @@ props.globals.initNode("/fdm/jsbsim/bushkit", 0, "INT");
 #reference Vne=158, max positive-g=5.8, max negative-g=5.8 
 props.Node.new({ "limits/vne":0 });
 props.globals.initNode("limits/vne", 158, "DOUBLE");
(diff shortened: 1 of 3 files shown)


dany93:
c172p.xml (FDM): OK

c172p-detailed-set.xml: OK

physics.nas:
(not easy to carefully compare)
- Tests ( neither easy, everything is so rapid...): it seems working as expected ("Repair" is magic for that!).
- File comparison: seems OK.

*IMHO, "Tab" indentation is to be avoided. I use "spaces" on purpose, because "Tab" mixed with "spaces" give different indentations depending on the editor. In JSBSim FDMs, the tables loose their alignment.*
- Tiny detail (but only formal, can be done later). Line 4 comment, preferably
#reference Vne=158, max positive-g=4.4, max negative-g=-1.76, break above 1.5 x max.

- Plus the small anomaly (my previous message) for the magnetos at start.

After the "Autostart" magnetos anomaly is fixed, I'd say it can be merged.

PR #181, opened 2015-05-13
Alternative FDM: damage/left wing, right-wing properties insertion

dany93:
Wing damage with over-speed or over-g's.
It had been inserted in the "original" FDM (via wlbragg commit) but the alternative FDM had been left as it was.
It gave unknown properties and crash of the simulator at launch with this alternative FDM.

### DIFF
--- a/c172p-fdm-dany93.xml
+++ b/c172p-fdm-dany93.xml
@@ -585,7 +585,18 @@
                 <property>atmosphere/rho-slugs_ft3</property>
                 <value>0.5</value>
             </product>
-        </function> 
+        </function>
+        
+        <function name="aero/function/total-wing-damage">
+            <description>Total damage on wings</description>
+            <product>
+                <sum>
+                    <property>wing-damage/left-wing</property>
+                    <property>wing-damage/right-wing</property>
+                </sum>
+                <value>0.5</value>
+            </product>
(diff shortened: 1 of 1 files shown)


c172p-fdm-dany93.xml
```diff
@@ -939,13 +950,23 @@
                     <value>0.0147</value>
                 </product>
             </function>
-            <function name="aero/coefficient/Clbk">
-                <description>Roll_moment_due_to_one_broken wing</description>
+            <function name="aero/coefficient/Cllwdmg">
```
onox: Are `aero/coefficient/Cllwdmg` and `aero/coefficient/Clrwdmg` summed and used somewhere?

dany93: They are naturally summed like every coefficient which is inside an ```<axis>``` (ROLL, here) channel.
They are used and set in physics.nas via the properties
/fdm/jsbsim/wing-damage/left-wing
/fdm/jsbsim/wing-damage/right-wing
which take the values between 0 (no damage) and 1 (totally broken).

onox: Yes, I must have been looking at an older version of the file it seems :frowning:  :sweat_smile:

PR #185, opened 2015-05-17
Instruments location

dg-505:
To me it seems more logical to have all cockpit instruments at the same place, so I moved the instruments located in c172p-detailed/Instruments to /Models/Interior/Panel/Instruments. Now all instruments are located in the same folder.


Juanvvc:
I completely agree. Thanks!

Issue #179, opened 2015-05-13
Strange 3D object above air vent lever


gilbertohasnofb:
See: http://s2.postimg.org/t488b92t5/fgfs_screen_001.png

Was this created on purpose? If so, what is it? It does look a bit strange to me.

Juanvvc:
It is part of the radio panel. I don't know why it is there, but it is like this also in the original c172p. I don't know why it didn't appear until now, probably a z-fighting issue with the right side of the panel.

I'm quite sure it is a bug and I'll remove the object. Thanks.

gilbertohasnofb:
Thanks Juan. I have the impression it appeared when the air vent lever was remodelled, but I may be wrong on that. Anyway, thanks for fixing it!

dg-505:
Solved. See pull request #187

Issue #160, opened 2015-05-09
Some instrument are poorly positioned


gilbertohasnofb:
Some instrument are poorly positioned. They are:

- the ADF indicator should go slightly down (notice the light grey area in the lower side of the gauge): http://s10.postimg.org/xx79eop55/fgfs_screen_001.png
- fuel and oil gauges are cut (notice how some numbers and words have parts cut, such as the F letter in the word Fuel , or the number 7 in 75 degrees: http://s12.postimg.org/wj8oszqot/fgfs_screen_002.png

dg-505:
Solved. See pull request #187.

PR #187, opened 2015-05-17
Position of fuel and oil gauges corrected and some other small bugfixes

dg-505:
Adessed issues #160, #179.
Other changes:
- The dial of the vac is illuminated with the instrument lights
- The magneto switch is no longer illuminated with the instrument lights


gilbertohasnofb:
Thanks for solving these issues! Right now I have no way of checking it myself, but I merged it anyway since these are very obvious 3D corrections that shouldn't cause any conflict at all. Also the instrument light changes sound very good to me. I will be able to test them (and also continue working on the plane) only in 2 days.

PR #190, opened 2015-05-19
Add pontoons, shut off all shadows, misc corrections

wlbragg:
This includes a couple other additions and corrections other than the pontoons. I found them while I was adding pontoons, sorry, I would have preferred to commit them separately.
They were misc corrections, missed additions and added the ability to turn off all ALS shadows.

The pontoons:

We need to add either auto selection depending on where you startup (land or water) or a warning when trying to select inappropriate selection.

We need someone to change the textures on the pontoons to match our themes.

We need to decide if the pontoons should be solid (no imbedded wheels) or choose to add the wheels that fold up inside the floats.

Check it out.


Juanvvc:
This is very cool! I wish also an opinion from the FDM developers. Pontoons will increase for sure the drag and weight, and probably change the COG.

If you are currently aiming only to visuals (and I'm perfectly ok with that), you might consider changing the particle system to show water drops after take offs from water.

Regarding textures, during the first iteration you can use the same texture than the fairings is using.

gilbertohasnofb:
> Regarding textures, during the first iteration you can use the same texture than the fairings is using.

That's a really clever idea. Anyway, just let me know if any textures are needed, ok?

Juanvvc:
I commented before checking the files. Pontoons use a complex texture, so it is not possible just reusing the texture on fairings.

After a visual inspection of the files, everything seems perfectly ok. I cannot test the aircraft today, but I guess we can accept this and fix any unexpected error in later commits :)

gilbertohasnofb:
In that case, where should we create the textures for them? There is hardly any space left at fuselage.png. On the other hand, wing.png has plenty of empty space and could fit nicely these textures. Would that be too messy?

onox:
I have trouble stopping after landing on the water.

gilbertohasnofb:
@onox could you please open a bug report for the tire switch bug you just described above? This way it won't get forgotten or lost.

Juanvvc:
@gilbertohasnofb I guess textures for the pontoons must be in a separated file. Check the current textures for the pontoons in floats.png: https://github.com/Juanvvc/c172p-detailed/blob/master/Models/floats.png

I my humble opinion, this texture is too grey. It could benefit from a curve to add brightness, even if they are not finally customizable for each livery.

wlbragg:
"landing on the runway, the model floated a bit above the runway"
onox, I can't reproduce this. Would you try it again and see if it is repeatable? There may be a specific order you had to do.

wlbragg:
"The texture is too grey"
Yes, I was hoping it wouldn't be too difficult to add new textures that comply with our liveries.

Something else from the modeling department, ha, where the two cables at the back end (tail section) attach to the pontoon and the aircraft, see the image http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=1350#p242283 There are some kind of cable fasteners the fasten the cable to the aircraft and the pontoon.. It would be nice if we could add those attachment points.
fastener

gilbertohasnofb:
@Juanvvc Oh, I wasn't aware we already had a float.png! In that case, I will improve it (make it less grey as you say). Also, could you guys help me make it livery dependent? So in case it's defined in the livery xml, the float.png will be a specific one, else it will use the default one. Does that make sense?

@wlbragg once the float.png is livery dependent I will create floats for all liveries, no worries :wink:

wkitty42:
i'm trying to cover several posts in this topic with just one post...

i'm not sure how it works IRL but it would be nice if we could land on land with these pontoons... i guess that means either
1. select pontoons without wheels for water landing only OR with wheels for possible landing on both
or
2. all pontoons have wheels so landing on water or land is easily done

in simpler terms, one set or two...

perhaps damage if no wheel pontoons are chosen and used when landing on land? grind'em down skidding on the pavement... what happens if pontoons are used on snow?

i like the idea of warnings if pontoons without wheels are tried to be used on land when selecting them... the question is how to place the craft in the water first and then select the pontoons...

if the FDM can be updated for the additional drag of the pontoons before the feature freeze and allow for testing, i say go for it... bugs can be fixed during the bug fix portion of the dev cycle...

onox:
Skidding and emitting sparks would be cool :smile:

dany93:
>If the FDM can be updated for the additional drag of the pontoons

For this, I will need a property like ..../pontoons = 0 or 1
The additional drag will be pure guess, unless someone would have better.

wlbragg:
Got it!
I'm doing both with wheels and without (selectable)
Yes, we'll can add damage and sparks eventually.
The friction needs adjusting both on water and on land, I or anyone else can/will get to it.:)

wlbragg:
"Also, could you guys help me make it livery dependent? So in case it's defined in the livery xml, the float.png will be a specific one, else it will use the default one. Does that make sense?"
Yes it makes sense, but I don't know what you need. I definitely want to make it livery dependent. I'll take care of adding the appropriate lines to the xml when we get to that point as it is still a little bit fresh in my mind as to "where all" some of this needs to go.

gilbertohasnofb:
@wlbragg as you probably already realised I know next to zero about xml, but once I see a working example I can generalize it for what I need (that's how I handled all xml livery files). So here is what I did: inside the folder Model/Liveries/OH-CTL there is a new file named float.png, which is the float for that livery (should be completely white except for a black tip, see: http://s8.postimg.org/lpb9z9pqt/fgfs_screen_001.png). Could you or anyone else help me to modify the OH-CTL.xml in order for that texture to become active for this livery? When this is done, I will then be able to do it for all other liveries myself. Does this make sense?

gilbertohasnofb:
@Juanvvc a brighter (less grey) version of the default floats has been uploaded. I also took the liberty to remove those two red lines from the default float, since that will clash with liveries that do not contain red on them (the float.png file located in the folder Model will be used for all liveries which do not have their own float.png). I also already uploaded livery specific textures for both OH-CTL and PT-IAO, but I need help with the xml files to activate them.

wkitty42:
you guys keep talking about grey pontoons but all i see are black ones... i'll try pulling another zip and see what i see...

gilbertohasnofb:
@wkitty42 do you mean the whole float is black? This is what you are supposed to see: http://s9.postimg.org/6ir4ia38v/fgfs_screen_001.png

wkitty42:
@gilbertohasnofb: yeah... maybe not black but really dark grey... darker than in that pic... i posted some pics earlier in my bug report... i am able to see the "boxes" that make up the top of the pontoons but it is just really dark all over... i don't think it is my monitor since it is calibrated for my setup and working conditions (low light) over here...

gilbertohasnofb:
The original texture was quite darker, this picture was done with the texture of my last commit. Can you pull it and see if it solves your problem? 

>  i posted some pics earlier in my bug report...

I can't find it.

wkitty42:
i think i did just pull your commit... i think my bug was 188 or 189 but will dig around and see...

PR #197, opened 2015-05-20
Damages: if v > 180 KIAS both wings (partially); physics.nas: popup m…

dany93:
…essages + some formal changes

 Damages: if v > 180 KIAS both wings (partially, can still fly but degraded: loss of lift)
 physics.nas: warning popup messages + some formal changes ( properties to kill engine)

See this issue message for the explanations https://github.com/Juanvvc/c172p-detailed/issues/134#issuecomment-102783068

### DIFF
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -1,22 +1,23 @@
 props.Node.new({ "/fdm/jsbsim/bushkit":0 });
 props.globals.initNode("/fdm/jsbsim/bushkit", 0, "INT");
 
-#reference Vne=158, max positive-g=5.8, max negative-g=5.8 
+# Specifications Vne=158 KIAS,
+# Normal category (1500 - 2400 lb): max positive-g=3.8, max negative-g=1.52,
+# Utility category (1500 - 2100 lb): max positive-g=4.4, max negative-g=1.76,
+# structure holds at least 150% max g's
 props.Node.new({ "limits/vne":0 });
 props.globals.initNode("limits/vne", 158, "DOUBLE");
-props.Node.new({ "limits/vne2":0 });
-props.globals.initNode("limits/vne2", 200, "DOUBLE");
 props.Node.new({ "limits/max-positive-g":0 });
-props.globals.initNode("limits/max-positive-g", 4.4, "DOUBLE");
+props.globals.initNode("limits/max-positive-g", 3.8, "DOUBLE");
 props.Node.new({ "limits/max-negative-g":0 });
-props.globals.initNode("limits/max-negative-g", -1.76, "DOUBLE");
+props.globals.initNode("limits/max-negative-g", -1.52, "DOUBLE");
 
 var g = getprop("/fdm/jsbsim/accelerations/Nz");
 var max_positive = getprop("limits/max-positive-g");
 var max_negative = getprop("limits/max-negative-g");
 
 var aero_coeff = "fdm/jsbsim/aero/coefficient/";
-#roll moment due to (diedra + roll rate + yaw rate + ailerons) ==> asymmetry to break one wing
+#Roll moment due to (diedra + roll rate + yaw rate + ailerons) ==> asymmetry to break one wing
 var roll_moment = getprop(aero_coeff~"Clb")+getprop(aero_coeff~"Clp")+getprop(aero_coeff~"Clr")+getprop(aero_coeff~"ClDa");
 
 var gears = "fdm/jsbsim/gear/";
@@ -43,9 +44,10 @@ var fairing3 = 0;
 #props.globals.initNode("/sim/rendering/leftwingdamage", 0, "INT");
 #props.Node.new({ "/sim/rendering/allfix":0 });
 #props.globals.initNode("/sim/rendering/allfix", 0, "INT");
+
 var resetalldamage = func
 {
-	setprop("/controls/engines/engine/magnetos", 1);
+    setprop("/fdm/jsbsim/propulsion/tank[2]/priority", 1);
 	setprop(gears~"unit[0]/broken", 0);
 	setprop(gears~"unit[1]/broken", 0);
 	setprop(gears~"unit[2]/broken", 0);
@@ -78,7 +80,7 @@ var nosegearbroke = func
 		setprop(contact~"unit[6]/z-position", -17.7);
 	
 	setprop(gears~"unit[0]/z-position", 0);
-	setprop("/controls/engines/engine/magnetos", 0);
+    setprop("/fdm/jsbsim/propulsion/tank[2]/priority", 0);
 	setprop(gears~"unit[0]/broken", 1);
 }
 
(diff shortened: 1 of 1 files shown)


wlbragg:
Dany, what is the reason for changing the kill engine to use "Float chamber" instead of magneto?

I'm not sure if that will be an issue or not. I think part of the reason I used magneto was because autostart cleared out the off condition (magnetos = 2). Will autostart still clear the off condition (tank[2]/priority = 1)?
Is there a benefit to using tank[2]/priority over magnetos?

I'm merging anyway. If it is an issue I can change it back.

wlbragg:
Dany, you used wing-damage > 0.5, for the stop engine condition, that represents a broken wing and not just a damaged wing, correct?

So it looks like you already accounted for everything here in this commit.:)

dany93:
A broken wing would have been wing-damage = 1. Both wings can have = 1 (left-wing =1 and right-wing = 1 ==> both broken, 100% total loss of lift and loss of aileron control).
wlbragg assumed just a "damaged" (??) wing to stop the engine close to the ground. He was just thinking of damages when hitting the ground (e.g. with a wing). But this stopped the engine also when trying an emergency landing with wings partially damaged in-flight.
I have set the in-flight maximum partial damage at 0.3 on each wing (or 0.12 on one wing) because the aircraft is too difficult or impossible to control above these values.
Hence my > 0.5 value to kill the engine, to allow emergency landing with the engine still running. I could have set wing-damage = 1 as well, with the same result as far as I'm concerned. Close to the ground is wlbragg domain. If he prefers a different threshold, that's compatible, just > 0.3 with my current wing damage values.

dany93:
>Dany, what is the reason for changing the kill engine to use "Float chamber" instead of magneto?

>I'm not sure if that will be an issue or not. I think part of the reason I used magneto was because autostart cleared out the off condition (magnetos = 2). Will autostart still clear the off condition (tank[2]/priority = 1)?
Is there a benefit to using tank[2]/priority over magnetos?

"Float chamber" seems more logical and less unrealistic to me than seeing the magnetos key turning down to zero. With "Float chamber", the user does not see why the engine stopped. However, each way is more or less artificial.
I think you used magneto at least because you had no other choice at this time (no float chamber).
I had not seen this "Autostart" effect and I know nothing about it. I hope this will not make a dysfunction.
I see no benefit to using tank[2]/priority over magnetos, other than my first phrase.

wkitty42:
i had reported about the brakes thing somewhere else... either in the forums or as a comment here...

i think it would be better if the reminder text were altered to point you to the "before takeoff" checklist which sets the brakes and runs you through the systems checks...

additionally the "normal takeoff" checklist should have you SHIFT-B to unset the brakes /after/ you set the throttle up... right now it doesn't tell you to unset the brakes...

dany93:
Yes, I remember your remark about the brakes (in the forum, I think).
In the context of "Autostart" (simplified procedure) there is no reason to point to the "before takeoff" checklist.

wkitty42:
@dany93: why not? that's the next step isn't it? set the brakes, run up to 1700 rpm, check the magnetos, and so on? then we go to the takeoff checklist where we set something i can't remember or get to right now, run up to max throttle roll out and pull back at 55kts...

on that note to pull back at 55kts, most of the time i find i'm already airborne before then and 99% of the time i don't have to pull back at all for the first 100 feet ALT or so... by then i've got the AP on for ROL stabilization and then switch to the heading bug and set the vertical rate to +300 up to +700 ascent until i reach my desired flight level...

dany93:
If users choose "Autostart", they are not interested by the complete procedure. I know that part of them (even confirmed) are like that. They prefer flying immediately, hopefully with realism.
For beginners, it must be like this IMHO.
The complete procedure is interesting, but only later.

For takeoff, the aircraft is balanced (horizontal stabilizer, trim) so that she climbs at around 70 KIAS with no action on the yoke. If we don't pull back during rolling on the runway, she takes off on its own around (or just above) this velocity.
This procedure can be used IRL if we have a long runway, but in practice the yoke is generally pulled at the recommended rotation velocity, would it be only to save the gear mechanics. Just after takeoff, a levelled flight is temporarily done in order to gain a safe velocity
Later, we have to correct the elevator trim (nose down) for cruise. That's a normal setting in manual control.

wkitty42:
@dany93: on the autostart, i agree with you... i just thought we were doing something a little different than others with autostart... i thought this because when you use the autostart on others, they actually start whereas with this one, we still have to set the brake and turn the starter... all i'm saying is lets either go full autostart (set the brake and start the craft) OR let's point them to the "before takeoff" checklist... they can still go full throttle if they want...

on the trim, yes... i'm still learning that... however, in a default load, with one 180 lb pilot, no cargo or passengers, and the default fuel load, i've been getting off the ground before 55 kts without doing anything other than one tick on the trim... just rudder steering and then switching to the ailerons once up and the horizon starts tilting...

dany93:
>additionally the "normal takeoff" checklist should have you SHIFT-B to unset the brakes /after/ you set the throttle up... right now it doesn't tell you to unset the brakes...

It *tells* you to unset the parking brake at the top of "Before Takeoff" (2nd SHIFT-B, should be written "UNSET"). Do not forget "Magnetos CHECK" must be done with the brakes ON.
Like this, we go into "Normal Takeoff" or Short Field Takeoff" with the parking brake unset.
It doesn't matter in FG if we start at the runway threshold.

However, IRL that's a bit different, you unset the parking brake to leave the parking, and roll to the runway holding point, to do the last checklist ("Before Takeoff"), before ATC authorization and alignment for takeoff.
In this more general and realistic case, the best for the parking brake unset is at the end of "Starting Engine", just before leaving the parking.

wkitty42:
@dany93: ok... i see what you are saying... 99% of the time, i'm appearing at the end of the runway to do all this... rarely do i start in a parking spot... mainly because too hard to steer with keyboard and don't know signs and layout to properly follow taxi methods to runway... yes, i guess there should be another issue just to double check that the flow works... i didn't realize the hijacking of this issue got so bad... my apologies to all...

dany93:
> mainly because too hard to steer with keyboard

Do you know that, with the mouse in "*controlling flight controls*" mode (by pressing "Tab" once, I presume that you use this), holding **left mouse button** gives you the control of the **rudder**? This enables you to accurately steer the aircraft on the ground. You just have to release it before rotation to recover the ailerons and elevator controls.

Juanvvc:
Please, wkitty42, remember the stall speed is around 45 knots. The checklists do not say "pull at 55 knots", but "Lift nose wheel at 55 knots". It is implied that if the aircraft wants to fly before 55 knots (and it will, remember it has enough lift to fly if it runs faster than 45 knots), you have to PUSH the yoke to keep the aircraft on ground. At 55 knots, just release the force your are applying on the yoke and the aircraft takes off. Sometimes, pulling is not necessary :)

Do not let the aircraft fly before it has a safe speed!

wkitty42:
@dany93 yes.. but it is cumbersome for my way of access...
@Juanvvc yes, i see that but i don't try to keep her down... if she wants to fly and be free, who am i to stop her ;) [/joke] but seriously... i let her come up and i catch her before the fall... turning on the AP to catch ROL and then setting the ascend back from some 2000 fpm to 10fpm to gain speed and then up to no more than 700 fpm... that's another thing, though... why does the AP end up at 1400+ fpm on the ascend when the altitude is set for 00000 and the ascend rate hasn't been set... the ascend rate should be zero 00000 until manually set... [/hijack]

dany93:
> why does the AP end up at 1400+ fpm on the ascend when the altitude is set for 00000 and the ascend rate hasn't been set... the ascend rate should be zero 00000 until manually set...

Not sure. See the KAP 140 Manual p. 86 (or 72/102). https://www.bendixking.com/HWL/media/PDF/PilotsGuides/PG-KAP140.pdf
When engaged, it takes the *current* vertical speed and VS HOLD mode (see, you are better than me in English).
The FG version works like this.

wkitty42:
@dany93 i've read that several times in the copy i have of that manual... just now what you say and what it says makes sense... that part never stuck with me before... pg 55 in my Rev.2 May/02 manual...

"2. AUTOPILOT ENGAGE/DISEN-
GAGE (AP) BUTTON - When
pushed, engages autopilot if all logic
conditions are met. The autopilot
will engage in the basic roll (ROL)
mode which functions as a wing lev-
eler and in the vertical speed (VS)
hold mode. The commanded verti-
cal speed will be displayed in the
upper right corner of autopilot dis-
play area. **The captured VS will be
the vertical speed present at the
moment of AP button press.** When
pressed again, will disengage the
autopilot. For software version 03/01"
**emphasis** added...

PR #207, opened 2015-05-25
added EFHF sticker, fix throttle labeling, some dirt

tigert:
Tweaks :-)


tigert:
Do you guys have an XCF file of the panel by the way? It's a bit difficult to work with binary files together, especially with a flat PNG without layers, let me know if there is one somewhere..?

gilbertohasnofb:
> Do you guys have an XCF file of the panel by the way?

Yes we do, but that's a different repository since the xcf files are huge. See: https://github.com/gilbertohasnofb/c172p-detailed-liveries

Really nice contribution, I love the sticker and the darker spots around the levers! But if you don't mind, I will manually add your modifications to the xcf myself instead of accepting this pull request. This is because you are using a different font for the throttle text and I think it doesn't look that nice this way. Would that be fine with you?

As for the "dirt", are these the darker spots around the levers of mixture, throttle, carb heater, etc. or is there some other dirt added? Please let me know so I can add them to the xcf exactly as you did.

gilbertohasnofb:
@tigert okay, so I incorporated all your changes but using the same font for the throttle label to the panel.png file in this repository and also to the xcf on the liveries repository. I hope this is all right with you!

PR #208, opened 2015-05-25
Battery master needs to be on to check fuel gauges

tigert:
This is not 100% accurate, as per the C172 operating handbook the fuel gauges are checked during the preflight check. However the complete preflight check is not included in the checklists, and the fuel gauge check is instead part of the "before starting engine" checklist, which makes sense in practice. So we need to move the battery master item a few lines higher to be able to see the fuel gauge values. 

In the handbook the battery is turned off afterwards to save the battery since the preflight takes a while, but turning it off here does not make sense, as you'll be starting the engine in a few moments anyway in the sim when you follow the list.

### DIFF
--- a/c172-checklists.xml
+++ b/c172-checklists.xml
@@ -80,6 +80,24 @@
       </marker>
     </item>
     <item>
+      <name>Battery Master Switch</name>
+      <value>ON</value>
+      <marker>
+        <x-m>-0.3500</x-m>
+        <y-m>-0.4000</y-m>
+        <z-m>-0.2300</z-m>
+        <scale>1.0000</scale>
+      </marker>
+      <condition>
+        <property>/controls/engines/engine[0]/master-bat</property>
+      </condition>
+      <binding>
+        <command>property-assign</command>
+        <property>/controls/engines/engine[0]/master-bat</property>
+        <value type="bool">true</value>
+      </binding>
+    </item>
+    <item>
       <name>Fuel Quantity</name>
       <value>CHECK</value>
       <marker>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
Cool, thanks!

> However the complete preflight check is not included in the checklists

What about adding a preflight check? Why is it not included in our checklists?

tigert:
I am confused. The change looks correct in this pull request, but for some reason I don't see it in master branch after you merged this..

Battery master ON should be right on line 82, right before "Fuel Quantity: CHECK".

tigert:
The complete preflight check could indeed be added.

The thing is, there should be a way to show what you are checking - there is little point in going through the walk-around inspection in the cockpit, without actually walking around and showing what to check.

https://www.youtube.com/watch?v=C-yqLXZtkjA < this is what it sorta involves. This is an older C172 with a Continental engine, so it has the oil dipstick on the other side of the cowling, which looks a bit different from the C172P.

But basically you first go in the cockpit, remove the gust lock pin* so you can actually move the control surfaces when walking around - and then you also turn on master battery for a moment and extend flaps, so you can later check them. Also, you want to check the beacon and strobe lights (and navigation and landing lights, especially if you are going on a night flight) because you need to check them for any broken bulbs. All this eats battery, so often you might need to do a first walk-around to just check the lights and then turn lights and battery master off and then check the control surfaces, oil and fuel levels etc like on that video.

So this could be modelled, but we should do this somehow that it actually a) works in practice and b) adds something useful as a learning / practice aid. Do you guys have ideas?

This could be nice if the checklist arrow button could switch camera angles and we could do those magenta circles to point at things to check. Or maybe the walker pilot character could point at stuff? :) It's quite big animation challenge though.

(* Gust lock, oh, we could model that too, I can take pics, but it looks roughly like this: http://www.weekendcfii.com/photos/c172preflt/C172_gust_lock.jpg)

onox:
@tigert You reverted the change in commit 475b20e9cf7706bcbaa672eed84de766fb064789

tigert:
Duh. True! Will add a new change :-P

Juanvvc:
I was me who removed (or decided not to include, I'm not sure) the pre-flight checklist, creating this issue. Please, check the comments to #210 for a longer explanation.

"Or maybe the walker pilot character could point at stuff? :)"

That would be amazing! It is doable and probably not even too difficult to automatize, but I guess it is more fun than really practical.

wkitty42:
is that "gust lock" the lock that is referenced in the very last checklist when shutting down the craft and going somewhere else? i thought they were talking about one of those steering wheel locking bars like some folks put on their cars :laughing:

tigert:
yeah, it is a pin that goes through the yoke bushing and the pipe, and also has a long extension and warning plate that goes over the magneto/start key hole when it is installed, to stop you from killing yourself by taking off with the lock in place. I can take a closeup of the thing for you today.

Let me know of any other closeups you wish to have from a c172, today is a good day since I will go to the airport and the hangar has a bunch of 172's :-)

wkitty42:
ok, if that lock in the very last checklist is the gust lock, perhaps we can have that wording added?

PR #212, opened 2015-05-26
Bush wheels and Float management: weight and drag. Merging "stall and spin" FDM in c172p.xml

dany93:
At this day, only in the c172p.xml FDM (see issue https://github.com/Juanvvc/c172p-detailed/issues/211#issue-80872837)

New: the "bushkit.xml" file.
I'm not 100% satisfied because these loads appear in "Fuel and Payload", but I could not find a better way.
The values for CG location, extra weight and drag are evaluated and will be changed when (if) we have better data.
At least, the tool exists and it works, changing the values is easy.
The glide ratio at 65 KIAS goes down from about 9 for the original gear to 6.5 with the amphibious.

@wlbragg: With bushkit.xml, I use /fdm/jsbsim/bushkit = 1 (to 4). Closest to the source as you suggested firstly. *Not* sim/model/c172p/bushkit_flag_0 (to _4) = 0 or 1. Sorry for having asked for these properties at start, and thank you.


dany93:
Added: Merging of *c172p-fdm-dany93.xml* (stall and spin) in *c172p.xml single file*.
c172p.xml now contains *Bush wheels and Float management: weight and drag* and the *stall and spin* features.

@wlbragg: Please if you can check that I have correctly kept all your changes which were in the initial c172p.xml, thanks.
I have a backup copy of the initial files.

onox:
Contacts points are ok :+1:, so if @wlbragg is ok with the changes then it can be merged :smile:

wlbragg:
I'm in the middle of pontoon damage and can't easily merge and check this until I commit my local changes. As soon as I get the pontoon damage commited locally I will merge this and check it out. I can hardly wait, thanks.

dany93:
The climbing rate is very low with Floats and with Amphibious.

Maybe I have put too much drag for these (the extra weight does not give such an effect). The extra weight can be quantified, but up to now the extra drag is purely estimated. Waiting for more information from pilots who know the c172. Easy to fiddle with this in the bushkit.xml file.

I can easily decrease this drag values, now if you want.

If the drag is plausible (see the glide ratio in my first post https://github.com/Juanvvc/c172p-detailed/pull/212#issue-80878085) this can lead to consider a 180 hp engine.
The 180 hp engine is easy to change, but it also needs a new propeller (less easy to match).
Also (if the 180 hp is convincing with a propeller which works well), I do not see how to keep 160 hp and 180 hp as *options* (possibly my lack of knowledge, but I've not found). I feel that it would be either two FDMs (and we fall back in the management difficulties) or only one FDM with 180 hp. My experience with DR 400 (120 hp then 180 hp) is that almost everyone flies with 180 hp.

wkitty42:
i'm fixing to go fly this craft for a little bit... one thing i have noticed in the last several days is that it can easily get up to 140kts after taking off and flying level... i always have to specifically remember to back off the throttle now... that's with standard gear and either of the float setups... i haven't tried the bush gear tires, though...

dany93:
>easily get up to 140kts after taking off and flying level... i always have to specifically remember to back off the throttle now..

See this response http://forum.flightgear.org/viewtopic.php?f=4&t=25157&p=239828#p239828. Where  I found 130 KIAS at 2700 RPM.
 And this one http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=945#p240086

POH p. 5-18 says 114 KTAS at max load (2400 lbs), 2500 ft, **2500 RPM**. 
C172p-detailed reaches 115 KIAS , which makes about 120 KTAS. Anyway, I guess you read KIAS.
Weren't you above 2500 RPM? Or, worst, in over-regime (more than 2700 RPM)?
At levelled flight, reaching over-regime at full thrust is normal (and forbidden).

wkitty42:
i understand that weight and drag is not in effect yet... it was just surprising that i was able to gain so much more speed... over-regime? i've seen it mentioned but i don't know what it actually is and there are (currently) no consequences for doing it...

as an example, i took off from EGAA (i think that's the ICAO in ireland) with the amphibious pontoons on... i left the throttle at max after i reached ~400 ALT and leveled off with the AP on... i reached just over 150 kts indicated on the left HUD... the ground speed was maybe 20 kts less in that test... advanced weather was enabled and set as well as real weather METAR from that location... this was mid-afternoon or so...

can a pontoon equipped 172 craft actually reach 150 kts indicated at max power within 5 minutes of takeoff and level flight at low altitude?

dany93:
>i understand that weight and drag is not in effect yet...
can a pontoon equipped 172 craft actually reach 150 kts indicated at max power within 5 minutes of takeoff and level flight at low altitude?

Don't you see an inconsistency between these two phrases?

>...i left the throttle at max after i reached ~400 ALT and leveled off with the AP on...
over-regime? i've seen it mentioned but i don't know what it actually is
__________________________________________________________
>....Or, worst, in over-regime (more than 2700 RPM)?
This response http://forum.flightgear.org/viewtopic.php?f=4&t=25157&p=239828#p239828. Where I found 130 KIAS at 2700 RPM.
And this one http://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=945#p240086

wkitty42:
i was referencing the old(er) flying sessions i had... in very few of them, before pontoons came along, was i ever able to get much over 120 kts... i don't even know which FDM i was using when i wrote my comment... i think it is the merged one after the removal of the other with spin which i've never flown separately...

the second part was a question about real life... basically "can a pontoon c172p reach 150 kts at all?"

dany93:
>the second part was a question about real life... basically "can a pontoon c172p reach 150 kts at all?"
Obviously not. Even without pontoons it doesn't (in levelled flight and without destroying its engine).

wkitty42:
fair enough although not obvious to me... it has been well over 35 years since i last flew in a small craft and i wasn't aware of the actual numbers involved... admittedly, while i have looked over some documentation, it hasn't become embedded in my ram yet... i still much prefer actual paper documentation that i can fall asleep reading... information seems to stick better :wink:

Issue #158, opened 2015-05-09
Better texture mapping of the baggage door


gilbertohasnofb:
The fuselage texture shouldn't be shifted in the baggage door (at least not by that much). See: http://s15.postimg.org/c026f7tez/fgfs_screen_005.png

dg-505:
See pull request #188

gilbertohasnofb:
@dg-505 has this correction been lost due to the pull request https://github.com/Juanvvc/c172p-detailed/pull/188 not being accepted?

dg-505:
Unfortunately this fix indeed has been lost. I can correct it again in the c172p-detailed.blend, but because I don't know how to export it properly to the .ac file I would only push the .blend file so that someone with more knowledge about Blender can export it.

gilbertohasnofb:
> I would only push the .blend file so that someone with more knowledge about Blender can export it.

That could work, we could ask @Juanvvc if he can do it. Also, you could try asking about the .ac export in our forum thread, certainly someone there would explain it to you. Thanks a lot!

dg-505:
Okay, I'm gonna do it later today, because at the moment I'm not at home.

gilbertohasnofb:
No rush, Jonathan!

dg-505:
Roger :smiley:

gilbertohasnofb:
I believe this has been corrected, right? I am closing this issue, but let me know if there is still any work to be done on thus UV map.

PR #188, opened 2015-05-18
Some 3D-improvements

dg-505:
Did some improvements wit the exterior model: More detailed struts for the landing gear, and corrected issue #158.

PR #217, opened 2015-05-28
Fixed issue #158

dg-505:
Because of the closing of PR #188 this fix has been lost.
I corrected it again in the c172p-detailed.blend. For some reason the new blend-file is half a MB smaller than the previous file. Because I don't know if any essential information got lost due to the fix, I included the c172p-detailed.blend1 as a backup file.
Maybe someone could export the .blend file to the c172p-detailed.ac.

Somwhere @Juanvvc mentioned that you guys use a special plugin for Blender to export the .ac files.
I already have such a plugin, but I don't know if this does the same thing like your plugin. Maybe you could give me a download link, so I can do the export myself properly before committing.


tigert:
One thing I have heard that might work is to have the older Blender which has the ac3d export/import included, and use that for the conversion. With the new version, then you need to enable "legacy mesh support" or something similar in settings to save the older .blend format.

onox:
I know of [majic79/Blender-AC3D](https://github.com/majic79/Blender-AC3D).

@zlsa is working on one for Blender 2.7 at [zlsa/io_scene_ac3d_v2](https://github.com/zlsa/io_scene_ac3d_v2).

wkitty42:
FWIW: on the blender-ac3d plugin, i currently have the io-scene one but i have no way to know if i have the right one or if i've used it correctly... i don't recall where i saw it mentioned... my blender is 2.69 for now...

Juanvvc:
I will merge this manually tomorrow. In addition, I'll add io_scene_ac3d we are using to the repository. thevirtualfer made some changes to this script to support groups. I'm not sure if we are still using these groups. If we are not, the "official" plugin will work. In we are using groups, it is only in the main .blend file. The other files can be exported with the "official" io_scene_ac3d plugin.

thevirtualfer:
Juan !

As far as i remember, the groups are used into the .ac file by the engine
itself.
The original plugin exports the model but doesn't create the necessary
groups into the .ac file.

Cheers, bro !


2015-05-30 20:31 GMT-03:00 Juan Vera del Campo <notifications@github.com>:

dg-505:
Thank you very much, Juan

gilbertohasnofb:
@thevirtualfer hey, welcome back! :smile:

onox:
@Juanvvc Have you seen [NikolaiVChr/Blender-AC3D](https://github.com/NikolaiVChr/Blender-AC3D)?

Juanvvc:
Manually merged. In addition, I have added to the repository the script I'm using to export the .blend file. It is thevirtualfer's version of one of the scripts linked in the wiki, but I'm not sure exactly which one. It seems there are many different scripts made by different authors.

dg-505:
Thanks, Juanvvc

PR #221, opened 2015-05-29
Restore name

IAHM-COL:
This merge brings back the aircraft name to c172p
Removing the -detailed part of it recursively and where-else is required

An important step to create the "release" for FGDATA


IAHM-COL:
Dear Ludomo
this is also not one of the "listed" issues, but definitely something needed to do in order to look forward to create the "stable" commit for stuart merge

I have cleaned off the past history. To accept your preference as the "curator" too.

Best,
Israel

Juanvvc:
Guys, what you think if we DON'T merge this into master, but create a new branch called "v3.6"?

This new branch will be "the freeze" we were discussing. Only bug fixing will be allowed in this branch.

Also, please, notice you changed some .ac files, and I guess the .blend files are no longer in sync. I don't care, since I guess the .blend files won't be released to FGDATA.

IAHM-COL:
1. I don't change the "content" of the ac files. Only the "name"
so the blend and the ac remains in sync ;) 
==keeping the ac content unaltered was one of the challenges of the renaming (PHEW)

2. I think given this project is officially replacing the c172p the rename should really become effective and going into master instead
That will make easier to keep the FGDATA version move forward in the future

(as oppose of attempt the renaming surgery every single time)

3. I understand there could be some romantic attachment over the -detailed prefix, but I think it keeps it simple to just let it go. 

4. I suggest not making a "release branch" 
instead making a release "TAG"

Best
Israel

IAHM-COL:
About removing the .blend files for importing into FGDATA, 
That's great idea --but I suggest it to be done on the "importing" proccedure instead of here in the repo

(I think that is what you meant, rite?)

wlbragg:
My 2 cents, I think this should merge into master BUT NOT UNTIL WE VERIFY THE NAME CHANGE IS WHAT THEY WANT. I am hesitant to make this merge until such a time that it is verified by stuart that this is the case.

IAHM-COL:
yes .
I agree with Wlbragg, too

but keep in mind, It is my opinion this merge should come as soon as possible

So
1) futures commits into master come with the new name. Not the old one.
2) we get time to test the rename went as expected. Most people accessing the renamed plane for testing purposes and detect critical bugs (example if some rename didnt go as expected and something was left behing)
so the earliest we merge this into master the sooner we are doing the pertinent testing

3) I dont see why the rename will not be needed --unless fgdata would change the aircraft name now.
but 

4) if so desired, in the future, this merge can be reverted really easy too.

so there is only benefits into bringing this early now.

wlbragg:
"Also, please, notice you changed some .ac files and I guess the .blend files are no longer in sync with them"
I don't think that is an issue either, but when it is all done I would like to name the blends the same as the ac's. If the name change is finalized.

I re-read the other pull request with the c172p history again. There is a point Israel brought out that I think really does need to be addressed, that is recognition of the c172p developers.

Juanvvc, you want to followup on the dev-list to get the answer to name change
I just tried to post there and I either sent about 5 duplicate emails or my mail manager wasn't lying to me when it said my outbound password is corrupted.:)

Juanvvc:
"I understand there could be some romantic attachment over the -detailed prefix, but I think it keeps it simple to just let it go."

Come on! There is nothing romantic about it and each time I see the directory I wonder if an adjective after a name is good English :)

My main concern is, if our aircraft is in included FGDATA, how we will continue the development. I believe the best way is maintaining an external directory for our development. This creates two versions of the aircraft with the same directory name: one in $FG_ROOT and the other in a directory listed in $FG_AIRCRAFT

In my experience, having the same aircraft in $FG_ROOT and $FG_AIRCRAFT is broken in FlightGear. This happened to me when I tried to have my personalized version of the C337 or the DA40, and I had to remove the aircraft in $FG_ROOT.

Hence, in my experience, if an aircraft if in $FG_ROOT, it is better to use a name such as "c172p-dev" at least for the directory name and -set file in $FG_AIRCRAFT. I agree we can change the name of all local files to c172p without any problem.

Juanvvc:
In case I didn't manage to explain myself (it has been an exhausting day):

I totally agree with removing right now the "-detailed" label in the name of our files, even in the master branch. It is the directory name and the -set.xml file my concern.

dany93:
>My main concern is, if our aircraft is included FGDATA, how we will continue the development. I believe the best way is maintaining an external directory for our development.

I hadn't realized that but *yes*, this confusion in the directory and -set name would be an important issue. Thank you juanvvc for having brought it up. A name like "c172p-dev" is necessary for the development.

IAHM-COL:
>I don't think that is an issue either, but when it is all done I would like to name the blends the same as the ac's. If the name change is finalized.

Dear Wlbragg
If I recall correctly, I renamed the .blend and the .ac (the file names)
See here:
https://github.com/IAHM-COL/c172p-detailed/commit/0e8cc049db718bdc716507df5a635c3443ab5610

For all files that undewent renaming
(anything -detailed in the tree of files basically!)

What I prevent doing it was editing the .ac contents.

(so as an example, the ac file my reffer internally to -detailed)
That's all

IAHM-COL:
>My main concern is, if our aircraft is included FGDATA, how we will continue the development. I believe the best way is maintaining an external directory for our development. This creates two versions of the aircraft with the same directory name: one in $FG_ROOT and the other in a directory listed in $FG_AIRCRAFT

Dear Ludomo. 
Here I am having quite some trouble understanding you.
*Im sorry*

My suggestion for keeping the development on the c172p aircraft (over this repository) is to effectively rename the project to c172p, then continue doing the work as it has been till now: you guys create branches to implement new feature (dual controls, pontons, etc) and merge them into master. 
The first difference is the project is no longer called -detailed. 

When you reach an stable status that is "good" for FGDATA, you tag the release (you create a #tag# v3.6, not a #branch# v3.6). Then replace the content of an fgdata fork with the content of the tagged zip file (removing the sourcres during this step if so wished) and sending the new commit over fgdata. 

After the released tag, nothing changes over c172 (now detailed). You keep working in this repository, which is, for all practical purposes, the development repository of c172p. 
You can even change the #repository name# to c172p-dev (or keep detailed), but the content is faithful to the content to be imported to fgdata (that is with the name change)

This way you prevent doing this surgery of file replacing and xml edition just before an importing to fgdata.

This way also, there will not be 2 versions of c172p over FGDATA or FGROOT. There will be only 1 c172p in FGDATA (Aircraft/c172p), whose content will be identical to the tagged release here. [Maybe source files removed].

onox:
I think we should rename the aircraft to c172p as @IAHM-COL says. Then when we think we can do a release, we make a release here on github and give the FGDATA devs the .zip file.

Users trying to fly the c172p from FGDATA will then always be flying our version. The only people who have to take care is us, because we have to remove/disable the directory in FGDATA so that we're sure that FG chooses the development version.

>My main concern is, if our aircraft is included FGDATA, how we will continue the development. I believe the best way is maintaining an external directory for our development.

I say we will continue development here? :smile: Just rename the directory in FGDATA to `c172p_old` or something.

dany93:
>Here I am having quite some trouble understanding you.

I think I'm understanding juanvvc.

I have a "personal", separate folder *out of git* where I do my changes and tests on c172p-detailed. When doing this, my git folder is renamed, in order to not be seen by FGRun (or a command line).
I admit this can be due to a lack of practice and not good enough knowledge of git, and that it may be done with a separate branch on my personal rep (github.com/dany93/c172p-detailed).

Whatever the method, we will have a c172p in FG_ROOT and another one in our "development" folder (which can be a "personal" one or a git personal branch). **How can we have FGFS make the difference?**
FGFS makes the difference with the "aircraft-name"-set.xml. When I have this confusion, I only rename this "aircraft-name"-set in my personal folder. Renaming only the "aircraft-name"-set.xml (c172p-dev-set.xml) in our development folder or branch can be a way for doing this without trouble with the paths inside which call the aircraft folder name. The aircraft folder keeps the name "c172p.xml", everything inside is the same, except "c172p-dev-set.xml".

*(Sorry if I waste your time because my poor knowledge of git)*

onox:
@dany93 Why don't you just let FGRun or command line point to your git folder?

dany93:
Because I do my tests and changes on a personal, separate folder, in order to avoid doing a mess in my git folder. This way, it remains as a reference or kind of backup.
My FGRun points to my git folder (among others), but it is renamed except when I want to test my changes copied in it before pushing (or when I want to run your last github version).

onox:
But as long as you don't commit anything, you can mess in your git folder as much as you want :stuck_out_tongue_winking_eye:

Juanvvc:
dany93 understood correctly my concerns: I prefer the development directory for our aircraft out of $FG_ROOT, and FligthGear doesn't like very much having the same aircraft in two different directories.

Is it possible to use $FG_ROOT/Aircraft/c172p as the development directory? I mean, without any additional commands such as "mv c172p_old c172p; git pull; mv cp172p c172p_old; cp -r $DEVDIR/c172p-dev c172p"

I think if would be possible with git with submodules. IAHM-COL, would it be a solution?

dany93:
@onox 
Two questions i this case:
- How can I delete all my changes if nothing works? Anyway, if I have something good to be committed, I have to sort out and clean my comments and unwanted lines from this git working folder.
- This same git folder is also the receiver for my git pull from your juanvvc:master... ==> Interferences? Possible conflicts?

onox:
@dany93 If you want to remove everything, you do `git reset --hard`. If you want to unstage certain files you do `git reset HEAD <file>`. Read what `git status` says.

If you do `git pull` you might have to merge yes. But if you keep your git folder clean, do git pull there, and then copy your files from your non-git folder to the git folder, then you will blindly override anything. Is that what you want? If you just make sure you commit your stuff before you do git pull, then your changes are set in stone. If you then messed up trying to resolve any merge conflict, then you just do `git reset --hard <commit ID of my last commit>` and you're back to where you were.

Another thing you can do is `git stash` and `git stash pop`. `git stash` will save all your changes.

@Juanvvc Why don't you use --fg-aircraft?

dany93:
Thanks, onox. Added to my notes.

wkitty42:
my 2 pesos, FWTW?
1. find out what the core devs want for the name. i think i read something in the dev list about replacing the current craft with this one but that was a couple of weeks ago...
2. maybe branch and then tag in the branch... that way fixes can be applied to that version of the craft without having to deal with everything newer in the main line... those fixes can then also be added into the main line if they are needed there...

informational: i do like @dany93 does... i have a git repo directory for this craft... i copy the contents (without the .git directory) into my ""flying"" directory (aka ~/myflightgear/Aircraft/c172p-detailed) and i test from there... this is the only craft i do like this... everything else, fgfs, fgdata, fgaddon, is used from their repos...

IAHM-COL:
>

dany93 understood correctly my concerns: I prefer the development directory for our aircraft out of $FG_ROOT, and FligthGear doesn't like very much having the same aircraft in two different directories.
>
>Is it possible to use $FG_ROOT/Aircraft/c172p as the development directory? I mean, without any additional commands such as "mv c172p_old c172p; git pull; mv cp172p c172p_old; cp -r $DEVDIR/c172p-dev c172p"
>
>I think if would be possible with git with submodules. IAHM-COL, would it be a solution? I want to be able to pull FGDATA (containing the frozen c172p) and the development directory of the c172p separately.


Dear Ludomo
I can't understand why you are choosing all the ways that can possibly make a managerial mess?
That can create undesirable forks, or complicated strategies?
Remembers, always, simpler is better (KISS strategy)
[Seeing unnecessary complicated solutions seem to be the FG style, to gratify our "thinkers" or something]

1. c172p as submodules?!
Not a good idea. 
Why?
The c172p is the only aircraft left in the FGDATA tree. 
When you clone FGDATA, it comes only with 1 aircraft. If you make it a submodules it will come with NO aircraft. [Except you explicitly initialize and update the submodule!]
I think leaving FGDATA without its aircraft, completely or only the UFO is not the way to go

2. Other aircrafts as submodules?!
Of course! the idea is fantastic and it works like a charm. But then you have the devels vetoing ideas just because # I # propose them. Very personal and not very technical way of chosing. but oh well.

3. branching before renaming instead of renaming master.
This is why you propose initially. It is doable, but totally inconvenient.
why?
Because you loose the possibility of decide LATER what is the stable commit. 
You need to decide who is stable, branch, then rename. (and it is a marriage about that commit)

If you rename, then you can choose the stable commit now or later, whenever is convenient, and the name is been as required for a good time!.

Think on the future too. If you do not rename master, 3 or 4 month you have a new "stable" to merge in FGDATA and the "renaming" has to be performed again, because the commit that executed the magic is back in the past, separated in an "stable" branch, and possibly doesnt even do a complete job anymore [Like new files not renamed by the commit, or new lines not corrected!]

4. Why I propose is fundamentally working as if c172p-detailed is a submodule, but not being incorporated this way to fgdata. 

The steps are simple. 
a. Rename the aircraft to c172p (as soon as possible)
b. keep working as you have now --you just drop the -detailed part
c. when the aircraft is FGDATA ready, create a tag. Give the devels the zip. (as ONOX also say)
d. Tell the devels that you will be O.K. if they want to delete the sources before importing to FGDATA 
(.blend, .xcf, others?)
e. Inform everyone that changes on the code of the c172p made over FGDATA are unmergeable to the developing repository hosted here, so it is extremely advisable to pull request here not directly to FGDATA for this aircraft
f. DO NOT branch for a release. TAG keeps it simple!

***

# Not renaming #

Not renaming, or appending a -dev, -detailed, or whatever you imagine to this aircraft generates an administrative mayhem. as described over number 3 above.

Just keep the repository name as c172p-detailed or c172p-dev to indicate this fact that this is the developing repository, but allow the content to be FGDATA compatible.

You told me, do not look in the past. Why are you trying to maintain the c172p-old active? this is clearly a great step forward into the future for this aircraft. Anyone willing to experience, look on the old aircraft should do a git clone and git checkout any historical commit. That's it. 
It can be done easily in FGDATA. And if you appreciate the technical elegance and beauty of my "merge history commits" merge, then you will get that benefit here too. 

(it is wise, sometimes drop the philosophy and let the technicalities speak. Noting as as example that over the other merge request, we end up loosing hours of discussion over philosophy as oppose to press a green button that was at least harmles, and certainly technically convenient and viable!)

I hope you understand
Best
Israel

wlbragg:
OK, I read and listened, Nowhere have I heard what the required fgdata name is to be, I don't think we get to make that decision. Until we have that answer, this is almost a waist of time except that once we do get that decision we can be ready to implement it.
Juanvvc, have you posted the question to the dev-list yet?
Personally, I also need to know what the status of the existing c172p is to become!
I know that a new branch "should" be completely unnecessary. We should use the tag option. If a branch is necessary then it is because FG and the way it is selecting aircraft repo's is broke or at best, poorly designed.

wkitty42:
FWIW: my thought about branching and then tagging sis specifically so we can go back and fix something if there's anything that needs be fixed... fix it in the branch and port the fix upstream if needed... then just leave the branch...

FWIW2: another project i work with has everything done in branches... there is no "master" really... the layout looks kinda like this...
>                              /--release2--/
>                /--release1--/
>      --start--/
>

in each release branch, we may branch to develop or fix a bug which is merged back into that release and may also be merged into the next release's dev code but once a release is made, it is tagged and left alone except for bug fix updates... this is why i suggested to branch and tag... it doesn't cost anything and it certainly isn't complicated...

IAHM-COL:
>FWIW2: another project i work with has everything done in branches...

ROFL 
So is this systemic?

IAHM-COL:
Look Wkitty
Do not take me offense, but your message, FWIW, really is not worth not even the 2 pesos.
The development structure of the c172p detailed is very well structured and definitely nicely managed.

Branches are created to implement new functionality and when the new functionality is "ready" , the branches are merged back into master.
That is how the system should work. 

More "original" ideas on how Source Code Management Could possibly be are frequently really very bad implementations. 

I would hate to see the detailed guys loosing the great track they have at this point, of doing this just the right way.

wkitty42:
@IAHM-COL there is no one way or method to use a repository... some operate like is being done here with this craft where everything eventually ends up back in master and others operate as i depicted earlier... each release branch was a dev branch... there were branches and/or tags for release candidates... in the end, when the release was made, that's what the dev branch became and the rest was deleted... then a new dev branch was made from the release code and the process started again... another project uses a combination of these methods... some stuff gets merged all the way back to the master and other stuff is only merged back into the base of the branch...

IAHM-COL:
@wkitty42 
I have no need to teach you how it is done

Juanvvc:
@onox, I want to use the --fg-aircraft option! @IAHM-COL , the issue is not how a repository is managed in git but how FlightGear manages the --fg-root and --fg-aircraft options. 

According to my past experience, the behaviour of these options is unreliable. I had a custom dv20, da40 and c337 in a directory listed in --fg-aircraft and the default (from git) in --fg-root. Well: I don't know why, but the c337 and da40 were loaded only from --fg-root, and the dv20 where loaded from --fg-aircraft. I'd believe a directory in --fg-aircraft has priority over --fg-root, but it is not always the case. I can't understand when a directory in --fg-aircraft takes precedence over --fg-root, but there is anything not-obvious there.
 
Please, notice this is independent of the label name. Based totally in my past experience with other aircraft, I believe our development would be easier if the directory in FGROOT (the stable version) and the directory in --fg-aircraft (the development version) are named differently.

Let's not make a fight of this. I will write immediately to the development list to clarify the name of the directory of the stable version. @wlbragg, sorry, I though it was you who were going to write to the list. Meanwhile, I'll run some tests to try to understand how FlightGear manages the --fg-aircraft and --fg-root options.

In any case, I'd seriously consider a branch for the v3.6 version. In fact, we already have a branch for the v3.4 version even if we are not using this name. There are lots of new effects that won't get enough testing for v3.6 and there is a good chance that some of them (the ALS lamps, for example) won't be even compatible with v3.6.

IAHM-COL:
>@IAHM-COL , the issue is not how a repository is managed in git but how FlightGear manages the --fg-root and --fg-aircraft options. 

Why this clarification? I am certain that the issue is not how a git repository is managed. And that's why I told Wkitty that his comment about branches was totally out of place.
Here the issue is whether the c172p-detailed will change the name to c172p in the master or in a release.

IMHO if this aircraft will replace the base, then the "name" should be according to what that base aircraft is: Which currently is c17p, but I understand you are asking the devels if they are considering a rebranding ... which is 100% fine too.

IAHM-COL:
>According to my past experience, the behaviour of these options is unreliable. I had a custom dv20, da40 and c337 in a directory listed in --fg-aircraft and the stable version (from git) in --fg-root. Well: I don't know why, but the c337 and da40 were loaded only from --fg-root, and the dv20 where loaded from --fg-aircraft. I'd believe a directory in --fg-aircraft has priority over --fg-root, but it is not always the case. I can't understand when a directory in --fg-aircraft takes precedence over --fg-root, but there is anything not-obvious there.

Dear Ludomo
I believe the way to handle this is simple

Just do not have 2 identical -set.xml names in two directories of your aircraft collective path.
That way you can load all the aircraft independently, by any different mean

Example, in your "custom" c337 change Cessna337-set.xml ot Cessna337_Ludomo-set.xml
That will do it

when there are two c172p's in your path (if that is how you plan to act), the simpler solution is to rename the FGDATA one ($fg-root) version as c172p-stable-set.xml [That is a local change pertinent to you]
And allow the development version (a.k.a this repository) to continue having the proper name 

(whatever the devel-cores decide it will be)

That is my personal advise for how you are managing the aircraft in your situation.

Best
IH-cOL

IAHM-COL:
I forgot to mention

I would love --really really-
If you would share your custom 337, dv20 and da40

I will make them widely available over FGMEMBERS too

But if these can't be GPLed, I would understand you keep'em private.

wlbragg:
"Meanwhile, I'll run some tests to try to understand how FlightGear manages the --fg-aircraft and --fg-root options."

Juanvvc, I will do the same as I too have had problems as of late using duplicate aircraft-set.xml names. FGRun has only been recognizing the first position priority even though you select the second position. It never used to do that, I would love to take the time to run regression on FGRun and see where it got broke.
However Israel is absolutely correct, we should be able to  keep our dev repo just the way it is, change the set file name and then use tags for releases. It has been working so well the way we have been using it and the tag method would be so clean and simple to use. I think that should be the preferred and most simplistic method available.
Please if it's not to late, see if you can find out what the plan is for the original c172p. where it will exist , if at all?

wlbragg:
I just read the response to naming this work. Looks like it will be c172p

I also looked over the files in Israel's pull request and everything I saw looked correct. But with out actually merging, well you know how that goes. So if agreed, because this no longer can be merged conflict free, why don't I pull it locally, resolve conflict, and push back as a new branch for more eyes on it. Then in can be merged next in line prior to any more conflicting commits?

onox:
PR continues in #234

PR #222, opened 2015-05-30
Do not apply brake if corresponding gear is damaged

### LINKED ISSUE #149: Remove braking from any damaged gear
Do not allow the use of brakes on any gear that is damaged.

onox:
Fixes #149

I think the proper fix though is to make sure JSBSim doesn't listen to the `/controls/gear` properties instead of (not) manipulating them.

For example, if you apply parking brake and then break all your gear by crashing, then currently JSBSim will apply the parking brake because the `/controls/gear/brake-parking` was already true before the gears got broken. But this issue only happens with the parking brake, not the normal brakes, because those return to 0 eventually.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -1,4 +1,35 @@
 ##########################################
+# Brakes
+##########################################
+
+controls.applyBrakes = func (v, which = 0) {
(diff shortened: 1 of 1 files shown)


wlbragg:
Yes, works for me if I understand the code correctly.

PR #224, opened 2015-05-30
Restore history

IAHM-COL:
This merge attempt at bringing the c172p commit history over the -detailed repository again. 
After ONOX closed the request, I separated the name change and the history addition in separate merge requests, and thus this needed to be generated.

Please consider. this has zero file changes. It has neglectible effect on repository size. But it has several technical advantages. 

Best,
Israel Hernandez

(PS: I will be sending new requests over the additional branches to incorporate history)


wlbragg:
I see no negative effects of this request, only positive, all it will do is allow c172p history to survive in this repo for as long as it exists and benefit future fgmembers configurations. There never was any "real" negative feedback from our group over this, maybe at best not seeing the necessity. But Israel assures me it is extremely beneficial to the technical side of his maintenance of fgmembers of which I have to deffer to his expertise in this area. Therefore I am merging this request and hope there is no negative side effects (I'm talking about social side effects) as it does no harm to the project and is value added applied to other issues. Please accept this on face value.

onox:
The new commits should have been rebased on top of the old ones, not the other way around... :disappointed:

wkitty42:
i can't find the original issue where this was but can we get this fixed? the history is all out of order now and it has me messed up... my OCD is really biting me in the arse about the ordering :disappointed:

IAHM-COL:
Hi Wkitty
I will be happy to help you finding what it is.
But what specifically are you looking for again? "What is the original issue" is what I wonder

The history is listed here
https://github.com/Juanvvc/c172p-detailed/commits/master
If that is what you need.

PR #225, opened 2015-05-30
Dual control --historical commits

IAHM-COL:
Restore History over dual-control branch


onox:
This is a development branch, and I've just rebased the 2 commits. No need to add old commits to the repository more than once.

PR #228, opened 2015-05-31
Complex startup start with engine running

wlbragg:
OK, did it as a pull to give ample opportunity to test.
I think this is the behavior we are looking for.

One thing I noticed that we could fix is, when you choose NOT Complex Startup, (uncheck Complex startup) then the act of pressing the s key sends an autostart() cmd right before "v = getprop("/controls/switches/starter")" in setlistener("/controls/switches/starter", func (line 117 engine.nas), which in turn activates "gui.popupTip("Engine already running.", 5);" because it starts running before you let up the s key (kind of residual). All we need to do is add a bool parameter to pass to function autostart (true or false), when we use autostart from the above function pass false to autostart, and use if true print gui.popupTip else skip it. Then whenever we use the default autostart (where we originally used it anywhere in the code) we pass true and we will get the popup like we always have in the past where it was appropriate.
I will do fix this as soon as I learn how to pass parms to functions in nasal, I wanted to get this out to test quickly so we can get it included in master before we freeze.


wlbragg:
Fixed the above and more. There were some ill effects if user tried dumb stuff not accounted for, like starter when engine already running, things like that. I am pleased with the robustness now. I'm not saying you can't find something that needs attention, but for the most part I think it does what we wanted.
One effect that i think is different, I have no idea how you all will like it or not, but that is the cranking on "Start up with engine running". It starts out cranking the engine until it starts. Kind of a cool effect.
If you Start up with engine running" and also have the "Complex startup" checked, then if you kill the engine it requires the primer and whatever else isn't set correctly to be accounted for. I think it is behaving well.

onox:
Does it make sense to allow enabling "Start with engine running" if "Complex start up procedures" is enabled?

onox:
The full autostart (pressing 's' instead of clicking 'Autostart' + pressing 's') seems to work :+1:

wlbragg:
"Does it make sense to allow enabling "Start with engine running" if "Complex start up procedures" is enabled?"
I asked the same questions a while back and I think it was gsagostinho that explained it and it made sense at the time. So I programmed it to match my recollection of that explanation.
You all can change it however you want, but I think it is working the way it was explained to me.

The two conditions really don't have to be tied I guess. You can choose to have the aircraft start up automatically when you load FG and yet still have to make sure everything is set correctly from that point forward. That is basically how it is now.

wlbragg:
So are we ready to merge this?

onox:
It can't be automatically merged. I'll merge it manually.

PR #231, opened 2015-05-31
Added missing var-keywords in the Nasal files.

andgi:
While leaving out the var keyword when introducing new identifiers is not an error it is not a good practice since the behaviour can change from creating a new local variable to overwriting visible one without any notice (e.g. if a later added global module happens to have the same name as the new variable). Using var always creates a new local (= present module or function) variable and will never overwrite a larger scoped visible one, should one exist.


Juanvvc:
I completely agree. Thank you very much!

PR #232, opened 2015-05-31
Hid FDM internal weights from the fuel and payload dialog.

andgi:
Don't add /payload/weight entries for weights that the user should not see or touch in the fuel and payload dialog.

Signed-off-by: Anders Gidenstam <anders@gidenstam.org>


onox:
Thanks! :+1:

dany93:
Done. https://github.com/Juanvvc/c172p-detailed/pull/233#issue-83424181
Thank you, Anders G, and sorry for not having found the property change by myself.

andgi:
No problem.

PR #234, opened 2015-06-01
Name change c172p

wlbragg:
I don't know that if this was the correct way, or even the best way to do this But it's here and ready to go. I even was able to rename my local aircraft folder from Custom/c172p-detailed to Custom/c172p and leave the original c172p in fgdataSVN/Aircrafts/c172p, that is what I call my fgdata. FGRun showed all the versions (old and new) in the single folder c172p on it's list of aircraft and let me select the "detailed" from that list. The only negative was a engine.nas nasal error which I am sure will go away once the original c172 is gone.

Nasal runtime error: undefined symbol: autostart
  at F:/FlightGear/Custom Aircrafts/c172p/Nasal/engine.nas, line 121
  called from: F:/FlightGear/fgdataSVN/Nasal/globals.nas, line 119

even though everything still worked!


onox:
I'll fix that Nasal error.

onox:
Didn't spot anything broken, so :shipit:

wlbragg:
Just in case I was dreaming about the name change on dev-list, I'm waiting for Juanvvc's blessing.

Juanvvc:
I'm ok with the name change. Whatever conflict we may have between the original/fgroot/fgaircraft directories, it is going to appear during the everyday use. I say: let's accept the merge and fix the conflicts when they appear.

wlbragg:
OK to both of you,

onox, I didn't notice the position change, I'm sure that will fix it.

I'll merge this here in a second.

PR #244, opened 2015-06-04
Fix #107

### LINKED ISSUE #107: Livery registration bug in default livery
blinking
User supplied immarticulation is incorrect and some letters/numbers blink.

Juanvvc:
Livery registration bug in default livery. Issue #107 

The problem was in multiplay params: the immat system used /params/multiplay/generic/int[0] to int[5], and these multiplay properties are also used by the crash and the lighting systems. For example, the flickering period was the same than the beacon, and if you pressed "apply" in the immat dialog wings collapsed.

I tried adding an offset to the params in c172p.xml, but it didn't work. I suspect the comment "These properties are aliased to MP properties in /sim/multiplay/generic/." in c172p.nas has something to do. Then, I applied an offset to the params in immat.xml.


onox:
Could you change the French comments in `Models/Immat/immat.xml` to English?

Juanvvc:
Of course. Also, I will add a dark background to the immat label inside the cabin.

wlbragg:
Ya, thank you!

Juanvvc:
Changes:

- Comments in English
- Comment about multiplay properties 14 to 19 in c172p.xml
- New dark background for the label on the panel

Changes open to discussion:

- New property /sim/c172p/immat-on-panel (default to true) to show or hide the immat label on the left side panel independently of the labels on the fuselage. Liveries can change this property.
- The label on the left side of the panel is moved slightly to the left.
- Removed the callsign label on the right side of the panel. I believe this label was redundant.

Juanvvc:
Oh, wait a moment. I configured the text editor with the wrong charset!

onox:
Is `<text>` in `Models/c172p.xml` supposed to be disabled?

onox:
Also, immatriculation is a French word, so better think of an English property instead of keeping calling this "immat" in my opinion.

Juanvvc:
In my opinion, yes, it should be disabled. It was the old callsign label.

The old label (it was a text, the label object was only active in the default livery)

fgfs-screen-005

The new label, optionally active in all liveries (currently: active in all liveries:

fgfs-screen-006

Regarding the name of the property, what about "callsign"? If I recall correctly, it was discussed somewhere but I can't remember which word was the best.

onox:
Registration number?

Personally I don't like having the callsign / registration number visible on the panel. It will also ruin screenshots IMO :stuck_out_tongue_winking_eye:

onox:
"callsign" is good. It's easier to understand than "immat".

onox:
Also please remove the commented/disabled old callsign in `Models/c172p.xml`, otherwise we get tons of disabled code everywhere :laughing:

Juanvvc:
It is an option of the liveries, and it could be switchable in the options dialog. In any case, having the registration number on the panel is completely common! Keep in mind these aircraft are flighted by many people that often fly other aircraft. Forgetting the registration of the aircraft you are currently in is normal, and these labels are very common. They are in all cockpits I've seen, both in real world and google images.

Any other opinion about these issues? Position/visibility of the checklist label and its name.

onox:
> Keep in mind these aircraft are used by many people that often fly other aircraft. Forgetting the registration number of the aircraft you are currently in is normal, and these labels are very common. They are in all cockpits I've seen, both in real world and google images.

Ah! Well better keep it then :smile: Could you add a checkbutton in the options dialog and remove the disabled text?

wkitty42:
FWIW: "immatriculation" seems to translate to "registration"... calling it registration or callsign both work ok in this case, i think...

on this aspect, the default plain white craft... can its default immatriculation be "DE-FAULT" or "DE-FALT" or "DE-FULT" to complete the effect when one loads that livery and has no callsign of their own defined?? good joke and humour for everyone when discovered :smile_cat:

onox:
How about N-OOB? :laughing:

wkitty42:
hahahahaha! is that N-00B (with zeros)? does the industry use 'O' (oh) and '0' (zero) or just one over the other?

that one is also funny, though :smile_cat:

onox:
If numbers are required, then we can choose N-008 :wink:

wkitty42:
HA! looks better with at least the B in there :lol:

PR #245, opened 2015-06-04
Gear damage by the forces exerted on gears

dany93:
Connected to this issue https://github.com/Juanvvc/c172p-detailed/issues/134#issuecomment-107867029

Indeed, the nose gear is hard to break without breaking the others. 
For this nose gear, I had to set the breaking force at 1400 lbs because it reached 1392 lbs at the end of a strong braking with the 36" bushwheels (1350 lbs threshold is needed with the default wheels).

I have included the" tool" for printing the exerted forces because it can be useful elsewhere and by someone else (under comments, of course).

@tigert if you can see for the breaking threshold due to side force (15 kts crosswind landing, full load).
I have set it at 1500 lbs (Y-force).
500 lbs can be enough for a very good crosswind landing, but that's easy to reach 600 - 700 lbs only by rolling not very straight. 700 - 1000 lbs seems possible, but demanding.

### DIFF
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -20,6 +20,18 @@ var aero_coeff = "fdm/jsbsim/aero/coefficient/";
 #Roll moment due to (diedra + roll rate + yaw rate + ailerons) ==> asymmetry to break one wing
 var roll_moment = getprop(aero_coeff~"Clb")+getprop(aero_coeff~"Clp")+getprop(aero_coeff~"Clr")+getprop(aero_coeff~"ClDa");
 
+# Force-on-gear = (compression-ft) x (spring_coeff) + (compression-velocity-fps) x (damping_coeff)
+var compr0 = getprop("/fdm/jsbsim/gear/unit[0]/compression-ft");
+var compr_vel0 = getprop("/fdm/jsbsim/gear/unit[0]/compression-velocity-fps");
+var force0 = 1800 * compr0 + 600 * compr_vel0; # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for NOSE
+var compr1 = getprop("/fdm/jsbsim/gear/unit[1]/compression-ft");
+var compr_vel1 = getprop("/fdm/jsbsim/gear/unit[1]/compression-velocity-fps");
+var force1 = 5400 * compr1 + 400 * compr_vel1; # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for LEFT gear
+var compr2 = getprop("/fdm/jsbsim/gear/unit[2]/compression-ft");
+var compr_vel2 = getprop("/fdm/jsbsim/gear/unit[2]/compression-velocity-fps");
+var force2 = 5400 * compr2 + 400 * compr_vel2; # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for RIGHT gear
+var gear_side_force = getprop("/fdm/jsbsim/forces/fby-gear-lbs");
+
 var gears = "fdm/jsbsim/gear/";
 var contact = "fdm/jsbsim/contact/";
 var lastkit=0;
@@ -190,7 +202,6 @@ var upsidedown = func
 
 var killengine = func
 {
-	#setprop("/controls/engines/engine/magnetos", 0);
 	setprop("/fdm/jsbsim/propulsion/tank[2]/priority", 0);
 }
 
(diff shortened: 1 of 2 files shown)


onox:
I think the nose gear is a bit too easy too brake.

Nasal/physics.nas
```diff
@@ -20,6 +20,18 @@ var aero_coeff = "fdm/jsbsim/aero/coefficient/";
 #Roll moment due to (diedra + roll rate + yaw rate + ailerons) ==> asymmetry to break one wing
 var roll_moment = getprop(aero_coeff~"Clb")+getprop(aero_coeff~"Clp")+getprop(aero_coeff~"Clr")+getprop(aero_coeff~"ClDa");
 
+# Force-on-gear = (compression-ft) x (spring_coeff) + (compression-velocity-fps) x (damping_coeff)
```
onox: Can you define a function for this so you can just call the function and give it the index, and the two numbers.

onox: You should be able to use the index like this: `getprop("/fdm/jsbsim/gear/unit", index, "compression-ft")`

onox:
:+1: for using the side force.

dany93:
>~~I think the nose gear is a bit too easy too brake.~~

You gave me a fright. I firstly read it, I later discovered your crossing out.
As I wrote it, I find it hard to break. I was a afraid that it would be a matter of nasal slowness on my computer (fps). And I began considering having to do it completely by xml.

Finally that is partly the way you are suggesting.
I see no reason that such functions cannot be defined in the FDM. That's elegant in the nasal file but in practice I do not know how to introduce the two numbers as a "field" (kind of vector) in this FDM function.
I see how to define 4 functions in the FDM:
- one for the roll moment,
- three for the three gears, with the compression and damping as numbers (coefficients) for each,

but I do not see how to define a "field" (kind of vector) for the compression and damping values with a single function.
Even in nasal that's not clear for me. I am rather basic in nasal.

Apart from the elegance, is there a practical advantage to do like this? I'd see one if it was done completely in xml.
But at this day we have not too much time....

onox:
What I meant is, you replace this:

```javascript
var compr0 = getprop("/fdm/jsbsim/gear/unit[0]/compression-ft");
var compr_vel0 = getprop("/fdm/jsbsim/gear/unit[0]/compression-velocity-fps");
var force0 = 1800 * compr0 + 600 * compr_vel0; # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for NOSE
var compr1 = getprop("/fdm/jsbsim/gear/unit[1]/compression-ft");
var compr_vel1 = getprop("/fdm/jsbsim/gear/unit[1]/compression-velocity-fps");
var force1 = 5400 * compr1 + 400 * compr_vel1; # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for LEFT gear
var compr2 = getprop("/fdm/jsbsim/gear/unit[2]/compression-ft");
var compr_vel2 = getprop("/fdm/jsbsim/gear/unit[2]/compression-velocity-fps");
var force2 = 5400 * compr2 + 400 * compr_vel2; # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for RIGHT gear
var gear_side_force = getprop("/fdm/jsbsim/forces/fby-gear-lbs");
```

with:

```javascript
var get_gear_force = func (index, spring_coeff, damping_coeff) {
    # Force-on-gear = (compression-ft) x (spring_coeff) + (compression-velocity-fps) x (damping_coeff)

    var compr = getprop("/fdm/jsbsim/gear/unit", index, "compression-ft");
    var compr_vel = getprop("/fdm/jsbsim/gear/unit", index, "compression-velocity-fps");
    return spring_coeff * compr + damping_coeff * compr_vel;
};

var force0 = get_gear_force(0, 1800, 600); # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for NOSE
var force1 = get_gear_force(1, 5400, 400); # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for LEFT gear
var force2 = get_gear_force(2, 5400, 400); # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for RIGHT gear
```

dany93:
> Are you doing the same kind of computation in c172p.xml?

No. This function is not calculated in c172p.xml. Only in physics.nas.

Changes done, it works, pushed.
Thanks, I would not have been able to do it by myself. I hope being able to understand it enough for later, if I have to change something....

onox:
Thanks, but now you define the function two times. On line 23 and on line 288. Can you remove the redefinition on line 288?

dany93:
I believed the second one (line 288) was necessary to update the variables inside the loop. Like any variable. Didn't realize that the ```func()``` called by ```get_gear_force``` does the trick.
Yes, this form is very concise!
Thank you for the advices.

Issue #246, opened 2015-06-04
Instruments and gauges have no shadow or flashlight effects applied to them


wlbragg:
Instruments and gauges have to include an effects inheritance code block in each of their xml files to inherit these effects. 
see http://forum.flightgear.org/viewtopic.php?f=4&t=25157&p=245632#p245632

wlbragg:
thx onox, I always forget the tags.

I didn't notice anything to outstanding. I'm not sure it works exactly the way you would want . t almost appears that the shadow may not align the same from instrument to gauge to panel.
But for the flashlight you can see it works really well
![flashlighgr](https://cloud.githubusercontent.com/assets/11069680/7992598/141abd2a-0ac6-11e5-87c5-231e2058b31f.jpg)

wlbragg:
This is finished and having heard of no ill effects, I'm closing it.

PR #247, opened 2015-06-04
Interiorshadoweff to gauges

wlbragg:
I didn't notice anything to outstanding. I'm not sure it works exactly the way you would want . t almost appears that the shadow may not align the same from instrument to gauge to panel.
But for the flashlight you can see it works really well
flashlighgr
.


wkitty42:
from what i see in the picture, it looks good... just need to be able to direct it or have it a little larger so we can see the necessary switches and levers... nice job, too!

Models/c172p.xml
```diff
@@ -258,7 +258,7 @@
     <object-name>CarbHeat</object-name>
     <object-name>CarbHeatMount</object-name>
 	<object-name>GenSwitch</object-name>
-	<!--object-name>LandingLightSwitch</object-name-->
+	<object-name>LandingLightSwitch</object-name>
```
onox: What's the difference?

onox:
Does this fix #246?

onox:
And how do I test the flash light?

wlbragg:
You don't test the flashlight until it is implemented in fgdata. But we need to make sure there are no negative side effects from the interior shadow effect on the gauges.

No differences in Landing light switch, i uncommented it so it takes the shadow effect.

wlbragg:
Yes it fixes issue #246

Issue #240, opened 2015-06-03
New Rating


dg-505:
Good morning everyone 

I suggest to give the c172p a new rating, because much has changed in the last few month.
Here is my suggestion:
- FDM: 4 or 5 stars. I don't exactly know how close to the reality the FDM is, but it seems very reasonable to me. Maybe someone with real flight experience with the c172p can ssy if it's more 4 or more 5 stars.
- Systems: also 4 or 5. There are some failure modes present (e.g. over-G damage), but as far as I know, there are no emergency checklists available.
- Cockpit: 5 Stars. We have a very accurate 3D-cockpit where most of the controls are animated and have hotspots, and the whole interior has photo-realistic texturing.
- External model: 4 Stars. The exterior model is already very detailed, but in my opinion it's not yet a 5-star, because I think 5 stars should be reserved to models which are detailed like the new P-51 Mustang: http://wiki.flightgear.org/North_American_P-51_Mustang.

As an overall status I suggest "production" or "advanced production" depending on how much stars are  achieved after discussing this issue.

For a reference see http://wiki.flightgear.org/Aircraft_rating_system

onox:
1. FDM: Unless the FDM is based on data from actual testing (wind tunnel, etc.), I would say it's (a solid) 4 stars.

2. Systems: I'm not sure what kind of systems we still need to model. Maybe the engine temperature primer stuff, oil, engine damage, etc.? I think 4 stars.

I agree with what you say about the cockpit and the external model.

tigert:
For systems we should test that with each breaker pulled out, nothing that takes power from it works. Also I think not all breakers might be functional yet?

I cannot say anything is wrong with the fdm with my limited c172 experience. I am pretty convinced about the stall/spin too, it recovers with textbook procedures and the aircraft behaves also very realistically when loaded to mtow (try that on hot day in mountains and you end up with an interesting day..)

I will test further, but definitely a 4 for the fdm, we would need to verify performance with flight manual figures to give it a 5 I guess, but it could well be there. 

All in all the cockpit feels very nice. I agree with 5.

dg-505:
@tigert: About the breakers I can say that most of them do their job properly. Only INST (for instrument lights), INT LT (for cabin lights), Turn Coordinator and Radio breaker 1 don't seem to work. (Maybe we could open another issue for that.)
@onox: About the systems maybe we could implement engine damage when overheating or when exeeding RPM limit, or something like this.
[...]

Issue #30, opened 2015-03-12
Check all interior objects are NOT selected if the viewer is further than 50m


gilbertohasnofb:
XML07 - Check all interior objects are NOT selected if the viewer is further than 50m (useful in multiplayer)

onox:
Note that LOD range animations are expensive. See http://wiki.flightgear.org/Modeling_-_Getting_Started#Level_of_Detail_.28LOD.29

PR #257, opened 2015-06-07
Updated c172p-set.xml (authors, status, rating)

dg-505:
Status: Advanced production

Rating:
- FDM :star: :star: :star: :star:
- Systems :star: :star: :star: :star: :star:
- Cockpit :star: :star: :star: :star: :star:
- Model :star: :star: :star: :star:

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -51,13 +51,13 @@ Many files from the original c172p directory are included in this file.
       <tag>fixed-gear</tag>
   </tags>
   
-  <author>David Megginson</author>
-  <status>early production</status>
+  <author>Original by David Megginson, gsagostinho (Gilberto Agostinho), wlbragg (Wayne Bragg), Juanvvc (Juan Vera del Campo), onox, thevirtualfer, Dany93 (Daniel Dubreuil), dg-505 (Jonathan Schellhase), IHAM-COL (Israel Hernandez), tigert (Tuomas Kuosmanen), AndersG (Anders Gidenstam), wkitty42 (Waldo Kitty), algefaen</author>
+  <status>advanced production</status>
   <rating>
-    <FDM type="int">3</FDM>
-    <systems type="int">4</systems>
-    <model type="int">5</model>
-    <cockpit type="int">4</cockpit>
+    <FDM type="int">4</FDM>
+    <systems type="int">5</systems>
+    <model type="int">4</model>
+    <cockpit type="int">5</cockpit>
   </rating>
 
   <!--variant-of>c172p</variant-of-->


gilbertohasnofb:
I might have missed something but how came our model has been downgraded from 5 stars to 4 stars?!

dg-505:
See issue #240

gilbertohasnofb:
Thanks, I didn't follow that issue very closely and I totally overlooked that the model was going to be downgraded. Well, if that's what the majority thinks, then fine with me, but by the standards of the other aircraft with 5 stars model I'd say ours should be up there with them.

dg-505:
Yes I know that other models have also 5 stars, but the wiki says "As a guideline, a "5" model is as realistic as possible given the available rendering technology." So to meet  this requirement, I thought 4 stars would match better.

gilbertohasnofb:
Sure, as I wrote I've seen you guys discussed it (sorry for missing it before), so of course I accept it. But if I may ask you, what improvements do we need to the 3D model to get a 5 stars? I have been discussing issues https://github.com/Juanvvc/c172p-detailed/issues/216 and https://github.com/Juanvvc/c172p-detailed/issues/6 with ludomotico and those will surely make the 3D model look better.

dg-505:
IMHO http://wiki.flightgear.org/North_American_P-51_Mustang. *That* model deserves 5 stars.

gilbertohasnofb:
Sure, that is an outstanding model. But I don't think the 5 stars is deserved by a single airplane only, and as I wrote before there are similar planes to ours with 5 stars models. But my question was meant to be  more practical actually: do you think we lack some specific 3D details? Do you think our model is not high-poly enough? I am just thinking in terms of improvements, that's why I ask.

dg-505:
1 example: 
unbenannt

gilbertohasnofb:
@dg-505 Jonathan, I really hope you are not getting me wrong. I understand your opinion and I understand this has been discussed and I am not trying to change the rating of the aircraft (hence why I already merged this PR). I am simply trying to brainstorm some ideas on how to improve the 3D models, particularly because you know quite a lot about 3D modelling and I know next to nothing! 

About the example you posted above, that's a good point and should be simple to fix and it won't make the model any heavier (it will be just a couple more of polygons to make the hole I guess). So I opened a new issue for that: https://github.com/Juanvvc/c172p-detailed/issues/259

If you do think of some other things like this example above, please consider opening issues for them so we can continue to improve this plane.

Cheers!

dg-505:
I'm not getting you wrong at all. I'm phoning, so I'm a bit curtly atm.

gilbertohasnofb:
No problems :smile: 

(I added your second picture to the same issue)

dg-505:
Roger :+1: :smile:

onox:
How many vertices does the model has? About 17000? That's very low poly.

dg-505:
The c172p.ac has 16,945 vertices and 16,406 surfaces.

gilbertohasnofb:
Of course we must keep in mind that since this aircraft will be the default FG plane from now on, we should keep it as low poly as possible. IMO the exterior is looking good (still has room for improvements without increasing too much the poly count) and the plane run well on modest computers.

Juanvvc:
I also prefer having a low polys exterior model, at least during the v3.6 version, and we can check how it goes. We can decide adding extra details after the freeze.

In any case, iirc, the fuselage has probably less vertices than the old model since I removed many repeated vertices. This is compensated by the cockpit, whit many additional objects and vertices. This is why issue #30 probably deserves a visit before the freeze: we have added many models that are probably still visible also in multiplayer.

gilbertohasnofb:
:+1:

Issue #134, opened 2015-04-30
Break wing inflight (+ gears on ground) due to excessive forces


wlbragg:
This is one I really want to do. We have the modeled effect and the consequent FDM effect in place. I need the force calculations from the FDM?

wlbragg:
I just noticed #123 is this the same or tied to this new issue I created?

dany93:
Wing(s) can break because an excess of speed or by over-g's (excessive load factor).
Both due to an excessive force in reality. But in FG not the same properties.
/accelerations/n-z-cg-fpf_sec
/fdm/jsbsim/accelerations/Nz, pilot-gdamped
I do not know if you need all of them.
I should say that it is enough.

onox:
You could take a look at how the [f-14a](https://github.com/Zaretto/fg-aircraft/commit/1dfba02ca6d997607590df3eaff65088a0d9d629) implements this.

wlbragg:
I'll do that, thanks.

wlbragg:
Dany added some new FDM effects tied to these forces. We are currently testing them using a non-branched source. So far so good, they may need some fine tuning. It's a real sinking feeling to have a wing just quit working.

wlbragg:
Wow, Dany, does our existing FDM Juanvvc:master have the same spin effects when you loose the flight envelope as you have in dany93? The one wing "damage", INCREDIBLE. I tried to limp along back to the airport and screwed around till I got into a spin at low altitude, oops! This is really nice.
Again my question, does our existing Juanvvc:master FDM  have the same spin effects when you loose the flight envelope as you have in dany93 FDM?

I'm ready to merge Dany's FDM into master along with this new FDM damage model. It is incredible gang. Any comments anyone?

dany93:
>does our existing Juanvvc:master FDM have the same spin effects when you loose the flight envelope as you have in dany93 FDM?

Yes. it is the same for the "spin and stall" effects. I have not changed the general FDM other than for damages on left or /and right wings.
Easy to test: fly with the FDM from juanvvc (no damage effects), cut the throttle, hold the stick back, up to stalls. Worsen it with more or less rudder (see the slip ball).
I wonder if you do not mix two effects:
- one which have existed for a long time, the spin and stall effects, with can give a spin roll on stall, depending on your flight asymmetry (that is the original feature of this FDM, from the beginning you had it). That is when you are close to the edge or out of the flight envelope (high angles of attack),
[...]

PR #262, opened 2015-06-08
Broken wings animation

wlbragg:
OK, here are the new broken wings animations. It includes a new function in physics.nas for both wings broken. I think there is room for improvement overall for the physics of wing damage and breaking.

They need to be tested.

dany, maybe you can do some magic with the FDM and the effects trigger being a little more realistic.


dany93:
I have tested your animations for wings damaged or broken, single or both.
They work very well. :clap:

However, I do not see well what you are meaning by "the FDM and the effects trigger being a little more realistic":

1 - The triggering conditions are (https://github.com/Juanvvc/c172p-detailed/issues/134#issuecomment-100706694).
V > 158 kt and flight asymmetry ==> one wing damaged (can still fly, degraded)
V > 180 kt and flight asymmetry ==> one wing broken
V > 180 kt and symmetric flight ==> both wings damaged (degraded lift)
V > 200 kt ==> both wings broken

g > 4.4 and flight asymmetry ==> one wing damaged (can still fly, degraded)
g > 5.5 and flight asymmetry ==> one wing broken
g > 6.6 both wings broken

This can be discussed, but will other conditions be more realistic?

2 - In the FDM, I cannot affirm that the damaged or, worst, broken aircraft behavior is closely like reality. For obvious reasons, I ignore what the reality is. What do you feel is too much not realistic and, above all, in what sense can it be improved?
 I do not think we have to (or that we can) go too far towards reality in this domain. The presence of this effect is already a good reminder of what is not to be done, and of what it might give. Somewhat like the stall and spin feature. Unless willing to add an excess of intricacy and heaviness (which, with accumulation, has a price in fps for marginal items and effects).

*Also, I'm not a magician, I must admit...*

wlbragg:
For example, just now I was testing the addition of engine rpm to water effect on windshield and I came down on the water a little hard, was it so hard that it should have broken the wings off completely or just collapsed them or damaged them or nothing at all?

So I guess my areas of concern would be
- how hard of a landing either breaks, collapses, does a little damage or no damage?

For the rest of it (forces in flight),
I really can't say I have done enough testing to have a real strong opinion either way. If you've been able to get damage and not break then I guess it is good for now. Maybe with more eyes on it after release we will get more feedback and adjust accordingly.

wlbragg:
I'm going to merge this now, we can adjust if needed later.

onox:
@wlbragg Does this fix #134?

wlbragg:
Yes

wkitty42:
i had a chance to (inadvertently) test this earlier.... along with @dany93 spin stuff... craft got too slow and the AP was trying to keep it nose up... so i turned off the AP to take control and level out to gain speed but all of a sudden, she was spinning and diving and that was all she wrote... so i went to heli view and saw the new damage on a wing tip where the front edge of the wing tip was angled up and back slightly... eventually it bent a more and then we made a slight crater in the ground... it isn't the first time i've run into an AP controlled stall and turned the AP off only to have the craft fall out from under me... i gotta get a (decent) joystick and maybe some pedals... or build something myself... hahaha...

wlbragg:
"she was spinning and diving and that was all she wrote"
Isn't that the coolest thing, (not so in RL). I am testing andgi's FDM changes and keep starting with engines off at altitude over the lake. I immediately have to turn to get the long lake run in front of me and if I turn to hard and don't drop the nose, bye-bye, I can make it happen every time if I want to. I also can start at a higher altitude so I have time to try to get out of the spin. :)

onox:
@wlbragg I have noticed a little problem:

fgfs-screen-019

First the outer parts of the wing broke off, but then the whole wing disappeared.

onox:
And both wings brake off very easy now :cry: I don't get it to be just damaged or collapsed.

wlbragg:
"And both wings brake off very easily now "
Well that is what I thought too, but I don't think anything really changed in physics.nas to account for it being easier to break.

Ouch, OK we need the exact order of what happened (if you even know) to track down what combination is missing. We're looking for things like,
damaged left, broke right.
or 
damaged left and right then broke left.
It is probably going to be a combination I didn't account for.

wlbragg:
I think I already found it. It is when you crash hard and both crash and both-wings-broken get triggered, I didn't account for it. I'll get it fixed, I think that is all it is.

dany93:
I don't know if this is the reason, but do not forget that, *from in-flight damages*, we have
g > 6.6 both wings broken
Which can happen on violent a crash on ground. It can be deactivated near the ground by a condition on height if this is a drawback, but is it unrealistic?
To test and analyze if this is the cause, just comment the lines in physics.nas
```
	if (g > (max_positive * 1.5))
	{
		rightwingbroke();
		leftwingbroke();
        gui.popupTip("Over-g Both wings BROKEN!!", 5);
```
	}

wlbragg:
That was part of the fix, I already had changed it to call bothwingsbroken. 
I looked at it again and am making one more minor adjustment. I basically give preference to bothwingcollapse (or crash) because it is triggered by all gear collapse (meaning you hit the ground hard) but only if all of the wings are not broken (wing-damage/* < 1). If any wing is broken we can't collapse (I suppose some day we could add collapse one, break or damage the other).

dany93:
It seems a good idea, better than the height.
priority bothwingcollapse (or crash) in case the aircraft hits the ground,
bothwingsbroken in flight.

onox:
I think that the wings should not be torn off from the aircraft if you do a hard landing. Only if you try to do sort of a kamikaze or if you end up upside down with a decent/significant ground speed.

wlbragg:
That should be how it is now, if you can hit the ground hard enough before you break either wing from forces, then it will collapse the wings.

onox:
I just tried to pull the yoke back hard and the wings broke away in air:

fgfs-screen-026

IMO that should not happend. The wing should be damaged (bended a little), but not completely torn off :confused:

PR #266, opened 2015-06-09
Bug 237

### LINKED ISSUE #237: Damage and instruments are not recorded in the flight recorder
Properties that reflect the damage should be recorded in the flight recorder. Instruments should be recorded as well.

Fixing it in branch [bug-237](https://github.com/Juanvvc/c172p-detailed/tree/bug-237).

* [x] Primer
* [x] Master Battery (already working)
* [x] Master Alternator (already working)
* [x] Magneto's (already working)
* [x] Starter (one position further than the magneto's)
* [x] Throttle (already working)
* [x] Mixture (already working)
* [x] Instrument lighting
* [x] Carb Heat
* [x] Pitot Heat
* [x] Cabin Heat slider
* [x] Cabin Air slider
* [x] Navigation Lights
* [x] Beacon
* [x] Strobe
* [x] Taxi Light
* [x] Landing Lights
* [x] Compass
* [x] Fuel selector switch
* [x] Elevator trim wheel
* [x] Heading bug
* [x] Doors
* [x] Passengers/baggage weight
* [x] Circuit breakers
* [x] Selected gear
* [x] Position of gear (for amphibious variant)
* [x] Smoke and sparks

It looks like the needles of the instruments are already working. Somebody should double check this.

Not implemented:

* Some extra environment conditions used by cabin temperature popup tip (Requires some refactoring to store its state in properties instead of Nasal variables)
* Time (There's not really a timestamp in the property tree. There's a `warp` property, but it doesn't work correctly during replay)
* Radio stack switches and dials (Will be done later)

onox:
Fixes #237 

It basically records almost everything: flight controls, systems, some parts of the instruments, doors, passengers and baggage, damage mods, effects (smoke and wakes), circuit breakers, lighting, etc.

There are some things it doesn't do though. I'm creating this PR to gather some feedback.

1. The instruments seem to listen to the state of the systems and environment instead of listen to the actual state of the instrument itself at the particular moment (for example the current indicated heading). Meaning that if you do Ctrl+R to go back to the beginning, it won't show you instantly the state of the instrument at that time. In other words, you won't see the heading indicator and attitude indicator spin correctly when you start the engines. The instruments will work, however, if you let the replay run for a while.

2. If you crash and then go back to the replay, the flight recorder seem to be fighting with Nasal sometimes. The Nasal files should be modified to create a listener that listen for the replay state (`/sim/freeze/replay-state`) and pauses any timers that modifies the damage properties. (Because the flight recorder will take care of it during replay)

3. The Nasal code for the cabin temperature should use properties for its internal state instead of Nasal variables, otherwise I cannot record the current cabin temperature.

4. I cannot record the current time. There's a warp property in the property tree, but it won't work correctly.

5. Currently I'm not recording any properties of the radio stack.

I hope to fix (2) and (3). I cannot fix (1) and (4). Don't have time now for (5). In the mean time I would like to get some feedback from you guys :smile:


wlbragg:
3 and 4 could be a bit of work, I know nothing about the procedure of flight recorder. If you need any help with it just let me know.

tigert:
I think the recorder code could handle these things? Now, I have no real experience in this area, but I would guess one way would be that the flight recorder could be given a list of aircraft-specific properties that should be recorded and reset according to the timeline, instead of doing this all in nasal? The recorder would have a built in list of default things that are common to most cases, but things like damage etc that are handled via custom properties could be added by aircraft developers. Maybe poke the devel mailing list for feedback on this?

wlbragg:
The simplest fix was set this listener as the first line in the  physics_loop function.

[code]
var nasalInit = setlistener("/sim/freeze/replay-state", func{
    if(getprop("/sim/freeze/replay-state")) return;
});
[/code]

wlbragg:
I also added the following properties for this in PR, see #270.
/environment/aircraft-effects/cabinairtempC
/environment/aircraft-effects/surfacetempC
/environment/aircraft-effects/cabinairdewpointC
If you don't need
surfacetempC
cabinairdewpointC
you can delete them.

wlbragg:
Leave this PR until onox see's and adds cabinairtempC?

wlbragg:
One other thing I noticed, it seams that the FR does not pick up the "Aircraft Name"?

onox:
> flight recorder could be given a list of aircraft-specific properties that should be recorded and reset according to the timeline, instead of doing this all in nasal?

That's what the FR is already doing, it's recording properties. That's why some of the Nasal should be paused/suspended during replay :smile:

onox:
@wlbragg What do you mean with "Aircraft Name"? The registration number?

wlbragg:
Let me do a replay real quick and see if it still does it.

wlbragg:
OK, it was Aircraft Version in the "Load Tape" that is "undefined". Maybe it doesn't matter.

onox:
@wlbragg I have added the three properties, so you can test them.

onox:
@wlbragg There's indeed an `<aircraft-version>` tag missing.

onox:
It seems (2) and (3) have been fixed, so this PR can be merged :smile:

wlbragg:
Looks good!

PR #270, opened 2015-06-09
multiple-issues-and-bug-fix

wlbragg:
onox, I think this is what you wanted. I added the following properties,

/environment/aircraft-effects/cabinairtempC
/environment/aircraft-effects/surfacetempC
/environment/aircraft-effects/cabinairdewpointC

If you only need cabinairtempC, you can delete the rest.

When I add  these new pull request for small changes like this I am using my main repo (master) which may include other pull requests that haven't been merged yet. I just now realized I should probably be using  Juanvvc/master, make a branch, and new stuff, push new branch, switch back to Juanvvc/master and do again with different new stuff. That way I could be pushing small single issue changes instead of building on like I am. Then I could also have another local branch that has all my combined changes in them.
Oh well, I'm still figuring this stuff out. I've learned a ton of stuff in many areas because of this project. It has been a wonderful exercise.


wlbragg:
Also in this PR is bypass damage_loop if recorder is true, see Bug 237 #266

Nasal/physics.nas
```diff
@@ -562,6 +571,9 @@ var poll_gear_delay = func
 
 var physics_loop = func
 {
+	var nasalInit = setlistener("/sim/freeze/replay-state", func{
```
onox: This doesn't look right. You're creating a listener during every iteration of the loop. What I meant is something like this:

```javascript
setlistener("/sim/freeze/replay-state", func (n) {
    if (n.getBoolValue()) {
        physics_loop_timer.stop();
    }
    else {
        physics_loop_timer.start();
    }
});
```

wlbragg:
Your right, that was dumb. 
But instead of using a listener at all, how about
if(getprop("/sim/freeze/replay-state")) return;

Because do we want to stop the entire physics loop or just the check_damage?

wlbragg:
OK, we ready to merge this?

onox:
Let me check for a minute.

onox:
@wlbragg Could you fix `Nasal/weather.nas`?

onox:
See the comments https://github.com/Juanvvc/c172p-detailed/pull/270#commitcomment-11595914

onox:
51d612f is ok :+1:

wlbragg:
I missed your comment above, "Nasal/weather.nas See #270 (comment)."

"So what you need to do is add at the top of the loop:
var cabinairtempC = getprop("/environment/aircraft-effects/cabinairtempC");
var surfacetempC = getprop("/environment/aircraft-effects/surfacetempC");
var cabinairdewpointC = getprop("/environment/aircraft-effects/cabinairdewpointC");"

Problem with this is then we never start the aircraft with outside temp and dewpoint. Say it was the first time you enter the aircraft in the morning and it is -10C, currently I load cabinairtempC with environment/temperature-degc and we have to defrost the windows before we can fly.

"At the bottom of the loop I think it's only necessary to do the setprop()'s if /sim/freeze/replay-state is false."
That is where and why I store nasal to props for recorder to use at the bottom.
setprop("/environment/aircraft-effects/cabinairtempC", cabinairtempC);
setprop("/environment/aircraft-effects/surfacetempC", surfacetempC);
setprop("/environment/aircraft-effects/cabinairdewpointC", cabinairdewpointC);
If recorder kicks in then these are set.
Am I missing something, or not understanding?

wlbragg:
OK, check it out and merge if ready.

onox:
It's not useful to call `setprop()` to store the variables if you don't read them at the beginning of the loop function and use them in your calculation.

I think your problem can be solved if you make sure you read `/environment/temperature-degc` and store it in the property?

onox:
Oh wait, you're already doing that when you call `initNode`.

onox:
Shall I just add a commit that shows what I mean? You can always do `git revert <commit ID>` later to revert my commit if necessary :stuck_out_tongue_winking_eye:

wlbragg:
I think your replying to something I deleted, I was confusing the new props for the ones I use to set the initial state of the internal cabin conditions. So I think you had it right in the first place. But i have an error, give me a second, I think it is because I try to use the new props before they are defined.

wlbragg:
OK check that out, I think this meets you criteria and mine.

onox:
I'll commit one small fix on top of your commits.

onox:
I have added the commit.

wlbragg:
OK, look DONE!

onox:
Nice, I'll update bug-237 to record the properties.

wlbragg:
:+1:

PR #272, opened 2015-06-09
Minor save state change, remove duplicate FloatStruts

wlbragg:
Save pontoon damage blend file with some extra objects hidden. One of which created an unseen duplicate FloatStrut.


onox:
Does this fixes a particular bug?

wlbragg:
Not a known bug, I just discovered it when looking at the pontoon VS aircraft pitch issue.

PR #276, opened 2015-06-11
No landing, taxi or ALS lights with left or both wings broken

wlbragg:


onox:
Tested and works, thanks :smile:

PR #277, opened 2015-06-11
JSBSim hydrodynamics-based water interactions for pontoons/floats.

### LINKED ISSUE #254: Request for feedback: Hydrodynamics-based pontoons/floats modelling
My hydrodynamics-based (as opposed to contact point-based) modelling of floats is approaching an, I think, acceptable state. You can get it here
https://github.com/andgi/c172p-detailed
in the default hydrodynamics-wip branch. The branch will be rebased onto the upstream master branch frequently so if you add your own commits to it update it with git pull --rebase.

As far as I know the main thing missing now is completing the integration with the damage system and, of course, improving the hydrodynamic reactions - the latter is a long term effort and compared to a contact point based approach without any additional FDM parts (c.f. those in the Boeing314) I'd say it's already ahead.

Testing is needed, both to to discover problems in handling quality but also missing pieces with respect to the effects (spray, sparks, damage, etc, etc) present in the current contact based version.

Handling tips:
* On take-off a slight back-pressure on the stick is advisable, none or too much will get you into trouble. This effect has emerged from the modelling but it is unknown if it is in any way similar to the RL seaplane problem of porpoising which has a strong dependency on the pitch attitude.
* If the nose is anything more than a very tiny bit below the horizon when you touch down you will be very sorry - and wet.
* When the hydrodynamic reactions freak out, consider carefully what you did before and when that happened and if that would have been advisable to do in a RL seaplane. Would damage have occurred first?

andgi:
The static buoyancy parts have been generated from the floats 3d model.
The various lift and drag coefficients have been tuned for "reasonable"
behaviour but has not been validated against real performance data.
    
The hydrodynamics system is not yet completely integrated with the damage
and effects systems of the aircraft.

Fixes #254


onox:
You changed some files in `Models/Effects/pontoon/`. Can you replace the four `compression-ft` properties in `Systems/flight-recorder/components/effects.xml` and record `/fdm/jsbsim/hydro/active-norm` instead?

andgi:
@onox Do you know the intention behind the  signals in Systems/flight-recorder/components/effects.xml?
They also need to be updated: contacts 17, 18 are the rear of the floats, contacts 19 and 20 doesn't exist anymore (these indices are now for the amphibious landing gears).

onox:
@andgi Because those properties were necessary for the smoke.

andgi:
@onox I see. I had not fixed the smoke and sparks particle systems. Is it ok to trigger it from any of the float contacts (fore, step, aft)?

onox:
Can't you just make sure the smoke looks exactly as in non-replay mode? Just look at the .xml files that generate the smoke and make sure these properties are recorded. That's what I did.

onox:
Also, some properties for wlbragg's damage mod are recorded in `Systems/flight-recorder/components/damage.xml`. Just FYI.

andgi:
@onox Yes, the damage.xml part is covered - no change in properties.
Ok. I'll just change the smoke to trigger on all float contact points and record them all.

onox:
Btw, for the wakes you are using `/fdm/jsbsim/hydro/active-norm`. So that one should be recorded. See https://github.com/Juanvvc/c172p-detailed/pull/277#issuecomment-111285332

onox:
@andgi Do you check that the pontoon smoke is not generated on land?

onox:
Actually it seems to be be generated anytime during replay.

andgi:
@onox That looks similar to the spray particle system, but it should not activate if /fdm/jsbsim/ground/solid is true and requires fdm/jsbsim/hydro/active-norm to be non-zero. That should not happen on solid ground. Is /fdm/jsbsim/ground/solid recorded? It is part of the spray's condition also in master.

But there is also a smoke particle system in Models/Effects/
pontoon/{left,right}_ground.xml. I'm not sure how it looks, when I tried to trigger it I only saw the spark particles (which uses the same trigger condition).

onox:
@andgi Sorry, it has nothing to do with ground actually.

onox:
If `/fdm/jsbsim/hydro/active-norm` is the only property required to enable/disable the pontoon spray, then it's not useful to record `/fdm/jsbsim/ground/solid`.

onox:
I think `/fdm/jsbsim/ground/solid` is computed during replay, so no need to record it.

onox:
@andgi Okay, I'm really messing up now. It seems you do no need to record `/fdm/jsbsim/ground/solid`!

onox:
@andgi I'll try to revert that commit.

onox:
@andgi I think you should revert that commit anyway, because `active-norm` is not a boolean. Plus you need to record `/fdm/jsbsim/ground/solid`. That should be enough to fix it.

wlbragg:
Go away for an hour and miss all the fun.
I'm having a bit of a hard time following the issues, anything I can do or answer?

I'm pulling this now to see if I can help with anything.

wlbragg:
I feel so stupid at times :(, I have never pulled a PR from github that wasn't a branch, I have no idea how to get it unless it is in the form of a branch. How do I get this pull request locally to test it?

wlbragg:
I see that we are getting wake in recorder regardless of where we are (solid or not). 

I tried adding /fdm/jsbsim/ground/solid to effects.xml in recorder/components.
Is that the correct place to add it?

Don't you have to record all conditions to get the desired outcome?
If so there is also /velocities/groundspeed-kt involved.

I first tried /fdm/jsbsim/ground/solid and that shut it off completely (no wake/spray). Then I tried to add  /velocities/groundspeed-kt and that had no change (still no wake).

I don't understand how the recorder works, I've never had to program it before.


In file F:/FlightGear/Custom Aircrafts/c172p/Systems/c172p-hydrodynamics.xml: li
ne 25
      Property floats-enabled-norm is already defined.

wlbragg:
The sinking float issue we discussed elsewhere is still there. I was testing this over MP and one of my aircraft was sitting on the water for several minutes while I got the other one landed on water next to it. By the time I got them close together the one that was sitting there had its float completely under water. However, applying a little power seemed to pop it right back up, also switching pontoon and back to amph may have floated it again.
It almost seems like it is maybe total lack of movement that causes it? May be a direction to look at.

wlbragg:
Speaking of MP, everything that I saw looked good.

@onox, there was no wake or spray visual over mp of the other craft, do I need to create an alias to get that to show?

wlbragg:
I answered m own question about " do I need to create an alias".
Something strange though.
When I tried to apply bumpspec to child models, I had to put the effect code block in the childModel.xml file for it to work.
Same thing is true of the params definition for MP aliases. If I try to put it in Models/c172p.xml it doesn't work, but if I put the params definition block in the Effects/pontoon/typeOfEffect.xml it works.

Speaking of multiplayer effects, I don't recall having the sparks and smoke showing over MP?
If not while I am at it I can add it also unless we weren't planning on adding these effects to be visual over MP?

wlbragg:
Well I am just about done for the night and it seems like I haven't gotten anywhere.

I got nowhere getting the wake/spray recording correctly (see above).
Now I am stuck on the MP wake/spray effects.

Seems there is another problem with getting the wakes/spray to work over MP.
As I said in my previous post, the alias didn't work on the local aircraft (as in no wake or spray on your own aircraft) unless you put the params definition in the effect.xml file. But even though the alias works for your own plane by putting it in the effect.xml file, it isn't working (does not show a wake or spray) on the MP aircraft.

I'm somewhat stumped by these findings.

wlbragg:
Another undesirable effect of this hydro effects is when I crash in any kind of spin in water, the plane continues that spin on the water. Very unrealistic, we need to find a way to stop the spin. It's like it loses any kind of friction and just enters an infinite spin.

andgi:
@wlbragg "Also, unrelated, in console there is the following report. Don't know if it is of any concern.
'In file F:/FlightGear/Custom Aircrafts/c172p/Systems/c172p-hydrodynamics.xml: line 25
Property floats-enabled-norm is already defined.'" -> No, this is harmless and due to floats-enabled-norm being used as input for c172p-hydrodynamics.xml but is computed by bushkit.xml.

andgi:
@onox I'll work on the spray and wake conditions and the recorder ASAP. Reverting the conditions is quickly done.

dany93:
>Another undesirable effect of this hydro effects is when I crash in any kind of spin in water, the plane continues that spin on the water.
If anyone figures any of this stuff out over night and run out of time to implement anything, leave me instructions and I can work on it during the 2nd shift! :)

Maybe by introducing a 0 factor in every roll moment of the FDM.
c172p.xml, lines > ~ 1113, ```<axis name="ROLL">```
Like (done for wing broken)
```
                    <table>
                          <independentVar>aero/function/total-wing-damage</independentVar>
                          <tableData>
                              0.0000	1
                              1.0000	0
                          </tableData>
                    </table>
```
With, instead of ```aero/function/total-wing-damage```, a property which would change when in the water. Or best, **which would become 0 under 10 or 30 m AGL**.
Something like (not tested)
inside each ```<product>```
```
                    <table>
                          <independentVar>/position/altitude-agl-ft</independentVar>
                          <tableData>
                              0.0000	0
                              20              1
                          </tableData>
                    </table>
```
My idea here is to cancel **every** roll moment below 30 ft AGL.

If you find it hard I can try it, but I'd also like to test the possibility of two engines https://github.com/Juanvvc/c172p-detailed/issues/273#issuecomment-111417586. What is your priority?

andgi:
@dany93 When wlbragg says "spins" I think he means a yaw axis rotation, and there is, in fact, no hydrodynamic yaw damping moment when the water speed is zero. I'll add the term back and try to tune it ASAP.

wlbragg:
Yes, I think yaw is the term, and I just experienced it without a crash scenario. I was sitting on the water (testing the windshield splash).  I was checking the splash pattern and direction (applied by RPM VS forward speed) by increasing the rpm rapidly and back to reducing it before I gained any forward speed and that put me into a right hand "yaw" spin that I couldn't get out of until I dropped the gear to apply friction. I was then able to gain control again.

wlbragg:
"I have not been able to reproduce the sinking float issue."

I managed to capture video of both the spin and sinking.
https://dl.dropboxusercontent.com/u/63093453/fg/c172p/video2.mp4

The frame rate is terrible so let me describe a little bit more of what I have uncovered.

More wind seems to = less instability or sinking.
Aircraft motion = less instability or sinking.
It almost seems that lack of motion is suspect.

wlbragg:
I have spent another couple hours on getting wake/spray to show across MP and "no go"!

Because of the fact that it required moving the params (alias) section to the child effect file to make the alias visible to the conditional code block, I decided to try moving an entire particle effect code block to Models.c172p.xml directly instead of using the child_file.xml, that had no effect, still didn't work.
But I can report no other issues in all this testing with either MP or hydro-fdm.

The only other thing I suspect is the var type definition of the properties at play.
I think we use int for bool in alias's?

So the aliases at play are
float n="6" alias="/velocities/groundspeed-kt"/
int n="23" alias="/fdm/jsbsim/ground/solid"/
int n="24" alias="/fdm/jsbsim/hydro/active-norm"/
which convert to 
sim/multiplay/generic/float[6]
sim/multiplay/generic/int[23]
sim/multiplay/generic/int[24]

andgi:
@wlbragg 'int n="24" alias="/fdm/jsbsim/hydro/active-norm"/' is no good at all. active-norm is a float property. The result is most likely that it is not sent over MP.

wlbragg:
Oh, you know, I thought that a possibility and tried a test to turn that to float, but I think it was incomplete on my part and I abandoned that path. I will go back and make sure I get that scenario covered and see if it works. Will be later tonight.

onox:
@andgi @wlbragg If we want the spray to be visible over MP as well, then we need to keep things simple, and just use one property that says whether to display the spray or not. But that means you have to compute that property in JSBSim or in a `<property-rule>`.

andgi:
Good point. I guess the same applies to float sparks and smoke (I have seen the smoke now!) too.

@wlbragg I have seen the bobbing too now but I don't understand how it can happen. It went away once I started the engine but came back when I stopped the engine. It is weird.

wlbragg:
OK, it was because I was using int instead of float.
So now I am stuck with a couple of questions.
@onox, while I understand why you say "keep it simple" I'm not sure doing what your asking is any simpler than what I have already finished.
Yes it is three aliases VS one property rule, but in this case the aliases are already configured and working for both the MP and local conditions (there already finished and the file has to have the condition and the 3 properties anyway). Your asking me to convert it back to using 3 properties and making a property rule VS just using the three aliases. I don't see any gain here. It seems to work well using the aliases. For now I say lets just get this in place and if you want to refactor it with FDM or prop rule later, I have no problem with that..

Now question 2, I created this work using the new hydro code base. I am doing it in a locally  merged version of master and the hydro base. I'm not sure what the best way to get this onto github would be. I can't really merge it into master because it uses /fdm/jsbsim/hydro/active-norm and master isn't there yet.
I did figure out how to pull hydrodynamics-merge and get it local, will I be able to push that back to the github PR for it as a commit?

wlbragg:
onox, if I did do this as a property rule, what would I do?
I'm not sure what you asking me to do, input the three conditions and output one and use the one output as an alias?

onox:
> but in this case the aliases are already configured

@wlbragg You mean it's already configured on your computer? Because I don't see it here in `<multiplay><generic>` in `c172p-set.xml`.

One reason is we might need the MP properties for other things in the future. Also, using three properties means you have to add three properties to the flight recorder instead of just one. I think it's better to keep the number of properties that the flight recorder has to record and keep in memory to a minimum.

onox:
> I'm not sure what you asking me to do, input the three conditions and output one and use the one output as an alias?

Yes :smile:

wlbragg:
OK, you know we're about out of time here. Are we going to merge this PR now or 3.8? If now, I am not going to set up for all the contact points, I will just do the hydro version. If not untill 3.8 than i need to set it up for contact points (which is even more work).
So what and when do you think we are going to do?

onox:
> I did figure out how to pull hydrodynamics-merge and get it local, will I be able to push that back to the github PR for it as a commit?

If you can add your commits on top of @andgi's latest commit (a69f9cb) you can create a PR at https://github.com/andgi/c172p-detailed.

Or we merge this PR, and you create a branch based on the latest master, and you add your commits to that new branch. Whatever @andgi wants.

onox:
We can merge this PR now, and you create a branch from master. That might be easier for you.

wlbragg:
Agreed!

onox:
Done! Please create a branch and a PR when you're ready. :smile:

wlbragg:
Will do!

onox:
@wlbragg Hang on a second, I'm going to revert that one commit from you that @andgi added.

wlbragg:
OK

onox:
@wlbragg Ok, make sure you're in master with `git branch`, pull, and do `git checkout -b spray-fix` or something.

onox:
Well done @andgi, thanks for the contribution. You can sit back, relax, and take some snus :wink:

wlbragg:
onox, why did you revert spray refactor?

onox:
Because `fdm/jsbsim/hydro/active-norm` is not a boolean.

wlbragg:
Oh, I'm changing it all anyway, but OK, no big deal. I'll see what I can do with prop rule.

onox:
Yeah, just create a PR so I can test and review it :smile:

wlbragg:
I'm still a bit unclear about how that is to work. Because I am used to using the aliases in the effect file condition. Now I am not going to use an alias there put I am still going to use the normal properties for conditions and use this new alias based on the same multiple properties factored into on alias only for MP?

wlbragg:
Can you pseudo code an example of this prop rule being applied to both the effects file condition and across MP ? Does that make sense?

onox:
You use the property-rule to output to `/fdm/jsbsim/hydro/spray` or something. You alias this property in `<multiplay><generic>` in `c172p-set.xml` and then you create a `<params>` in the effect xml file and do `<property alias="/params/..."/>` in the same file.

onox:
If that doesn't work, try relative paths. Something like ../.../params, etc. That's what they do in various xml files of the instruments. (Although /params works for the main model xml file, so /params might work in other xml files too). In worst case scenario, you can use `/sim/multiplay/generic/...`

wlbragg:
It clicked in my head as you were replying. OK, I think I get it. Expect PR by Morning.

onox:
:+1:

andgi:
Thanks!

I'll open a new pull request with added yaw damping - that should prevent spinning on the water.

PR #279, opened 2015-06-13
MP wake effects

wlbragg:
I made a property rule, everything worked perfectly with it except over MP. All the generic params looked correct but there just wasn't any wake. So I decided to do it with all the aliases. However as it turns out it was only going to save one alias because I needed groundspeed_kt separate anyway or I would have had to make a bunch of different prop rules and an alias for each which put us right back to where we were anyway. Look at the wake, spray, middle files, they all take different  groundspeed_kt params so it needed to be stand alone. That only left hydro/active-norm and ground/solid of which I also tried a prop rule fro the two. Again, locally everything worked fine, but not over MP. So we didn't save "one" alias. We can try to refactor later, maybe something will stand out that is causing the problem.  I honestly don't remember where this left off for recorder. I think it still needs ground/solid added to it. I'll look at that in the morning along with pontoon MP ground effects. This shouldn't have taken anywhere near as long as it did to finish, but testing MP is difficult especially when it doesn't want to cooperate.

### DIFF
--- a/Models/Effects/pontoon/left-spray.xml
+++ b/Models/Effects/pontoon/left-spray.xml
@@ -2,27 +2,40 @@
 
 <PropertyList>
 
+	<params>
+        <!-- IMPORTANT: int properties 14 to 19, both included, are reserved for the immat system. Do not use them! -->
+		<ground_solid>
+			<property>sim/multiplay/generic/int[23]</property>
+		</ground_solid>
+		<groundspeed_kt>
+			<property>sim/multiplay/generic/float[6]</property>
+		</groundspeed_kt>
+		<hydro_active_norm>
+			<property>sim/multiplay/generic/float[7]</property>
+		</hydro_active_norm>
+    </params>
+
     <particlesystem>
         <name>left-float</name>
         <texture>spray.png</texture>
         <emissive>false</emissive>
         <lighting>false</lighting>
-		
-        <condition>
-            <and>
-				<not>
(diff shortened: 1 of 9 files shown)


onox:
I'll take a look at this.

onox:
Finished changing some code, now testing...

onox:
My changes are working over MP :smile:

onox:
Need to check flight recorder...

onox:
@wlbragg It's done. You can test the changes.

wlbragg:
"I don't understand how the product of spray-wake-speed-kt drives the condition of value < 20"

After looking at it again, the product of all those conditions is still the output of groundspeed-kt

/fdm/jsbsim/ground/solid = 0
/hydro/active-norm = 0 or 1 (this one is confusing because can't it be 0.?? which will influence the output product)?
/velocities/groundspeed-kt = 20.0

0 * 0 * 20 =

could be 
0 * 0.?? * 20 =

or

0 * 0.5 * 20 = 10 (when it needs to be output as 20 for the effect to work?)

andgi:
@wlbrag /hydro/active-norm decreases as more of the aircraft's weight is carried by the wings, but then the groundspeed should be high so hitting the lower boundary probably won't happen until very near the lift off instant.  Also, /hydro/active-norm already includes the intention of /fdm/jsbsim/ground/solid via the floats-enabled-norm property which is 0 if all float contact points are over solid ground.

PR #283, opened 2015-06-13
Mp record float ground effects

wlbragg:
@onox, please check carefully, I verified the MP but not the recorder.

Also we lost pontoon damage in the hydro changeover (I think). I may not be able to fix until late tonight. It should be in physics.nas (if contact points compression = ....) those points may not be there anymore, need to use 14-19 I think.


wlbragg:
Also, I just notice the merge trail, sloppier than I would have liked but too much work to have cleaned it up.
Please check closely.

onox:
The changes, look alright. Let me check if I can do something about the commits history.

onox:
It seems you continued from a commit (b721cfc) from feature branch

PR #287, opened 2015-06-13
Added water sound from the pontoons/floats at speed.

### LINKED ISSUE #264: Sound effect for floats
Sound effect for floats: needs the sound itself and then to be programmed.

andgi:
Fixes #264 

Added sound effects for the pontoons/floats using the existing `water-fast.wav` sound.
The volume is rather subtle. What do you think?

### DIFF
--- a/c172-sound.xml
+++ b/c172-sound.xml
@@ -979,7 +979,7 @@
    <condition>
     <greater-than>
      <property alias="/params/spray_wake_speed_kt/property"/>
-     <value>2</value>
+     <value>5</value>
     </greater-than>
    </condition>
    <position>
@@ -1005,7 +1005,7 @@
    <condition>
     <greater-than>
      <property alias="/params/spray_wake_speed_kt/property"/>
-     <value>2</value>
+     <value>5</value>
     </greater-than>
    </condition>
    <position>
@@ -1023,6 +1023,28 @@
     <max>2.0</max>
    </volume>
   </right-pontoon>
+
+  <pontoon-in-water>
+   <name>pontoon-in-water</name>
+   <mode>looped</mode>
+   <path>Sounds/water-still.wav</path>
+   <condition>
+    <greater-than>
+     <property alias="/params/spray_wake_speed_kt/property"/>
+     <value>0</value>
+    </greater-than>
+   </condition>
+   <position>
+    <x>0.0</x>
+    <y>0.0</y>
+    <z>-2.0</z>
+   </position>
+   <reference-dist>10.0</reference-dist>
+   <max-dist>25.0</max-dist>
+   <volume>
+    <offset>0.3</offset>
+   </volume>
+  </pontoon-in-water>
 
 </fx>
  
--- a/c172-sound.xml
+++ b/c172-sound.xml
@@ -972,22 +972,27 @@
         <max-dist>50.0</max-dist>
     </baggage-door-closing>
 
-  <left-pontoon>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
See my comments at #264

onox:
@andgi It seems 1f9d6f0 ended up in the PR, could you do a rebase?

c172-sound.xml
```diff
@@ -289,6 +292,80 @@
...
+     <value>0</value>
+    </greater-than>
+   </condition>
+   <position>
+    <x>0.0</x>
+    <y>0.0</y>
+    <z>-2.0</z>
+   </position>
+   <reference-dist>10.0</reference-dist>
+   <max-dist>25.0</max-dist>
```
onox: Currently this sound is not hearable from the external views. Can you increase `<mix-dist>` to 50.0 to make it hearable and then decrease `<reference-dist>` to for example 5.0 to decrease the volume?

onox: Also make sure you lower the volume if `/sim/current-view/internal` is true. You could use `<volume>` with a `<property` for this, for example.

andgi:
@onox The still water sound is intentionally silent in external views. I don't think it would be audible at >20-25 meters away.

onox:
Ok. Is it still possible to reduce the volume if `/sim/current-view/internal` is true and the two doors are closed?

andgi:
Sure, I think I saw other sounds doing that in the file. I'll get to it later today.

wkitty42:
@andgi i don't guess there's any way to determine the zoom level is there? when we're out in our flat bottom boat, we easily hear the sound of the water slapping the hull from at least 10 feet away... but it also depends on how rough the waves are... by that i mean how high they are from the breeze blowing them... generally no more than 4-6 inches from top of wave to bottom of trough... but then, maybe this would be too much "reality" ? :wink:

onox:
@wkitty42 FlightGear changes the field of view when you press x or shift+x. Personally, I think it would be better to simply move the whole position of the current view towards the target you're looking at. That will also change the distance between the view and the position of sounds, thus some sounds will automatically become hearable/louder when you get closer. In that case it would also work for other aircraft as well. But this is something that needs to be fixed in FlightGear / fgdata itself.

wkitty42:
@onox i guess FOV is the same as when i hold the mouse button and roll the wheel? that's what i was calling the zoom... but yes, i agree, now in retrospect, that it is something that would need to be handled by fgdata...

andgi:
Sound configuration isn't my thing. There are now separate interior and exterior sound effects (using the same sounds). The interior sounds attempt to take open/closed doors into account but it doesn't work well, as, apparently, the door properties can only be used to add or subtract from the total volume. I think an external function (e.g. property-rule) might be needed to compute the volume - then it should also be possible to merge the interior and exterior sound effects again.

onox:
I think it's better if we just use a `<property-rule>`? Note that the door position when opening is between 0.5 and 1.0. (0.0 to 0.5 is the handle moving). Can we use a mathematical function to quickly increase the volume if the value is near 0.5 and slower increase when near 1.0?

wlbragg:
Love the sounds, I hadn't heard (pulled) any if this until just now (an hour ago). Really cool, it make a big difference, seeing, hearing, now we need to add feel and smell.

gilbertohasnofb:
> Love the sounds

Glad to hear that :smile:

PR #290, opened 2015-06-14
Bug 248

### LINKED ISSUE #248: Damage repair does not recharge the battery
load the craft normally with the engine off... make sure that damage is enabled in the craft's options... press the autostart just to ensure that everything is ready to go... do not press the 's'tart button! also, make sure the instrument lights are turned all the way up to 100%... now, go get lost editing or providing support or whatever... as long as you're away long enough for the battery to run down... now, when you get back, you cannot start the engine but the instrument lights are all still on! if the battery is dead, the instrument lights should be off, too...

so, then click the damage repair button... now try to turn the engine over... no go.

When pressing "Repair" button:
* [x] Recharge battery
* [x] Enable all circuit breakers

onox:
Fixes #248

### DIFF
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -61,8 +61,27 @@ var init_electrical = func {
     print("Electrical system initialized");
 }
 
-var reset_battery = func {
+var reset_battery_and_circuit_breakers = func {
+    # Charge battery to 100 %
     battery.reset_to_full_charge();
+
+    # Reset circuit breakers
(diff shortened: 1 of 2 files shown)


wkitty42:
yes, this seems to work... i let the battery drain a little bit, turned off all the switches, clicked the [repair] button and the battery value went to 1... i turned on the battery switch and it dropped to .99xxxx and continued dropping... i clicked the [repair] button again and it went back to 1.0 and immediately was dropping again... this looks good and proper to me :smile_cat:

PR #296, opened 2015-06-14
MP and record damage ground effects

wlbragg:
These are the spark and smoke effects from collapsed gear, wing tips and tail drag.
Also refactored gearAGL.xml to ground-effects.xml. Created channel groundeffects in c172p-ground-effects.xml and ground-effects.xml. 
I have one more set to do and that is off road tire smoke.


Models/Effects/damage/left_gear.xml
```diff
@@ -40,7 +47,13 @@
...
     </offsets>
 
-    <condition>
+	<condition>
+	    <greater-than>
+            <property alias="/params/left_gear_ground_effect_speed_kt/property"/>
+            <value>8</value>
+        </greater-than>
+    </condition>
+    <!--condition>
```
onox: Can you delete the commented code?

Models/Effects/damage/left_gear.xml
```diff
@@ -28,8 +28,15 @@
 
 <PropertyList>
 
+  <params>
+    <!-- IMPORTANT: int properties 14 to 19, both included, are reserved for the immat system. Do not use them! -->
```
onox: It's not really necessary to put this warning here in all these effects files as long as it's put in `c172p-set.xml`. That's where you decide which multiplay property to use.

Systems/flight-recorder/components/effects.xml
```diff
@@ -35,6 +35,36 @@
         <property type="string">/fdm/jsbsim/contact/unit[8]/WOW</property>
     </signal>
 
+	 <!-- Gear smoke ans spark-->
```
onox: Small typo :smile:

onox:
@wlbragg Can some of these `/fdm/jsbsim/contact/unit[i]/WOW` properties be removed from `Systems/flight-recorder/components/effects.xml`?

onox:
I think `/fdm/jsbsim/contact/unit[6]/WOW`, `/fdm/jsbsim/contact/unit[7]/WOW`, and `/fdm/jsbsim/contact/unit[8]/WOW` can be removed. They are only used during non-replay mode.

onox:
And `/fdm/jsbsim/contact/unit[3]/WOW`, `/fdm/jsbsim/contact/unit[4]/WOW`, and `/fdm/jsbsim/contact/unit[5]/WOW` as well.

onox:
Good that you renamed `gearAGL.xml` to `ground-effects.xml` :+1:

wlbragg:
Yes, I forgot to remove the comments, I'll remove them in my next commit over this branch with the off road smoke. I'll also clean up the recorder file as per your suggestion.

onox:
Thanks.

wlbragg:
OK, one question. Look at Effects/tyresmoke/tyre-smoke-????.xml

...xml
 <condition>
	  <or>
	    <property>/gear/gear[1]/tyre-smoke</property>
	    <greater-than>
            <property alias="/params/left_tyre_smoke_ground_effect_speed_kt/property"/>
            <value>100</value>
			
        </greater-than>
	  </or>
    </condition>
...
the property
...xml
  <property>/gear/gear[1]/tyre-smoke</property>
...
is only = 1 on a hard ground landing and only for a short time. It was for touchdown tire smoke.
I really wasn't sure how to handle it, I hated to write even more aliases for it.

onox:
@wlbragg It was possible to use `/gear/gear[1]/tyre-smoke` in the JSBSim functions?

onox:
Also you named everything `<left_tyre_smoke_ground_effect_speed_kt>` :smile:

onox:
Btw, the old code says:

```xml
<greater-than>
    <property>fdm/jsbsim/ground/rolling_friction-factor</property>
    <value type="double">1</value>
</greater-than>
```

So, wouldn't it be better to replace `<property>/fdm/jsbsim/ground/rolling_friction-factor</property>` in `Systems/c172p-ground-effects.xml` with:

```xml
<gt>
    <property>/fdm/jsbsim/ground/rolling_friction-factor</property>
    <value>1</value>
</gt>
```

Or am I missing something?

onox:
In that case you can just use the number 20 instead of 100 in the `Effects/tyresmoke/tyre-smoke-*.xml` files.

onox:
About `<property>/gear/gear[1]/tyre-smoke</property>`. What you can do to get "OR" functionality, is use:

```xml
    <sum>
        <property>/gear/gear[0]/tyre-smoke</property>
        <product>
            ...
            <gt>
                <property>/velocities/groundspeed-kt</property>
                <value>20.0</value>
            </gt>
        </product>
    </sum>
</function>
<clipto>
    <max>1</max>
</clipto>
```

In the effect file you then have to remove the `<greater-than>` test because it has been moved to JSBSim.

onox:
Actually, JSBSim has an `<or>` element: http://jsbsim.sourceforge.net/JSBSim/classJSBSim_1_1FGFunction.html

wlbragg:
Thank onox, I reviewed you comments and will make changes this afternoon/evening.

onox:
@wlbragg In case you have noticed that we've just merged some big branch to master, you don't need to merge master to this branch because Github says we can still merge this PR (+ it will not put an ugly "Merge master" commit in this feature branch :smile:).

wlbragg:
"Also you named everything ```xml<left_tyre_smoke_ground_effect_speed_kt>```"
Where are you seeing this?

wlbragg:
OK, I think I got it all. I can't believe I butchered it as bad as I did last night. I must have been really tired :(

wlbragg:
I just got through testing this on MP and everything looks good. It is really cool to see all those effects carry over to MP.

wlbragg:
Can we merge this?

onox:
Let me check this.

onox:
That last commit looks OK, so yes, :shipit:

onox:
> "Also you named everything <left_tyre_smoke_ground_effect_speed_kt>"
Where are you seeing this?

I checked the diff again and it occurs in just one location, so I must have been very tired yesterday :tired_face:

wlbragg:
What do you think about bringing the tire water spray over from Aircraft/Generic/Effects/ to local so we can add it to MP effects? Right now I think that is the only effect that isn't setup that way. I never did like the idea of centralized "common use" items. It might not be a bad thing if you could always guarantee they would hold all the advantage of a local object. But they hardly ever do for such little gain.

onox:
That's true. In which files are those water spray effects referenced?

onox:
I'm not completely sure which effect you're talking about right now.

wlbragg:
Models/c172p.xml lines 909

wlbragg:
It's just water spray off tires when raining.

onox:
Ah, yeah, we should add that as well :smile:

wlbragg:
Will do

onox:
Btw, the maximum float and int properties can have index 19. So we have float[18] and float[19] available. Also, following properties shouldn't work:

```xml
274         <int n="20" alias="/fdm/jsbsim/wing-both/broken"/>
275         <int n="21" alias="/fdm/jsbsim/wing-damage/left-wing"/>
276         <int n="22" alias="/fdm/jsbsim/wing-damage/right-wing"/>
```

onox:
To be honest, I'm not sure why we aren't using a `string` property like `string[0]` for the registration number. Now it's taking up precious properties :stuck_out_tongue_winking_eye:

onox:
Although, we can (ab)use properties in `/engines`, or `/rotors` if we really want. See http://wiki.flightgear.org/Multiplayer_protocol They get transferred automatically, so you don't need to do any aliasing in `c172p-set.xml`.

wlbragg:
Oops, 20-22 do work so what does that mean?

onox:
They do work?

onox:
They shouldn't because they aren't in the list.

wlbragg:
I guess I'd better double check.

wlbragg:
20 is not working

wlbragg:
There not working. OK, time to refactor.

onox:
Then 21 and 22 shouldn't be working either :grinning:

wlbragg:
They aren't and 13 which is pontoon right broken (which is the next INT we use is working.

onox:
That's beause 13 <= 19? :stuck_out_tongue_winking_eye:

wlbragg:
10-4

wlbragg:
Yes, just saying. I think I have some room to play with on some of the earlier damage effects, hopefully I can get enough freed up!

onox:
Ok. Just know you can (ab)use the unused `/engines`, or `/rotors` properties if you want.

wlbragg:
Looking at it again I think I am going to have to. I really thought I might be able to free some up but I don't see any.

onox:
Can we try to move the immat properties to string[0] or string[1]?

wlbragg:
Where are they?

onox:
`/sim/multiplay/generic/string[0] `

onox:
Does this registration number thing even work over MP?

onox:
`Nasal/immat.nas` is only used locally.

wlbragg:
No I don't see it over MP.

wlbragg:
Unless it's used in pilot list or MP chat.

onox:
You need to define it first in `c172p-set.xml`.

wlbragg:
Which means they are free because it wasn't set up?

onox:
Add this to `<multiplay><generic>`: `<string n="0" type="string">Hi there! :)</string>`.

onox:
They are free yes, you can use them.

onox:
Although, sometimes @andgi's MP broadcast system uses some of the string[] properties, so need to be careful with the index.

wlbragg:
"@andgi's MP broadcast system uses some of the string[] properties"

What is that and which ones?

onox:
I looked in my dual-control branch again, and it seems you need to specify which string[] property the MP broadcast system can use, so I think we can safely use string[0] for the registration number.

wlbragg:
Oh, and then we can user 14-19?

wlbragg:
They were using one INT per registration digit? How?

onox:
The problem right now is that `Models/Immat/immat.xml` uses int[14] to int[19]. If we can find a way to just use string[0] then we can reclaim int[14] to int[19].

wlbragg:
Right, looks like hey used each int to carry like a single digit ASCII char  code or something to that effect?

onox:
Why is it using textranslate animation? Can't we just `<text>` or something?

onox:
Like how the callsign was placed on the panel near the radio stack. (Old location)

wlbragg:
This would require rewriting or replacing immat.nas

onox:
That's not a problem for me. The `<animation>` stuff is (for me).

onox:
If you are given a string property, are you able to place the text of it on the fuselage?

wlbragg:
point me to the file and line#

onox:
`Models/Immat/immat.xml` line 70 for the `<animation>`'s

wlbragg:
Does canvas do text?

onox:
Hmm... good idea. I think it can do text yes :smile:

onox:
Shall we just temporarily nuke/freeze the immat system and reclaim int[14] to int[19]?

wlbragg:
Works for me, we may get yelled at by the first person that uses it!

onox:
Oh, I don't worry about that.

onox:
Let's create a new issue for this.

onox:
Too many comments in this PR already :stuck_out_tongue_winking_eye:

wlbragg:
OK, I have to go, can you kill the immat and I will move the aliases over in the morning?

onox:
Yes. See #299

wlbragg:
OK, thanks, we'll go from there.

onox:
Good idea from you to use Canvas :smile: :+1:

Issue #80, opened 2015-03-13
add functionality to the circuit breakers 


gilbertohasnofb:
As the user tigert posted in the forum thread:

"[...] you can see the circuit breakers in the diagram [see link below], those are listed with their destinations, so they can also be modelled correctly. Breakers are basically current-limited switches that pop open when the rated current is exceeded so you dont melt the wires and get a fire if some equipment breaks. You can also pull them from the knob by hand on some and you can push them back if you suspect the problem is solved."

http://s18.postimg.org/n12yt1qk9/Screenshot_from_2015_03_13_17_33_45.png

gilbertohasnofb:
@Juanvvc wrote in the forum: "Breakers pushed. Not all breakers are functional yet. The master and flaps breakers do nothing yet. The pitot heat breaker does nothing, and probably will do nothing forever"

Also, I believe the interior lights breaker should affect instrument lights, and the turn coordinator breaker isn't working either.

Juanvvc:
Master and flaps breakers were already functional later that night :) Yes, interior lights breaker is not functional yet. The turn coordinator breaker did work when I tested it: remove the breaker with master OFF, switch master ON and the turn coordinator doesn't move. Insert the breaker and the tc starts moving.

The non-funcional breakers are interior light breaker and iirc, some of the exterior lights. I think landing light breaker does work correctly.

gilbertohasnofb:
> pull the breaker with master OFF, switch master ON and the turn coordinator doesn't move. Push bacj the breaker and the tc starts moving.

Thanks, that's certainly working, as well as everything else you mentioned.

gilbertohasnofb:
@Juanvvc I was thinking about this a little bit more and I start to wonder why the turn coordinator does not immediately stop to work when the breaker is pulled. According to the schematics posted in the initial comment, the breaker is the only connection between the TC to the primary bus, and surely pulling it out should make the TC inoperable, shouldn't it?

Juanvvc:
The turn coordinator is one of the main instruments in Flightgear and it is implemented in C++, not in XML or Nasal. The internal gyro decays 0.5% every second, so it takes 200 seconds to stop completely once powered. This is more or less realistic and anyway we cannot do anything to change it except modelling a new turn coordinator in Nasal. This is not going to happen.

gilbertohasnofb:
[...]

PR #301, opened 2015-06-15
Bug 80

onox:
Fixes bug with the marker beacons test switch in the audio panel in #80. Also renamed the tooltip because "Radio" is confusing when the circuit breaker is connected to the audio panel instead.

fgfs-screen-085
fgfs-screen-086


gilbertohasnofb:
Tested, all working well, thanks for the PR!

PR #304, opened 2015-06-15
Bug 299

### LINKED ISSUE #299: Reclaim int[14] to int[19] MP properties
The registration number system (immat) uses 6 precious properties. We need to reclaim the properties for the more important effects.

Use Canvas to paint the registration number from `/sim/multiplay/generic/string[0]` on the fuselage. Draw text on the fuselage or use http://wiki.flightgear.org/Howto:Dynamic_Liveries_via_Canvas

`/sim/model/c172p/immat` is true for the livery "default" only. All other liveries do not display the registration number on the wings and fuselage.

Current problem is we need to replace the "textranslate" technique to display the registration number in the cockpit with either Canvas or the old method (`<text>` IIRC). We can use local properties for this since only the white plain ugly "default" livery displays it on the fuselage.

* [x] Use different local properties for displaying the registration number in the cockpit. For example, we could use `/sim/model/c172p/regnum1`, etc. (lets not keep using French). Modify `Models/Immat/immat.xml` and `Nasal/immat.nas`.
* [x] Disable the animations that display the registration number on the fuselage in `Models/Immat/immat.xml`. (If we leave it enabled, we would need to use the same local properties as those used for the cockpit panel, which looks weird over MP. So better just disable it) (All liveries except "default" do not show the registration number on the fuselage anyway)
* [ ] Use int[14] to int[19] to show water spray effects over MP

* ~~Use Canvas to paint the registration number on all liveries. Use custom colors and fonts tailored for the specific liveries.~~

onox:
Fixes #299 

Disables the usage of int[14] to int[19] so that these properties can be used for smoke and sparks, etc.

### DIFF
--- a/Models/Immat/immat.xml
+++ b/Models/Immat/immat.xml
@@ -6,24 +6,6 @@
 
   <animation>
     <!-- Immatriculations -->
-    <object-name>ImatWing1</object-name>
-    <object-name>ImatWing2</object-name>
-    <object-name>ImatWing3</object-name>
-    <object-name>ImatWing4</object-name>
-    <object-name>ImatWing5</object-name>
-    <object-name>ImatWing6</object-name>
-    <object-name>ImatFuselageL1</object-name>
-    <object-name>ImatFuselageL2</object-name>
-    <object-name>ImatFuselageL3</object-name>
-    <object-name>ImatFuselageL4</object-name>
-    <object-name>ImatFuselageL5</object-name>
-    <object-name>ImatFuselageL6</object-name>
-    <object-name>ImatFuselageR1</object-name>
-    <object-name>ImatFuselageR2</object-name>
-    <object-name>ImatFuselageR3</object-name>
-    <object-name>ImatFuselageR4</object-name>
-    <object-name>ImatFuselageR5</object-name>
-    <object-name>ImatFuselageR6</object-name>
     <object-name>ImatPanel1</object-name>
     <object-name>ImatPanel2</object-name>
     <object-name>ImatPanel3</object-name>
@@ -34,24 +16,6 @@
 
   <animation>
     <type>noshadow</type>
-    <object-name>ImatWing1</object-name>
-    <object-name>ImatWing2</object-name>
-    <object-name>ImatWing3</object-name>
-    <object-name>ImatWing4</object-name>
-    <object-name>ImatWing5</object-name>
-    <object-name>ImatWing6</object-name>
-    <object-name>ImatFuselageL1</object-name>
-    <object-name>ImatFuselageL2</object-name>
-    <object-name>ImatFuselageL3</object-name>
-    <object-name>ImatFuselageL4</object-name>
-    <object-name>ImatFuselageL5</object-name>
-    <object-name>ImatFuselageL6</object-name>
-    <object-name>ImatFuselageR1</object-name>
-    <object-name>ImatFuselageR2</object-name>
-    <object-name>ImatFuselageR3</object-name>
-    <object-name>ImatFuselageR4</object-name>
-    <object-name>ImatFuselageR5</object-name>
-    <object-name>ImatFuselageR6</object-name>
     <object-name>ImatPanel1</object-name>
     <object-name>ImatPanel2</object-name>
     <object-name>ImatPanel3</object-name>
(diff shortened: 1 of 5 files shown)


gilbertohasnofb:
> Do not draw registration number on wings and fuselage

@onox Am I misunderstanding something? I just created two liveries for the custom callsign and obviously they need the registration text at least on their fuselage (ignore wings).

onox:
#299 explains that we need the 6 properties for more useful things.

gilbertohasnofb:
Please.... see #299. I am quoting you:

>> My ideal solution would be to keep things as they are: one special livery for custom callsign, all others with fixed callsign on their exterior. I can improve this custom callsign livery, make it more interesting than simple white. But I am 100% against removing the hard coded text from the HD liveries.

> OK.

Why would I go for the trouble of creating the new liveries right now then??

onox:
I don't know? I think almost everyone wants to use the nice HD liveries with the hard coded text instead of using an (almost) white livery with custom callsign.

gilbertohasnofb:
> I think almost everyone wants to use the nice HD liveries with the hard coded text instead of using an (almost) white livery with custom callsign.

Sure onox, but have you seen that the new liveries are not almost white? They are both based on real planes once again. But anyway, it's up to you.... The thing is you should have let me know before I did this work. I just don't like to waste time... 

So please delete the two new custom callsign liveries (two subfolders and two xml files) if you are removing the possibility of having callsign. And also remove the menu option for selecting a registration number...

onox:
Sorry, that you wasted some time, but I thought it was clear that we were gonna take back int[14] to int[19].

The menu option still allows you to display your own number in your own cockpit.

onox:
Besides, I can try to use `/sim/multiplay/generic/string[0]` with some Nasal to get MP custom callsign again.

gilbertohasnofb:
> The menu option still allows you to display your own number in your own cockpit.

Not really. As I mentioned, I removed the callsign from all panels except for the "default" livery, which is now removed. So no panel is showing call signs any longer.

onox:
:open_mouth: Shall I revert your commit then?

onox:
@gilbertohasnofb I'm confused, I still see my callsign in the cockpit:

fgfs-screen-091

gilbertohasnofb:
No, not at all! The whole point of that commit is:

- it doesn't make sense to have different callsigns in the panel and then in the exterior of the plane!
- some panels have hard coded callsigns
- it was decided in the forum that only the custom callsign livery ("default") would have BOTH panel and exterior call signs

So no, if you want to remove the call sign from the exterior, you may as well remove from the panel and get one more property to use for something else. I really don't think they fit with any of the liveries left.

gilbertohasnofb:
@onox you must have turned it on in the menu (show registration in the panel). It was supposed to be off by default to all liveries except "default"

onox:
Okay.... so you want to keep immat-on-panel always disabled in all cases?

onox:
@gilbertohasnofb I can show the callsign on the wings and fuselage, but it will be your local callsign. Not that from a remote player.

wlbragg:
Wow, we need a teleconference.
Can we trick it to use a bogus defalte on the MP craft and the chosen one on the local craft?

gilbertohasnofb:
@onox this is getting quite confusing, onox... look:

- if we do not keep a "custom callsign" livery, then I don't think we need ANY of these things (that is no callsign on any part of the plane -- wings, fuselage, panel, etc.)
- if we keep a "custom callsign" livery, then it's up to you where they'd be displayed. Minimally IMO is the fuselage, then panel and wings are less important

wlbragg:
I don't want to loose "custom callsign" livery!

onox:
@wlbragg I could try in `<nasal>` to use the actual callsign of the remote player.

wlbragg:
Even if we have to come up with a work around for MP.

wlbragg:
Yes!

onox:
Okay, I'll try that one then.

gilbertohasnofb:
So why don't you guys remove:

- immat on panel
- immat on wings

and leave the callsign in the fuselage at least?!

gilbertohasnofb:
Note that if you remove immat on panel, you also have to modify the menu and remove that option "show registration in the panel"

wlbragg:
That is not the issue, it is the way it is put down, we are arguing apples and oranges. But if we still the 14-19 props then we will get wired results if we display the immat on the outside of the aircraft.

gilbertohasnofb:
@wlbragg I will let you guys discuss the technicalities then since this is something I don't know.

wlbragg:
In other words it is an all or nothing proposition with the props we are trying to get back, I sounded like you thought they use one for wing, one for tail and one for panel and we could get back at least one or two by eliminating those displays. They don't it is all 6 for each display.
I think onox is arguing that if we don't remove the outside (MP) visual display then it is going to show local registration on MP aircraft. So he wanted to remove it altogether, which does effectively kill custom registration.

onox:
Yeah, but I have added the `<animation>'s for the wings and fuselage back. The difference though is that in `<nasal>` in `Model/c172p.xml` I will use put the remote players callsign instead of his/her custom registration number (that would require usage of int[14] to int[19]).

gilbertohasnofb:
> In other words it is an all or nothing proposition 

I see now, indeed I was completely misunderstanding it.

onox:
Actually I can probably just use `/sim/multiplay/generic/string[0]`. No need to use the callsign :smile:

onox:
That would be a win (we get to use int[14] to int[19] for something else) win (we can have MP custom reg numbers) situation.

wlbragg:
:smile:

wlbragg:
It must have originally been broken down to individual chars to allow for positioning of each char on the model or there wasn't string available!

Let me know when your finished and I will review, merge, pull and do my thing.

wlbragg:
I looked at the nas code this morning and tried to do what your doing now but I kept getting it wrong.

onox:
It works :smile:

wlbragg:
:clap:

onox:
I have pushed the commit. Please test :smile:

wlbragg:
@onox, doing it now.

@gilbertohasnofb

>I just created two liveries for the custom callsign

Are you going to PR the new custom liveries?

onox:
I think our good ol' @gilbertohasnofb  just added it directly to master :grinning: :sob:

wlbragg:
It happens to the best of us!

onox:
For liveries and sounds I don't really blame him :smile:

gilbertohasnofb:
> For liveries and sounds I don't really blame him :smile:

Thanks for understanding it, onox! :smile: 

And just a comment: sorry for my grumpiness today, I'm having a bit of a stressed period here and I think I may have unleashed it on you. My apologies for the tone of my previous messages.

wlbragg:
srtingim
:clap: :clap: :clap:

I noticed the pontoon gear aren't being sent over MP (I guess), I'll fix that while I am at the other.

Great job, thank you soooo much!

onox:
@wlbragg Yeah, it's not sent over MP. But now you can first fix int[20], int[21], and int[22], by moving them to int[14], int[15], and int[16] for example.

onox:
I'm glad we have figured out how to do it with just 1 property instead of 6. It just felt so inefficient and silly.

gilbertohasnofb:
Good job, guys :smile:

onox:
Thanks :smiley:

wlbragg:
How do you make a function output and integer?

wlbragg:
I have two MP floats available for 4 different effects.

wlbragg:
can I typecast a float to an int by merely setting an int property with a float?

onox:
Hmm.. don't know :confused:

onox:
You can also use `/engines/engine[9]/n1` and `/engines/engine[9]/n2` for example. See http://wiki.flightgear.org/Multiplayer_protocol

onox:
Just make sure you do the aliasing in `<engines><engine n="9">` instead of `<multiplay><generic>`.

wlbragg:
See new PR #305, I posed the questions there.
Alright, lets switch any conversation to new PR.

wlbragg:
Got it, you think that is the way to go then?

onox:
Yes.

wlbragg:
OK

PR #305, opened 2015-06-15
Refactor wing damage and broken over MP

wlbragg:
I have two MP floats available for 4 different effects.
How do you make a fcs_function output an integer?
or
Can I typecast a float to an int by merely setting an int property with a float?


onox:
@wlbragg Check https://github.com/onox/V-22/blob/master/v22-set-common.xml#L256-L269

wlbragg:
Per conversation on Bug 299 #304 
>You can also use /engines/engine[9]/n1 and /engines/engine[9]/n2 for example. See http://wiki.flightgear.org/Multiplayer_protocol

wlbragg:
@onox, I tried to guess at the pontoon gear portion and failed miserably, I either need some education on that system or you are going to have to implement it.

So then I went on to tyre rain spray and also failed to tye the effects to MP. I stepped back and just moved the community files over to our local folder and noticed I had made a really dumb mistake in implementing the rain spray over MP. But I don't know that that is the only thing that was breaking it.

A couple questions I had for it was if it needed to be setup in the FDM (fcs_function) the same as the other ground effects and also the replay bypass, does that need to also be implemented?

wlbragg:
@onox, I don't know why I can't get even the rain tire spay to work using the MP engines for props. I looked at the example code and just can't see what I am doing wrong. I guess if you want to take a crack at it, go ahead.
Same with the pontoon gears, I failed with that also.
You can merge what is here if you want as it is correcting a few things and setting it up for the rain tire spray to be handled locally.

onox:
If I look at the files in `Models/Effects/tyrespray/rain-port.xml`, I see they use the following properties:

`gear/gear[0]/spray`
`gear/gear[0]/sprayspeed-ms`
`gear/gear[0]/spray-density`

`gear/gear[1]/spray`
`gear/gear[1]/sprayspeed-ms`
`gear/gear[1]/spray-density`

`gear/gear[2]/spray`
`gear/gear[2]/sprayspeed-ms`
`gear/gear[2]/spray-density`

This is for the default/26"/28" tires, right? Which files are being used for the amphibious?

If you look at http://wiki.flightgear.org/Multiplayer_protocol  you will see that they are not transferred automatically. So we need to do some aliasing in `c172p-set.xml`. I think we can get rid of `gear/gear[?]/spray` and just use the other two (`sprayspeed-ms` and `spray-density`).

So what I noticed is that these regular tires use `/gear/gear[0]`, `/gear/gear[1]`, and `/gear/gear[2]`. The amphibious uses `/gear/gear[19]`, `/gear/gear[20]`, `/gear/gear[21]`, and `/gear/gear[22]`. Is there a particular reason why these high indices (19 to 22) were used?

Now, if we look at http://wiki.flightgear.org/Multiplayer_protocol we can see that gears 0, 1, 2, 3, 4 are transferred automatically (just the `position-norm` and `compression-norm` properties).

So what I suggest is:

1. For the two big (rear) wheels of the amphibious, use `/gear/gear[3]` and `/gear/gear[4]`. That will fix the 

2. For the two front gear of the amphibious use `/gear/gear[5]` and `/gear[6]`. We need to alias these in `c172p-set.xml`. The model xml files should thus use something like `/engines/engine[9]/n1` and `n2` for example.

3. Try to get rid of `gear/gear[?]/spray` and just alias `sprayspeed-ms` and `spray-density`.

onox:
I'll fix item 3 for you (the `spray` properties)

onox:
@wlbragg I have fixed the tyre spray over MP. You can do the gear amphibious position stuff + any other remaining effects. You were talking about off road smoke?

onox:
Btw, I don't know how the amphibious gear `/gear` interacts with certain properties in JSBSim. For example, I don't see how JSBSim know the position of the pontoon gear (retracted or down). So, I don't know if you need to make changes in the FDM or in files in `Systems/`.

Maybe @andgi knows

andgi:
@onox I have not changed how the pontoon landing gears work. As far as I know they are "normal" JSBSim landing gears but are also moved up and down (with the aircraft variant) by physics.nas and possibly also c172p.nas to get them out of the way when they should not be present at all.

onox:
But how does JSBSim know that the pontoon landing gear are at `/gear/gear[19]` to `/gear/gear[22]`?

andgi:
What do you mean by "know"?
It knows their indices based on where the contact elements are in c172p.xml, where it is also set whether they are landing gears or contact points and brakable or not.
Some of the properties in the /gear/gear[x]/ and in /fdm/jsbsim/gear/unit[x]/ or /fdm/jsbsim/contact/unit[x]/ property subtrees are tied together by FG's JSBSim interface (JSBSim.cxx).

dany93:
@onox 
Internal Properties, look at /fdm/jsbsim/gear/unit[N]/z-position
N = 0, 1, 2:  tyre gears (default or bush wheels)
N = 19 to 22: amphibious (front and rear)
In physics.nas, wlbragg sets the z-value for each gear.
When a set must be used, its z-position is different of zero (< 0)
When not used, it does not "disappears", but its z-position is = 0. For the FDM, it no longer has contacts with the ground (or the water, for no drag).

wlbragg:
>You were talking about off road smoke?
That one is already finished and working.
I know this because I broke it twice trying to get the spray working (dumb mistakes).

I just looked over your commits and you went down the same path I took (even using the end of engines array 8-9), but I don't see everything I had tried to do, but yours obviously works and mine didn't . I am going to pull your PR and double check everything and see where we diverged, I have my changes memorized by now because I did it twice.

Thanks for you efforts.

wlbragg:
OK onox, that is why they pay you the "big buck".
I was on the right track but I think I "WAY" over thought this.

>you went down the same path I took (even using the end of engines array 8-9)

and using rpm for nose and n1-n1 for left and right, just seems logical doesn't it?

Because these engine props are automatically sent over MP what you did is all that was required?
But I don't understand how you are using the spray-density for the on-off condition. I could see it if it was combined in a function with the original gear/gear[0]/tyre-smoke property.
So what happened to gear/gear[0]/tyre-smoke?

wlbragg:
I just brought it up on MP and it is working perfectly, well done.
I really would like to understand though how you got rid of gear/gear[0]/tyre-smoke property?

wlbragg:
OK, looking again, you can't have spray without spray-density so you use it as a condition for off-on. But in my mind spray-density should be an assigned value, right off the bat it should have a user assigned value for how much density you want your spray to have, so what gives?

wlbragg:
>Is there a particular reason why these high indices (19 to 22) were used?

Because at the time they went into the system that was the next available contact positions and I didn't want to rework all that came before it.

wlbragg:
I give up, I'm spending hours on this stuff and getting nowhere.
I went back to basics and still can't get a thing with the pontoons.

For the front wheels I tried
```<engine n="6">
    <n1 alias="/gear/gear[19]/sprayspeed-ms"/>
    <n2 alias="/gear/gear[20]/sprayspeed-ms"/>
  </engine>
  <engine n="7">
    <n1 alias="/gear/gear[19]/spray-density"/>
    <n2 alias="/gear/gear[20]/spray-density"/>
  </engine>```

and in the model.xml

```<params>
        <tyre_spray_speed_ms>engines/engine[6]/n1</tyre_spray_speed_ms>
        <tyre_spray_density>engines/engine[7]/n1</tyre_spray_density>
    </params>```

I believe this is exactly what you did yet I get nothing.

wlbragg:
Before that I went through the whole procedure of making a function to combine float and wheels output to one prop to use in model.xml
When that didn't work I went backwards to see if it was something other than the function.

I also tried to figure out the gears, I am stuck on "mapping" gear 19-20 to 4-5 and
aliasing gear 17-18 to engine 6-7, I am sorry but it just isn't clicking and I am tired of spinning my wheels. I guess it is over my current level of knowledge.

onox:
What's gear 17-18?

onox:
I think 21 and 22 should be moved to 4 and 5, since 21 and 22 are the rear gear according to `Models/c172p.xml`.

andgi:
@onox contact points 13-18 are structure contact points for the floats, they are only used on land.
Some of them trigger sparks and smoke particle systems when the floats slide over solid ground.

onox:
@andgi Where are the gears defined in JSBSim? I see a lot of `<contact>`, but no `<gear>`.

andgi:
@onox They are all called <contact> but the type is BOGEY for gears and STRUCTURE for structure contact points.

onox:
Ah, and the position of the `<contact>` element defines which index it gets?

onox:
Then I guess gear[3] and gear[4] cannot be used because it uses indices 3, 4, 5 for damage mod contact points.

wlbragg:
>What's gear 17-18?

Me being totally overwhelmed :)

onox:
haha :smile:

wkitty42:
:lol:

onox:
Let me try to use `/engines/engine[6]` and `/engines/engine[7]` for the position-norm of the amphibious gear.

onox:
I'll alias `compression-ft` as well.

wlbragg:
OK,  that is right [3] and [4] are used for contacts, I never thought of that.
That was part of my trouble.
Keep in mind we should be able to function gear tires and pontoon tires together into a prop and use that new prop in model.xml?
We could have done that with gear tires and pontoon tires for sparks and off road smoke (at least for left and right) because they should sit really close to the same location (maybe).

onox:
I'm almost done with the MP properties for the positioning of the pontoon gear.

wlbragg:
>We could have done that with gear tires and pontoon tires for sparks and off road smoke (at least for left and right) because they should sit really close to the same location (maybe).

That is not right we "could do" it for off road "dust" with the amphibious gear but not spark and damage smoke, they probably would have required separate files anyway.

wlbragg:
See, that is why I wanted to bang my head against the wall, this isn't rocket silence, but I am just not getting it. As soon as I see what you did, and how simple it was, I'll probably jump off a cliff so as not to perpetuate my linage. "Not to take away from your knowledge and effort", only my stupidity!

onox:
Don't jump, we might need your help with other cool stuff in the future :stuck_out_tongue_winking_eye:

wlbragg:
onox, I guess we just got scolded on the forums by stuart, but to be fair he gave us an hour and took it back in 8 minutes, reminds me of a boss I once had! Thank you ludomotico for bailing us out!

wlbragg:
:clap:

onox:
Why does he want the stuff now? fgdata is going to branch on 17th of July, so I don't understand why he is in such a hurry.

onox:
Also, milestone is tomorrow. Stuart has to wait.

wlbragg:
I Nice work onox, I don't think I would have been able to figure that out.

I am clueless now on any more changes, are we to fix bugs now on 3.6 branch and add any new stuff here?

onox:
I'm just going to delete v3.6. Stuart will have to wait until tomorrow. Also, I don't see the point in branching now. We should all focus on 3.6, not on 3.6+3.8. Plus http://wiki.flightgear.org/Release_plan says we need to tag tomorrow and branch on 17th of July.

wlbragg:
And this didn't make it into 3.6 so is this a bug or feature?

onox:
Can you test this PR? We're just going to add it to master. Stuart is no position to negotiate.

wlbragg:
But what if he decides to punish and not put it in fgdata?

wlbragg:
Yes, testing now!

onox:
Then they'll have a crappy old version and angry user base.

onox:
:rage4: :rage:

wlbragg:
Thing of beauty!

onox:
Thanks :smile:

wlbragg:
Unless we are pushed here for time, you want me to take one more crack at the rain spray for pontoons? If so what approach do you suggest?

onox:
I think we should forget about the rain spray for the pontoons for now :cry:

wlbragg:
OK

onox:
> (What's the markup for verbatim? &lt; etc is getting old..)

@andgi The syntax is a one backtick \`. Code can be enclosed by three backticks \```. If you add "xml" or "javascript" immediately after the three opening backticks then you get syntax highlighting.

wkitty42:
have you guys seen the posts in the dev list or this topic on the forum? they seem to be related to using MP data slots for other than intended operations... like above where you guys speak of using open engine[x] slots to transfer data for the tyre speed... i think this needs to be reconsidered and to use the generic data slots available as mentioned in the dev list post linked below... apparently another craft, 707 or 747 or some such, is maybe doing something similar and their engines appear to be in flames when they really aren't...

related post on dev list via gmane: http://permalink.gmane.org/gmane.games.flightgear.devel/78471
forum topic: http://forum.flightgear.org/viewtopic.php?f=68&t=26648

i was asked, directly, who was using unused slots for transferring unrelated data and i did not answer that directly... this has the potential to be really ugly :cry:

onox:
Kitty, you're seeing problems that aren't there. We are already using all of the generic MP properties (`sim/multiplay/generic/`), but there are not enough of them. 

> I do not think you could use any of the 27 IDS
> for engines, 9 engines, 3 props... since these would
> be written to engines/engine[n]/n1, n2, rpm. Who
> exactly is using these?

Yes we can. These properties are in the sIdPropertyList[], thus they get transferred over MP. Does the c172p has 9 engines? No, it has 2 (160 HP + 180 HP), thus we can safely use almost all of these properties for other things.

The real problem is that FG just doesn't properly version aircraft. Can someone still fly a 10-year old version of an aircraft for FG 1.0? Probably yes. Is it going to look good on computers of people who run FG 3.4/3.5/3.6? Probably not.

If you want to be helpful, make sure that we get more generic ints and floats. Also, it would be nice to have at least 100 booleans. The silly XDR format currently uses 4 bytes (that's 32 bits) for a single boolean, even though a boolean needs just 1 bit to carry all its information. If the MP system does it properly you can have at least 128 booleans (8 bits * 16 bytes) compared to 4 boolean right now.

geoffmcl:
@onox, I do not think anyone is suggesting that a user can join MP with a 10 year old aircraft ;=() but why not...

The case in point, I was using release 3.4 as the remote, and latest 3.5 as the local. That's less that 6 months old... when I got the wing damage model view... as I now note others have seen...

All I would say is we do need to have some **CARE** about backward compatibility, especially for the **DEFAULT** aircraft. 

Would you not agree?

And I do not yet know the exact cause of the local view of a 707 with engines in flames when the remote, I think also 3.5 in this case, was sitting parked with engines off...

It may have nothing to do with your use of the (maybe too many) engine slots for other purposes... but that tickled my ear...

And while I would never use **silly** to decribe someone else's voluntary hard work, namely the XDR format of MP, it is indeed wasteful especially for BOOLEAN! And strings for that matter...

Also, extending the list of mp props ID/Values probably is not a hard job but I have read **nothing**, no RFC or hint on the FG dev list, or fgms. Only now, joining here, I get a big earful of what's wrong with MP ;=))

One immediate thought is why not pack your boolean bits into an INT? mp would faithfully carry the value, and set say `/sim/multiplay/generic/int[n]`, for you to pick up, split back to bits, and set other appropriate boolean props...

Anyway, please take care not to deliberately alienate others by what I would call harsh rhetoric ;=(( on any public list!

And note there is now Issue #336 which directly addresses this 'broken' wing problem over MP...

onox:
The c172p doesn't need any boolean currently, but it is probably very useful for other aircraft.

geoffmcl:
@onox stupid me i read your -
> Also, it would be nice to have at least 100 booleans.

I hope other aircraft can make their desires known, and I or someone would look into it...

wkitty42:
@onox 
> Kitty, you're seeing problems that aren't there.

it isn't me that's seeing this stuff... i don't do MP... it is others and now the reports are spreading over the forums about the problem :crying_cat_face: 

@geoffmcl is the main one on the dev list i've messages with on the list... i'm glad to see he has dropped by and i hope that he can help us get this fixed so that all craft will look as they are supposed to look no matter if they are v3.4 or v2.x... i won't mention the trouble others would have working on the project and trying to figure out why certain data slots are being used for different options than their intended purpose...

onox:
@wkitty42 It's not really something that can and needs to be fixed. It's like asking to make Python 3 backward compatible with Python 2.

The only problem that currently exists is that some users use a development version of the c172p on FG 3.5 and then expect everything to be alright when they meet other users who haven't upgraded yet (because FG 3.6 hasn't been released yet)

andgi:
@geoffmcl It is well known that the current MP protocol has many shortcomings and some very inefficient content encodings. It has been that way since close to forever. Changing, basically, anything in it has the drawback of breaking backwards compatibility - something that was last done back in ~2005 between 0.9.8 and 0.9.9 or, perhaps, it was between 0.9.9 and 0.9.10. An added complication is that reworking the MP protocol properly requires a substantial amount of concentrated (and rather thankless considering that it would be hard to reach a consensus on the requirements and breaking backwards compatibility is never popular with the users) effort. Additionally, there is the concurrent and long running effort to move to HLA instead, meaning that any work on improving the present system may well be wasted, anyway.

@onox There are ready components in Aircraft/Generic/DualControl/dual-control-tools.nas for encoding and decoding bits in MP enabled int properties.

wlbragg:
>Only now, joining here, I get a big earful of what's wrong with MP

I can't speak for onox, but I can say if you heard any mention of MP shortcomings from me, here and now VS anytime in the past, that would be because this is my first exposure to the FG MP protocol. I think we manage to get out of it what we needed though.

I am unclear as to why there would be a problem with other "non" c172p aircraft because we used any MP properties available to the c172p, whether we used an engine[9] prop for tires smoke should have no effect on any 707.

>apparently another craft, 707 or 747 or some such, is maybe doing something similar and their engines appear to be in flames when they really aren't.

Sound like they didn't have onox programming their MP properties! :smile:

wkitty42:
@wlbragg yeah, the 707 problem is likely unrelated to this project... the symptoms, though, are the same which leads me to believe, at this time, that they are doing similar by using other MP data slots to transfer the information...

wlbragg:
> the symptoms, though, are the same 

What symptoms?
If your referring to the "broken" original c172p over MP, that has nothing to do with us using MP properties that weren't being used and everything to do with not taking into consideration that the two aircraft might be used together over MP (at least on my part).

Or is there something else, "symptoms", I missed?

wkitty42:
the symptoms being that data sent over MP is making craft look strange when an unused slot is used to transfer data for something else in the craft...

all i'm saying is that it would appear that the 707 is using an unused slot to transfer something unrelated to that slot in the same way that we are... that is causing the 707 to appear with engines on fire in the same way that old c172p craft are showing with both wings broken...

no big deal... i do also understand the not taking into consideration part... i've done it myself many times over the last 50+ years :wink:

wlbragg:
>old c172p craft are showing with both wings broken

OK, I just wanted to make sure I was understanding the potential problem correctly.

FYI:
It is not the extra "unused" props we stole that are causing this issue. It is the fact that users are trying to mix old versions and new versions of the exact same folder/model together under MP, it will never work "as is", it wasn't designed to. Can it be made to, probably, but to what cost in time and effort. We went into this about halfway through knowing this was replacing the original, when have we ever continued to supported the original when it was being replaced?
I guess that is why I am confused by why we are even discussing this farther, do we do this kind of backwards compatibility for anything in FG? 
I guess this is a little different because it is the default aircraft, but not really. We shouldn't even attempt to mix the old with the new, as far as I am concerned.
The way I see it, if you want to fly the "old" c172p, you don't do it in MP and you don't do it in versions that it has been replaced by the "new" c172p (3.5+).
I came to that conclusion and opinion the second we were told to name the aircraft the same as the original, that is why at the time, I insisted on getting that question answered, because I knew it would make a difference in how we moved forward with this upgrade.

wkitty42:
i think something is being misunderstood... the problem is that we, running fgfs 3.5, see older <=3.4 c172p craft with broken wings on our screens when they show up in our viewing area... i don't know what those older fgfses see when we show up... depending on our configuration, we may be floating if we have bush gear or pontoons on...

i understand the reason for the regression version and i fully agree with the reasoning...

i also understand not supporting older versions but this isn't about that... it is about how those old craft appear to us on our screens... we're not looking to support them... only to view them properly when they are in our view range...

onox:
Please continue the discussion in #336 if you want. This is not the right place.

wkitty42:
understood! :wink:

PR #312, opened 2015-06-18
Amphibious rear wheel animation

wlbragg:
I need some help with the positioning of the rear wheels for spin. I have a wobble and I don't know how to get rid of it.


Models/c172p.xml
```diff
@@ -5626,15 +5627,15 @@
...
+		<x-m>0.9631</x-m>
+		<y-m>-1.2560</y-m>
+		<z-m>-2.0037</z-m>
 	</center>
  </animation>
  <animation>
 	<type>spin</type>
 	<object-name>RRwheel</object-name>
-	<property>gear/gear[22]/rollspeed-ms</property>
+	<property>gear/gear[22]/rollspeed-ms</property>B
```
onox: What's this "B" doing here?

onox:
Shouldn't the `rollspeed-ms` animations put before the positioning animations?

onox:
@wlbragg I notice that the upper metal part of the front gears don't really move up when the gear move over little bumps in the road. Maybe we could use http://wiki.flightgear.org/Tracking_animation to solve this.

onox:
I found this https://github.com/tomgey/blender-io_scene_fdm, but it requires the objects in Blender to be in a hierarchy.

onox:
I'll push a test branch in a minute.

fgfs-screen-141
fgfs-screen-142

onox:
Pushed the two commits to bug-311. Needs more work before we should open a PR.

PR #321, opened 2015-06-20
Fix damaged amphibious gear to match default

### LINKED ISSUE #320: Amphibious gear damage model doesn't match default model.
Move original "working" amphibious wheels over to pontoon-damage.ac

wlbragg:
...and fix broken rudder select logic.

This includes correcting a previous unknown bug that caused the rudders to disappear in certain select animation logic's.

Also I added the amphibious wheels, tarm, etc to bumbspec and specular effects, but it really doesn't seem to have any effect on these objects. But I didn't compare them in various lightning conditions either.

Fixes #320


onox:
I see the rudder disappears on the non-damaged pontoon:

fgfs-screen-145

Is that how it's supposed to be?

wlbragg:
No, that is supposed to be fixed, that is what was wrong with the original. I'll check it again, maybe I mess up the logic.

onox:
I was using your bug-320 branch.

wlbragg:
OK, I think I got it. Please check carefully!

onox:
Now the rudder never disappears. Intentional?

fgfs-screen-148

wlbragg:
The intention was that the rudders never disappear, so that is intentional.

Has to be one of the gear no matter what. I think I originally had just broken bush gear cause damage to the propeller, then added floats and that might be why they all ended up there.
So good catch, thanks.

This one is vexing me?

get a good propeller

not
 or
    left damaged
    right damaged
  or
not 

means any conditions have to ! be true

and
  not
     left damaged
  /not
  not
    right damaged
  /not
/and

means all conditions have to ! be true

This is the same?

I just checked your way in the sim and it appeared to work correctly.

PR #323, opened 2015-06-22
Corrected best glide speed for C172P

wezzer:
Best glide speed for C172P according to Pilot Operating Handbook is 65 knots. Updated all the speeds to the tutorial.


dany93:
Oh, thank you @wezzer, you are perfectly right!

PR #328, opened 2015-06-23
Bug 85

### LINKED ISSUE #85: Improve stall horn sound
improve stall horn sound, use this as reference: https://www.youtube.com/watch?v=4tLWIvnVnCQ#t=42

onox:
Fixes #85


onox:
@gilbertohasnofb Can you check the pitch (by listening) in FlightGear when stalling?

@dany93 I postpone the stall horn a bit if you have flaps. Is that ok? See https://github.com/Juanvvc/c172p-detailed/pull/328/files

gilbertohasnofb:
@onox the pitch seems fine to me. So let's just wait for @dany93 's answer and then we merge this :smile:

dany93:
@onox wrote
> I postpone the stall horn a bit if you have flaps. Is that ok? 

Yes, that seems logic. Difficult to know by how much it should be, but yes.

gilbertohasnofb:
So in that case, let's merge this!

wkitty42:
is there some way to increase the volume of the stall horn? i can barely hear it over the engine of the craft... in the sim i have to turn down the Effects slider and crank the others all the way up to be able to hear the ATIS and radio comms... the beeps from the markers and the AP are loud enough but the radio comms and the stall horn i have problems with...

gilbertohasnofb:
@wkitty42 Really? Last time I checked I didn't feel it to be too soft at all, it was quite decent. I will check again later.

dany93:
I was wondering too if the stall volume was high enough (my ears??)

@onox 
> I postpone the stall horn a bit if you have flaps. Is that ok? 

After thinking, I'm wondering if the flaps really increase the stall AoA. They lower the stall speed, yes.  But when you put the flaps down, you can pitch down. Remind that what we call the AoA is the angle from the wings without flaps, even if they are down.... When the flaps are down, the effective AoA increases, but what about the usual AoA at incipient  stall??
I cannot find a clear and only response in the Internet.

onox:
> After thinking, I'm wondering if the flaps really increase the stall AoA. They lower the stall speed, yes. But when you extend the flaps, you can pitch down. Remind that what we call the AoA is the angle from the wings without flaps, even if they are extended.... When the flaps are extended, the effective AoA increases, but what about the usual AoA at incipient stall??
I cannot find a clear and only response in the Internet.

@tigert @wezzer Could you comment on this?

wkitty42:
one of the things that i was working on was to not follow the stall-warning property directly... some of my research said that the stall horn stats sounding 5 to 10 kts before the critical stall speed is reached... AoA is just one factor... that's why i was looking at something dividing CL by CD as hinted at in the forum... i have something that is close to the stall warning in plot but it starts sooner and is a little higher in value but in my testing, i wasn't able to get the sound to start playing when using this... i've shelved it for now... it is nothing as complex as that used in bug-85... if you want to look at it, that's ok... i'll send it over some how...

wezzer:
Hi guys. Quick comment on stall angle of attack and flaps: In general the angle of attack in which the stall occurs is less with flaps extended. As flaps are extended the wing chamber (or curvature) increases and the wing produces more lift with same angle of attack. But the stall Aoa is reached earlier. Here is an illustration from Boeing: http://www.boeing.com/commercial/aeromagazine/aero_12/images/attack_fig2.jpg

But now, if Flightgear defines AoA with wing chord line without flaps, I cannot say for sure. I will take a look on this tomorrow with my old study books.

onox:
So I would actually need to make the stall horn kick in sooner than later (as it does currently).

PR #330, opened 2015-06-23
hydrodynamics: soften the behavior

dany93:
cf this issue https://github.com/Juanvvc/c172p-detailed/issues/326#issuecomment-114518061
For merging  this in the current 160 hp version.


onox:
I'm okay with 0.005, I have not tested 0.004.

PR #331, opened 2015-06-23
Add tooltip and menu item for showing/hiding yokes

### LINKED ISSUE #322: Create tooltip and GUI menu option for removable yoke
Create tooltip when hoovering above the removable yoke mount and also a GUI menu option for it. The menu option "Hide/show yoke" should not be inside the aircraft options menu (where we have engine type, repair, wheel type, etc.), but as a standalone option in the C172P drop down menu itself, which is typical for most aircraft in FG.

onox:
Fixes #322


gilbertohasnofb:
Thanks!

PR #333, opened 2015-06-23
Tuned the friction of the float contact points.

andgi:
It is now possible to land on a concrete runway without nosing over as
the FAA seaplane operations handbook indicates should be possible without
severe damage. There is currently no damage modelled for this float
maltreatement but it could be modelled by increasing the float leak rate.

Signed-off-by: Anders Gidenstam <anders@gidenstam.org>

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -336,8 +336,8 @@
             <y>-56.889</y>
             <z>38.5</z>
           </location>
-          <static_friction>0.8</static_friction>
-          <dynamic_friction>0.7</dynamic_friction>
+          <static_friction>0.6</static_friction>
+          <dynamic_friction>0.5</dynamic_friction>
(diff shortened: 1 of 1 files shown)


onox:
Thanks! :+1:

wlbragg:
Float leaking is something I was hoping we could do.
When you say "Currently no damage modeled for this float maltreatment ", you mean FDM wise as in "float leaking", correct?
Not that there isn't any visible physical damage, that it should have already.

andgi:
The FDM has support for float leakage which is activated when the damaged or broken properties are set, but it has two problems: i) it doesn't take into account that floats are partitioned; and ii) I'm unsure how to make "repair" empty the float again (I think it will stay filled).

A soft landing on concrete would create more subtle damage, not as visible as the "damaged" state, but introducing slow leakage when in the water - the sort of "looks fine today - sunk tomorrow morning" issues.

PR #337, opened 2015-06-24
Update help text to reflect this is the core c172p

stuartbuchanan:
This is just a trivial change to some of the help text that no-longer applies now this replaces the fgdata c172p


onox:
@stuartbuchanan ehh.. I just see you created a PR for version-3.6 branch. But we're actually still using master for bugfixing, so you just target that branch next time.

onox:
I'll cherry pick your commit and add it to master.

PR #340, opened 2015-06-25
Bug 338

### LINKED ISSUE #338: Add back panel only variant
Add back the high resolution panel-only display for multi-computer setup from the old c172p.

onox:
Fixes #338


gilbertohasnofb:
Tested, but found it a bit buggy. Sometimes I press P on the panel view and the 2D panel does not show. Then I have to press c once for it to appear (but the sound stops) and then c again for the sound to restart. Also, when this 2D panel is being shown, sometimes P makes it disappear but does not make it reappear back. Last thing: when the 2D panel is being shown, if you right click and move the mouse (that is, move the viewpoint), it immediately disappears.

onox:
Yes I know about those issues. Problem with pressing c is that it sets `/sim/current-view/internal` to false.

If you look at `keyboard.xml` in fgdata, you see that pressing P toggles `/sim/panel/visibility` and c toggles `/sim/current-view/internal` among other things.

gilbertohasnofb:
Well, in that case I will merge this. Should we open a (some) new issue(s) for these bugs I mentioned or that's related to how FG works rather than our plane?

gilbertohasnofb:
And one more thing: when I merge something via GitHub's interface, I always get a `You're all set—the bug-338 branch can be safely deleted.` Should I delete branches after merging or should we keep them?

onox:
>Should I delete branches after merging or should we keep them?

That depends if you expect some additional fixes or not. Most often you want to delete branches otherwise you're get a very long list of them.

For this branch, please don't delete it yet, I'm looking in the C++ code to see how FG uses `/sim/panel/visibility`.

gilbertohasnofb:
> Most often you want to delete branches [...] For this branch, please don't delete it yet,

:+1:

onox:
Okay, I found it, you need to set `/sim/panel/visibility` to true and then set `/sim/current-view/heading-offset-deg` to 0.

PR #341, opened 2015-06-25
Set heading offset to 0 when toggling panel

### LINKED ISSUE #338: Add back panel only variant

onox:
Fixes #338. See PR #340 for the buggyness of the P key binding.

The panel does not show itself if /sim/current-view/heading-offset-deg
is not 0. Therefore, add an additional <binding> to set it to 0.

You only need to press P, there is no need to press c. First time you press P you need to wait a little because FlightGear needs to load the panel.

### DIFF
--- a/c172-help.xml
+++ b/c172-help.xml
@@ -20,7 +20,7 @@
   <line>For checklists, see under Help->Aircraft Checklists</line>
   <line/>
   <line>To show the high resolution panel for a multi-</line>
-  <line>computer setup, press P and then c.</line>
+  <line>computer setup, press P.</line>
   <line/>
   <line>Takeoff: no flaps, full throttle, rotate at 55 KIAS</line>
   <line>Climbout: no flaps, full throttle, 80 KIAS</line>
--- a/c172p-keyboard.xml
+++ b/c172p-keyboard.xml
@@ -19,6 +19,20 @@
 
 <PropertyList>
 
+    <key n="80">
+        <name>P</name>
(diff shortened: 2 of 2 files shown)


onox:
@gilbertohasnofb Can you test this? :smile:

gilbertohasnofb:
@onox absolutely! Just a minute.

gilbertohasnofb:
Yes, now pressing P does the trick. The only thing is that the panel still disappears if I try to move the view (right click and move mouse).

onox:
@gilbertohasnofb You want me to force the heading offset to 0 while `/sim/panel/visibility` is true?

gilbertohasnofb:
@onox I think that would make sense, right? If someone is using that panel then he or she is certainly not panning the view around the 3D cockpit.

onox:
I just tried to force it to 0, but it seems you can still make the panel disappear (intermittently) while dragging the mouse. The mouse is fighting with the `<filter>` to set the heading offset they each want. Which means I would need to start overriding some of the mouse bindings. Also I don't know how this works with multiple screens/computers. So it's not worth it for now.

gilbertohasnofb:
Fair enough, @onox. That's a minor thing anyway as I can't imagine someone attempting to pan while having the 2D panel.

onox:
Yeah, those who actually use this panel can always open an issue if it really bothers them.

gilbertohasnofb:
Absolutely.

Issue #293, opened 2015-06-14
Fix attitude indicator and heading indicator


onox:
Investigate if we can use the trick with using a `<filter>` which is enabled if `/sim/freeze/replay-state` is false to passthrough the values `/instrumentation/attitude-indicator/indicated-pitch-deg` and `/instrumentation/attitude-indicator/indicated-roll-deg` to two new values.

The instruments and the flight recorder must then use these two new values.

Currently, the values are not recorded because the flight recorder seem to lose when fighting with the actual instruments in FlightGear. The result is that during replay the instruments are operating and mostly display the correct attitude. However, if you crash or go back to when starting the engine, it seems the attitude indicator instrument doesn't work 100 % of the time in replay mode.

onox:
I'm not sure whether I should commit the fix for this issue. I can work around it by using `/instrumentation/heading-indicator/indicated-heading-deg` in the instruments and then adding a `<filter>` that connects `/instrumentation/heading-indicator/indicated-heading-deg` to `/instrumentation/heading-indicator/indicated-heading-deg-actual` if `/sim/freeze/replay-state` is false.

The work around does work (it shows the correct pitch/roll/heading during replay), but it feels kinda hacky. Furthermore, the instruments show the correct state 9 out of 10 times after ending replay mode, but sometimes the instruments are different (I tested by doing barrel rolls and crashing and then entering replay mode).

This is (I think) because the instruments (AT and HI) are still active and changing their state even during replay. If they would fully compute their state from the properties in the property tree, then this wouldn't be a problem. But they only use some properties (like the `spin` property which I have already added to the flight recorder). And the rest of it is (probably) read from C++ variables.

I have fixed the AT and HI in the cockpit and HSI in the high-res panel. The ATI in the high-res panel doesn't use parameterization, which means I would have to copy a complete `.xml` file just to change some properties.

I'm not sure if there's code outside the c172p project that uses the indicated properties of the AT and HI, and using these `-actual` properties feels too hacky to use for the critical 6 instruments.

PR #342, opened 2015-06-25
Bug 273

### LINKED ISSUE #273: 160 HP engine is under-powered for pontoons and amphibious
As several people noticed, the 160 hp-powered c172p is under-powered, which makes it "lazy" at full load. Floats and amphibious worsen everything (added drag).
I tried to fiddle the propeller pitch angle, at no avail. The original one is not far from the best.

I have a test version with a **180 hp engine**.
On my c172-180hp branch https://github.com/dany93/c172p-detailed/tree/c172-180hp

 Easy for the engine.

The propeller is much more questionable. It is the result of compromise between climbing rate, short field takeoff, and cruise or max velocity.
For the propeller, I use the NACA 640 data http://naca.central.cranfield.ac.uk/reports/1938/naca-report-640.pdf. They worked rather well for DR400, DC3, Cap10B. But everything is never certain... And I refuse to cheat the values. I check their consistency, I (re-)calculate the efficiency when I extract them from the curves.

**Positive:**
- With default gear at full load (2400 lb), I have a 830 - 850 ft/mn climbing rate (80 KIAS).
- With amphibious at full load (2400 lb), 648 ft/mn (80 KIAS). In accordance with @tigert's data  https://github.com/Juanvvc/c172p-detailed/issues/269#issuecomment-110268962

**No so positive:**
- The propeller (76" diameter) does not reach 2700 RPM, which means that the 180 hp are not reached during climbing. But if I decrease the pitch angle, the max air velocity will be lower than I expect. My attempt with a 78" diameter propeller makes me feel that it is not the solution (remains to be seen better).
- Levelled flight 2000 ft, 2500 RPM: with standard gears115 KIAS (hardly higher than with 160 hp). With amphibious 110 KIAS (I would have expected less, and do not know what to do).

My reasoning is that, even if this propeller is still questionable, the result is better than with the 160 hp engine. More pleasant.

onox:
Fixes #273 

Make sure you have the latest updates from git via `git fetch`, after that do `git checkout bug-273`.

Please test that you can:

1. Manually start each engine

2. Exhaust smoke (make sure to enable particles) when starting an engine

3. Autostart each engine

4. Controls work (throttle, mixture, magneto's, starter) via keyboard, mouse, and/or joystick

5. Check the instruments (oil temperature/pressure, EGT) and systems (attitude and heading indicators)

6. Master switches

7. Press Ctrl+R and check if the replay is ok. Attitude and heading indicator might not work 100 %, but that's another issue (See #293)

8. Follow the checklists for each engine and check green/yellow states are as expected

9. Try to do some tutorials


gilbertohasnofb:
@onox Okay, I tested mainly everything and all seems to be well. Here are my comments:

1. Everything fine.

2. Shows on both engines both wish manual start and autostart.

3. Works fine.

4. Tested only with mouse and keyboard, all fine. Someone with joystick please test this.

5. Seems fine to me.

6. Work as expected.

7. I didn't see any problems with them.

8. Didn't test this.

9. Didn't try this.

I wanted to be able to do 8. and 9. but I really don't have the time for it right now. Also, I will be away for the next days, I should be back on Sunday or Monday. So good luck to you guys and see you back soon!

dany93:
I have tested 1 to 7. Everything works. (except I have forgotten  oil temperature/pressure, EGT, and my JS has no mixture control).
Huge work. Thank you, onox.

wlbragg:
@onox, go ahead and do the honors of merging it!
It is desperately needed and is way better than what we have now.
If there are any bugs, we still have time to find them, especially with more users using it.

onox:
I was just doing some of the tutorials. It seems KHAF is not in a very good condition :cry:

PR #346, opened 2015-06-26
Always have shadow volume if ALS is enabled

### LINKED ISSUE #148: New solid 3D shadow needed
We need a new solid 3D shadow. I finally confirmed why our 3D shadow is solid. It requires a bit of trans in the ac. However our shadow displays varying levels or shades of gray depending on the amount and type of object producing shadow. So the simplest one piece 3dshadow model we can make is what is needed. So I am putting it on the todo list.

onox:
Fixes #148 

* Remove the 2D shadow check box in the Aircraft Options dialog
* No more annoying messages about ALS after loading FlightGear
* Always enable the '3D' shadow volume when ALS is enabled
* Tuned the z-m and pitch-deg offsets a bit. The shadow is not exactly 2D, so it's difficult to get it on the ground as much as possible.


wlbragg:
@onox, please leave the check box to turn off the 3D shadow. I think the version check should be removed, I don't mind it defaulting to shadow on and I don't care if you remove the 2D shadow. But I do want the option to turn off the shadow, I don't like it on over water and i think other users may want the option to turn it off.

onox:
Ok, I'll add back the check box for the 3D option.

wlbragg:
If it was a perfect shadow and behaved as such I might not be as insistent on having the ability to turn it off. But IMO there are times when it just doesn't work and I turn it off.
Thank you!

Don't we use a version check for something else? My only concern with removing it is if someone tried to use the new c172p with anything < 3.5. I thought we checked for that.

>No more annoying messages about ALS after loading FlightGear

You should never see that message unless your trying to use it in a version that doesn't support it?

onox:
> You should never see that message unless your trying to use it in a version that doesn't support it?

ALS is always turned off here after starting FG, I need to turn it on manually.

@wlbragg But the master branch version is going to end up as part of fgdata 3.6, thus the check is quite useless. I think it's sufficient to mention in README.md that the master branch requires >= 3.5.

PR #350, opened 2015-06-29
gathering system file calls in c172p.xml

dany93:


dany93:
- c172p.xml : move 2 lines
- The bushkit weights had been split into two gears, but one (right 36") had been forgotten.

PR #364, opened 2015-07-02
preliminary, much work in progress web panel for testing

tigert:
Give this a try please, and let me know if it works!

I am trying to add more instruments but the idea is to have two monitors or computers and you can run the instrument panel on a separate one.


gilbertohasnofb:
@tigert I don't understand why this PR has some already accepted commits, such as `add yoke show/hide to aircraft menu`. Also, I believe the commit `added EFHF sticker, fix throttle labeling, some dirt` has been declined, as I remember you used a different font in the panel than me. I manually merged your modifications to the panel, so the font used in the labels stayed the same, and so accepting this PR will undo that work...

Also, I believe it's better to make PRs with fewer elements on it, as testing becomes very difficult. For instance, I know nothing about testing FDM, but I can test a lot of other things (menus, check textures and 3D objects, test features or procedures, etc.).

onox:
I'm afraid I have to close this PR for the single and only reason that the commit tree looks very unclean.

@tigert Please make a copy of `WebPanel/` and the changes you made to `gui/dialogs/c172p-menu.xml. After that please create a new fork of our repository and then add back `WebPanel/` of which you made a backup.

Issue #351, opened 2015-06-29
startup errors...


wkitty42:
using the latest master (i just updated to make sure) i'm getting these error messages... the one about line 25 has been showing for several weeks... the other one (repeated three times) is rather new...

Enabling ATI viewport hack
Failed to tie property fcs/ to object methods

In file /home/myuser/myflightgear/Aircraft/c172p/Systems/c172p-hydrodynamics.xml: line 25
      Property floats-enabled-norm is already defined.
Failed to tie property fcs/ to object methods
Failed to tie property fcs/ to object methods
Loading local weather routines...

Issue #348, opened 2015-06-28
Main gear damage is unrealistic


tigert:
While it is nice and adds the "oops" feeling to a heavy landing, the main gear in a c172 is very very strong and our damage modelling is actually unrealistic.

https://www.youtube.com/watch?v=O1yWkbJuqkE#t=420

We discussed this at the club and remembered this video from NASA Langley research center, which is scary, sobering and very much shows how strong a c172 is (to a point, like everything is).

See at 7 minutes and onward what the gear can take. We could model a burst tire though, and then change the drag on the contact point to make it skid?

dany93:
Several people (you among others, for the nose wheel if I recall well) found that the wheels were too hard at breaking.
Everything is in `Nasal/physics.nas` (lines 200 - 350, it depends on your version).
```
# GROUND DAMAGES

    force0 = get_gear_force(0, 1800, 600); # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for NOSE
    force1 = get_gear_force(1, 5400, 400); # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for LEFT gear
    force2 = get_gear_force(2, 5400, 400); # MUST be the same coefficients as spring_coeff and damping_coeff in the FDM for RIGHT gear

    gear_side_force = getprop("/fdm/jsbsim/forces/fby-gear-lbs");
    
#    # For tests: forces (LBS) exerted on gears (along Z and Y), uncomment these lines.
#    if(force0 > 1000) print("Nose Z-force =", force0); #future breaking forces for gears
#    if(force1 > 1200) print("Left Z-force =", force1); #1500 - 2000 lb seems plausible. Mind full load and cross wind landing
#    if(force2 >1200)  print("Right Z-force =", force2);
#    if(gear_side_force > 500) print ("left side-force =", gear_side_force);
#    if(gear_side_force < -500) print ("right side-force =", abs(gear_side_force));

	if(force0 > 1400)
		nosegearbroke();

	if(force1 > 2000 or gear_side_force > 1500)
		leftgearbroke();

	if(force2 > 2000 or gear_side_force < -1500)
		rightgearbroke();
```
force0 is for the nose (in lbs). Currently breaks at 1400 lbs.
force1 and force2 for each main gear. Currently 2000 lbs each.
If you know how much they hold, we enter the values. Very easy.
If you do not know how much, you change the values and do tests. Then tell us.

[...]

Issue #336, opened 2015-06-24
MP problems experienced when a remote user runs the old c172p model.


andgi:
Apparently (I don't use MP these days), there are some problems with the MP/AI model when a user with the new c172p(-detailed) receives data from a user that uses the old c172p model (e.g. using the current or a previous FG release).

It is possible to avoid some, but perhaps not all, of these problems by coding the conditions and expressions of the new animations defensively. If a required property is not sent its value will be nil, so if the nil value activates the normal non-damaged state and the default bushkit configuration the missing value would not be a problem. It's more difficult for MP properties that the new c172p uses for a different purpose than the old, but I don't know if there are many of these.

Since it is likely that there are quite a mix of FG versions on the MP network it is probably worth investing some time trying to remove some of the problems (e.g. that broken wings seems to be the default state).

An alternative and perhaps simpler solution is to attach some Nasal code to the c172p model that initializes all animation properties to a safe value. If the remote is a new c172p these values will be overwritten immediately, but if it is an old c172p the ones not sent will now have a safe value. It would still leave problems with MP properties that are used differently by both models, though.

onox:
I don't think it's really unavoidable. It's a one time pain that you have to go through. 3.8 will probably be less painful.

To give an example, a lot of aircraft that use registration numbers (aka "immatriculation") use (waste) six generic ints (int[14] to int[19). What we did is use string[0] instead with about [six lines of Nasal code](https://github.com/Juanvvc/c172p-detailed/blob/master/Models/c172p.xml#L61-67) in `<nasal><load>`. Not only is our solution not wasting six integers, it's also more flexible; it's not necessarily limited to six digits.

andgi:
Yes, some problems may be unavoidable but it is probably possible to avoid the the most visible ones, like broken wings. E.g. if the wing state is sent over a property/properties unused by the old c172p a few lines of Nasal in the model file could ensure that old c172p:s are not represented with broken wings.

onox:
We could add a version number.

andgi:
Yes, but wasting a MP property for that doesn't feel good. It might be enough to check the value of some property/ies that the new c172p sends and the old doesn't to separate them.

wlbragg:
I'm not sure we should try to stay compatible with the original c172p. I thought the whole idea was that this replaces the original.

@onox, if we are going to try to maintain the original along side the new, you and I probably need to do am MP session, one old one new and log the problems. Then see if it is practical to fix.

[...]

PR #365, opened 2015-07-03
Bug 280

### LINKED ISSUE #280: Tune g limits for wing damage
The current g limits might be too low and thus make it too easy to tear off the wings.

See https://github.com/Juanvvc/c172p-detailed/issues/134#issuecomment-110781583 and https://github.com/Juanvvc/c172p-detailed/issues/134#issuecomment-111212020 in #134.

onox:
Fixes #280, #336, #348, and #351. Additionally it:

* has more deterministic execution of damage logic
* fixes a small problem in 88d5db5
* frees up 5 MP properties
* removes the possibility to damage the aircraft when switching to a new bush kit
* the old code in `Nasal/physics.nas` had a problem that you could manage to get `settledelay` greater than 5, causing the code to continuously keep executing `poll_gear_delay()`

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -65,6 +65,19 @@
      setlistener(regnum_path, func (node) {
         set_registration_number(rplayer, node.getValue());
      }, 1, 0);
+
+     # Gear
+     rplayer.initNode("sim/multiplay/generic/int[6]", 0, "BOOL");
+     rplayer.initNode("sim/multiplay/generic/int[7]", 0, "BOOL");
+     rplayer.initNode("sim/multiplay/generic/int[8]", 0, "BOOL");
+
+     # Bush kit
+     rplayer.initNode("sim/multiplay/generic/int[9]", 0, "INT");
+
+     # Damage
+     rplayer.initNode("sim/multiplay/generic/int[14]", 0, "INT");
+     rplayer.initNode("sim/multiplay/generic/int[18]", 0, "INT");
+     rplayer.initNode("sim/multiplay/generic/int[19]", 0, "INT");
    </load>
 
    <unload>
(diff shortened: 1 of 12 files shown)


wlbragg:
I'm OK with merging this as is if we need to. We can come back to it later if necessary.

dany93:
Do you have these messages from the branch master:bug/280 ?
```
Failed to tie property fcs/ to object methods

In file /home/dany/FG/github/c172p/Systems/c172p-hydrodynamics.xml: line 25
      Property floats-enabled-norm is already defined.
Failed to tie property fcs/ to object methods
..................................
Failed to tie property fcs/ to object methods
```
The first ones are rather a question to @andgi
I found once that the other ones (10) can come from switches for which JSBSim do not like having no name.

andgi:
@dany93 The `floats-enabled-norm` warning is due to me having the habit of declaring the input properties for JSBSim systems in the respective system file. In this case the other system that sets the input property is loaded first and thus has already created the input property. The warning as such is harmless. Declaring the inputs is also required for use in JSBSim/standalone if the input property in FG is created by a JSBSim external method (e.g. the -set file).

OTOH, having no name or an already used name on a component is not a good practice.

onox:
I'll add names to the switches.

onox:
@andgi I notice that when I add name attribute to the `<switch>` components, that JSBSim alters the type (int becomes double for example). Any way to prevent that?

onox:
Never mind, it seems that if I use `name="right-wing-damage"` instead of `name="wing-damage/right-wing"` that JSBSim does nothing with it (which is what I want).

onox:
I have nothing more to commit, so please give your :+1: or :-1: :smile:

dany93:
OK.

wlbragg:
See bottom of #280

wlbragg:
Well done onox, lots of work here.

@dany93, are we closing #280 or are you still questioning the wing limits?

dany93:
>@dany93, are we closing #280 or are you still questioning the wing limits?

No, thanks, the changes that I proposed (and you merged) solved my concern.

PR #366, opened 2015-07-03
trying again :-)

tigert:
This should be less of a huge commit mess hopefully.


onox:
PR has to wait a couple of days until after 3.6 release. See https://github.com/Juanvvc/c172p-detailed/issues/339#issuecomment-118189846

onox:
@tigert Do you know about interactive rebasing? You can do `git rebase -i HEAD~2`. You will see two lines starting each with the word "pick". You can change the lower "pick" to "fixup". git will then meld that commit into the previous one. So you get one commit instead of two. After that you have to force push (once) though. You can read about it [here](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/conceptual-overview).

PR #368, opened 2015-07-04
Fix some usage of old engine properties

### LINKED ISSUE #367: GUI popup tip problems with 180 HP engine
Some code still assumes there's only one engine or use the wrong prefix (`/engines/current-engine/` instead of `/engines/active-engine/`).

onox:
Fixes #367. Please wait until user "someguy" on the forum has reported it fixes his problem.


onox:
User "someguy" reported it has been fixed in bug-367.

Issue #269, opened 2015-06-09
Seaplane operations reference data


tigert:
# Background

This is from OH-CTL "CAPRE Aqua Model 2200" floatplane conversion STC , which doesnt have a free license, so I won't put it out for download. Instead I will be quoting the relevant performance and operational points for you below.

## Limitations

  * Engine and Limits: 
    * Lycoming 0-360-A4A 180hp
    * Maximum normal operating power 2550 rpm (150hp)
    * Takeoff (five minutes) 2700 rpm (180hp)

  * Propeller and Limits
    * McCauley 1A200/DFA
    * Static rpm at maximum permissible throttle (ground testing I think):
      * Not over 2200 rpm, not under 2100 rpm
      * Diameter between 78 and 82 inches (overhaul sanding will eat metal away)

  * Powerplant instrumentation
    * Red Line at 2700 rpm
    * Green arc 700 rpm to 2550 rpm

  * Max gross weight floatplane: 2350 lbs
  * Center of gravity floatplane: 
    * 29.8 to 45.5 at 2350 lbs.
    * 36.5 to 45.5 at 1825 lbs or less
    * straight line variation between above points

  * No acrobatic maneuvers, including spins, approved.

  * Altitude loss in a stall recovery: 300ft

### Caution

  * Retract water rudders for fast taxiing, takeoff and landing
  * Floatplane max flaps 40 degrees (hm, isn't this max of 172 anyway? most recent 172s with higher MTOW have this limited to 30 degrees anyway)

(more to follow in a while)

tigert:
There should be a handle to retract/extend the water rudders, placed in the cockpit floor - need to find a photo of this for reference. It has a placard that reads "Water Rudder" and "Extended" in the forward edge and "Retracted" in the aft, and a double-headed arrow in between.

tigert:
## Emergency Procedures
  * No changes in float plane version

## Normal Procedures (suppolement to Section 4 of Pilots Operating Handbook)

### Before entering floatplane
  1. Inspect floats, struts, cables, pulleys, and fittings for dents, scratches, fraying cables, and attachments to the aircraft,
  2. Inspect each float compartment by removing the rubber caps on each standpipe and pump out any accumulated water. Reinstall caps. 
    * I don't know if we need to simulate that, 

### Before starting engine
  1. Water rudder operation - CHECK visually
  2. Water rudders - DOWN for taxiing
  3. Make sure you are CLEAR before starting engine

### Takeoff
[...]

PR #371, opened 2015-07-05
checklist: Add items for float planes

onox:
Add items from #269 to checklists for float planes.

@tigert @wezzer What should be the takeoff speed and landing speed for float planes?

### DIFF
--- a/c172-checklists.xml
+++ b/c172-checklists.xml
@@ -1868,7 +1868,7 @@
     </item>
     <item>
       <name>Airspeed</name>
-      <value>50-60 KIAS (flaps DOWN)</value>
+      <value>60-70 KIAS (flaps DOWN)</value>
       <marker>
         <x-m>-0.3608</x-m>
         <y-m>-0.3083</y-m>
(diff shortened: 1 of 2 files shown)


wezzer:
I believe the speeds are pretty much the same, but operating procedures are a bit different. @tigert has already provided the operating procedures here: https://github.com/Juanvvc/c172p-detailed/issues/269#issuecomment-110267638

onox:
PR can be merged if the items are OK.

onox:
@gilbertohasnofb Could you test this PR as well?

gilbertohasnofb:
Same question: can this branch be deleted or is there any other float plane operation related issue to be implemented?

onox:
I didn't implement the "Balked landing (go-around)" checklist (it would be one item) and item 3 and 5 of the takeoff checklist. They are bit useless because "the sim starts to jump the nose up and down" (in item 5) doesn't apply anymore; you could hold yoke full aft and just takeoff when airspeed becomes high enough.

wkitty42:
is there any way to limit these float craft checklist items so they only show then a float craft is in use?

onox:
Unfortunately no.

wkitty42:
ugh... that bites... oh well... it was worth asking about...

Issue #316, opened 2015-06-19
Ctrl+C shouldn't show certain objects as clickable


gilbertohasnofb:
Ctrl+C shouldn't show certain objects as clickable. At most their names, but certainly not as clickable:

- compass
- airspeed gauge's needle
- EGT gauge's red button (currently I believe it does nothing)

onox:
@gilbertohasnofb It seems a pick animation is used for the compass and airspeed in order to show a tooltip. Would you suggest to just remove the pick animations? The other instruments don't have them anyway.

gilbertohasnofb:
I don't know what is a pick animation, but I would suggest removing the yellow contours around the airspeed needle and compass, and maybe even from the EGT button as of now it doesn't do anything (AFAIK).

onox:
Removing the yellow contours means also no more showing the tooltips of the compass and airspeed indicator.

onox:
EGT's red button changes the `/engines/engine[0]/egt-bug-norm`. Click and hold and then drag your mouse up or down.

gilbertohasnofb:
> Removing the yellow contours means also no more showing the tooltips of the compass and airspeed indicator.

I don't think we need a tooltip for the airspeed indicator and compass as virtually no other gauge has it (we don't have it for the altimeter, vertical speed, heading indicator, tachometer, etc.). Or we should add it to all gauges. The point is to be consistent.

Also, are you sure the yellow contour is necessary for showing tooltips? I can see the airspeed tooltip when my mouse is hoovering above the airspeed indicator but NOT above the needle.

> EGT's red button changes the /engines/engine[0]/egt-bug-norm. Click and hold and then drag your mouse up or down.

You are right. In that case, that button should behave identically to a knob and this should be clearer (right now you have to click slightly to the right of the button to go up and slightly to the left to move it down). Also, the cursor should be the same as the other knobs, with two horizontal arrows such as :left_right_arrow:

onox:
It seems I can make the tooltip for the airspeed indicator appear by removing the "Needle" object from the pick animation. (Now it's applied only to the "Face" object)

onox:
I've decided to remove the tooltip for the ASI for the reasons you gave above.

gilbertohasnofb:
Okay onox, I will test the PR.

gilbertohasnofb:
Thanks onox.

wlbragg:
[...]

Issue #313, opened 2015-06-18
Create two bars across the cabin for the pontoon and amphibian variants


gilbertohasnofb:
Create two bars across the cabin for the pontoon and amphibian variants, such as seen in these images:

http://islandaero.com/images/70-172dbp.jpg

http://www.parkrapidsaviation.com/userfiles/image/IMG_4538%281%29.JPG

Possibly related to #206

tigert:
Not the perfect view but visible from different angle still: http://tigert.1g.fi/kuvat/MIK/2006-07-30-parnu/_DSC2644.jpg

tigert:
Looks like there is a flat metal bracket that the bolts attach to and which is bolted to the glare area of the fuselage?

 |__\\//__|

gilbertohasnofb:
@tigert Very nice, thanks for the additional information!

According to your picture, it looks like we will need to raise our compass as well as the two bars connect below it. To be honest, I always thought our compass should have some sort of mount below it.

Also, I suggest we only implement this after solving #6 and #118 as they may influence the position of the panel.

Finally, once this work is done I can create some nice textures for these bars. For instance, in the case of OH-CTL in the image tigert posted above, the bars have their painting quite scratched, but some liveries (such as the shiny 4X-CHV) could have them in better conservation.

onox:
http://tigert.1g.fi/kuvat/MIK/2006-07-30-parnu/_DSC2651.jpg :open_mouth: 

~~We~~ @thevirtualfer can add still much more detail to the model :smile:

gilbertohasnofb:
@onox I don't understand the connection between the topic and the picture you posted

onox:
There's no connection :stuck_out_tongue_winking_eye:

gilbertohasnofb:
Oh :smile:

onox:
Here's another picture of the two bars: http://tigert.1g.fi/kuvat/MIK/2006-07-30-parnu/_DSC2673.jpg

gilbertohasnofb:
Very nice! Now we can see better how the compass look like and how it's mounted in the cockpit!

wkitty42:
FWIW: i've seen these support braces in some craft's pics and not in others... i haven't paid attention if they are only floatplanes where i've seen them, though...

[...]

PR #375, opened 2015-07-05
Re-add tooltip to magnetic compass.

geoffmcl:
There was discussion in #313 and #316 that seemed to reach the conclusion
that this 'helpful' display is not used. It is...

In fact it should be used on each start-up, and perhaps during flight to
set and keep the gyroscopic compass 'accurate'...

Tried to read through the history of this change...

wkitty42 commented 16 days ago #313
https://github.com/Juanvvc/c172p-detailed/issues/313#issuecomment-113345108

mag-compass - why is it `clickable`? Because we need the tool tip to tell us what it is on!

Yes, you can not change anything - it is a magnetic compass - but without zooming in, and repositioning to read it is difficult to read, as it is in RL ;=))

And some of us fly **without** the hud ;=)))) so need that tooltip. And was very amused that others consider the magnetic compass as an `ornament` LOL!

onox commented 16 days ago # 316
https://github.com/Juanvvc/c172p-detailed/issues/316#issuecomment-113579784

So think this is a move in the wrong direction ;=((

A little off topic, but I also do not exactly know **WHY** all the mag-compass stuff was moved from `Instruments-3d` into the `c172p` folders. 

Yes, I can see maybe wanting to add this 'flashlight', but still why copy the mag-compass.ac, and mag-compass.rgb? These could have continued to live in the `common` folder...

I hope the idea will be to continue to try and **share** common instruments, where possible. That is why `Instruments` and `Instruments-3d` continue to exist in fgdata/Aircraft...

Anyway, this is mainly about restoring the magnetic compass tooltip to show its value to facilitate setting the gyroscopic compass before start, and perhaps re-set the gyro during flight if gyro drift is implemented...

### DIFF
--- a/Models/Interior/Panel/Instruments/mag-compass/mag-compass.xml
+++ b/Models/Interior/Panel/Instruments/mag-compass/mag-compass.xml
@@ -21,7 +21,8 @@
  </animation>
 
   <animation>
-  <type>material</type>
+  <!-- <type>material</type> set this if you do NOT want a tooltip -->
+  <type>pick</type>
   <object-name>Ring</object-name>
   <emission>
    <red>1.0</red>
@@ -29,5 +30,14 @@
    <blue>0.0</blue>
    <factor-prop>/sim/model//material/instruments/factor</factor-prop>
   </emission>
+  <hovered>
+    <binding>
+        <command>set-tooltip</command>
+        <tooltip-id>magnetcompass</tooltip-id>
+        <label>Magnetic heading: %3d</label>
+        <property>instrumentation/magnetic-compass/indicated-heading-deg</property>
+        <mapping>heading</mapping>
+    </binding>
+   </hovered>
  </animation>
 </PropertyList>


gilbertohasnofb:
@geoffmcl I just tested this. I am fine with bringing the tooltip back as it helps to read the compass. I am also fine labelling the compass when Ctrl+C is activated. What I dislike (and was part of the subject in those discussions) is the yellow contour around the compass ball when Ctrl+C is pressed: it gives the illusion that the compass is clickable and thus is confusing. Do you think you can remove this yellow contour but leave the tooltip and label in this PR?

geoffmcl:
@gilbertohasnofb no do not think the yellow contour is confusing... it makes it clear when you hover you just see `Magnetic heading: 116`... 

If a new user thinks that is an invitation to change where is magnetic north, then I think [s]he should read a little before entering an aircraft cabin... and try adjusting everything that turns ;=)))) sort of learn by touching, regardless of the consequences...

So I would not think to do this... but am also fine if someone else wants to add it later, as an addition... just please get back the tooltip...

gilbertohasnofb:
> no do not think the yellow contour is confusing... [...] but am also fine if someone else wants to add it later, as an addition... just please get back the tooltip...

Well, normally the yellow contour is around things that can be switched, clicked, moved, etc. It's not the case with the compass. Moreover, this has been discussed before and the conclusion was that it was confusing indeed, which is why we remove it. Simply reverting it back to get the tooltip working again will bring this issue back to where we were in the past, and I am not happy about it. I'd much rather have this properly solved at once. So: would you be able to remove this yellow contour from this PR yourself, so then I can merge it? If help is needed, we can as @onox to have a look on it and give us a hand. So then we have a single PR to bring the tooltip and label (but nothing else) back.

geoffmcl:
@gilbertohasnofb re-read most of the discussion and did not read that confusion was the reason to remove this magnetic compass tooltip... if you wanted a fix that it only displayed the tooltip and not the yellow contours, then that would have been the time to find it rather than removing a very much needed tooltip...

And certainly agree that removing the `screws` from the Ctrl+C display, as in current fgdata, was a very good thing... about 6 or 7 `picks` magically disappeared...

I know of no way to keep the tooltip and remove the yellow... on a `pick` or a `knob`...

So I certainly hope the magnetic compass tooltip is restored **before** the next merge with fgdata...

BTW do you know when that will be? An update minutes ago showed it still had duplicated liveries... possibly other file errors...

We are only 12 days from the fgdata freeze (except for bugs) stage, and it would be good to get more moved over soonest...

I guess there would be some noise from the dev-list if this magnetic compass did not have a tooltip to make reading it easy... that's all...

gilbertohasnofb:
> did not read that confusion was the reason to remove this magnetic compass tooltip..

You are misunderstanding me. I argued about the yellow contour, that's all. The thing is we ended up removing everything: yellow contour, label and tooltip. I agree 100% with you that the tooltip is something we should put back. So look, I will simply merge this and reopen an issue to remove the yellow contour of the compass, so we avoid discussing this ad eternitum, okay?

> BTW do you know when [the next merge with fgdata] will be? An update minutes ago showed it still had duplicated liveries... possibly other file errors...

I really don't know, it depends on Stuart, I guess. As for the duplicates and file errors, I can't understand why they are happening. They were instructed to delete everything from the c172p directory and replace with our repository. How can this be happening still is beyond me...

geoffmcl:
@gilbertohasnofb tar muchly... have fetched upstream and rebased... ready for the next bug ;=))

I guess someone needs to tickle Stuart, or other fgdata commiters, on the dev-list, to get this sorted... release time ticks forward...

And will keep searching for a way to have a tooltip without the yellow contours... you will hear me load and clear if I find something ;=))

Have put up a `blog` mainly about the c172p detailed, but have not yet published the page in my `index` - http://geoffair.org/fg/fgc172p.htm - Searching for a few more pictures to add... If you get a chance to view/read you will see I am quite **concerned** about the mp `broken` wings bug...

gilbertohasnofb:
Nice, I will check your blog. About Stuart, I tried contacting him some days ago about another issue with fgdata but got no answer so far, he seems to be away right now. 

> And will keep searching for a way to have a tooltip without the yellow contours... you will hear me load and clear if I find something ;=))

I reopened #316 for this reason, okay?

onox:
@gilbertohasnofb Could you create a new issue for the compass with a 3.8 release milestone?

gilbertohasnofb:
@onox You mean about the yellow contour or about what? The tooltip has been reintroduced with this PR.

onox:
The yellow contour thing (being able to have tooltips without requiring a pick/knob animation).

gilbertohasnofb:
@onox I will do that right now.

PR #381, opened 2015-07-08
Remove shadow from TrimWheel object, fixes #250

### LINKED ISSUE #250: trim wheel texture disappears
When rolling the trim wheel, the texture disappears and reappears at certain settings... this happens with all liveries... it is not limited to just one livery... the numbers at the points where the texture disappears or comes back are not the same across all liveries, though... I thought it might have been number related or maybe length/wrapping but I don't know now...

See issue #241 for some reference pics I posted... more can be posted if needed...

onox:
Fixes #250


gilbertohasnofb:
Tested under several weather conditions (and therefore different light conditions), it seems to be working fine.

Issue #82, opened 2015-03-13
save the state of the trim wheel


gilbertohasnofb:
Similarly to saving the switch and radio states, do the same for the trim wheel. Proposing behaviour:

- save the state of the trim wheel between sessions
- if autostarting (either via the menu or by unchecking the options "complex start up" and "start sim with engine running") then trim should be set to take off position, as to not complicate things for the beginners

onox:
@gilbertohasnofb Do we really want this? The flight controls (surfaces) aren't saved either.

gilbertohasnofb:
@onox Yes, I think it would be great. Flight controls are not relevant to be saved, but the trim is for the reason that the trim setting for landing is quite different than the trim setting for take off, and a pilot should always check this (it's part of the checklist). If we save the settings, we get a much more realistic procedure: just like in RL, we can't overlook where we last left the trim else the plane won't really lift.

dany93:
True. But can you let things easy for beginners? Those who press "S" and that's all...
e.g. have the autostart take a preset or the -set value.

gilbertohasnofb:
Well, I think we can have the autostart set the trim to the neuter position. As well, I already proposed to have an option in the menu which toggles saving the state between sessions or not (radios, knobs, toggles, trim wheel, etc.). To be honest, I think realism must be our main concern and if we are giving the beginners a chance to overcome the situation (e.g. auto start sets trim as well) then I am not really concerned if a beginner will screw something up.

gilbertohasnofb:
And the benefit for people looking for realism outweighs the necessity of the noobs, IMO.

dany93:
Do not forget that this aircraft is the first contact for FG beginners.
People looking for realism can find it, but you have already made very good efforts with the options to allow beginners use it quickly. I think it is important to keep this.

gilbertohasnofb:
I understand, Dany, but do you think that making the autostart automatically set the trim wheel isn't enough? 

This is what I think: if a truly beginner doesn't know how to handle the start up procedure, the levers and switches, etc., then he can use autostart and problem solved, the plane is ready to fly. If someone else bothers to learn how to properly handle the plane, then he has a realistic atmosphere. Both are happy like this. I think that catering for beginners is fine, but there is a limit for that: I really dislike the idea of making the plane less realistic for someone who truly wants to experience realism just so that a beginner can fly it without autostart (even though he also doesn't know what a trim wheel is).

dany93:
[...]

PR #382, opened 2015-07-08
Bug 285

### LINKED ISSUE #285: Save and restore state of switches
last time i was in the sim, i shut down everything and ensured that all switches were in the OFF state... i'm not sure where this problem is but i just loaded the sim with the same craft and location as last time and the battery, alternator, and avionics switches are all ON instead of OFF where they had been set... the starter switch is also set to BOTH for the magnetos... that was also specifically turned off... additionally, the mixture knob is pushed all the way in to 100% instead of being pulled out to 0% where it was also left...

onox:
Fixes #285 and #82

### DIFF
--- a/Nasal/avionics.nas
+++ b/Nasal/avionics.nas
@@ -83,7 +83,11 @@ aircraft.data.add(
     "/controls/lighting/landing-lights",
     "/controls/lighting/nav-lights",
     "/controls/lighting/strobe",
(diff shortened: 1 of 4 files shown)


gilbertohasnofb:
I will test this now.

gilbertohasnofb:
Can we delete this branch? (sorry @onox, but I'd rather always ask about it every single time than to delete some important branch without knowing it!)

PR #388, opened 2015-07-09
Fix rotation of frequency knobs of KX-165 instruments

### LINKED ISSUE #374: Radio knob bug
When switching between the stand by frequency and in use frequency, I believe that the knob that selects the frequency shouldn't move. In our plane, it seems that each position of the knob correspond to a specific frequency (so it moves when the switch happens), but in real life I am almost 100% sure that the knob does not move as it's position is not tied to a frequency (it's only the number of clockwise or counter-clockwise clicks that matter). See: https://vid.me/FSBD

@tigert @Juanvvc could you please confirm if I am correct about the knob's behaviour?
@onox I didn't add a milestone because I am not sure how much work would be to fix this and if we would make it in time for the 3.6 release.

onox:
Fixes #374 

* Check that knobs' positions are restored after restart of FG
* I choose about 4 deg per click (90 clicks per 360 deg). Kinda arbitrary
* Clockwise (mouse scroll up) is increasing, counter-clockwise (mouse scroll down) is decreasing


onox:
@gilbertohasnofb Could you test this PR?

gilbertohasnofb:
===Main panel===
EGT: state saved
NAV1 radial: __NOT__ saved
NAV2 radial: __NOT__ saved
ADF heading: __NOT__ saved
Heading indicator's heading bug: __NOT__ saved
Heading indicator's heading offset: __NOT__ saved
Altimeter Hg: __NOT__ saved
Attitude indicator's horizon offset: __NOT__ saved
Instrument light: state saved

===Radios===
Nav1: state saved
Nav1 volume: __NOT__ saved
Nav2: state saved
Nav2 volume: state saved
Com1: state saved
Com1 volume: state saved
Com2: state saved
Com2: state saved
ADF: state saved
Transponder setting: state saved
Transponder numbers: state saved
DME switch: state saved
DME N1 HLD N2: __NOT__ saved

Also found a bug while testing this: if the Com1 volume is set to 0, then both Com1 AND Nav1 are turned off. Same for Com2 and Nav2. Turning the volume of Nav1 to 0 does nothing to it, same for Nav2.

* * *

About scrolling, some knobs are still inverted. They are: heading indicator (both knobs), altimeter, Nav1/2/Com1/2/ADF volume knobs, transponder numbers and setting, ADF radio knob, instrument light, microphone setting (COM1 COM2 EXT),

onox:
@gilbertohasnofb I only fixed the four dials of the KX-165 instrument. Could you create a separate issue for the other dials (the state saving + inverted scrolling problem) and then merge this PR?

onox:
> Also found a bug while testing this: if the Com1 volume is set to 0, then both Com1 AND Nav1 are turned off. Same for Com2 and Nav2. Turning the volume of Nav1 to 0 does nothing to it, same for Nav2.

I think this is how it's supposed to work. The volume knob of the comm radio is used to turn on/off the instrument.

onox:
I'll see if I can save the volume of the nav1 and nav2 dials. Hang on a second... :smile:

onox:
I added saving the state of the two volume dials of the NAV radio's as well. You can merge the PR.

gilbertohasnofb:
>  Could you create a separate issue for the other dials (the state saving + inverted scrolling problem) and then merge this PR?

Yes. I will test this PR once again then I will do as you say.

> I think this is how it's supposed to work. The volume knob of the comm radio is used to turn on/off the instrument.

Are you sure about it? I thought the volume of the com1 controls only the com1 and NOT the nav1. Else why would we have a volume knob for nav1? So I thought that com1 with volume 0 turns off only com1, and nav1 with volume 0 turns off only nav1.

gilbertohasnofb:
Tested, now all KX-165 knobs are saving their states except for the DME mode (N1 HLD N2) knob. Wouldn't you like to save this as well before we merge this PR?

gilbertohasnofb:
> Could you create a separate issue for the other dials (the state saving + inverted scrolling problem) 

See #390 and #391

wkitty42:
On 07/10/2015 08:14 AM, Gilberto Agostinho wrote:
>     I think this is how it's supposed to work. The volume knob of the comm radio
>     is used to turn on/off the instrument.
>
> Are you sure about it? I thought the volume of the com1 controls only the com1
> and NOT the nav1. Else why would we have a volume knob for nav1? So I thought
> that com1 with volume 0 turns off only com1, and nav1 with volume 0 turns off
> only nav1.

i think he's talking about the on/off part of the switch... it is actually a 
separate set of mechanical contacts in many cases... closed provides power for 
one or more devices and each device then uses its own volume control so that 
each one can be set to their own level of loudness...

i looked real quick for a schematic of such but wasn't able to find one in a few 
minutes... i have several such volume controls with switch contacts here... the 
volume being an interior potentiometer (variable resistor) enclosed in the 
control's metal housing... the switch being two contact points mounted on the 
back of the housing and activated by a cam lug on the shaft... when the switch 
is turned, the cam lug pushes the two contacts together completing the 
circuit... one side of the switch contacts carries positive power to the switch 
and the other contact may have two or more wires to carry the power to other 
circuits when the switch points are making contact...

onox:
> Wouldn't you like to save this as well before we merge this PR?

Not really, DME != KX-165 :wink:

gilbertohasnofb:
@onox 

> Not really, DME != KX-165

My bad! This was added to #390 then

@wkitty42 Well, I am still confused. Why would the volume knob of com1 have anything to do with nav1? And why the volume knob of nav1 would NOT turn nav1 off when set to 0?

wkitty42:
On 07/10/2015 01:12 PM, Gilberto Agostinho wrote:
> @wkitty42 <https://github.com/wkitty42> Well, I am still confused. Why would the
> volume knob of com1 have anything to do with nav1? And why the volume knob of
> nav1 would NOT turn nav1 off when set to 0?

personally speaking, i don't know... i was only explaining how the idea works if 
the previous explanation of the COM1 switch controlling the NAV1 power is 
true... looking at them, they appear to be one unit and only one one knob is 
marked as power...

see this close up image
   http://www.pilotshop.com/catalog/graphics/11-00644.jpg

from this page
   http://www.pilotshop.com/catalog/avpages/kx16514.php

and this one from bendixking...
 
http://www.bendixking.com/Products/Navigation-Communication/Glideslope-Receivers/KX-165

gilbertohasnofb:
@wkitty42 you are absolutely right, I see now how only the COM knob has the "off" label and how both COM and NAV are in a single unit. So our radio works as expected and I was just being ignorant, thanks for the messages!

wkitty42:
@gilbertohasnofb you're ok, man! no longer ignorant in this, either, since you learned something :wink: :smile_cat:

gilbertohasnofb:
:+1:

gilbertohasnofb:
About knobs: I don't know if this is a bug or not, so I am asking here first. The outer dial in the COM and NAV radios selects only the decimals, and they do not carry on for the units (that is, from 105.95 we move to 105.00 and not 106.00). The same is not true for our ADF knob, but it doesn't work with decimals (the outer changes 1 KHz and the inner 100 KHz). Is this the proper behaviour of these knobs?

onox:
Knobs only change either the MHz part or the KHZ part. About the ADF, do I need to change the outer and inner knob? See https://github.com/Juanvvc/c172p-detailed/issues/369#issuecomment-118611834 and https://github.com/Juanvvc/c172p-detailed/issues/391#issuecomment-120503093

onox:
@gilbertohasnofb The outer changes 100 KHz and the inner does 1 KHz. So it seems to be correct.

gilbertohasnofb:
@onox Okay, thanks for the clarification. So it seems to be correct.

PR #393, opened 2015-07-10
[WIP] web panel files for post-freeze.

tigert:
I wanted to push what I have been hacking on. Hopefully this stuff is useful and can be added after the freeze is over. It's still work in progress and I will be updating things when I get a chance.


onox:
:warning: Don't merge this pull request until after v3.7 has been tagged (monday), otherwise I'll have to force push :warning:

tigert:
Actually lets put this on hold, I should do something that is feature complete first and then we can include it, interested folks can try my fork anyway, so it should go in once it has the basic functional set.

PR #402, opened 2015-07-11
Bug 400

### LINKED ISSUE #400: Consistent line endings and formatting
It seems we use a mixture of LF line endings (Linux and OS X) and CRLF (Windows). We also use a mixture of tabs, 1, 2, 3, and 4 spaces. This makes the code chaotic and uglier than it needs to be. It also creates ugly diffs making reviewing changes more difficult. Tabs also look very wide on Github.

I suggest we go with just LF and 4 spaces. Editors like notepad++ are able to use LF and insert spaces when you press the tab key. For the line endings I would like to add the following `.gitattributes` file:

```
* text eol=lf

*.png binary
*.jpg binary
*.rgb binary
*.wav binary
*.blend binary
*.xcf binary
*.zip binary
```

I can fix the `.xml` files with `XMLLINT_INDENT="4 spaces here" xmllint --format --output FILE FILE`. I won't change the `.nas` files because that needs to be done manually and is just too much work for now.

onox:
Fixes #400 

* Use LF line endings for text files
* Formatted `.xml` files with 4 spaces indentation using the `xmllint` tool

I have processed most of the files, except those in:

* `Models/labels/` 
* `Models/Interior/`
* `Models/Immat/`
* `Models/Pedals/`
* `Models/Liveries/`
* `Models/Effects/`
* `Models/Human/`

I'll fix those remaining files during 3.8. `.nas` should use 4 spaces as well. There's not really a formatting tool for this, so just get rid of those tabs while you working on some `.nas` files.

How to test:

1. Do the usual `git pull`

2. Then `git checkout bug-400`

3. And then `git log -w -p`. Note that you need to use `-w`. It should show only minor changes (changed comments and removal and insertion of blank lines)

4. Just check that you don't see any crazy things.


gilbertohasnofb:
Ok, I went through the whole log and all seemed fine to me, except for one single modification:

- inside the `<electrical>` block in Systems/systems.xml you exchanged `<path></path>` for `</path>`, is that correct? There is no `<path>` before that `</path>` statement in that file!

onox:
`<path></path>` has been replaced by `xmllint` with `<path/>`, not `</path>`. That's perfectly fine.

gilbertohasnofb:
I see. So all is fine, thanks!

gilbertohasnofb:
Just learned about `<tag />` in xml, didn't know it substitutes `<tag></tag>`.

onox:
@wlbragg @dany93 FYI Please use 4 spaces instead of tabs from now on to keep things consistent, otherwise those diffs are going to look ugly again :disappointed:

dany93:
I  do not insert "tab" since a long time ago just because of that.
My Geany indentation is configured to add "4 spaces" even when I press tab

wlbragg:
That's probably mostly coming from me and, yes I will watch for that and make sure I use spaces.

onox:
> My Geany indentation is configured to add "4 spaces" even when I press tab.

Ok, good :smile: I wasn't sure and since you also do some of the programming, I just included your name, just in case.

dany93:
I see what you mean, in physics.nas.  But all that I had written (damages on wings and gears, related to the FDM), you have transposed it in .xml code.

Issue #354, opened 2015-06-30
180 engine not registering on HUD


wlbragg:
The throttle setting in the HUD does not move when in the 180 engine condition, it only moves when using the 160 engine.

onox:
Yeah about that HUD... it's generic, it's default, it's ugly, it's only needed if an aircraft doesn't have a nice cockpit. So... can we not just disable that thing?

dany93:
Even if it doesn't have realistic justification, it easily gives some information (GPS coordinates, ground velocity, altitudes if I recall well). These can be useful in special cases, like development. I people do not like it, they just have to not set it.

wlbragg:
Sorry onox, didn't mean to bring up a sore subject. You know what the only thing I use it for is, checking where the throttle is set at.

wkitty42:
please **don't** disable this generic HUD... i cannot fly very well without it... no matter how detailed the panels are, i have to zoom in to see them but with the HUD i can easily tell what's going on... i'm sure that i'm not the only one doing this, too... i use it with all craft i fly... zooming in and back out and trying to get my view back to normal is a huge PITA...

tigert:
The HUD is likely the thing that everybody has their own favorite use for it. I verify that I have the correct QNH set (since it seems to auto-adjust it!) and also I use it to verify wind direction when practicing the water landings on the c172p float versions, since it shows the groundspeed (I can align myself to the waves and check with groundspeed whether I am going downwind or upwind. Lazy things, yeah. But sometimes handy... :-P

wkitty42:
eagerly awaiting this fix... there's only a few more days (three or four if i'm counting properly) and it definitely needs to be in the 3.6 release or there's gonna be a lot of complaints when folks choose to use the 180HP engine...

gilbertohasnofb:
I found a fix to this, but it's probably not very elegant. See the PR above.

onox:
The fix doesn't work for 160 HP engines and I find it a bit hacky.

How about we just display the throttle and groundspeed on the screen and then just disable the HUD? Like @andgi did in his hydrodynamics branch?

That way we don't have this ugly HUD which people aren't supposed to use (we have a beautiful cockpit) (so far only our keyboard-only user wkitty42 complained about it) but you can still see the throttle and groundspeed.

gilbertohasnofb:
Fixed with PR #407.

onox:
[...]

PR #407, opened 2015-07-12
180 HP controls the throttle of both engine[1] and engine[0] properties

gilbertohasnofb:
I found a fix for the bug #354 , but it's probably not very elegant (though it works!). The thing is that the throttle displayed on the HUD is always from property `/controls/engines/engine[0]/throttle`, but the 180 HP engine currently controls only `/controls/engines/engine[1]/throttle`. If we set it to control the throttles of both `engine[1]` and `engine[0]`, then the HUD display shows the correct throttle position and the engine also behave normally (as changing `engine[0]`'s throttle rightfully does nothing to the 180 HP). 

@onox please feel free to decline this fix if you think it's too much of a hack, okay?

### DIFF
--- a/Systems/engine.xml
+++ b/Systems/engine.xml
@@ -332,7 +332,7 @@
             <property>/controls/engines/current-engine/throttle</property>
         </input>
         <input>
-            <value>0.0</value>
+            <property>/controls/engines/engine[1]/throttle</property>
         </input>
         <output>
             <property>/controls/engines/engine[0]/throttle</property>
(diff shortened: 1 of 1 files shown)


onox:
This doesn't work.

onox:
@gilbertohasnofb Please try my fix for both 160 HP and 180 HP engines.

gilbertohasnofb:
Thanks @onox, now it works perfectly!

PR #419, opened 2015-07-14
co-pilot view, pilot model now depends on weight

### LINKED ISSUE #416: Create Co-Pilot View
Create Co-Pilot View

gilbertohasnofb:
Creates a Cockpit Right Seat View. For this to work well, I also made the pilot model dependent of its weight (so that one can set 100 pounds of weight to the right seat and less than 100 pounds to the left seat and have an empty pilot chair). 

Fixes #416.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -110,12 +110,14 @@
             var co_pilot = rplayer.initNode("pax/co-pilot/present", 0, "BOOL");
             var left_passenger = rplayer.initNode("pax/left-passenger/present", 0, "BOOL");
             var right_passenger = rplayer.initNode("pax/right-passenger/present", 0, "BOOL");
+            var pilot = rplayer.initNode("pax/pilot/present", 0, "BOOL");
 
             setlistener(pax_state_path, func (node) {
                 var state = node.getValue();
                 co_pilot.setBoolValue(bits.test(state, 0));
                 left_passenger.setBoolValue(bits.test(state, 1));
                 right_passenger.setBoolValue(bits.test(state, 2));
+                pilot.setBoolValue(bits.test(state, 3));
             }, 1, 0);
 
             # Make sure human occupants are always visible in remote aircraft
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -5220,7 +5220,7 @@
         <path>Human/male-copilot.xml</path>
         <offsets>
             <x-m> 0.9 </x-m>
-            <y-m> 0.2 </y-m>
+            <y-m> -0.2 </y-m>
             <z-m> -1.45 </z-m>
         </offsets>
     </model>
@@ -5229,7 +5229,7 @@
         <path>Human/male-copilot.xml</path>
         <offsets>
             <x-m> 0.9 </x-m>
-            <y-m> -0.2 </y-m>
+            <y-m> 0.2 </y-m>
             <z-m> -1.45 </z-m>
         </offsets>
     </model>
@@ -5269,7 +5269,7 @@
         </condition>
     </animation>
 
-    <!-- Co-pilot not visible if not enough weight on his seat or if uing Cockpit Right Seat View-->
+    <!-- Co-pilot not visible if not enough weight on his seat or if using Cockpit Right Seat View-->
     <animation>
         <type>select</type>
         <object-name>Copilot</object-name>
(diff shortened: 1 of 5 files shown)


gilbertohasnofb:
@onox I hope this is the right time for a PR with new features since our master has been tagged as `3.7` already.

onox:
You need to update `Models/c172p.xml` and `Nasal/c172p.nas` as well for the pilot to be visible over MP.

gilbertohasnofb:
Okay, I will do that!

gilbertohasnofb:
Done and I am about to commit, but shouldn't I also change https://github.com/Juanvvc/c172p-detailed/blob/bug-416/Nasal/c172p.nas#L209 ?

gilbertohasnofb:
@onox Just ignore that, I obviously need to update those lines as well. I am committing these modifications now, but I'd still like to to work on the rear window views before testing and possibly merging this.

gilbertohasnofb:
Passenger views added. Also, on lines 5223 and 5232 of `Models/c172p.xml`, the right and left passenger models were at inverted positions (that is, the model called Right Passenger was located on the left seat and vice versa)! To fix this, someone inverted the passenger models on lines 5294 and 5304, which made it all more confusing, so I took the liberty of fixing this mess :smile:

@onox this is ready for testing

onox:
Can you set the view numbers to 100, 101, etc.?

onox:
The helicopter view is gone.

onox:
@gilbertohasnofb I don't how you fixed (if fixed) the LeftPassenger and RightPassenger problem. The left passenger object is called RightPassenger and I don't see you changed the Blender and ac files.

onox:
@gilbertohasnofb Could you move the `<view>`'s to a separate file? Also you cannot use view number like 1, 2, 3, etc. because you will override the standard views. So use something like 100, 101, etc.

See https://github.com/onox/V-22/blob/master/Sim/views.xml

You would have to include it in `<PropertyList>` in https://github.com/Juanvvc/c172p-detailed/blob/master/c172p-set.xml#L19

In the new c172p-views.xml file you can then include the walker. Same trick:

```xml
<PropertyList include="Aircraft/Generic/Human/Include/walker-include.xml">
```

gilbertohasnofb:
> Can you set the view numbers to 100, 101, etc.?

Will do that.

>  Could you move the <view>'s to a separate file? 

Sure, I will do that. 

> In the new c172p-views.xml file you can then include the walker. 

I will have a look if I understand this and try to implement it, okay?

> I don't see how you fixed (if fixed) the LeftPassenger and RightPassenger problem. The left passenger object is called RightPassenger and I don't see you changed the Blender and ac files.

The model position is defined in `Models/c172p.xml`, such as in:

    <model>
        <name>LeftPassenger</name>
        <path>Human/male-copilot.xml</path>
        <offsets>
            <x-m> 0.9 </x-m>
            <y-m> -0.2 </y-m>
            <z-m> -1.45 </z-m>
        </offsets>
    </model>

So I fix it by changing the `y-m` and my tests showed that everything we all right. Which problems with it are you still seeing?

onox:
> The model position is defined in Models/c172p.xml, such as in:

Ah, ok, that explains it.

gilbertohasnofb:
@onox 

> You would have to include it in `<PropertyList>` in https://github.com/Juanvvc/c172p-detailed/blob/master/c172p-set.xml#L19
> 
> In the new c172p-views.xml file you can then include the walker. Same trick:
> 
> ```xml
> <PropertyList include="Aircraft/Generic/Human/Include/walker-include.xml">

Hmm, that didn't work. What worked for me was to simply have `<PropertyList>` in the `c172p-views.xml` file and have the following in the `c172p-set.xml` file:

    <PropertyList include="Aircraft/Generic/Human/Include/walker-include.xml">
    
        <sim include="c172p-views.xml">

* * *

Also, there is a little problem with the walker. The pilot model (if present due to weight and "show models" option is on) is programmed to be shown in every view except in the following: if in the cockpit view, if in the walker view, if in the walker orbit view. That means that the pilot will be seating in the cockpit AND the walker will be outside if we chose another view while the walker is enabled (let's say model view, or cockpit right seat view). Also, the walker is male while the pilot is female. Are these things known bugs? Should I open new issues for them?

onox:
> Hmm, that didn't work

That's probably because you didn't add the `<sim>` tags to the `c172p-views.xml` file. But the solution you have chosen now will probably work as well and is maybe even better.

I'll look at the other little issues in your comment tomorrow.

gilbertohasnofb:
> That's probably because you didn't add the <sim> tags to the c172p-views.xml file. But the solution you have chosen now will probably work as well and is maybe even better.

Yeah, now that you mention it I completely understand what was wrong, but if you are satisfied with the result then I will leave it as it is.

> I'll look at the other little issues in your comment tomorrow.

Sure, take your time. If you want, I can create issues here in our repository for them. But they are not directly related to this PR, and this work here is basically done, so once you have time you can test it and merge if you like the result.

onox:
@gilbertohasnofb I have moved the co-pilot view to the last position, so that when you press Shift+V you can go from the pilot view immediately to the co-pilot view. If you're ok with that, you can merge the PR.

You can open new issues for the two issues with the walker.

gilbertohasnofb:
@onox 
> I have moved the co-pilot view to the last position, so that when you press Shift+V you can go from the pilot view immediately to the co-pilot view. 

Great idea!

> You can open new issues for the two issues with the walker.

I will do that straight away!

gilbertohasnofb:
See #423 and #424.

PR #436, opened 2015-07-23
Correct Linux absolute paths in ac file. Convert RGB to PNG

IAHM-COL:


Juanvvc:
Thanks! If I recall correctly, flightgear always loads the texture files from the local directory. This is why everything was apparently working correctly: the absolute paths were just ignored. In any case, this behaviour may change in the future and this was obviously a bug. Thanks!

IAHM-COL:
Thanks for the explanation Ludomo
You are probably right. It is intriguing thou: I have had cases that it
does not "simply works" out of the box.

One example; I've seen props lacking the blur or the texture by problems
like this.
Also, in rare occassions this can lead to verbose output errors indicating
that a texture is missing, where it is just misdirected.
while correcting these I found a few of exceptional cases where the texture
was to be found somewhere else in the Aircraft path: such cases --luckily
not the c172p ;) -- definitely the absolute path to someone else filesystem
does not work.

Where appropriate I also converted RGBs to PNGs.

Thanks for merging.

IH-COL

PR #449, opened 2015-08-29
Bug 117

### LINKED ISSUE #117: sound of engine and wind should be louder with door opened
The sound of engine and wind should be louder with door opened. Also, the same would be true if the opening of the windows is implemented (see: https://github.com/Juanvvc/c172p-detailed/issues/116)

onox:
Fixes #117 

* Louder stall sound
* Volume gets limited to 1.0 in Simgear, so reduced maximum volume


onox:
@gilbertohasnofb @wkitty42 Please check the sound. Note that this is all we can do for now. It would be nice to have better (individual) control over volume of various sounds, but that will require changes in SimGear by the core devs.

gilbertohasnofb:
@onox I will test this tonight and merge it if no one is able to do that before

gilbertohasnofb:
@onox Volumes seem better to me.

PR #451, opened 2015-09-01
Bug 450

### LINKED ISSUE #450: FATAL:'-' found in propertyname after ''
Following error occurs:

687474703a2f2f7332382e706f7374696d672e6f72672f7970716167793162782f6a75616e7676635f30352e706e67

because of `<reference>-.75</reference>` in `Systems/ground-effects.xml`.

onox:
Should fix #450 

I have also added some `<property>` tags to `Systems/ground-effects.xml`, to make things a bit more explicit and less depend on guessing in FlightGear's C++ code.

### DIFF
--- a/Systems/KAP140.xml
+++ b/Systems/KAP140.xml
@@ -250,8 +250,12 @@
         <name>static port pressure rate computer</name>
         <debug>false</debug>
         <type>derivative</type>
-        <input>systems/static[0]/pressure-inhg</input>
-        <output>autopilot/internal/pressure-rate</output>
+        <input>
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
I think it would be ideal if @thevirtualfer could test and merge this PR as he is seeing this error and I am not, so I have no idea how I would test it.

Juanvvc:
I'm also not seeing this error, but the pull request seems ok to me.

If the string "-.75" is OK on some systems but not others, it is a compatibility issue and, probably, the FlightGear core team must be informed.

PR #452, opened 2015-09-14
Fix straight-in-landing tutorial

### LINKED ISSUE #442: bug in the straight-in landing tutorial
Report by Thorsten:

> Would you kindly take another look at the in-air tutorial start (straight-in landing)?

> It seems the tutorial starts the engine in a weird state - throttle is set to go, but I needed to press 's' once to actually make RPM respond to it. It seems the starter is somehow still on and the tooltip cries for the primer, and after pressing 's' again I get a 'starter off' message and everything runs fine.

wlbragg:
Fixes #442 

I think this is all that is needed. Tested it using "Complex start up procedures" checked and "Start with engine running" not checked.

### DIFF
--- a/Tutorials/landing.xml
+++ b/Tutorials/landing.xml
@@ -41,15 +41,19 @@ You start the tutorial cruising at around 1000ft and 90kts. To begin your descen
         <set>
             <property>/controls/switches/magnetos</property>
             <value>3</value>
-        </set>
-        <set>
-            <property>/controls/engines/current-engine/throttle</property>
-            <value>0.4</value>
-        </set>
+        </set>		
         <set>
             <property>/controls/switches/starter</property>
             <value>true</value>
         </set>
+        <set>
+            <property>/engines/active-engine/running</property>
+            <value>true</value>
+        </set>
+		<set>
+            <property>/controls/engines/engine/use-primer</property>
+            <value>false</value>
+		</set>
         <set>
             <property>/instrumentation/adf/frequencies/selected-khz</property>
             <value>340</value>
--- a/Tutorials/landing.xml
+++ b/Tutorials/landing.xml
@@ -149,7 +149,10 @@ You start the tutorial cruising at around 1000ft and 90kts. To begin your descen
             <property>fdm/jsbsim/bushkit</property>
             <value>0</value>
         </set>
-		
+		<set>
+            <property>/controls/engines/current-engine/throttle</property>
+            <value>0.4</value>
+        </set>
     </step>
 
     <step>
--- a/Tutorials/landing.xml
+++ b/Tutorials/landing.xml
@@ -34,6 +34,7 @@ You start the tutorial cruising at around 1000ft and 90kts. To begin your descen
     </presets>
 
     <init>
+        <message></message>
         <set>
             <property>/controls/switches/magnetos</property>
             <value>3</value>
@@ -61,7 +62,6 @@ You start the tutorial cruising at around 1000ft and 90kts. To begin your descen
     </init>
 
     <step>
-        <message>Configuring the aircraft.</message>
         <set>
             <property>/controls/flight/flaps</property>
             <value>0.0</value>
@@ -106,10 +106,6 @@ You start the tutorial cruising at around 1000ft and 90kts. To begin your descen
             <property>/controls/engines/current-engine/throttle</property>
             <value>0.4</value>
         </set>
-    </step>
-
-    <step>
-        <message>Configuration finished.</message>
     </step>
 
     <step>
--- a/Tutorials/landing.xml
+++ b/Tutorials/landing.xml
@@ -35,10 +35,6 @@ You start the tutorial cruising at around 1000ft and 90kts. To begin your descen
 
     <init>
         <set>
-            <property>/controls/flight/flaps</property>
-            <value>0.0</value>
-        </set>
-        <set>
             <property>/controls/switches/magnetos</property>
             <value>3</value>
         </set>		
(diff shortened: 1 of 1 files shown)


onox:
You have some tabs in the code (line 53 and 56). Could you replace those with (4) spaces for consistency?

onox:
You seem to have fixed those annoying error messages. But the tutorial doesn't do the pre-start and doesn't set the throttle anymore. Isn't that what the tutorial should do? Have you tried to set those things?

If that didn't work, maybe we should do some post-init with Nasal manually?

wlbragg:
>Have you tried to set those things?

No I didn't test for any of that and wasn't aware of the necessity to set all those other conditions. I can see if that is possible.

Can you elaborate on the throttle setting though, it is suppose to be set? Like you discovered, the throttle seems to have a mind of its own, from the init I guess. I took out the line that tried to set it, I am not sure it was having any effect.
The tutorial immediately instructs the operator to throttle to 50%, I figured that would be up to the user to adjust at that point. But I can put that line back in and see if it did adjust it.

onox:
> Like you discovered, the throttle seems to have a mind of its own, from the init I guess.

Yes, that's why I was thinking of finding a way to execute some post-init code.

wlbragg:
OK, let me at least take a look at aircraft configuration props and see if it can "easily" be managed.

per Juanvvc
>I will be enhancing the tutorials and missions during these weeks.

I don't want to step on anything Juanvvc is planning on doing. Does this mean he was going to work on this?

onox:
I don't know. @Juanvvc What are your plans?

Juanvvc:
My plans are revisiting all existing tutorials and make a few new "tutorials" (actually, missions) I will work during the next few weeks in this.

Of course, you can merge any bug fix or make any modification in any tutorial and I'll build my proposals on yours. I'm just saying I will work on tutorials for the next weeks, in case you want to focus your efforts elsewhere :)

wlbragg:
@Juanvvc, @onox,
This last commit isn't finished or perfected but shows a path forward.
It appears that fg's init overrides the tutorial init. If we set certain conditions that are being overridden in the first "step" as opposed to the "init" section, then it changes. Test what I did. It sounds like it starts the engine twice because of the engine change over. I started my test with the 180.

wlbragg:
Throttle worked also adding it to firs step. We could even create a "first" step independent of the tutorial instruction to add the init resets, kind of a second init so to speak. Then step 2 would be the normal tutorial instruction starting point.

onox:
Interesting idea. It takes a couple of seconds though before I suddenly see the throttle being changed to 40 %. There's some sort of a delay. Can you try to move all this bushkit/engine stuff to a new (clean) `<step>` between the `<init>` and the first `<step>`? That way you have the "initialization" separated from the actual tutorial.

You might need to add a dummy `<exit>` condition:

```xml
<exit>
    <condition>
        <true/>
    </condition>
</exit>
```

Try first without it.

And try to remove those tabs, I can still see them in the diff on Github :smile:

wlbragg:
Will do, first I have to watch our presidential debate going on tonight so I know who not to vote for :smi

onox:
Aren't they all republican? Then it's easy to know who not to vote for :trollface: :runner:

wlbragg:
Wanted to reply to you last comment with an emo, but I couldn't find the OUCH emo!

OK, see if this will work. I originally added a message at the beginning of the step "Configuring the aircraft." But I didn't see any way to erase a message one it is rendered to the screen and the aircraft was configured. It doesn't take that long to setup the aircraft so in the long run I figured it wasn't necessary to show a message. I think it works reasonably well. I like it when starting with the 180 engine, the tutorial will execute the mechanics sound when configuring the engine, really cool. We probably should also add the repair prop just in case.

wlbragg:
Change my mind and added the message. Looked like it allows for one message per step? So I added a second step to indicate configuration is finished.
I'm not convinced the messages are needed, remove if you wish.

onox:
> I couldn't find the OUCH emo!

Don't listen to me too much, @wlbragg, you can vote for whoever you want :smiley: 

> So basically I think any saved param will need to go into the step.

Yeah, I think that is correct.

Personally I would just use an empty message. (If you don't add `<message>` then you see "Tutorial step 0")

One more thing: if I have everything turned off before I start the tutorial, then after loading the tutorial, everything is still turned off, and when I then turn on the battery, the starter tries to start the engine. Mixture isn't set either.

@Juanvvc If you want to use this branch to do a complete overhaul of the tutorials, that's fine with me if @wlbragg is ok with it.

wlbragg:
Works for me!

Juanvvc:
I think a comment in the code explaining why the aircraft is configured during the first step and not in the init block is necessary, for future references ;)

Ok, I'll use this branch to modify tutorials. This means we don't merge it yet, right?

wlbragg:
>This means we don't merge it yet, right?

Correct, unless there is an urgency for 3.6 bugs.

Also, since you plan to do more work here, I'll let you go ahead an clean this tutorial up along with the rest. If you need any help, don't hesitate to ask.

wlbragg:
>One more thing: if I have everything turned off before I start the tutorial, then after loading the tutorial, everything is still turned off, and when I then turn on the battery, the starter tries to start the engine. Mixture isn't set either.

I just caught what you were saying here, so basically, we still either need an autostart call or account for everything from a cold craft forward?

onox:
> I just caught what you were saying here, so basically, we still either need an autostart call or account for everything from a cold craft forward to where the tutorial takes control?

Yes. I'll close this PR for now. It can be reopened later by @Juanvvc.

wlbragg:
So what has caused this change where we now have to do more setup to all.props when we apparently didn't have to do quite this extensive amount before. I say this because of the original tutorial and its lack of complexity.

onox:
Because we didn't bother to properly initialize the plane? Would you prefer to do the tutorial overhaul in a new branch? (So we can quickly merge this branch with maybe 1 or 2 extra changes)

wlbragg:
Yeah, we can do that. I take it you mean then we could go ahead and get this one merged with what we have already?

onox:
Yes. If you agree, could you remove that empty step that just has a message + give the init step an empty message (`<message></message>`) (instead of `<message>Configuring the aircraft.</message>`)? Then you can merge it.

wlbragg:
OK, we still need to add all the possible aircraft configurations that may be changeable to put the craft into the best possible profile during the start of the tutorial. But at least this gives us a path forward.

PR #457, opened 2015-10-06
Move and correct pontoon normal and reflect mapping.

wlbragg:
Please merge this to avert a conflict when we merge the skis later.


wlbragg:
As far as I can tell, this last commit should be the proper way to implement the external reflection and transparency shaders. glas object was never added to the effect so I added it. I am unsure how much of an effect, if any, it is having. I tested it at length and have inconclusive results.

I would appreciate if everyone could visually compare these shader changes to the way the existing layout renders. Let me know if there is any noticable difference.

Pay close attention to
propblur
external glass reflection

onox:
Do these `landinglightcover*` objects have their normals point inward? Should they even have the `glass` effect attached to them? We should attach `model-combined-transparent` to all the glassy/transparent objects that are visible from the external views, and attach `glass` to the inner cockpit glass that is visible from the cockpit views. Someone should check the normals in Blender.

wlbragg:
>Do these landinglightcover* objects have their normals point inward?

No they are pointing outward.

xml~~~
<inherits-from>Effects/glass</inherits-from>
~~~
Or actually "glass.eff" as it is being used in this context is a local file and it is "model-combined-transparent" (meant for external reflection and other). Look at the file Effects/glass.eff

"glas" the object, is the external glass layer and should inherit this also.

onox:
Okay, that is causing the confusion. What about renaming `Effects/glass.eff` to `Effects/glass-exterior.eff` (`model-combined-transparent`) and `Effects/glass-interior.eff` (`glass`). That would be less confusing I think.

AFAIK currently we have two glass layers. One of them is is used for the interior glass effect and the other is used for the shadow. The exterior glass effect is currently not used for the actual cockpit glass (one of the two layers). It seems we need three layers or ask Thorsten to somehow combine the interior glass effect with the shadow effect.

wlbragg:
>Okay, that is causing the confusion. What about renaming Effects/glass.eff to Effects/glass-exterior.eff

The only thing glass.eff was being used for was glas (front and back windows) and the external side windows and it wasn't really doing anything noticeable because I think internal-glass-shadows was overriding it.
According to Thorsten, what I did with interior-glass-shadows was never an intention of his when I did it.

> we have two glass layers. One of them is is used for the interior glass effect and the other is used for the shadow. The exterior glass effect is currently not used for the actual cockpit glass (this would require a third layer). It seems we need three layers

That's correct and there was also an effect glass.eff that was trying to use model-combined-transparent on glas which wasn't working because interior-glass-shadows took precedence.

I don't think we should keep the interior-glass-shadow unless we do a third layer. I would rather have the nice external reflection. So that is what I did.

One thing I noticed and Thorsten will have to clarify, I used model-transparent on the external glas object (and sides) it gave you a really nice reflection on the front window. The reflection showed the round heater vent tubes when looking down and at an angle.
Using model-combined-transparent you lost that reflection but gained the sky-cube reflection. I thought the model-transparent looked way better.

Long story short, I am more confused than ever about what is available and what works with what. But as far as I am concerned it is all cleaned up and working the way we want with the exception of loosing the interior-glass-shadows.

Part of my reasoning for dumping the interior-glass-shadows was, it wasn't necessarily a planned feature by Thorsten and it really didn't show that well except in certain lighting conditions. I think we could add a 3rd glass layer is we want it back.

I also moved all four bumpspec files out of Effects and into to a new folder called "Effects/external", Effects/internal's counterpart.

wlbragg:
@onox, alright, I add a 3rd glass layer to the cockpit and added the interior shadow back in. We can keep it, get rid of it, test it or blow it up, your call.

You can still merge "correct-pontoon-norm-and-reflect" as I added the 3rd glass layer to "skis" branch for testing.
I know better than to even suggest this, but we could add a 4th layer for the sky cube (to go with the nice model-transparent reflections). I would still like to ask Thorsten or anyone that might know, can you get the nice model-transparent reflection and the model-combined-transparent skycube together at the same time on one layer.

wlbragg:
>One thing I noticed and Thorsten will have to clarify, I used model-transparent on the external glas object (and sides) it gave you a really nice reflection on the front window. The reflection showed the round heater vent tubes when looking down and at an angle.
Using model-combined-transparent you lost that reflection but gained the sky-cube reflection. I thought the model-transparent looked way better.

Without using model-combined-transparent I was seeing the interior reflection from the outside. So I will put back model-combined-transparent and get the sky-cube back.

onox:
Now I'm even more confused: I see the interior glass effect (fog), but not the shiny reflection of the exterior glass effect. Even stranger is that the normals of both "glas_interior" and "glas" are pointing outwards. But if that's the case, then the exterior glass effect should work and the interior one shouldn't.

Am I missing something? :confused:

wlbragg:
>Even stranger is that the normals of both "glas_interior" and "glas" are pointing outwards. But if that's the case, then the exterior glass effect should work and the interior one shouldn't.
Am I missing something?

Yes, the same thing I am missing. If you point the normals on interior to the inside, you loose the rain and fog. I tried it, I thought it made sense to have them pointing inward also, but it wasn't how it worked. The exterior reflections need adjusting or something. I agree, they don't seem to be there are at the least aren't very noticeable. But if you look closely and at just the right angle you can see some of it, I think?

So yes,maybe this could stand some more scrutiny. But I can say we haven't lost anything (other than the interior glass shadow that we were never really supposed to have). We have gained the exterior "falling rain" effect or animation.

Suggestions?

wlbragg:
Here is the visual difference between shadow
shadow
 and external rainfall.
rain

wlbragg:
Oops, wrong button.

thevirtualfer:
Falling rain. My vote.

PR #467, opened 2015-10-19
Skis

wlbragg:
See latest commit for status. I'm happy with it and DONE!


onox:
I did some quick testing:

* Movement of ski's appears to be ok :smile: 

* But the wheels are spinning too fast :sob: 

@wlbragg I only did a quick look, but are you duplicating things? I think I also saw extra properties for the roll speed of the wheels. Isn't it easier to just pretend it's a 26" bushkit + ski's? That way you don't have to duplicate any logic/blender stuff for the wheels.

wlbragg:
I am using the 36" wheels + skis. Any additional code is for the ski wheels which can be slowed independent of the 36" wheels.

wlbragg:
>But the wheels are spinning too fast

Which wheels?

wlbragg:
@onox, don't know if this is worth it or not. You decide. If we keep it I'll delete all the commented out trash.

onox:
Hmm.. it's just that when landing, the wheels look like they spin faster than the aircraft moves on the ground. Can you reproduce that?

Instead of `<factor>` you can use `<expression>`.

wlbragg:
I'm testing now, but in the meantime, specifically which gears are you referring to and in which bushkit configuration?

wlbragg:
@onox, I pushed another change, added flight recorder and MP for the new wheel rotation filters. I also adjusted the front wheel rotation speed. Now you should be able to check the speeds easier using the recorder. I didn't test the MP yet and don't know that I accounted for all of it.

Please test the wheel speed when you get a chance (this applies to anyone). Just make sure you report specifically which bushkit config and which wheel as they are all different now.

onox:
I meant the big main wheels of the ski's configuration.

onox:
Now I see you have added [`/engines/engine[10]`](https://github.com/Juanvvc/c172p-detailed/blob/12b80dbaae5793b4ec6663a6cbff5175b6ecec55/c172p-set.xml#L539-L542). But yeah, as I said before, try to (re-)use `/engines/engine[7]`. Also, `/engines/engine[10]` won't be transmitted over MP. See http://wiki.flightgear.org/Multiplayer_protocol

Sorry for being such a pain in the *** :stuck_out_tongue_winking_eye:

wlbragg:
I think your getting confused, engine10 is for the small wheels on the actual ski itself. Engine7 is the default/26"/36" wheels. I am still using engine7, but I use it as input to a function where I get the products of it and a wheelspin factor from a switch statement. I output that product to the model.xml's wheel spin animation. I chose this approach because you can't change the factor in the animation statement on the fly. But I can make my own rotation speed x factor in a function and pass it to the model animation. That way each wheel size can have its own spin speed using just a single animation statement. It's calculated in the function.

onox:
> I think your getting confused on what I did and why. Engine10 is for the small wheels on the actual ski itself. Engine7 is the default/26"/36" wheels.

OK

Have you tried using `<expression>`? Because the current `Models/c172p.xml` file won't work correctly over MP (because you're using local properties).

wlbragg:
>Because the current Models/c172p.xml file won't work correctly over MP (because you're using local properties).

But I could reconfigure that so it was an MP alias, right?


>Have you tried using <expression>?

OK, I'll see if I can figure that one out.

wlbragg:
Alright then, it is finished, for sure, no matter what, I'm done. :smile: 

Really, it should be ready to go. Everything is accounted for. MP works, recorder works. I used engine7 for default/26"/36" wheel spin along with "expression" to do the factor for the different configurations (works like a charm), you know your tools well onox!.
I used rotors/main/blade[0]/position-deg for the little wheels on the skis. 
Bottom line, I didn't have to use the 4 extra generic floats.
It took forever but I really am happy with the results of this mod.
Please merge when you feel confident it wont blow something up.

onox:
Alright then! I'll do the merging :smile:

onox:
@wlbragg I have done the merging, but there was a merge conflict in `Models/c172p.blend`. Please check which version (ski's or the glass fix) went into master. If it was ski's, you may need to fix the glass again.

wlbragg:
@onox, I'm not sure what caused the conflict. The only permanent change to the blend should have been changing "glas" to single side and that is there so I think we are good to go.
Sorry for the ugly history, lots of work revision and testing in that merge.
Thanks for all the help, I feel really good about the outcome, your direction is the main reason.

gilbertohasnofb:
@wlbragg what do you think about a new splash screen for the ski variant? I can take care of that if you all like the idea.

wlbragg:
I think that would be great!

onox:
@wlbragg The conflict is because it is binary file and it was modified in both branches.

When I do `git log --oneline Models/c172p.blend` in `skis` branch I get:

```sh
520387d Remove interior glass shadows, make "glas", "leftwindow", "rightwindow" single sided to allow external rain visual.
7485f86 Make "glas" object single sided to allow "external" rain to be visible in cockpit view mode.
1d43781 Add 3rd glass layer, add interior glass shadow.
721f6f7 Black texture back side of propeller
```

In the "Move and correct pontoon normal and reflect mapping" I get:

```sh
911e115 Make "glas" object single sided to allow "external" rain to be visible in cockpit view mode.
e061b25 Make "leftwindow", "rightwindow" single sided to allow external rain visible
721f6f7 Black texture back side of propeller
```

It seems in the `skis` branch you reproduced what you did in the other branch + added a 3rd layer of glass.

onox:
> Sorry for the ugly history, lots of work revision and testing in that merge.

Next time, for a feature branch (`bug-xxx`) you can do [interactive rebasing](https://www.atlassian.com/git/tutorials/merging-vs-rebasing/). It is very useful to squash commits or reorder them. For example, you can do `git rebase -i HEAD~10` to edit the last 10 commits. When you choose to "edit" a commit, git will drop you back to the console. You can then make modifications, do `git commit` to commit your changes, and then do `git rebase --continue` to continue to the next step. At all times you can do `git rebase --abort` if you want to cancel the rebasing.

So for example, if you create a commit, but then notice some mistakes, you can commit your fix like `git commit -m "Bla bla bla"`. And then do `git rebase -i HEAD~2` and modify the last commit to "fixup". Git will then meld the "Bla bla bla" commit into the previous one and throw away the useless commit message :smile: 

Be sure to play in a new git repository (`git init`) if you have never done interactive rebasing before.

gilbertohasnofb:
@wlbragg cool! I will do that and propose something here.

wlbragg:
>It seems in the skis branch you reproduced what you did in the other branch + added a 3rd layer of glass.

That sound right, then I decided the 3rd layer was unnecessary and it should have been deleted somewhere. Bottom line what we should have ended up with is a single sided "glas", "leftwindow" and "rightwindow" and NO 3rd glass layer.

gilbertohasnofb:
@wlbragg @onox Do you guys like some of these options here?

a

b

c

d

e

wlbragg:
I like them all. If I had to order them I guess it would be
#5 but the aircraft as close up as in 1 or 4.
#1 
#2
I think #2 would be better if it were just about to touchdown on a snow covered land (not airport) in the pines

I know, now I'm being picky.

@gilbertohasnofb, go fly around PAJN Juneau set to winter. Not with the airport in the picture (unless it is in the distance). But with the river and the mountains in the background.

onox:
I like 2 and 1.

gilbertohasnofb:
Hmm, okay, so what about some of these?









I personally like 1 and 2 from the previous batch, and 6 and 9 from this one. Let's just try to agree on one of them.

wlbragg:
1 and 2, I like to see the skis sitting in the angled up position and also the closeup.

gilbertohasnofb:
@wlbragg do you mean 1 and 2 of the newer set of pictures (overcast), or 1 and 2 from the first batch (open skies)?

@onox How do you think these new images compare to the previous ones? Which one would you pick for a splash screen?

wlbragg:
1 and 2 from the first batch (open skies).

gilbertohasnofb:
Ok, in that case I will go with no. 2, as we all three like it. Here is how it looks:

splash9

@onox Where should I push this now that the skis have been merged? To the master?

wlbragg:
Nice, thank you!

onox:
I think you can directly add it to master this time.

gilbertohasnofb:
@wlbragg thanks!
@onox Ok, done.

onox:
@gilbertohasnofb Don't you need to edit `c172p-set.xml`? :open_mouth:

gilbertohasnofb:
Oops, I surely do. Just pushed it.

onox:
:+1:

PR #468, opened 2015-10-19
Heat air refactor

wlbragg:
I added a way for users to adjust the factors that control the frost and fog. You can now make it however you want. If someone gives us better settings than we currently have, at some point we could potentially remove the option.

### DIFF
--- a/Nasal/weather.nas
+++ b/Nasal/weather.nas
@@ -34,10 +34,59 @@ props.globals.initNode("/environment/aircraft-effects/surfacetempC", airtempC, "
 props.Node.new({ "/environment/aircraft-effects/cabinairdewpointC":0 });
 props.globals.initNode("/environment/aircraft-effects/cabinairdewpointC", dewpointC, "DOUBLE");
 
+#factor used to increment internal temprature over time using *(cabinheatset*cabinairset)
+#smaller number will slow effect
+props.Node.new({ "/environment/aircraft-effects/internalairtempfactor":0 });
+props.globals.initNode("/environment/aircraft-effects/internalairtempfactor", .04, "DOUBLE");
+
+#factor used to increment internal surface temprature over time using *(cabinheatset*cabinairset)
+#smaller number will slow effect
+props.Node.new({ "/environment/aircraft-effects/internalsurfacetempfactor":0 });
+props.globals.initNode("/environment/aircraft-effects/internalsurfacetempfactor", .03, "DOUBLE");
+
+#factor used to increment internal air temprature from air inflow (no heat) over time using *(cabinairset)
+#smaller number will slow effect
+props.Node.new({ "/environment/aircraft-effects/outsideairflowtempfactor":0 });
+props.globals.initNode("/environment/aircraft-effects/outsideairflowtempfactor", .03, "DOUBLE");
+
+#factor used to increment internal surface temprature from air inflow (no heat) over time using *(cabinairset)
+#smaller number will slow effect
+props.Node.new({ "/environment/aircraft-effects/outsideairflowsurfacetempfactor":0 });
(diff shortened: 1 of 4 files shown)


gilbertohasnofb:
Hmm, I have some few problems with this idea:

- first and most important, I find it confusing. I don't think we need to give the users 8 confusing variables to play with as it is more likely to do harm than good (people using silly values and asking for help, for instance). In my opinion, we should simply try to come up with good parameters ourselves and stick to it

- that said, if you all want to go ahead with this idea then we need to fix the GUI button, as of now it is smaller than the text (at least on my screen):



- also, in the image above, all values start at 0. Is this how it is supposed to be? That is, the user is actually setting an offset of the parameters we use, and using 0 would be the default behaviour?

wlbragg:
No, this was all correct and filled in with the default values in my setup. Although I did also have the default values hard coded for reference in case one wanted to reset it back to default. But they weren't displaying and I forgot to fix it. I'll cancel the pull request and fix the layout. What is your system?
Also, I agree with you as far as a "average" user is concerned. I also don't think this should necessarily be available to the gen public, although you don't give people enough credit, they don't have to mess with it. I mainly did this so those that are not satisfied with the factors as they are now, they can try their hand at adjusting them.
If you really think this is trouble, then we can just keep it a separate branch or a hidden feature and only those in the know will have access to it.

onox:
I don't think this GUI should be added. You can decide though whether to expose the factors to the property tree:

1. Read the factors somewhere at the top of the `weather_effects_loop` function.

2. Read `frostlevel` (`/environment/aircraft-effects/frost-level`) and `foglevel` (`/environment/aircraft-effects/fog-level`) also near the top of that function. Currently they are at the top of the file instead. As you can see in the file (`Nasal/weather.nas`), the values of frostlevel and foglevel are written to the property tree, but they aren't read from it.

3. Fix `moisture` as well. I think it's the only Nasal-only variable currently.

onox:
Also, I'm not sure about the `if (!getprop("/fdm/jsbsim/weather"))` check in the code. I think the levels should always be computed, and instead put the check in the model XML files somewhere in an appropriate place. I'm not sure whether that is possible, maybe we need an extra function (filter) that toggles between the actual level and 0.

This way you just hide the frost and fog, instead of completely setting its state to 0.

gilbertohasnofb:
> although you don't give people enough credit

It's true that I tend to give very little credit to the average user, but I think this is really based on observation rather than a prejudice of my part. But in the end, I don't even think this is an issue of what the end user wants (or think he wants), it's rather about us deciding to make a simple and clean interface vs. giving tons of options that can be overwhelming. I think the latter has been a staple of so many open source projects and Linux distributions (the idea of choice being the holy grail, even if it will make the GUI confusing, overwhelming, having a gigantic list of buttons and toggles, etc.) and thankfully a movement of clean GUI with fewer options is finally catching up. That is not to say I don't appreciate the effort here, I honestly do (you must know how much I admire all the work you put into this plane), I just personally don't think this particular PR would be a good idea. I would say what @onox proposed (expose these variables in the property tree) seems like a good idea which will allow those who are interested to tweak their sim, while keeping the general clicking-every-button-and-toggle-I-can-see crowd safely flying :wink:

wlbragg:
No worries @gilbertohasnofb, like I said, I really add this option so anyone that felt like working on it could try to adjust the factors to make this more true to life, ie: devs, not necessarily users. So if everyone/anyone wants to keep the GUI simple, no problem. Lets just leave the branch there and hopefully someone will take the time to adjust the factors and report back (or maybe I'll get to it at some point). If not then this is all in vain.
I think there is enough factors in the entire frost and fog generation code to make it simulate "real life" with some tweaks to those factors.
I have no problem with not merging this with the master branch.

PR #469, opened 2015-10-21
fixing bug 463

### LINKED ISSUE #463: Volt-meter constantly at red line/max rather than in green area
volt-max

During my test flight I discovered that the voltmeter continously peaks to the max. allowed value ~32V while the PT shows only 28.
So I guess it's about an animation forgotten/unlinked/wrong set up.

gilbertohasnofb:
This PR should fix #463


onox:
Animation is better now, thanks.

legoboyvdlp:
Alright, that's great! (But I thought it was supposed to be after 3.6 comes out?) :P

onox:
This PR has been merged into master, which will become 3.8 in a couple of months.

Issue #472, opened 2015-10-25
Add grain and lightmap effect to interior


wlbragg:
I think we need to get this added. I just added an implicit light map and it looks good. We may not need an explicit one unless we have a really good light person that wants to make a map.

Also the grainMap, I am not sure exactly how the grain texture is suppose to look so I asked Thorsten for the example he used in his documentation.His c172p example actually looked really good.

Here is the implicit lightmap effect.
![lmap](https://cloud.githubusercontent.com/assets/11069680/10714429/4f8c2670-7abd-11e5-9283-bf087d6e777d.jpg)

wlbragg:
Same thing with a bit heavier lightmap setting and a quick and dirty grain effect.
Obviously we can do much better.
![gmap](https://cloud.githubusercontent.com/assets/11069680/10714545/de3f0366-7ac1-11e5-82ed-ef78ef7d371c.jpg)

onox:
What about that irradiance map?

gilbertohasnofb:
@wlbragg In the second picture, the interior texture is missing (note how the door is texturized, but not the other "walls").

wlbragg:
@onox, the first picture is the Irradiance. It is only to additional lines of code, We can add that right now.

The second picture is using a grain overlay (and the Irradiance) on the "chassis" only, I hadn't applied it to the door yet.
I don't fully understand how to create the grain texture yet, in other words It isn't set correctly and it may not be made correctly. It should add detail to what is already there and in my case it basically replaced it (we can do much better).

PR #473, opened 2015-10-26
Irradiance lighting effect.

wlbragg:
After considerable testing and tinkering, this is about the best we can do without a custom made light map and grain map. However the grain map won't work well without breaking out some of the objects into manageable pieces. For example the chassis has parts that should not have the grain applied to it like certain tight curves and the top of the dash. It doesn't project well in those places.


onox:
Could you use 4-space indentation? After that do `git rebase -i HEAD~3` (assuming you're going to add a third commit) and then change the last two "pick"'s to "fixup". (Keep "pick" for "Irradiance lighting effect")

Models/Effects/interior/c172p-interior-light.eff
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<PropertyList>
+  <name>c172p-interior-light</name>
+  <inherits-from>Effects/model-interior</inherits-from>
+  <parameters>
+    <irradiance-map-type type="int">2</irradiance-map-type>
+    <irradiance-map-strength type="float">1</irradiance-map-strength>
+
+    <!--texture n="3">
```
onox: I think you need to write `<texture` instead of `<!--texture` here?

onox:
Btw, does this fix #472?

wlbragg:
Sorry onox, actually we aren't even using that block of code (that's why it's commented out). That is if you want a custom light map. All we are using is the two lines that start with irradiance. I don't know why I left that block of code there, just in case someone wanted to try to make a custom light map.

wlbragg:
Yes it will close #472.

Yes, I will fix tab spaces, sorry, forgot to check it.

The block of code that is commented out is only needed if we do a custom light map.

wlbragg:
> invalid upstream HEAD-3

onox:
Hmm.. I see you merged `master` into `irradiance` branch. Usually you don't have to do that.

wlbragg:
Sorry, I was unclear as to whether or not I should do that. I knew there was at least the one change to c172p.xml tail contact.

onox:
The only reasons to merge `master` into a feature branch is to resolve conflicts before merging the branch into `master`, or if lots of commits were added to both branches and you want to test the results.

But for these small PR's, Git can merge your branch just fine.

wlbragg:
Thanks

PR #474, opened 2015-10-26
Tail skid contact height

dany93:
Z-contact by the tail skid in the FDM was too low (z=8), which made it "touch" the ground prematurely (see external view by rolling at very high pitch-up angle). 

tail-skid-contact

More importantly, it prevented from reaching a high pitch-up angle at landing (slow speed when grazing the ground, down to stall).

By changing for z=15 ==> Accordance between FDM and 3D view contacts, and very soft touching down at slower speed, higher AoA becomes possible at landing by progressively pulling the yoke so to pitch up and hold the plane.

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -260,7 +260,7 @@
             <location unit="IN">
                 <x> 226.7 </x>
                 <y> 0 </y>
-                <z> 8 </z>
+                <z> 17 </z>
             </location>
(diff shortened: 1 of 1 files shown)


wlbragg:
Thanks dany93, I verified your adjustment. I wonder when and why it got so far off from the original setting.  At one point in time 8 worked. Oh well, works now.

PR #477, opened 2015-10-27
Use mouse scroll wheel to zoom in and out for 'lookat' views

### LINKED ISSUE #476: Allow mouse wheel scrolling to zoom in and out
For `lookat` views we can use mouse wheel scrolling to zoom in and out. Changing the FOV does not make much sense for `lookat` views. We could set limits to 5 meter and 150 meter. 5 meter distance is just a little bit closer than the tip of the wing. It could give nice camera views from left or right.

onox:
Fixes #476 

It works nicely with the external Helicopter View, Model View, and Walker Orbit View. Chase View gave weird results (distance as well as altitude of camera changes), so disabled it for that view.

### DIFF
--- a/Systems/views.xml
+++ b/Systems/views.xml
@@ -45,6 +45,10 @@
                 </not-equals>
                 <not-equals>
                     <property>/sim/current-view/name</property>
+                    <value>Chase View Without Yaw</value>
+                </not-equals>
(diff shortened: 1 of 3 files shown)


wlbragg:
Looks good.

PR #478, opened 2015-10-30
Attenuate engine and wind volume if one or both doors are closed

### LINKED ISSUE #117: sound of engine and wind should be louder with door opened

onox:
This time actually fix #117 

Factor also controls sounds related to engine start, but I find it almost impossible to hear any difference for `Sounds/engstart.wav` and `Sounds/cranking.wav`. Maybe because these sounds are very soft themselves (in Audacity), so the factor does not have much of an effect on it.

Test:

1. Both doors closed

2. One door opened. As soon as the door is a little bit open, the sound should become a bit louder.

3. Both doors opened. Second door should have less effect.

4. Try external view. Very close (zoom in with mouse scroll wheel) and more distant (up to 150 meter). Engine sound should become softer.

For those who want to play with the amount of boost:

Currently, boost is about 20 % for one door open. 32 % for both doors open, and 40 % in external view. The 32 % value is related to the 20 % value, but the 20 % and the 40 % can be directly tuned (on line 142 and 146 in `Systems/c172p-sounds.xml`).

You will also see the value '0.7143' in the file ` c172-sound.xml` below. This is basically 1/1.4 (because of the 40 % boost).


onox:
Please test, but do not merge yet, because factor doesn't change during replay mode yet.

gilbertohasnofb:
So I tested this, and here are some points:

- first of all, it all works for me!
- I think we really should boost the engine sounds. Our engine at idle is really soft, while in RL it should be quite loud: https://www.youtube.com/watch?v=cqgngQ0lAgM
- I also think the difference between opened and closed doors could be higher, but that very likely is tied to the soft sound files I guess

Would you like me to boos those sound files?

And by the way, I saw this this morning and it reminded me of you :wink: 
git

wlbragg:
:laughing:

onox:
Sure, you can try to boost the sounds.

gilbertohasnofb:
> Sure, you can try to boost the sounds.

The wasp sound has already as much amplitude as it can without being clipped, so I can't raise its volume. Maybe we have to raise the volume in the xml file?

onox:
> Maybe we have to raise the volume in the xml file?

That's not possible, maximum volume in xml file can be 1.0. The only thing we can do is boost the volume factor (the 20 % and 40 % values) even more, and decrease the `<factor>` in the xml file so that the idle engine sound will be lower. But it will force the user to increase the hardware volume of his/her speakers.

gilbertohasnofb:
> so that the idle engine sound will be lower
No, the opposite I would say! Our idle engine is already a soft kitten

onox:
If you want to increase the idle engine sound, then you would have to decrease the volume boost factor. Thus the difference between 'doors closed' and 'doors open' will be less.

gilbertohasnofb:
I understand the problem, I am just saying that our idle engine does sound very soft to me. But in that case we can just leave it as it is.

onox:
But what if you try to increase the volume in FG's or your OS' audio settings?

onox:
I'm trying to add a `<position>` to the wind sound because currently you can hear it even if you're far away from the plane.

gilbertohasnofb:
> But what if you try to increase the volume in FG's or your OS' audio settings?

I am of course speaking in relative terms, as I am aware I have a volume slider on my OS :wink: 

> I'm trying to add a <position> to the wind sound because currently you can hear it even if you're far away from the plane.

I thought that would be expected. This wind sound is from the environment and not produced by the engine, and therefore it should be heard omnipresently and regardless of the distance, right?

onox:
> I thought that would be expected. This wind sound is from the environment and not produced by the engine, and therefore it should be heard omnipresently and regardless of the distance, right?

Yes, but if an airplane is going very fast, then it sounds like as if the wind on the ground (where the camera is) is going very fast as well. Thus I think it's better to specify a `<position>` for the wind.

gilbertohasnofb:
I see. But won't that cause some side effects on external views of the plane?

onox:
You won't hear the wind if you're observing the plane from a large distance (> 100 meter).

gilbertohasnofb:
Sounds fair enough.

onox:
@gilbertohasnofb Is the volume of the wind ok now?

gilbertohasnofb:
I can barely hear any wind to be frank, and I only hear it when I am in the cockpit. But to be honest that is fine!

onox:
@dany93 @wlbragg @tigert  What do you think of the volumes? It's mainly the difference between the various sounds that's important I guess.

wlbragg:
I don't have a good comparison to go by, but I can report that what I hear sounds pretty good. Volume is good on my Windows machine but low on the Linux box. My Linux box is probably the issue and not the sound files or project.

PR #481, opened 2015-11-01
Multi thunder

wlbragg:
I think even 2 more thunder sound modules and this would be optimal, especially if there were 2 or 3 different thunder sounds. It's really a nice feature, you guys did a great job. Very immersive, I really like the rain sound too.

Issue #443, opened 2015-08-18
Thunder sound


gilbertohasnofb:
According to: http://wiki.flightgear.org/FlightGear_Newsletter_August_2015#Atmospheric_Light_Scattering

> Since the simulation writes the position and range of the lightning strike relative to the current eye position into the /environment/lightning node of the Property Tree (lightning-pos-x, lightning-pos-y, and lightning-range), aircraft modellers can use this information to trigger a thunder sound after a delay. 

So it would be cool if we implement some thunder sounds in our plane. I can come up with the sound and maybe @onox could give me a hand implementing it.

* [ ] Volume of thunder proportional to distance
* [ ] Directional sound

(Both could be supported if we can spawn sounds at absolute positions).

gilbertohasnofb:
@wlbragg Sound uploaded to branch `bug-443`.

thevirtualfer:
Hi guys.

I lost my password.

When i inform my username and e-mail address, the page says that the
e-mail/username information submitted could not be found.

How to go ahead ?

Cheers, to all of you.

2015-08-19 18:12 GMT-03:00 Gilberto Agostinho <notifications@github.com>:

onox:
You only have to enter your e-mail address when you click on that lost password link on the log-in page. And you should know your e-mail address because Github is sending notifications (c172p-detailed related) to it.

gilbertohasnofb:
@onox when you have time, could you please have a look on this? I already uploaded the sound to branch `bug-443`

onox:
@gilbertohasnofb Don't we need to be able to set the position dynamically for this? That's not possible right now in simgear.

wlbragg:
@onox 
I'm not sure what your referring to, According to Thorsten the position of the strike is accesible from the prop tree. We just need to calculate the distance to the strike and use that to compute the sound delay and maybe the sound volume also.

gilbertohasnofb:
@onox I think @wlbragg is correct, Thorsten told me the same thing in the forum. But if for some reason we can't use the position and distance of the lightning strikes, no worries, we can either randomize it a bit or just simply fix it as a straight ahead thunder and far away.

Also, it would be nice to add some delays between the lightning strike and the thunder sounds, just like in real life. If we could use the distance to create this delay it would be so cool!

onox:
I understand you can compute the distance, but wouldn't it be nice if you could hear that the thunderstrike happened behind you or to your right, for example?

[...]

PR #484, opened 2015-11-01
Bug 443

onox:
From PR #481: "I think even 2 more thunder sound modules and this would be optimal, especially if there were 2 or 3 different thunder sounds. It's really a nice feature, you guys did a great job. Very immersive, I really like the rain sound too."

Fix some issues in #443


gilbertohasnofb:
Don't merge this yet as the new sounds are not programmed in. I mentioned before that ideally they could be randomly selected (to avoid a 1->2->3->1->2->3 pattern), but if that's not possible we can simply use the 123 solution. Could any of you guys check this out when you have a chance? Thanks!

onox:
In that case we need 9 sound definitions?

gilbertohasnofb:
Hmm, I see that `<path>Sounds/thunder.wav</path>` can't be randomly selected, right? So in that case we can just go with the 123123 and see how it sounds. Maybe the effect will be good enough like that, particularly if we take into account the distance of the lightning when dealing with the volume (thus making each sound even more individual).

onox:
OK, try to modify `<thunder2>` and `<thunder3>` and test how it sounds.

gilbertohasnofb:
Will do that!

gilbertohasnofb:
I think it sounds great, better than before. And the sounds are similar enough so that the 123 pattern can't be easily heard. As for making the volume proportional to distance, would you be able to have a look at it?

onox:
You can do that as well :smile: 

`lightning_distance_norm` gives a value between 1.0 and 2.0. If you remove the '1 + ` part, then you get a number between 0 and 1.

Next step: replace `<sum>` with `<product>`.

onox:
Hmm... I doubt the XML sound system supports `<expression>` though...

gilbertohasnofb:
Hmm, so in that case would you have some other ideas of how to implement this? Do we need to use Nasal then?

onox:
You can just add 3 small `<filter>`'s to `Systems/sounds.xml`.

wlbragg:
I don't think the 123 pattern will happen anyway. because if you watch the debug properties it doesn't appear to stay in order (even though it should) I think various other actions may cause whats playing to execute at a different rate than the 9 seconds so it may go 1 2 3 then 2 may finish first and now it is available and so it get picked up, then 1 becomes available then it goes again then 3. So you might have 1 2 3 2 1 3. At least that's what I thought I saw last night.
I sure am glad you noticed my botched merge and commit. I don't know what I did to get to that state. I noticed it last night when I did it, that it looked strange but was to tired to look into it.
I haven't finished reading all the posts in this thread yet so if I am way out of order bare with me. This time difference can cause issues all on its own.

wlbragg:
Thanks @onox, I just finished catching up on this thread. I wasn't sure how to get the volume because I don't know what the volume tag is expecting as input (as in the range of value 0-1 or 0-100 or what. I also didn't know what how it was calculating the factor and in our case using two products. I added the 1+ to make it a value between 1-2 to see if I could hear a difference easier to see if it was working. 

>particularly if we take into account the distance of the lightning when dealing with the volume (thus making each sound even more individual).

Yes I thought the same thing, it should really help mix it up.

wlbragg:
@gilbertohasnofb Do you know how to add a filter?

gilbertohasnofb:
@wlbragg not really, I was trying to read about it but still couldn't find anything about `<filter>`. If you are able to do that, feel free to implement this.

wlbragg:
OK, I'll take care of it. You can look at what I do then you'll know, there really isn't much to it, unless you've never done it, ha!

gilbertohasnofb:
Sounds good! :+1:

onox:
@wlbragg @gilbertohasnofb I have fixed the issues I wrote about in the comments above. Currently `lightning_distance_norm` is very linear, but shouldn't the sound energy dissipate across a surface (2D) or volume (3D)?

gilbertohasnofb:
I believe it should be squared (2D) as the intensity should be inversionally proportional to the surface area of a sphere around the sound source.

gilbertohasnofb:
> The sound intensity `I` (energy) falls inversely proportional to the square of the distance

onox:
What about this: `25000000/math.pow(lightning_distance, 2)`? This will give:

```
25000000/(5000**2) = 1.0
25000000/(15000**2)
```

Loudest (100 %) at 5000 meter, 10 % at 15 km.

onox:
I'm not completely sure about squaring the distance. The volume decreases rather quickly if I change the minimal distance at which volume is 1.0.

onox:
This expression is better: `1/((10000.0/5000.0)**2)`. Gives 25 % for 10 km, 100 % for 5 km.

gilbertohasnofb:
Sounds reasonable (no pun intended), but maybe we want to bring that figure of 5km down to 1km, so that if a strike is very nearby it will be quite loud.

onox:
Should we in that case switch to a different sound (just a bang, instead of lots of rumbling)?

onox:
I did some measurement of 68 thunder strikes:

min: 4783.409328325995
mean: 8874.076194374664
median: 8802.61975248733
max: 13938.43289182003

gilbertohasnofb:
In that case, I would say use the last formula you came up with. And I would keep the same sound files, I really like the rumbling.

onox:
One more thing we can do is: you can hear thunder on the ground up to about 16 km. We can make that 20 km and limit playing sounds only for thunder < 20 km.

gilbertohasnofb:
I think a limit of distance is indeed reasonable.

wlbragg:
Looks good guys. The idea of a loud bang for lightning that is right on top of you isn't a bad idea. We get tons of lightning here and a close strike will about knock you out of your chair and then no other sound.
But I am good with it either way.

I'm getting ready to pull it and test.

onox:
I have some changes to commit.

wlbragg:
I agree, a limit of distance is reasonable, believe it or not it doesn't travel as far as you might think. We can't really hear it much over 10 miles under perfect conditions.

wlbragg:
OK, I haven't pulled it yet.

onox:
Committed. I have removed the clamping (to 50 seconds) because of the distance check.

onox:
I'm a bit surprised I just cannot find the speed of sound anywhere in the property tree :open_mouth:

gilbertohasnofb:
> I'm a bit surprised I just cannot find the speed of sound anywhere in the property tree :open_mouth:

Yep, I also spent some long time looking for it!

wlbragg:
Really, how do we do sonic booms without it, ha?

onox:
Can we find the atmosphere pressure and density somewhere in `/fdm/jsbsim/`?

gilbertohasnofb:
I think you can find it in `/environment`

wlbragg:
I don't know how I missed it but when I was cleaning up some of the commented out code, I cleaned c172p.nas but left the main unused "terrain_survol_loop" function. I deleted the reference to it but not the function. It needs to be deleted.
@onox, if your still to commit more to this branch, you can delete the "terrain_survol_loop" function.

onox:
OK

onox:
I guess density needs to be in kg/m^3.

wlbragg:
Absolutely incredible, this is really cool, to bad it is so difficult to fly in the thunderstorms, I keep breaking my plane.

I have noticed on occasion that it still cuts off the sound prematurely. I can't tell why though. It may be that the 9 second timer needs to be extended a little bit to allow for system bottle necks?

onox:
> I have noticed on occasion that it still cuts off the sound prematurely.

The problem is that the listener callback can be invoked multiple times shortly one after another, and in both cases they might think a sound is 0 because of the delay.

onox:
So I think you need to delay the decision which sound to use until you actually want to play the sound.

wlbragg:
Yeah, extending it didn't help.
OK, I see what your talking about, now how to do this?

onox:
Call `settimer()` with appropriate delay. Choose a sound in its callback and then call `click()` with delay 0?

wlbragg:
Use the same delay_seconds as a timer before calling the 
if (!thunder1) 

?

onox:
Yes.

I have nothing to commit currently. Still busy with the speed of sound.

wlbragg:
Oops, I posted same time you did. I need to think about this, my brain is fried today.

wlbragg:
OK, that make sense, at least at the moment. :confused:

wlbragg:
OK, testing it now.

wlbragg:
Much better, I think that completely fixed it.

wlbragg:
Yes, no more cutouts. Can I push my bug-443 now?

onox:
Remove all tabs? :wink:

wlbragg:
I removed "terrain_survol_loop"
I wasn't sure what to describe for the commit.

onox:
"Removing unused function terrain_survol_loop"?

wlbragg:
I've been listening to it this whole time and not so much as a hiccup.

Unfortunately it also fixed the randomness, it is now following the 123,123,123 order. But I don't think that it is really noticeable because you actually get, 12, 312, 31, 231, 23, 123 etc.

wlbragg:
>"Removing unused function terrain_survol_loop"?

No, changing the delay to how we did it. I wasn't really sure how to describe it. :smile:

onox:
But the function still exists?

wkitty42:
On 11/01/2015 05:59 PM, onox wrote:
> Should we in that case switch to a different sound (just a bang, instead of
> lots of rumbling)?

there would be a really loud bang plus some loud rumbling that gets quieter as 
the sound dissipates...

gilbertohasnofb:
> there would be a really loud bang plus some loud rumbling that gets quieter as the sound dissipates...

If you guys want to break the sound down to its components (that is, a sound file for the loud bang, another for the rumbling, another for low frequencies, etc.) it will get a bit complicated. This sound we have is a RL recording, and then either we will have to synthesize the components separately (tons of work, not good result) or we need to filter parts of it away (can get full of artifacts). My suggestion is to keep using our current sound which, IMO, is quite decent.

onox:
Let keep it simple for now.

wlbragg:
>But the function still exists? Any intention to remove it? 

I removed it in the last push, I thought?

wlbragg:
I'm an idiot, I guess I forgot to actually delete it, only in my mind!

onox:
Well? I still don't see it here on Github :laughing:

wlbragg:
You don't see the push at all?

onox:
Nope. Do `git status`.

onox:
@wlbragg Everything OK over there on the other side of the pond? :wink:

wlbragg:
I just pushed it again with the "ground detection" deletion.

onox:
Yes. I see it now.

onox:
Merge the PR?

wlbragg:
Fine with me!

Issue #482, opened 2015-11-01
Strange shadow of a line


gilbertohasnofb:
Yesterday I realized that there is some strange shadow of a line which si much darker than any other shadow in our cockpit. See:

![line](https://cloud.githubusercontent.com/assets/5700795/10868935/9948c75a-809e-11e5-9499-26984038b87a.jpg)

Is this a bug or a feature?

dany93:
I wonder if this would not be a bug, same kind as https://github.com/Juanvvc/c172p-detailed/issues/428
ALS shadows in the cockpit are pleasant at first view, but I feel they are sometimes weird, not easily correlated with the upstream objects and the light source direction.

gilbertohasnofb:
I don't know, Dany, but I don't think it's related to #428. This artifact appeared some short time ago and it's always present as far as I can tell.

I also think the ALS shadows in the cockpit are one of the most immersive effects we have, so if you spot something that is weird try telling @wlbragg about it so we can try to improve it. But IMO the effect is fantastic right now!

wlbragg:
I'll see if I can at least figure out what it is that is causing it.

gilbertohasnofb:
@wlbragg one more thing: it seems to appear when the sun is behind the plane

dany93:
> I also think the ALS shadows in the cockpit are one of the most immersive effects we have,...
...IMO the effect is fantastic right now!

I agree. And I'm pleased with it, whatever.



> so if you spot something that is weird try telling @wlbragg about it so we can try to improve it.

Same place (LFML), same orientation (90 deg), same time (11:30).

**With Rembrandt:**
![rembrant-shadows-lfml](https://cloud.githubusercontent.com/assets/11994061/10869839/a4f2098c-80bb-11e5-8001-7a032d89ce9d.png)

**with ALS:**
![als-shadows--lfml](https://cloud.githubusercontent.com/assets/11994061/10869841/c0a340ce-80bb-11e5-8b67-9aa193155638.png)

Not a critic, I do not like criticizing when I cannot help to get a solution. So, thanks for the work done and the result, anyway.

wlbragg:
Interesting @dany93, that could be my poor excuse of a cube map. Good comparison though to try to make it better.

I don't know if we are capable of getting the yoke to shadow like Rembrandt.
I think not, because we are only using images and not real objects, we just get solid/not solid (no 3d definition).

dany93:
[...]

PR #490, opened 2015-11-02
Correct addition of irradiance map.

wlbragg:
The irradiance code I pushed broke the interior shadow. You can't have the same object assigned to two different effects which is what I did. I needed to have the irradiance lines added to the original interior effect which is what this commit does.


wlbragg:
Because of the interest in the c172p and this is a limited correction to an error, I went ahead and merged this.

onox:
Did this PR fix #482?

wlbragg:
I don't know, I need gilbertohasnofb to check for the line he saw. But it was a mistake on my part and needed to be corrected regardless.

gilbertohasnofb:
@wlbragg I've had my hands full the whole day and I still couldn't check it. I will try to do it still today but unfortunately later, okay? Sorry about that.

wlbragg:
@gilbertohasnofb, no worries, good work your doing!

gilbertohasnofb:
@wlbragg Finally made some tests and it seems that problem is fixed! Under the same conditions that line isn't appearing at all. The shadow also looks much better somehow, I love the effect :smile: 
I will close #482

wlbragg:
Good, glad to here this. It must have been a side effect of the faulty effect I corrected.

gilbertohasnofb:
Thanks for it, my impression was that the whole shadow cube and the interiors were looking much better, although I can't tell exactly why!

Issue #303, opened 2015-06-15
Implement external preflight inspection


gilbertohasnofb:
Here is a list of ideas of what we can implement. 

* [x] clicking on the fuel tank door (red circle above each wing) will open the fuel and weight dialogue
* [x] ~~clicking on ailerons, rudder and elevator (if and only if on the ground and with the engine off) causes them to wiggle, as if testing them~~ maybe too silly, one can simply use their joystick/yoke/mouse in control mode to look at the surfaces moving from outside to simulate this
* [x] clicking on the pitot tube will add or remove the pitot cover. If forgotten, the airspeed indicator should be considered as blocked and the airspeed reading should be much lower than normal
* [x] clicking on the wheels (if and only if on the ground and with the engine off) should add or remove a block and the aircraft can't move if the block is on (kind of like parking brakes). Something like this yellow blocks but on all gears: http://www.cessna172club.com/forum/avatars/1314.jpg
* [x] create cords for securing the aircraft to the floor (click to add remove if and only if on the ground with engine off), aircraft can't move if they are on. See: https://upload.wikimedia.org/wikipedia/commons/1/1f/CAP_cadets_secure_a_Cessna_172_%28090317-F-2156C-003%29.jpg
* [x] create option in the aircraft menu to [ ] secure aircraft (add all things above or remove all things)
* [x] ~~also simulate fuel contamination by water? (how it affects the engine)~~ split into #504
* [x] ~~implement oil level. As ludomotico wrote: "You must check the oil level always, and refill about once out of ten flights. I don't know if flighgear implements engine failures if the oil lever is too low, or they must be simulated by the aircraft."~~ split into #499 

* * *

Old issue:

gizmo4487 made a feature request in the forum:

> An external preflight inspection should be added, since the plane supports Walker view. It would add to the realism and be very entertaining for real pilots.

That is a nice idea indeed. I think there was some discussion about it some time ago, but I can't remember where it lead to. Let's leave an opened issue for this (no milestone attributed so far since this has still to be discussed).

Juanvvc:
We need two things for this:

- a working walker :) The walker needs the latitude and longitude of the model to be initialized before switching to the walker view. Currently, you have to switch the walker mode on in the menu and then switch to the walker view, or you spawn in the middle of the ocean. Initializing latitude and longitude is trickier than it seems, and the code is buried somewhere inside the walker nasal code. I tried some weeks ago to fix this, with no success.
[...]

PR #491, opened 2015-11-02
Bug 303

gilbertohasnofb:
@onox I just pushed my last commit trying to fix what we discuss at https://github.com/Juanvvc/c172p-detailed/commit/5112b2fdd83be4485a56c3b65178b011b2d493f8 . What do you think if we already test (and hopefully merge) this PR for the pitot tube and for the chokes? Then I will move on to create the wing straps (should be much smoother than it was today, as it will use the same parking brake logic we already implemented).

Also, I will open new issues to remind us to add these items to the checklist and tutorials later.

Note that I also set the pitot tube cover and chokes are set to false by default, so the first time someone opens the sim they will be able to fly without a problem (and will do the same for the wing straps). For those of us who want realism, then we can start setting them and their state is saved.

Let me know what you think, and thanks a lot for all the coaching, I really appreciate it :smile:


gilbertohasnofb:
Spelling fixed.

@onox I am still having one problem which I can't spot the mistake. The menu toggle goes gray when the speed is above 5 knots (as expected), so you can't add chocks while in movement. But clicking on the wheel at any speed is adding the chocks! And I have such a clear conditional, see:

```
    <animation>
        <type>pick</type>
        <object-name>NoseWheel</object-name>
        <action>
            <button>0</button>
            <repeatable>false</repeatable>
            <condition>
                <and>
                    <not>
                        <property>sim/model/c172p/chock</property>
                    </not>
                    <less-than>
                        <value>5.0</value>
                    </less-than>
                </and>
            </condition>
            <binding>
                <property>sim/model/c172p/chock</property>
                <value>true</value>
            </binding>
        </action>
```

Can you help me spot the mistake?

onox:
Maybe you need to put `<expression>` around `<less-than>` block.

Juanvvc:
I believe, but I'm not sure, that the condition block must be inside the binding block. Like in "bind this only if this condition is met"

gilbertohasnofb:
No, it's not that.

gilbertohasnofb:
> I believe, but I'm not sure, that the condition block must be inside the binding block. Like in "bind this only if this condition is met" 

Yep! That should be it!

onox:
Hmm yeah, `<expression>` shouldn't be necessary for things like comparisons. So it's probably what @Juanvvc said.

gilbertohasnofb:
@Juanvvc that solved it, thanks ludo!

Models/c172p.xml
```diff
@@ -6587,5 +6587,273 @@
...
+    </animation>
+    <animation>
+        <type>pick</type>
+        <object-name>PitotTube</object-name>
+        <action>
+            <button>0</button>
+            <repeatable>false</repeatable>
+            <binding>
+                <condition>
+                    <and>
```
onox: You have this logic here multiple times (in different bindings), so better put it in a `<logic>` element then.

gilbertohasnofb:
@onox one more thing, do you think I should add another condition for allowing the removal of chocks and pitot cover: not only ground speed below 5 knots, but also the wow conditions? I find it unrealistic that anyone will ever be in the air with ground speed below 5 knots and trying to play with chocks and pitot cover, but I could do that just to be able to sleep well :wink: 

And when I have a `<and>` or `<or>` block, can I have more than two expressions inside? That is, is the syntax below correct?:

```
<and>
    expression1
    expression2
    expression3
</and>
```

or do I need to do this?:

```
<and>
    expression1
    <and>
        expression2
        expression3
    </and>
</and>
```

Models/c172p.xml
```diff
@@ -6587,5 +6587,273 @@
...
+                <label>Put Pitot Tube Cap</label>
+            </binding>
+        </hovered>
+    </animation>
+    
+    <!-- Wheel chock -->
+    <model>
+        <name>chock1</name>
+        <path>/Exterior/chock/chock.xml</path>
+        <condition>
```
onox: Tried removing the `<condition>`?

Models/c172p.xml
gilbertohasnofb: Okay!

onox:
You can put multiple things in `<and>` or `<or>`.

gilbertohasnofb:
:+1:

Models/c172p.xml
gilbertohasnofb: I will try

gilbertohasnofb: It works, just pushed a new commit.

Models/c172p.xml
gilbertohasnofb: Done, see my new commit.

onox:
@gilbertohasnofb Do you have anything to commit? I would like to merge `master` to `bug-303` to resolve the conflicts.

gilbertohasnofb:
No, I am all done here, you can go ahead :smile:  The conflict will be easy to solve: I believe its related to the 3D baggages I added in a different branch and which we merged yesterday.

onox:
You can pull :smile: Thanks for the work you have done @gilbertohasnofb

gilbertohasnofb:
Thank YOU for all the patience with me :smile: I think I am starting to get the hang of it

onox:
You want to keep the branch? Since #303 is still open.

gilbertohasnofb:
Well, bug 303 still has other things to be done, so I'd say let's keep it. I will try to work on the wing straps (for securing the plane) when I have a chance, and I was thinking about working on that branch.

onox:
Ok :+1:

Juanvvc:
Thanks for all your work! This aircraft is becoming more and more amazing!

gilbertohasnofb:
My pleasure!

wkitty42:
what @Juanvvc said :smile_cat:

gilbertohasnofb:
:smile:

PR #501, opened 2015-11-04
oil door to normal map

### LINKED ISSUE #500: Add oil door to normal map
Add oil door to normal map

gilbertohasnofb:
Closes #500 

fgfs-screen-001


legoboyvdlp:
I must say it looks beautiful, Gilberto. Nice work! Is it correct that it is openable?

onox:
Currently no. This PR just adds it to the normal map, which is used to create the illusion of some depth.

gilbertohasnofb:
Yeah, onox is correct. An operable door would need a redesign of the 3D model, and given that @thevirtualfer is planning to do a major overhaul of the model I don't think we will add the door right now.

legoboyvdlp:
I see. That's alright.

Issue #314, opened 2015-06-18
Add glass reflection visible from external view


onox:
Duplicate the glass and give it a very small distance from the interior glass, so that there is no z-fighting. Add an effect based on `Effects/model-combined-transparent`.

Tune the parameters to get something like: http://tigert.1g.fi/kuvat/MIK/2006-07-30-parnu/_DSC2655.jpg

gilbertohasnofb:
Is this possibly related to #13 or not at all?

onox:
This is about the large windshield and the glass in the doors. I just noticed there is already a "glas" object. ("glas_interior" and "glas_interior_sides" are for the interior).

onox:
@wlbragg It seems object "glas", "leftwindow", and "rightwindow" have a `Effects/interior/c172p-interior-glass-shadow` effect applied to them. What does this do? If I remove that effect and try to apply `Effects/glass` to them, I'm not getting any good reflection at all. Are the normals pointing outward?

wlbragg:
>rightwindow.. have a Effects/interior/c172p-interior-glass-shadow effect applied to them. What does this do?

interior-glass-shadow
applies the cubemap shadow to glas, leftwindow and rightwindow.

interior-glass-reflection-front 


interior-glass-reflection-front
applies the cubemap reflection to glas_interior.


glas_interior and  glas are duplicate, offset, front windows (only one effect per window)

leftwindow, rightwindow

glas_interior
leftwindow_interior
rightwindow_interior
glas_interior_sides

onox:
No I mean this:

![fgfs-screen-338](https://cloud.githubusercontent.com/assets/3361585/8259705/de55bb40-16be-11e5-8b1a-2f3777758361.png)

Look at the rainbow reflection on the canopy. That should be handled by `Models/Effects/glass.eff`, but when I tried to apply it to the "glas" object, I didn't see the cool reflection.

onox:
Is this cube map shadow thing an ALS effect to be seen from inside the cockpit or from external views?

onox:
> outside rain effect and sky reflection is already applied through
bumpspec "Effects/model-combined-deferred"

Transparent objects need to use `Effects/model-combined-transparent` (which is used by `Models/Effects/glass.eff`) not `Effects/model-combined-deferred` as far as I know.

onox:
> glas_interior and glas are duplicate, offset, front windows (only one effect per window)

[...]

Issue #507, opened 2015-11-04
Texture or normal issue with the pontoons


wlbragg:
I noticed this issue with the pontoons while flying in Alaska.
![texture1](https://cloud.githubusercontent.com/assets/11069680/10950481/7b8abed8-82fd-11e5-8c6e-9f80bb7e3fbc.jpg)

gilbertohasnofb:
Does this appear all the time for you or was it just in that particular flight (under specific light conditions)? The textures are symmetrical, so unless they were wrongly applied they should be identical (and I thought they were). Also, I don't even know where is the normal map for the floats. I will make a test and see how they show on my computer, ok?

wlbragg:
I think it is only on the floats and not on the amphibious.

wlbragg:
I found the problem, it's because of the bumbspec effect being applied to the pontoons, I'm not sure it should be.

gilbertohasnofb:
Can't you just copy the effect from the amphibious to the pontoons to solve this?

wlbragg:
I was just looking at that and they are both using the same thing I think? All there is is pontoon.xml
I am looking at the blend.

wlbragg:
I know I should know this seems how I did it, but where did we get the texture for the pontoons?

wlbragg:
It's a problem in the blend, I'll fix it.

gilbertohasnofb:
`/Models/floats.png`, and they are livery dependent.

wlbragg:
I mean originally, where did they come from, who made them, what aircraft?

gilbertohasnofb:
Hmm, I don't know, maybe from the beaver? The default one is almost untouched except for some additional textures I added in blank spaces, but the ones for the HD liveries are heavily modified by me.

gilbertohasnofb:
@wlbragg Has this issue been fixed in one of those PRs mentioned above?

wlbragg:
No this has not been fixed.
Now that I have a little more experience with Blender, I'll take another look at it and see if I can figure out what is wrong with it.

gilbertohasnofb:
Thanks!

PR #511, opened 2015-11-04
Pontoon uv fix

wlbragg:
This corrects #507


wlbragg:
This also should close #314

wlbragg:
I can't believe I did that, I had it just the way I wanted and then I mad a couple last minute changes and totally mucked it up. You should notice the difference now.

wlbragg:
Hey, I am really sorry about all this BS. I just went full circle to end up pretty much back where I started.  This is looking the way it is suppose to so I good with were it's at. Bottom line was there was some material entries in the blend file that needed corrected.

wlbragg:
fix3

onox:
> It will inherit model-combined-transparent and there was a noticeable difference, it's because I screwed it up right before I committed it. I'll recommit and you'll see the change. I'm glad your paying attention.

But why is it now -deferred instead of -transparent? I don't see any difference if I toggle between the two.

onox:
Continued in PR #512

Issue #206, opened 2015-05-25
Create landing gear lever for amphibian model


gilbertohasnofb:
Create landing gear lever for amphibian model: http://www.wipaire.com/Wipaire-Dev/piwigo/i.php?/upload/2012/06/18/20120618091859-cfcacdce-sm.jpg

Basically, it needs three things:
- a simple 3D lever on the floor of the cockpit
- textures for the lever (could be added to panel_parts.png)
- animation and programming of the lever

wlbragg:
Different model Cessna but the same concept I think. thttp://www.trade-a-plane.com/jhimage?id=1374753&keywords=1981%20Cessna%20185F%20-%20Listing%20%23%3A%202068369+%2318&width=573&height=430

gilbertohasnofb:
I think the picture I posted above is also from a different model (it's surely not a 172, even though that's what I was searching), and it looks fairly similar to what you posted. Nevertheless, if we don't find any RL information, then this idea of putting a long lever on the floor sounds quite plausible. And if anyone complains this is not realistic because his/hers amphibian 172 has a lever in a different place, we can always argue that this is not any amphibian system but rather "wlbragg's system" :wink:

Juanvvc:
That lever is quite similar to the flaps lever in a Piper PA28. Probably, we can share the model with the Piper Archer CX project.

gilbertohasnofb:
> Probably, we can share the model with the Piper Archer CX project.

That's a clever idea!

tigert:
I think that is the flap lever in those pictures. I would guess very old C172 models had mechanical flaps. Let's research a bit before modelling :-)

tigert:
For the Wipaire amphibious floats, the gear retraction handle and status lights seem to be combined into a round instrument-face it seems: Page 102 and 104: http://www.wipaire.com/pdf/Parts_Manuals/PM_2350_Cessna172.pdf

That has good reference photos for the modelling in any case, also side view for checking measurements and angles etc.. Closeups also for the gear mechanism and such.

wkitty42:
i did some looking to see if i could find some interior panel pictures with one of these controls on it... i didn't find anything, though... i did, however, look through the above linked PDF and another on the same site for the same product (2350 floats) and i found some info in the other one concerning weight... i didn't look close enough to see if there was anything about drag, though...

[...]

PR #512, opened 2015-11-05
Float enhance

wlbragg:
This can be merged anytime. It has a select animation so it only appears when floats or amphibious are selected.It also has the pontoon-uv stuff merged into it.

fix2


wlbragg:
The pontoon lighting effect changes are in the #511 PR. I also merged that into this branch,  why you might ask, I don't know, because I was getting to confused with what I did and didn't do. :confounded:

wlbragg:
OK, this is finished, or so I think. Please review and merge at your earliest convenience. 

One thing though, I put all the new Models/c172p.xml code a the end of the file instead of breaking it down and inserting it were it more appropriately belongs. I can do that before we merge if I get a "yes, we're going with this".

wlbragg:
Oops not quite, still needs flashlight effect applied.

gilbertohasnofb:
> Oops not quite, still needs flashlight effect applied.

Let me know when it's applied and I will test and merge this.

wlbragg:
Done

gilbertohasnofb:
Ok, I am testing it right now.

gilbertohasnofb:
Okay, here are some thoughts:

- I think you should bind the `g` key to the landing gear in `c172p-keyboard.xml`
- bug: the gear lever is being shown in all variations, see:
fgfs-screen-001
- I think those two bars should stick to the panel and not the fuselage, more or less around the compass:
tigert-photo

wlbragg:
>I think you should bind the g key to the landing gear in c172p-keyboard.xml

It is I think. "g" also controls the gear.

I'm working on the other now.

gilbertohasnofb:
> It is I think. "g" also controls the gear.

Really? It didn't work on my first test. As soon as you update this PR I will test it again, including the `g` shortcut.

wlbragg:
Pushed Gearswitch fix, still looking at the support bars.

wlbragg:
Yeah, your right on the support bars, that'll take a bit. It's not easy at my Blender expertise level.
I'm working on it now.

gilbertohasnofb:
Okay, so the keyboard binding worked for me now (I am having problems with the FailureManager, I think that it was the reason why it didn't work the first time). Also, the gear switch is fixed on all the other variants, thanks!

As for the editing the object in Blender, I guess you could just change the position a bit and that's all, just stick it right into the panel around the compass and that should be it for now :smile: 

But here is a "problem" I have: with the amphibian on a runway (and gears down, of course) as soon as I start moving it turns so much to the left, regardless of my rudder input. This also happens regardless of the water rudder position. Have you seen anything like this?

wlbragg:
I felt like I had to give it hard right the one time I tried it after putting in the instrument, how can that effect the steering? I thought it was just the way the original amphibious might have worked, maybe not. We need to look at the properties it's using to put the gear up and down and make sure they are the same as the default c172p "g" mapping is.

As far as the crossbars, I want them in there correct. My compass isn't showing and the GPS needs to be taken into account?

gilbertohasnofb:
> I felt like I had to give it hard right the one time I tried it after putting in the instrument, how can that effect the steering? I thought it was just the way the original amphibious might have worked, maybe not. We need to look at the properties it's using to put the gear up and down and make sure they are the same as the default c172p "g" mapping is.

I don't know, I just tried now twice: autostart and power on, and the plane goes crazily to the left even with full rudder to the right. I have not flown the amphibian almost at all so I don't know if this behaviour has always been there, or even if i am doing something stupid here!

> My compass isn't showing and the GPS needs to be taken into account?

Why your compass isn't showing?? As for the GPS, it's the GPS that should move, not the bars :wink: So when the GPS is ready, we can just position it somewhere else in the cockpit for the float and amphibious in case of collisions

wlbragg:
My compass is showing now, weird. OK, it's going to take a bit to get these bars where I want them. It may not be until late tonight.

I can switch branches and see it the amphibious pulls for a first test.

gilbertohasnofb:
> I can switch branches and see it the amphibious pulls for a first test.

I just tested with the `master` and the left pull is also there... :sob:

wlbragg:
It's master, not what I did.

wlbragg:
Yeah, I posted the same time as you.

gilbertohasnofb:
Okay, so should I open a new issue for that and then we try to solve it in this branch here?

wlbragg:
I guess a new issue would be best.

wlbragg:
I really don't care, I'm going to concentrate on getting what I started finished first.

gilbertohasnofb:
> I really don't care, I'm going to concentrate on getting what I started finished first.

Sure!

> I guess a new issue would be best.

Here it is: https://github.com/Juanvvc/c172p-detailed/issues/515

Models/c172p.xml
```diff
@@ -6955,4 +6888,252 @@
...
+            <button>0</button>
+            <repeatable>false</repeatable>
+            <binding>
+                <command>property-toggle</command>
+                <property>controls/gear/gear-down</property>
+            </binding>
+        </action>
+    </animation>
+
+  <!-- TEXT OSG -->
```
onox: Can you give this 4-space indentation like code above it?

wlbragg:
I think we are squared away on this.

model-combined-transparent and model-combined-deferred bot inherit model-combined and use the same cube mapping from what I could tell. I'm not sure where the differences are. I switched to deferred solely because I use another aircraft's example and that is what they were using. Transparent is giving us what we want and I think it is the correct one to use so I changed it.

OK, back to work. I am going to finish the support bracing and push again and I think we will be done here?

wlbragg:
@onox, I'm not sure with our compass positioned where it is the adding the cross piece will be visible at all from the inside. You will only see it from the outside, looking down from the top, close up.
Also note, we are moving it from its current location to right behind the compass.
Here are some un-textured images of the potential new placement. I am in the process of finalizing the position. It is difficult at best so I only want to do this once. Obviously I still need to move it back a little bit or down (which would give it even a bigger gap between the two supports), so it is not touching the compass. I have to quit for the night. @onox and @gilbertohasnofb, please review and advise where (precisely) you think we need to position the ends that are going into the dash and I will finish it up tomorrow late morning.

fgfs-screen-007
fgfs-screen-009
fgfs-screen-006

wkitty42:
FWIW: the pontoon craft both tried to go left just like the standard craft... the push was maybe slightly greater but not so much that one could not take off down the runway... remember that i flew the amphibious almost as soon as it appeared and if this steering problem had been apparent, i would have said something then...

onox:
@wlbragg It seems you got the compass fixed! :+1: 

There is one problem that you need to fix: you're using `gear/gear[4]/position-norm` up to `gear/gear[7]/position-norm`. But for the amphibious you need to use 19 up to 22 instead. 4 .. 7 are JSBsim contact points.

You can see that currently the gear lights in the cockpit don't show the correct state in the flight recorder. So just try to use 19 .. 22 in `Models/c172p.xml` and then replay mode should work.

onox:
@wlbragg Actually, make sure you use `alias="/params/gear-amphibious/front-left-position-norm"`, etc.

wlbragg:
Gotcha!

wlbragg:
Hopefully that is all it needs.

onox:
Let me check. Anything else that needs to be done or can I merge it after testing?

wlbragg:
I think I am done.

onox:
@wlbragg Which issues does this PR fix?

onox:
#507, #313, and #206?

wlbragg:
#313, #314, #206

Leave #507 open for now.

wlbragg:
#314 was addressed somewhere, I don't remember where.

#507, is back, I don't get it. It was working and now it's not. I need to make sure I got it fixed.

onox:
I prefer not to close #314 until we can get a strong reflection like in https://github.com/Juanvvc/c172p-detailed/issues/314#issuecomment-113593767.

wlbragg:
OK, I'll keep at it till I figure it out.

gilbertohasnofb:
@wlbragg Looks beautiful

wlbragg:
Thanks gilbertohasnofb, it's all in the texturing! :wink:

Issue #424, opened 2015-07-15
Both walker and pilot visible from certain views


gilbertohasnofb:
The pilot model (if present due to enough weight and if the "show models" option is on) is programmed to be shown in every view except in the following: if in the cockpit view, if in the walker view, if in the walker orbit view. 

This means that the pilot will be seating in the cockpit AND the walker will be outside if we chose another view while the walker is enabled (let's say model view, or cockpit right seat view). 

We should try to find a better way to identify if the walker is present rather than using the Walker view as we have in the current conditions for the pilot model to be displayed:

```xml
<not-equals>
    <property>sim/current-view/name</property>
    <value>Walk View</value>
</not-equals>
<not-equals>
    <property>sim/current-view/name</property>
    <value>Walker Orbit View</value>
</not-equals>
```

PR #513, opened 2015-11-05
fixing some walker bugs

### LINKED ISSUE #423: Walker has the wrong gender
Our walker has the wrong gender: while our pilot model is female, the walker is male.

gilbertohasnofb:
Fixes #423 and #424.

I had to make a local copy of the `walker-include.xml` in order to change the gender of our walker into the correct one. The model is still shared though.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -6119,7 +6119,7 @@
         </offsets>
     </model>
 
-    <!-- Pilot not visible in the cockpit AND walker modes or if not enough weight on his seat -->
+    <!-- Pilot visible only when not in the cockpit, the walker is not activated and if there is enough weight on his seat -->
     <animation>
         <type>select</type>
         <object-name>Pilot</object-name>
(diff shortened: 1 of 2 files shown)


wlbragg:
I'll leave this for @onox to review and close out, I know next to nothing about the walker code, but it looks good.

gilbertohasnofb:
@wlbragg no worries!

onox:
There's no need to include the whole XML file, you can just override settings. Put the following in `<sim>` (you can put it before `<aircraft-data>` for example):

```xml
<walker>
    <gender>1</gender>
    <altitude-ft>-5.0</altitude-ft>
</walker>
```

Modify `c172p-set.xml`. Remove the included walker file: `git rm Systems/walker-include.xml`.

Then commit with `git commit --amend`. Check results with `git log` (or `tig`). You should see 1 commit, but with different commit id. Force push.

onox:
(Removed milestone so that only issues show up in the milestone page)

gilbertohasnofb:
I tried overriding it before but it didn't work out, hence why I copied the whole xml. I will try again as soon as I can, using your instructions above

gilbertohasnofb:
@onox Seems to have worked out.

PR #521, opened 2015-11-06
Adjust exterior glass reflection.

wlbragg:
I think it is the shape of our glass that is limiting this effect. Even with that said, I think this is pretty good.
See #314


onox:
There are still issues with the glass as mentioned in https://github.com/Juanvvc/c172p-detailed/issues/314#issuecomment-154606750

It's either Blender, a conflict between multiple effects or a regression in the shaders.

I'll merge this PR, code is OK and there's not much else we can do about the glass.

PR #523, opened 2015-11-07
Move glass rain effect out of nasal and into FDM/XML. Lower rain speed.

wlbragg:
There is more work I want to do to this but please go ahead and merge this after review. All this did was take it out of nasal and put it in FDM/XML and lower the rain speed.

I'm not sure how long it will take me to get the rest of what I want to do finished and in the meantime I'd like these changes added to master.

The rest of the work I want to do includes smoothing the standard flow of rain over the windshield during different stages of airspeed. Right now the top end of the airspeed which I lowered to 40 is maybe still a little to fast but not bad. However, at the low range of speed there are areas that the rain flow over the windshield accelerates at an unacceptable rate. This has nothing to do with taking it out of nasal, it is a flaw in the basic structure of the original computations. I have an idea of what is necessary to smooth it out. This I think is a mandatory fix that needs to be done. Another needed "fix" is rain movement or speed shifts at prop off/on
After that I am still considering adding some effect due to prop wash or rmp which is the tail end of prop on/off. And finally, maybe, wind speed calculated to have an effect if high enough.

But to reiterate, the current change from nasal to FDM/XML is a definite improvement over what we had so I would like to have it merged now in case I don't get to the rest of it right away.


Systems/c172p-glass-effects.xml
```diff
@@ -0,0 +1,70 @@
...
+
+<system name="c172p-glass-effects">
+
+    <channel name="rain">
+
+        <fcs_function name="glass-effects/splashx">
+            <function>
+                <difference>
+                    <value>-0.1</value>
+                        <product>
```
onox: You can unindent this whole block by 4 spaces.

Systems/c172p-glass-effects.xml
```diff
@@ -0,0 +1,70 @@
...
+                            <value>2.0</value>
+                            <sqrt>
```
onox: It seems the whole `<sqrt>` block is the same in the `splashz` function as well. Is this always true? If yes, then you should move it to a separate `<fcs_function>` above the current two functions.

Systems/glass-effects.xml
```diff
@@ -0,0 +1,38 @@
...
+<PropertyList>
+
+    <filter>
+        <name>splashX</name>
+        <type>gain</type>
+        <input>
+            <property>/fdm/jsbsim/glass-effects/splashx</property>
+        </input>
+        <output>
+            <property>/environment/aircraft-effects/splash-vector-x</property>
```
onox: Have you tried moving `<property>/environment/aircraft-effects/splash-vector-x</property>` to within `<fcs_function>`? (At the same level as `<function>`)

onox: Have you tried adding `<output>/environment/aircraft-effects/splash-vector-x</output>` to within `<fcs_function>`? (At the same level as `<function>`)

Systems/c172p-glass-effects.xml
wlbragg: Oops!

Systems/c172p-glass-effects.xml
wlbragg: OK, good idea.

Systems/glass-effects.xml
wlbragg: I thought about that but I didn't think you could get out of /fdm/jsbsim/ space to /effects/ which is what the shader I looking for?
I'm looking for a way to override the shader but am getting conused.

wlbragg: I didn't know you could output like that, much better.

onox:
@wlbragg @dany93 @tigert The raindrops behave fairly steady above 40 knots, the raindrops are not affected that much when accelerating or decelerating (unlike when the max speed was 140), but below that speed the behavior gets a bit weird. Is this the behavior we want?

wlbragg:
>The raindrops behave fairly steady above 40 knots

Yes, I think that part is good. The high rate of movement we had at 140 was unrealistic according to some video we have seen. That is why I dropped the high end down.

>but below that speed the behavior gets a bit weird. Is this the behavior we want?

No, that is what I have left to do. It is probably more noticeable now that our range is 0-40 VS 0-140. We're in the "bad zone" longer (5-15?). I am going to try to do the following.

0 airspeed and no prop spin (0 rpm) = static unless wind blowing, then it may effect the raindrops . The wind affecting raindrops will be the last thing I do if at all.

Prop rpm will affect the rain but only slightly at  low rpm and more at higher rpm. (no airspeed).

Airspeed will be clamped to the "sweet range" (>15).

I think if I can implement all this it will hopefully behave the way we want.

dany93:
@wlbragg wrote:
>Sitting stationary no prop rpm, the rain should flow down off the windshield?
Sitting stationary with prop rpm, the rain should flow out and up off the windshield.
Moving the airspeed should increase the rain flow out and up off the windshield.

For me, that's correct.

Issue #504, opened 2015-11-04
Simulate fuel contamination by water


gilbertohasnofb:
Simulate fuel contamination by water and how it affects the engine.

Split from #303.

Juanvvc:
This is a very unlikely event: I've never found water in the fuel deposits after about 150 flights. You must check always if there is some water in the deposit, of course, but I think this issue must be of a very, very low priority for the project :)

If you want to educate pilots, we can define a hotspot on the valve. If the user clicks on the valve, a message will appear on the screen "Fuel tank checked: no water"

gilbertohasnofb:
> This is a very unlikely event: I've never found water in the fuel deposits after about 150 flights.

I was thinking about trying to find how unlikely it is and trying to randomize it around that chance. And then if the fuel is contaminated, just clicking on a button would fix it. But if it is really that unlikely then maybe I'd focus on other stuff. Nevertheless, I am reading about it now and then I will see, as if I do get the a reasonable value for the chance of contamination, it would be very easy to implement anyway

> As a first approach and if you want to educate pilots, we can define a hotspot on the valve. If the user clicks on the valve, a message will appear on the screen "Fuel tank checked: no water"

That's kind of cool :smile: I think I will work on that

gilbertohasnofb:
@Juanvvc little question: if contamination is detected, what is the procedure to follow?

gilbertohasnofb:
Also, one more question ludo: from which places under the wings can one take fuel samples? How many holes on which wing for this?

Juanvvc:
To detect water in the tanks, you get a sample of the liquid inside the tank (about 50ml). If it is blue, then it is all fuel. If it is transparent, then there is water at the bottom of the tank. Remember water is more dense than fuel and it goes to the bottom. If you get water, you just continue draining the tank until you get a blue liquid.

There are many draining valves at the bottom of the wings and engine. Check this video: https://www.youtube.com/watch?v=dqCqHlAn3b0 (minute 8:50) In the video, they say you must get samples from all the valves but I sincerely think that is too much trouble... at least, in an aircraft that flies frequently. I guess if an aircraft is landed for a long time, a big amount of water condensation in the tanks is much more likely. The valves in the wings allow checking for water in the fuel tanks, and the valves in the engine are for checking the carburetor.

onox:
Even though unlikely, I think this is still a cool (low priority) feature :smile:

gilbertohasnofb:
[...]

PR #534, opened 2015-11-10
fuel contamination et al.

### LINKED ISSUE #528: Separate left and right tiedowns and their hotspots into different objects
Separate left and right tiedowns and their hotspots into different objects

### LINKED ISSUE #530: No indicated airspeed
Pulled "Latest commit 4358ae0"

Indicated airspeed remains at 0 on the instrument.
`instrumentation/airspeed-indicator/indicated-speed-kt` is (none) or nil

Not so important, but I'd like to know:
I wished to come back to my previous version, but I could not.
My local branch is with this bugged version.
[My remote](https://github.com/dany93/c172p-detailed) is still with the previous one (ebb386b)
Attempts: I did
`git merge --abort`
then 
`git reset -- hard`
then
`git pull' (from origin/master, which is my dany93 remote)
With no success ("Your branch is ahead from 'origin/master' by 89 commits")

I did **no git push**. My remote branch is still with my previous version.
[EDIT] After gilbertohasnofb response, **I did a git push** to my remote (dany93). Everything is OK.
But if you can (thank you) I'm still interested in the easiest procedure to come back to the previous version in case of a bugged one. I guess there is an immediate way.

How can I recover to my previous version (or one which works) by the easiest way?
`git reset --hard ebb386b` ??

I hope that my attempts did not do too much mess and I do not dare to go farther...

gilbertohasnofb:
Implements fuel contamination, creates separate left and right tie-downs so during pre-flight inspection one must untie each wing separately, fixes some small bugs.

Closes #528 and #504
Fixes #530


dany93:
Bug #530 is not fixed.
- I check "Allow securing aircraft", check "Put pitot tube cover", I have no airspeed indicated. **OK**.
- If I uncheck "Allow securing aircraft" without explicitly unchecking "Put pitot tube cover"(pitot cover unchecks by itself), I have no airspeed.
- By checking again  "Allow securing aircraft" then checking "Put pitot tube cover", I can have the airspeed indicated with the pitot cover ON.

gilbertohasnofb:
@dany93 ok, I will check that again in a short while.

gilbertohasnofb:
Okay, I think I fixed it this time, would you please give it a try once again, @dany93?

dany93:
Weird... Still not fixed, same anomaly with indicated airspeed.
I have done
`git fetch upstream`
`git checkout bug-504`
Is it correct ?
Can someone else test this?

gilbertohasnofb:
I think you must do a `git pull` after switching to branch `bug-504`

dany93:
You are right.
**That's OK**, the indicated airspeed works correctly now in accordance with the pitot tube presence, whatever my checks / unchecks.

Sorry.
That was probably because I still had "bug-504" branch in my local list. `git pull` is not needed for the first checkout of a branch. Even so, my files seemed to be updated at the time of last checkout. git is often hard to understand...

gilbertohasnofb:
> git is often hard to understand...

Tell me about it! But I am glad the bug has been solved.

gilbertohasnofb:
@onox do you think you can take a look at this when you have time? I can try to help with solving the conflicts if you want.

wlbragg:
I fixed the conflicts.

gilbertohasnofb:
@wlbragg Thank you so much!

onox:
@gilbertohasnofb Testing...

onox:
**The good:**

When I put on the pitot tube cover, the needle doesn't move and when I disable the "Allow securing aircraft" button, it moves again.

Colored text in "Fuel Tank Sampling" window is a nice detail.

**The bad:**

With the pitot tube covered, the airspeed indicator doesn't always point to 0 knots. Instead it seems to indicate the wind speed from the weather dialog. (Try to create a wind of 40 or 50 knots)

**The ugly:**

When you click on the wing under the sample, it returns a result even if there's no fuel in the tank at all. The "Return sample to the tank and close" button doesn't seem to return the fuel to the fuel tank. It seems the fuel isn't taken from the tank when you open the "Fuel Tank Sampling" window, but rather when you click the "Discard sample and take another one".

Perhaps you could extract fuel when there's fuel in the tank. Store the extracted fuel in a variable. When you click the "Discard" button you set the variable to 0 and try to extract fuel again. When you click the "Return" value, you add the value in the variable back to the fuel tank.

"Close" would have to discard or return the fuel.

Juanvvc:
The ASI takes two inputs: the dynamic air pressure from the pitot tube and the static pressure from the static valve (next to the pilot's door) Then, the ASI shows the DIFFERENCE of these two inputs, since the difference is directly related to the airspeed. If the pitot tube is covered (by the red cover or by an insect, doesn't matter), the input from the static valve may be still different to the input from the tube, the difference is non-zero and the ASI does not indicate 0 knots but something different.

As a result: If the pitot tube is covered, the ASI indicates wrong speeds but not necessarily 0 knots.

On the other hand, the amount of fuel you took to sample the tanks is very, very small compared to the tanks capacity. The reason to return the sample to the tanks is to avoid lead contamination in the environment, rather than saving fuel.

gilbertohasnofb:
@Juanvvc actually, the problem with the AI is that when the property `serviceable` is set to `false`, the displayed airspeed isn't set to `0` but rather freeze on the last known position. I am not working on forcing it to 0 if the pitot cover is added, which will solve the problem.

As for the fuel, even though the amount is very small, 50ml is still noticeable in the fuel dialogs. Anyway the UI I created isn't really intuitive as onox pointed out, so I am fixing all this right now.

gilbertohasnofb:
Thanks for testing this, @onox!

So:

- I figured out what was happening with the pitot tube: if I simply set the wind to 50 knots and the pitot tube was on, the AI was correctly displaying 0 knots. But if I manage to put the pitot tube on with that wind, the AI will freeze on its last indicated airspeed (which can be very high, since the pitot tube can be put on at __ground speeds__ lowed than 1 knot). So: if pitot tube is on, I force the AI not only to be unserviseable, but also set it to 0 knots now. I tested it a lot and it seems that should do.

- Yeah, I see how the UI I created for fuel sampling can be confusing. I tried making it clearer now, so see what you think. Now you have to click on a button to take a sample, which diminishes the fuel. You can discard this sample or return it (fuel is returned, as well as any water on it). The buttons are enabled depending if you have a sample in your hands or not: that is, you mut discard or return a sample in order to take another one, and you must have a sample in hand in order to discard or return it.

Thanks for the input, I think it's looking better now. See what you think and let me know if I should try to improve something else, okay?

gilbertohasnofb:
Thanks for testing this, @onox!

So:

- I figured out what was happening with the pitot tube: if I simply set the wind to 50 knots and the pitot tube was on, the AI was correctly displaying 0 knots. But if I manage to put the pitot tube on with that wind, the AI will freeze on its last indicated airspeed (which can be very high, since the pitot tube can be put on at __ground speeds__ lowed than 1 knot). So: if pitot tube is on, I force the AI not only to be unserviseable, but also set it to 0 knots now. I tested it a lot and it seems that should do.

- Yeah, I see how the UI I created for fuel sampling can be confusing. I tried making it clearer now, so see what you think. Now you have to click on a button to take a sample, which diminishes the fuel. You can discard this sample or return it (fuel is returned, as well as any water on it). The buttons are enabled depending if you have a sample in your hands or not: that is, you mut discard or return a sample in order to take another one, and you must have a sample in hand in order to discard or return it. Also, the take sample button is only enabled if there is at least 50ml of fuel in the tank.

Thanks for the input, I think it's looking better now. See what you think and let me know if I should try to improve something else, okay?

onox:
As @Juanvvc said "ASI indicates wrong speeds but not necessarily 0 knots", this is something that would properly be fixed by fixing the C++ instrument in FG, but setting it to 0 knots is an acceptable workaround :+1: 

Dialog is much better now! :smile:

gilbertohasnofb:
Great, thanks! I will work on the oil management now, as well as in a function to make the engine stutter (I already made some tests and I think I can come up with something nice).

As for the ASI, I understand what @Juanvvc said. Maybe we could use a filter to model that? So that we can convert true airspeed -> broken airspeed using some sort of formula (I assume that with the pitot tube cover on, the airspeed indicated will always be less than if the cover was removed, right?). Here is a new issue for it in case we come up with a solution: https://github.com/Juanvvc/c172p-detailed/issues/538

onox:
I think issue #538 should be fixed in FlightGear itself, so that it gets fixed for all aircraft.

gilbertohasnofb:
Well, I am simply forcing the ASI to be unserviceable. As I had mentioned before, failing the pitot system wasn't working at all. So yeah, if they would be fixed in FG itself so that aircraft developers can simply use a property such as `pitot-tube-blocked` = `true`, that would be ideal.

dany93:
What about Equipment > System failures ?
Prop sim/failure-manager/Instrumentation/airspeed-indicator

gilbertohasnofb:
I think that property simply freezes the AIS at the current reading, which is the same as the property I am using (I am forcing the AIS to 0 knots after I make it unserviceable).

Issue #496, opened 2015-11-02
Movement of raindrops on side windows should depend on airspeed and AoA


wkitty42:
The movement of the raindrops on the side windows should depend on the airspeed and AoA. Also, if the aircraft is not moving at all, then there should be little or no rain at all because the side windows are sheltered by the wings (acting like a roof).

Rain spots from blown rain is ok and maybe some running if there's enough drops accumulated together to overcome the adhesion factor and succumb to gravity making them roll down the window but certainly there shouldn't be any running down the door windows because of the wings acting like a roof.

gilbertohasnofb:
I think this is one of the limitations of the techniques we use for rain drops, shadows, etc. Although I wouldn't completely agree that a side window would not have rain in a thunderstorm, let's suppose that the rain is coming vertically down and lightly so the window in a parked aircraft should be more or less dry. So now if the aircraft start moving, we would need to factor the angle of the rain coming, the speed of the aircraft, the intensity of rain of course, and fake the transition from dry window to thunderstorm wet window. This is basically impossible to be done I think, and even if it could I think it wouldn't worth the effort to be frank.

gilbertohasnofb:
I will leave this opened until we hear more opinions (@onox, @wlbragg), but my 2 cents is that we shouldn't bother.

wlbragg:
Yeah, I wish we cold be a little more precise but I don't think we can. Maybe a little bit of programming could improve on it, but it wouldn't be much improvement for the amount of effort it would take. About the only think we could do would be to remove the rain from the side windows. I guess we just have to imagine that we get no gentle rain in the FG world, only gales! :smile:

wkitty42:
On 11/02/2015 06:23 PM, Gilberto Agostinho wrote:
> I think this is one of the limitations of the techniques we use for rain drops,
> shadows, etc.

yes, i understand that... that's why i mentioned about the possibility if the 
cube map might be able to be broken apart or (new thought) even have a second 
cube map with holes where the door windows are that could be switched out for 
the full cube map when rolling above a certain speed while on the ground...

> Although I wouldn't completely agree that a side window would not
> have rain in a thunderstorm, let's suppose that the rain is coming vertically
> down and lightly so the window in a parked aircraft should be more or less dry.

right... i didn't say that the door windows would be completely dry... that's 
why i mentioned wind blown rain on them... what i'm mainly looking at is the 
drops striking the door windows and running down them as if the wings weren't 
even there and sheltering them...

[...]

PR #537, opened 2015-11-10
Interpolate rain on glass effect

wlbragg:
I'm not totally satisfied, but please let me know what you think.

The side windows are running on as alternate set of shader instruction. That is why it works as it does and you can't really do much with it. Thorsten changed tho logic on it for use awhile back because it didn't work well using the same code as the front. To see what it used to look like you can comment out 
~~~xml
<surface-mapping-scheme type="int">2</surface-mapping-scheme>
~~~
in Models/Effects/interior/c172p-interior-glass-reflection-side.xml

~~I also noticed the back window is basically backwards, the only way to change it would be to make it a separate object and code its own logic for the rain.~~
I think this would need a new shader effect written for it.

It still flows down on deceleration which I guess I need to ask Thorsten about.


onox:
I think there are still some problems that need to be fixed before we can merge this:

1. The effect on the side part of the front window is now very soft to almost none. The rain on the side windows is more diagonal (not vertical), which is good, but it looks a bit weird.

2. When I'm in the air with a reasonable speed and then reduce power to idle, the rain starts shivering. Maybe we can use a low-pass filter or make one of the input filters (xa/xr/za/zr) a noise-spike filter.

3. Sometimes the rain even goes backwards.

@gilbertohasnofb @dany93 @tigert What do you guys think of the branch as it currently is?

@wlbragg Is this PR fixing any issue we're tracking?

gilbertohasnofb:
I haven't seen the effect going backwards as @onox did, but like him I also felt that the effect is too soft on the sides of the front glass. The side windows are fine by me, so would it be possible to maybe tweak them separately?

wlbragg:
>The effect on the side part of the front window is now very soft to almost none.

I didn't notice, is that at full rmp/airspeed "only" that it appears to  soft?

>Is this PR fixing any issue we're tracking?

#496

@gilbertohasnofb, no, you can't tweak them separately. There is no mechanism for that. Only a switch to tell that one is a side window and to handle it differently. The only values that the shader effect are spash-x, y and z. Any object with the a rain shader applied to it will pick up that on value and apply it through the shader.

>the rain starts shivering

I noticed that during one iteration, but I thought I got rid of it in my last pass.

>Sometimes the rain even goes backwards.

I can't verify this but the backwards rain may have always been there. It is happening on pitch up and deceleration. I don't know why. I guess on this one we need to ask Thorsten.

onox:
Perhaps someone should ask Thorsten on the forum if he can work on the effect a bit more to make it more robust and easier to tune?

@wlbragg I still notice tabs in your code sometimes. Are you switching between tabs/4-spaces for different projects? You know you can configure any editor to insert 4 spaces automatically when pressing the tab key?

Sorry for nitpicking about this, but consistency is important, otherwise the diffs become unnecessary chaos.

onox:
> I didn't notice, is that at full rmp/airspeed "only" that it appears to soft?

I think when I started the airplane. After takeoff the rain was gone on the side parts of the front window.

About the backwards rain: IIRC it is when reducing power to idle and at low IAS.

wlbragg:
@onox, I'm pretty sure the shuddering is the airspeed and the rpm logic fighting for control, or better put, swapping control. We need to change that logic to allow only airspeed or only rpm through greater ranges. Something like airspeed > 20 it is airspeed exclusively then < 20 RPM can take over. I'm not sure how to code that using the filters as we are.

wlbragg:
Something like
~~~xml
        <expression>
            <else>
                <greater-than>
                    <property>/environment/aircraft-effects/splash-xa</property>
                    <value>20</value>
               </greater-than>
               <property>/environment/aircraft-effects/splash-xr</property>
            </else>
        </expression>
~~~

wlbragg:
Never mind I think I found it, 
~~~xml
<ifthen>
~~~

gilbertohasnofb:
@wlbragg where can I find information about `<ifthen>`?

wlbragg:
I tried 
~~~xml
<filter>
        <name>splash-x</name>
        <update-interval-secs type="double">0.1</update-interval-secs>
        <type>gain</type>
        <gain>1.0</gain>
        <input>
            <expression>
                <ifthen>
                    <gt><p>/velocities/airspeed-kt</p><v>20</v></gt>
                    <p>/environment/aircraft-effects/splash-xa</p>
                    <p>/environment/aircraft-effects/splash-xr</p>
                </ifthen>
            </expression>
        </input>
        <output>
            <property>/environment/aircraft-effects/splash-vector-x</property>
        </output>
    </filter>
~~~

But I only get airspeed.

wlbragg:
I've been referring to 
http://jsbsim.sourceforge.net/JSBSim/classJSBSim_1_1FGFunction.html

gilbertohasnofb:
Thanks!

onox:
But this isn't JSBsim. So you need to use multiple `<input>` elements, each containing a `<condition>`. See `Systems/views.xml` for an example.

wlbragg:
Thanks

wlbragg:
I don't think I am going to get it any better than this. I have the side windows working as before they got "soft". I have the rear window moving in the correct direction and the front window rain has bot h RPM and then airspeed factored in. The way it works now is RPM-no airspeed, at idle up to max RPM, slight movement. Airspeed kicks in at 20 kt. I think we're to a point of diminishing returns to go any farther. I spent a lot of time testing and adjusting to finally decide on this configuration. If you try to account for to many conditions then you start running into little quirks that spoil the whole effect.
Try this out and let me know what you think.

Also I have a typo in the commit description and have no idea how to correct it.

wlbragg:
After a little more testing I noticed a slight stutter when you pitch up at a precise point. I'm not convinced that this has always been there. Also the pitch up still has a moment of slowing to stopping to rolling the rain backwards, again, I'd be willing to bet that has always been there too.

onox:
I will look at this PR tomorrow. Don't have much time today.

wlbragg:
No hurry, no worries!

wlbragg:
I am glad I had one more go at this. I lowered the airspeed needed to switch over from RPM to airspeed and increased the upper speed limit of the rain. It is really close to what we had originally plus the addition of RPM. About the only thing I might add would be to increase the affect of the RPM. But the change over from RMP to airspeed control needs to be synchronized.
I'm actually pretty happy with where it is now.

It appears that the new numbers have affected the angle and flow over the back window. I think we need to ignore the back window, you can't change it without affecting the front.

@onox,

~~~xml
 <entry><ind>   50 </ind><dep> -2.1  </dep></entry>
 <entry><ind> 170 </ind><dep> -2.1  </dep></entry>
~~~
Can the 170 entry be removed? If removed, will the 50kt entry "=-2.1" hold true all the way up to max airspeed without the upper 170 limit?

onox:
Yes, it should. If the last line is (50, -2.1) then the output will be -2.1 if you reach 50 knots. If you go faster, 100 knots, 200 knots, 10_000 knots, the output should still be -2.1.

wlbragg:
OK, this is as good as I am going to get it. I like the RPM rev up except one point it reverses for a brief second, I can eliminate that part but then the overall effect isn't as nice. After the airspeed takes over it seems to work well.

Issue #536, opened 2015-11-10
Create function for coughing engine


gilbertohasnofb:
Create function for coughing engine

gilbertohasnofb:
Closed via https://github.com/Juanvvc/c172p-detailed/pull/540.

Issue #535, opened 2015-11-10
Apply cough sound when engine killed


gilbertohasnofb:
Apply cough sound when engine killed

gilbertohasnofb:
Closed via https://github.com/Juanvvc/c172p-detailed/pull/540.

Issue #41, opened 2015-03-12
carb heat does not drop RPM as expected


gilbertohasnofb:
PRC01 - carb heat does not drop RPM as expected

gilbertohasnofb:
I am reopening this as it's better so separate it from #360.

As for the RPM drop once the carb heat toggle is pulled, how much percent are we talking about? I read that the RPM drop is around 80 RPM at cruise RPM, which is around 3,5%. Can anyone confirm these numbers to me? @Juanvvc, @tigert?

gilbertohasnofb:
Other people mention much smaller drops, such as 25 RPM, 35 RPM, 50 RPM.

gilbertohasnofb:
A technical question, maybe @onox can help me with this one: if I create a `<filter>` which lowers the value of the property `/engines/active-engine/rpm`, I do see a RPM drop in the gauge, but that's not a real drop in power: the engine still effectively runs at the non-filtered RPM. Any recommendations on how can I force a loss of power as well as indicated RPM?

dany93:
Engine RPM cannot be imposed, it is the **result** of throttle position, propeller moment due to the air, all of this giving a power depending on the max power of the engine.
I don't know if that is the best solution but I'd try with `/fdm/jsbsim/propulsion/engine[<N>]/air-intake-impedance factor`.
This is a bit cheating. If someone with a better knowledge has a better idea?  (@andgi?)

Juanvvc:
The "carburetor heat" does not modify directly the RPM, but takes the air from a different place and heats the airflow by wiring the the air circuit close to the hotter parts of the engine. This way, the air feeding the engine is hotter, less dense and as a result the mixture is leaner than with the normal air intake.

I guess a first approximation to the issue is just leaning the mixture by a factor (0.8?) if the carburetor heat is on. Anyway, even if dany93 believes it is cheating :) and according to the documentation ( http://wiki.flightgear.org/JSBSim_Engines ), /fdm/jsbsim/propulsion/engine[<N>]/air-intake-impedance seems exactly the property to control not only to simulate the use of alternate air intakes like the carburetor heat, but also simulating intake icing. I guess icing is going to be the next thing to simulate after the carburetor heat.

gilbertohasnofb:
Thanks for the ideas, guys, I will try using `air-intake-impedance` property and see what I can manage.

gilbertohasnofb:
[...]

Issue #471, opened 2015-10-24
Gather a list of features and bugs we fixed in 2016.1


onox:
We should gather a list of all the new features, major changes, and bugs we fixed. Please create a list (use `* ` to create bullet points) here.

* Ski's
* Amphibious plane improvements:
    - Added two metal bars in the cockpit
    - Added panel for gear deploy and retract
    - Added water rudders retraction mechanism
* Saving state of switches is now optional
* Zooming with mouse scroll wheel in Helicopter View, Model View, and Walker Orbit View
* Additional views for co-pilot and passengers
* New sounds:
    - Rain (ambient and windshield) and thunderstorms
    - Wing damage
* Windshield rain overhaul
* Opening the baggage door displays a window which allows weight to be added. Baggage will be displayed if the weight is high enough
* ASI now shows the IAS instead of the CAS
* Adjusted stall horn volume to sound to about 5-10 knots above stall speed 
* Carburetor heat toggle now drops RPM as expected
* Oil consumption and management have been implemented
* Refactored fog/frost system: air flow now depends on airspeed and engine RPM. Supports overhead air vents, windows, and doors. Humans add extra heat and humidity.
* Preflight inspection improvements:
    - Added "Preflight Inspection" checklists and tutorial
    - Pitot tube can be clicked to add/remove a red "remove before flight" label
    - Nose wheel can be clicked to add/remove chocks
    - You can tie down the wings and tail to the ground. It will hold the aircraft even with full engine power.
    - Each fuel tank can randomly get contaminated with water. Samples from the fuel can be taken to test for contamination.
* Added GPS 196 device
* Fixed glass reflection in external views
* Improved lighting, including better looking navigation lights (and depending on viewing angle) in ALS
* Improved textures of instruments, primer, and cabin air/heat levers in cockpit
* Smooth instead of instant animation of yoke, pedals, and parking brake
* Added menu item to open a web panel in your browser, showing most instruments

gilbertohasnofb:
Hooray wrote in the forum:

> you should consider maintaining a changelog (if you aren't already) and addin this to the source tree itself, as well as to the wiki.
> You could also show a simple GUI dialog with the changelog for people wanting to learn more

What you guys think? @onox?

onox:
We can add a `NEWS.md` file with the release notes of 3.6 (and 3.8 just before creating the release), but a complete changelog is unneeded IMO.

gilbertohasnofb:
[...]

PR #540, opened 2015-11-12
Oil management implementation and other stuff

### LINKED ISSUE #499: Implement oil level management
Implement oil level management (splitting issue from #303, which is getting too large already)

@Juanvvc wrote:

>You must check the oil level always, and refill about once out of ten flights. I don't know if flighgear implements engine failures if the oil lever is too low, or they must be simulated by the aircraft.

@tigert wrote:

> Oil amount is usually measured in "quarts" (imperial unit pretty close to a liter), minimum 6 in that engine I think (depends on engine size, so the 180hp might have a bigger oil sump, need to verify from the handbook). If you plan for a longer cross country trip, you want to up it close to 8, since the lycoming engine does eat something like a quart in every 10 hours. For this reason you have a bottle or two in the trunk, and if it is much below 7qt when you check you add one quart before the flight.

### LINKED ISSUE #539: Fix typo in the fuel contamination dialog
Fix typo in the fuel contamination dialog: exchange "translucid" for "transparent" (sorry, that was my Portuguengrish :stuck_out_tongue_winking_eye: )

gilbertohasnofb:
- oil management implementation (consumed according to RPM)
- oil can be replenish using a special GUI activated by clicking on the oil sump door on the aircraft nose (contains warnings about low and critical levels)
- if oil level is getting too low (below 5.0 quarts), oil temp rises and oil pressure drops.
- if oil level critically low (< 4.92 quarts), engine is killed. Just before this point, at cruise RPM the oil temp should be super high and the pressure almost null
- different engines have different oil sump sizes just like the POH (160 HP has a 7 quarts sump, while 180 HP has an 8 quarts sump. Both get critical below 5 quarts).
- changing engines replenish oil level
- added master switch for oil management (off by default) on main aircraft dialog
- created engine cough function
- apply sound when engine coughing
- when oil level is just above the engine failure threshold (4.925 quarts > oil level >= 4.920 quarts) the engine starts coughing
- water contamination: if water in either tank is slightly contaminated (water on either > 0.0 and <= 0.2) then engine coughs. If water > 0.2 engine fails.
- fixes typo in the fuel contamination GUI
- oil has weight, with point of mass located slightly below the oil door
- carburetor heat toggle now diminishes RPM for both engines if pulled

Closes #499, #536 and #535
Fixes #539 and #41


Juanvvc:
Your work is really impressive! Congratulations!

Have you considered the weight of the oil?

gilbertohasnofb:
@Juanvvc Thanks a lot, ludo! :smile:

As for the weight, you are right, I forgot about considering it. So: at 70oC (160 Farenheit), which seems like the average temperature for the engine oil, the density of engine oil is 0.8477 g/cm^3, or 847.7 g/L. Given that 1 quart = 0.946353 L, the weight of 1 quart of oil is 802.22 g or 1.7686 lbs. So I just pushed an implementation of this, it seems to work well here for me.

gilbertohasnofb:
Of course we could of course try to get a table of oil weight according to the temperature, but this would complicate things quite a lot: when temperature changes, it's no the weight that changes but rather the volume of the oil. I'd say that using a fixed value such as I proposed above is the best approach, particularly when we consider that the variation we are talking about is minimal when comparing to the plane's weight.

Models/c172p.xml
```diff
@@ -2539,6 +2540,10 @@
                 <command>property-toggle</command>
                 <property>controls/anti-ice/engine[0]/carb-heat</property>
             </binding>
+            <binding>
```
onox: I would use `/controls/engines/current-engine/carb-heat` via an alias here (and on 2524) and add the property to `<params><controls>` somewhere at line 258. This prepares the aircraft for having controls transmitted over MP (I already did the same preparation for throttle and mixture)

Also add two filters that write an output to `/controls/anti-ice/engine[0]/carb-heat` and `/controls/anti-ice/engine[1]/carb-heat`. You can basically take the two filters for mixture (line 366 to 404), modify them, and insert the somewhere on line 405. Plus add a little comment block:
```
<!-- ============================================================== -->
<!-- Carburator Heat                                                -->
<!-- ============================================================== -->
```

You can replace `<!-- Fuel Contamination and low oil level-->` with a block like above as well. I use the three line comments to visually group the filters.

Nasal/engine.nas
```diff
@@ -88,6 +88,55 @@ var primerTimer = maketimer(5, func {
...
+        setprop("/engines/active-engine/low-oil-temperature-factor", low_oil_temperature_factor);
+    } else {
+        setprop("/engines/active-engine/low-oil-pressure-factor", 1.0);
+        setprop("/engines/active-engine/low-oil-temperature-factor", 1.0);
+    };
+
+});
+
+setlistener("/sim/signals/fdm-initialized", func {
+    oil_consumption.start();
```
onox: Can you instead add this line to the listener at the bottom of the file? Currently you create three listeners for the same property. You can group the code so that there's only one place where you listen for `/sim/signals/fdm-initialized`.

Nasal/engine.nas
```diff
@@ -88,6 +88,55 @@ var primerTimer = maketimer(5, func {
...
+    var coughing = getprop("/engines/active-engine/coughing");
+    var running = getprop("/engines/active-engine/running");
+    var delay = 3.0 * rand();
+    if ((coughing) and (running)) {
+        settimer(func {setprop("/engines/active-engine/killed", 1);}, delay); # the <logic> block in engine.xml which controls the engine killed status will force it back to killed == 0 if oil level is still all right
+    };
+});
+    
+setlistener("/sim/signals/fdm-initialized", func {
+    engine_coughing.start();
```
onox: Same thing here :wink:

Nasal/engine.nas
```diff
@@ -88,6 +88,55 @@ var primerTimer = maketimer(5, func {
...
+
+# ========== engine coughing ======================
+
+
+var engine_coughing = maketimer(3.0, func {
+    var coughing = getprop("/engines/active-engine/coughing");
+    var running = getprop("/engines/active-engine/running");
+    var delay = 3.0 * rand();
+    if ((coughing) and (running)) {
+        settimer(func {setprop("/engines/active-engine/killed", 1);}, delay); # the <logic> block in engine.xml which controls the engine killed status will force it back to killed == 0 if oil level is still all right
```
onox: Some very minor remarks:

1. The parentheses around the variables in the `if ()` are not needed here, so you can remove them. I would only add them if you have a complex boolean expression, for example, one that mixes `and` and `or`.

2. You can put the `delay` inside the body of the `if`. Try to make variables as local as possible. In this case the delay is only going to be used if the engine is coughing and running.

Systems/carburetor.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0"?>
+
+<!--
```
onox: Perhaps you could add a copyright block like we have in the other `Systems/c172p-*.xml` files? Also, we have (apparently) sort of a convention to prepend `c172p-` to files containing JSBsim systems in the Systems/ directory.

Systems/carburetor.xml
```diff
@@ -0,0 +1,38 @@
...
+        <fcs_function name="Carb heat on cuases RPM drop on engine 0">
+            <function>
+                <ifthen>
+                    <property>/controls/anti-ice/engine[0]/carb-heat</property>
+                    <value>0.35</value>
+                    <value>0.15185</value>
+                </ifthen>
+            </function>
+            <output>/fdm/jsbsim/propulsion/engine[0]/air-intake-impedance-factor</output>
+            <output>/fdm/jsbsim/propulsion/engine[1]/air-intake-impedance-factor</output>
```
onox: Shouldn't this (second `<output>`) be moved to the `<fcs_function>` below?

Systems/carburetor.xml
```diff
@@ -0,0 +1,38 @@
...
+***********************************************************************************
+Carburator icing and carb heat toggle
+***********************************************************************************
+ -->
+
+<system name="carburetor">
+
+    <channel name="icing">
+    
+        <fcs_function name="Carb heat on cuases RPM drop on engine 0">
```
onox: Typo: 'cuases' -> 'causes' (in function below as well)

Systems/oil.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+
+<!--
+***********************************************************************************
+Extra weight due to oil quantity
+***********************************************************************************
+ -->
+
+<system name="oil">
```
onox: Maybe you should combine `Systems/oil.xml` and `Systems/carburator.xml`. You could name the file `c172p-engine.xml`.Just them the functions in different channels.

Nasal/c172p.nas
```diff
@@ -270,6 +270,18 @@ var reset_system = func {
 }
 
 ############################################
+# Engine coughing sound
+############################################
+
+var coughing_engine_sound = func{
+    if ((getprop("/engines/active-engine/running")) and (getprop("/engines/active-engine/killed"))) {
```
onox: Unneeded parentheses here again.

Nasal/engine.nas
```diff
@@ -88,6 +88,55 @@ var primerTimer = maketimer(5, func {
...
+    var oil_level = getprop("/engines/active-engine/oil-level");
+    var rpm = getprop("/engines/active-engine/rpm");
+    # quadratic formula which outputs 1.0 for input 2300 RPM (cruise value), 0.6 for 700 RPM (idle) and 1.2 for 2700 RPM (max)
+    var rpm_factor = 0.00000012 * math.pow(rpm,2) - 0.0001 * rpm + 0.62;
+    # consumption rate defined as 1.5 quarter per 10 hours (36000 seconds) at cruise RPM
+    var consumption_rate = 1.5 / 36000; 
+    var low_oil_pressure_factor = 1.0;
+    var low_oil_temperature_factor = 1.0;
+    
+    if ((getprop("/engines/active-engine/running")) and (getprop("/engines/active-engine/oil_consumption_allowed"))) {
```
onox: Unneeded ... :wink:

Nasal/engine.nas
```diff
@@ -88,6 +88,55 @@ var primerTimer = maketimer(5, func {
...
+        oil_level = oil_level - consumption_rate * rpm_factor;
+        setprop("/engines/active-engine/oil-level", oil_level);        
+    }
+    
+    # if oil gets low, pressure should lower and temperature should rise
+    if (oil_level < 5.0) {
+        low_oil_pressure_factor = 11.25 * oil_level - 55.25; # should give 1.0 for oil_level = 5 and 0.1 for oil_level 4.92, which is the min before the engine stops
```
onox: So what are you trying to do here? Upperbounding `oil_level`? If you replace `oil_level` in the expression with `std.min(oil_level, 5.0)`, then you can remove the whole if check and `else` block.

onox:
Thanks for implementing this @gilbertohasnofb! :smile: Code looks good overall, just had a couple of small remarks. Still need to test if everything work from a functional perspective.

One more thing: about the checklists, they still assume `/controls/anti-ice/engine[0]/carb-heat`. You should use `/controls/engines/current-engine/carb-heat` instead (see one of the remarks about this property in `Models/c172p.xml` above).

gilbertohasnofb:
@onox thanks for the review and for all the patience with my mistakes. I will check all your comments in a short while and try to fix everything ok?

wlbragg:
@onox = "Mr. Patience"! :smile:

gilbertohasnofb:
Particularly when it comes to my code :wink:

Systems/carburetor.xml
gilbertohasnofb: I am using the new alias inside the `<ifthen>` block and I also now have two __proper__ functions, sorry for the mess. But what happened before was that the `carb-heat[0]` is being used by both engines, and I forgot to clean this messy part of the code (I was troubleshooting why the carb heat wasn't working for the 180 HP and I left this mess here). I now made this code look pretty, but I don't know where could I fix the problem (that is, where can I associate the 180 HP to `carb-heat[1]`)

Nasal/engine.nas
gilbertohasnofb: > So what are you trying to do here? Upperbounding oil_level (to 5.0)?

Not really. The upper bound of oil_level is 7.0 for the 160 HP and 8.0 for the 180 HP. What I am doing here is the following: between the critical oil level of 5.0 and the failure level of 4.92, I am generating two properties (`low_oil_pressure_factor` and `low_oil_temperature_factor`) which will make the pressure lower and temperature rise as the oil gets critical. They are used in two filters in `engine.xml`. Is that fine?

gilbertohasnofb:
@onox okay! So I fixed absolutely everything you wrote but added two comments below your notes. I also retested the plane and everything seems to be working well. Let me know what you think and if I should do any more corrections here, ok?

Systems/c172p-engine.xml
```diff
@@ -0,0 +1,67 @@
...
+        </fcs_function>
+        
+    </channel>
+    
+    <channel name="carburetor">
+    
+        <fcs_function name="Carb heat on causes RPM drop on 160 HP engine">
+            <function>
+                <ifthen>
+                    <property>/controls/engines/current-engine/carb-heat</property>
```
onox: I think you need to use `/controls/anti-ice/engine[0]/carb-heat` here, and in the function below you use `/controls/anti-ice/engine[1]/carb-heat`.

The idea is that the user (pilot) controls `/controls/engines/current-engine/carb-heat`. The value of that property is then forwarded to either `/controls/anti-ice/engine[0]/carb-heat` or `/controls/anti-ice/engine[1]/carb-heat` depending on the selected engine. (So that JSBsim uses the correct value). Furthermore, for displaying and checklists you use `/controls/engines/current-engine/carb-heat` again (via the params in the model XML file). (So that animation system uses the correct value).

So for input (controls) and output (display) we use `/controls/engines/current-engine/carb-heat`, but for processing (JSBsim) we use `/controls/anti-ice/engine[0]/carb-heat` or `/controls/anti-ice/engine[1]/carb-heat`.

gilbertohasnofb: Makes sense, I am fixing this straight away.

Nasal/engine.nas
onox: I don't see the number 7.0 and 8.0. When I look at the if-else block I see you test `oil_level < 5.0`. So let's try 3 values for the pressure: 4.99, 5.0, 5.01:

For oil_level = 4.99 we get: 11.25 * oil_level - 55.25 = 0.8875000000000028
For oil_level = 5.00 we get: 11.25 * oil_level - 55.25 = 1.0
For oil_level = 5.01 we get: 11.25 * oil_level - 55.25 = 1.1124999999999972

However, for the last two cases, Nasal will execute the code in the else block (because in that case `oil_level < 5.0` evaluates to false), so we actually get 1.0. This is the same as just limiting `oil_level` using `std.min(oil_level, 5.0)`.

So your code is effectively this:

```javascript
# If oil gets low, pressure should lower and temperature should rise
var oil_level_limited = std.min(oil_level, 5.0);

# Should give 1.0 for oil_level = 5 and 0.1 for oil_level 4.92,
# which is the min before the engine stops
low_oil_pressure_factor = 11.25 * oil_level_limited - 55.25;

# Should give 1.0 for oil_level = 5 and 1.5 for oil_level 4.92
low_oil_temperature_factor = -6.25 * oil_level_limited + 32.25;

setprop("/engines/active-engine/low-oil-pressure-factor", low_oil_pressure_factor);
setprop("/engines/active-engine/low-oil-temperature-factor", low_oil_temperature_factor);
```

Systems/c172p-engine.xml
gilbertohasnofb: Okay, so I fixed it but there is a problem: as I mentioned before, the 180 HP engine's carb-heat is being controlled by `/controls/anti-ice/engine[0]/carb-heat` rather than `/controls/anti-ice/engine[1]/carb-heat`. So now this code is pretty but the carb heat does not drop RPM in the 180 HP engine. 

I also noticed that the flight recorder only saves the property `/controls/anti-ice/engine[0]/carb-heat` and not `/controls/anti-ice/engine[1]/carb-heat`. So what should I do, how can this be fixed?

Nasal/engine.nas
onox: Also, try to put comments above the code instead of after the code to prevent long lines. Normally you keep code and/or comments limited to 72 to 79 characters.

gilbertohasnofb: onox, the property for the oil-level is not being limited in the Nasal code. But the dialog that allows the user to add oil does not accept more than 7.0 quarts for the 160 HP and 8.0 quarts for the 180 HP, effectively limiting it. I do understand your code above and it basically does the same thing, I just don't understand why the `if` block before was bad. Anyway, I can exchange it.

gilbertohasnofb: > Also, try to put comments above the code instead of after the code to prevent long lines. Normally you keep code and/or comments limited to 72 (multi-line comments) to 79 characters.

Ok

Nasal/engine.nas
```diff
@@ -88,6 +88,46 @@ var primerTimer = maketimer(5, func {
...
+});
+
+# ========== engine coughing ======================
+
+var engine_coughing = maketimer(3.0, func {
+    var coughing = getprop("/engines/active-engine/coughing");
+    var running = getprop("/engines/active-engine/running");
+    if (coughing and running) {
+        var delay = 3.0 * rand();
+        settimer(func {setprop("/engines/active-engine/killed", 1);}, delay); # the <logic> block in engine.xml which controls the engine killed status will force it back to killed == 0 if oil level is still all right
```
onox: Move the comment above the code. Make it a multi-line comment if the comment is longer than 72 characters (including the whitespace in front of the comment)

onox:
In the commit?

Systems/c172p-engine.xml
onox: Let me think about why the 180 HP engine uses `/controls/anti-ice/engine[0]/carb-heat`.

Nasal/engine.nas
gilbertohasnofb: Ok, I shortened it as well.

Systems/c172p-engine.xml
gilbertohasnofb: :+1:

gilbertohasnofb:
Yes, in your commit. How exactly the `<property alias=.../>` work? As for the rest, I think my last commit fixes them.

Systems/c172p-engine.xml
onox: The FR should use `/controls/engines/current-engine/carb-heat`. you can add it to line 47 in `Systems/flight-recorder/components/engine.xml`.

onox: (It's a bool I think). Also remove lines 42 .. 45 in `switches.xml`.

onox:
The `alias` attribute tells FG to look for the actual property in `<params><controls><carb-heat>`. This is the place where you define the actual property that is used for reading the state of the carb-heat device. In the future it will be changed to some MP property.

Note that the params are just used as read-only properties. They are used to know how to display our aircraft. (What's the state of ailerons, the yoke, the switches, the throttle, the carb-heat, etc.?) You can leave the bindings in the model XML file untouched. You're either the pilot (you have control) or you're not (your controls won't be transmitted over MP).

gilbertohasnofb:
@onox Thanks for the explanation. As for your last two comments regarding the FR, they should be fixed with my last commit. Also, for some reason your last two comments are not showing here in GitHub, but I was able to read them on my e-mail (note how we don't have `onox commented on a diff` for the last two).

onox:
Can you see them if you go to the "Files changed" tab?

gilbertohasnofb:
Not really.

onox:
What makes you think JSBsim uses `/controls/anti-ice/engine[0]/carb-heat` for the 180 HP engine? I cannot find anything about "carb" in JSBsim.

gilbertohasnofb:
Empiric test: try controlling the `[0]` and `[1]` properties with the 180 HP engine in the sim and you'll see that the RPM drop only happens with the `[0]`.

onox:
@gilbertohasnofb I see something that is probably not going to work: https://github.com/Juanvvc/c172p-detailed/blob/oil-management/Systems/engine.xml#L557-L561

You have the same input and output property. Also, the condition isn't necessary because the gain that you use will be 1.0 if the condition is false.

So what I would do is remove the whole filter, and move the gain to line 201 (Engine Oil Temperature filter)

onox:
Do the same for that pressure filter you added. Move the gain to 221.

onox:
@gilbertohasnofb You made a mistake with the carb heat: https://github.com/Juanvvc/c172p-detailed/blob/oil-management/Systems/engine.xml#L437

The value needs to be 1. That is probably explaining the problem you were having.

gilbertohasnofb:
> I see something that is probably not going to work

It actually worked well in my tests, but I can change if you propose something better. The condition is indeed unnecessary and I will remove it, but else how can I tackle this without using the same property as input and output? Maybe if that Nasal block already outputs a modified temperature and pressure values, so they work as input rather than gain parameters? What do you think?

onox:
No, just move the `<gain>` that you used to the location that I said.

gilbertohasnofb:
> No, just move the <gain> that you used to the location that I said.

Okay

>  You made a mistake with the carb heat: https://github.com/Juanvvc/c172p-detailed/blob/oil-management/Systems/engine.xml#L437

:sob:

gilbertohasnofb:
Done. Sorry for all the mess in the code, onox, I hope it's not giving you more trouble to find and correct mistakes than to implement the whole thing from scratch :disappointed:

gilbertohasnofb:
I just tested and everything seems to work fine: the carb heat now works on the 180 HP, and the gains are working properly in those filters.

onox:
@gilbertohasnofb Your code isn't a mess at all, it's organized, which is good :smile: There were just some small mistakes in the code. This is why we do peer-review.

onox:
I'm doing functional testing now.

onox:
One more thing: see https://github.com/Juanvvc/c172p-detailed/blob/oil-management/Systems/engine.xml#L490-L507 and https://github.com/Juanvvc/c172p-detailed/blob/oil-management/Systems/engine.xml#L465-L472

It seems the code doesn't care about the state of the fuel selector, does it?

gilbertohasnofb:
> It seems the code doesn't care about the state of the fuel selector, does it?

Yep, you are right, if the fuel selected is clean (regardless of the other tank's contamination) then it shouldn't cough. I will work on it now.

onox:
Oil management and carburetor work. :+1: (Just need the fuel selector needs to be taken into account for the contamination)

Some small things:

1. There's no tooltip for the carburetor control

2. Perhaps we should disallow changing the fuel and oil (or make the sliders read-only) if the aircraft is moving or in the air.

3. IMO we should always have fuel contamination and oil management. They are supposed/deserved to be always simulated, just like the vacuum and electrical systems for example.

4. I think the tooltip for "Allow securing aircraft" can also be removed. It's not that hard to click on the items to remove them :stuck_out_tongue_winking_eye:

onox:
Meant checkbutton, not tooltip for "Allow securing aircraft" :laughing:

gilbertohasnofb:
Cool! About your comments:

- I am working in the fuel selector now
- as for the check button for the carburetor, I think that's unecessary, it's just a tiny drop on RPM. Once we implement carburetor icing, we can add an icing check button, but even then the carb heat toggle should drop RPM in my opinon
- making the oil slider read only if the aircraft is in the air (same conditional as for the tie-downs and chocks) sounds great, and I will disable the fuel sump select animation altogether if the aircraft is in movement
- I don't know if I agree with you about enabling oil management and fuel contamination for everyone. I agree it's more realistic to have them enabled, but I still think that we can leave that for the user to decide. Let's allow beginners to live without a problem, and it's also very easy to click on a check button if one wants them, isn't it?
- as for the allow securing aircraft, same as above: it's easy to click on them if you know what's going on, but if a newbie activates a chock or tie-down by accident, or enabled them in the menu and don't know what "chocks" or "tie-downs" mean (I didn't before working on this), then they will get all frustrated that the aircraft doesn't move. And for us who want the feature, it's just one single click in the menu to activate it forever.

onox:
> as for the check button for the carburetor, I think that's unecessary, it's just a tiny drop on RPM. Once we implement carburetor icing, we can add an icing check button, but even then the carb heat toggle should drop RPM in my opinon

I did not talk about a checkbutton for the carburetor, but about a tooltip when you hold your mouse on the carburetor control in the cockpit.

gilbertohasnofb:
Sorry, in that case yep, I will do that!

onox:
Perhaps the checklists should be modified to add some bindings. On some of the checklists they want the carburetor to be cold/heated.

gilbertohasnofb:
@dg-505 will work on the checklists and on our wiki after this get implemented (see https://github.com/Juanvvc/c172p-detailed/issues/493#issuecomment-155068849 ). I'd say let's try to fix these last things, merge this PR and then work on the checklists.

onox:
:+1:

gilbertohasnofb:
@onox I fixed almost everything but I am not able to make a slider read-only under a certain condition. The `<enable>` block works for radios and check buttons but not sliders, apparently. Do you have any suggestion?

gilbertohasnofb:
I am using a condition inside a `<visible>` block to make the slider disappear if the aircraft is in movement, so that might suffice I think.

onox:
Uhh... hang on.

gilbertohasnofb:
Okay. If no better method, the slider is disappearing now under that condition so it's functional, but not as pretty as having it disabled.

onox:
No hang on :stuck_out_tongue_closed_eyes:

gilbertohasnofb:
Ok!

onox:
`<enable>` didn't work?

gilbertohasnofb:
Nope.

onox:
In my V-22 I have all `<slider>`'s read-only: see https://github.com/onox/V-22/blob/master/Dialogs/fuel-payload-settings.xml#L85

gilbertohasnofb:
My mistake again, onox, `<enable` does work: what happened is that I had applied the modifications to the oil dialog of engine 160 HP only and forgot about the 180 HP dialog, and I had that engine on the plane. Now it should be all good.

onox:
You have tabs in the code?

gilbertohasnofb:
Do you mean if I used tabs in this whole PR at all? I don't think that's possible as my editor is configured to use spaces for tabbing.

onox:
No, only in your last commit. https://github.com/Juanvvc/c172p-detailed/commit/e4d28f167cccacc3f773c40183699e0f60340cd2

onox:
Drag your mouse over the whitespace in Github and you will see them.

onox:
It's only on the two lines containing `<enable>`.

gilbertohasnofb:
Sorry, I have no idea how they appeared there.

onox:
OK. Now remains the tooltip and fuel selector and then we're done :smile:

gilbertohasnofb:
They are already done! See: https://github.com/Juanvvc/c172p-detailed/commit/9d0703fd800d3de83ee3a0f1485359ac3b9ae3ab

onox:
Commit did not appear here on the "Conversation" tab, but I see the commit in the "Commits" tab  :open_mouth:

gilbertohasnofb:
Strange, I do see it here. Maybe it is related to your comments which also were not appearing here today? Well.

onox:
I see it now after refreshing. I'll make one change to the tooltip.

gilbertohasnofb:
Sure.

onox:
I'll make the text "Carburetor: COLD" and "Carburetor: HEATED". Is that good?

onox:
OK. Shall I merge the PR?

gilbertohasnofb:
> I'll make the text "Carburetor: COLD" and "Carburetor: HEATED". Is that good? 

That's nice

> Shall I merge the PR?

Yes!

onox:
@gilbertohasnofb Could you update #471?

onox:
And thanks for implementing the oil management and carburetor! Good work. :+1: :smile:

gilbertohasnofb:
I am glad you liked the work, thanks! And sure, I will update #471.

onox:
I'm not sure that works. But the use of `<mapping>` is documented on the [wiki](http://wiki.flightgear.org/Tooltips).

gilbertohasnofb:
Okay, thanks! I will read that.

onox:
Btw, `<ifthen>` only works in JSBsim, not anywhere else.

gilbertohasnofb:
Ah, thanks for pointing it out! I thought it was an xml thing.

onox:
@gilbertohasnofb btw, congratulations with your [#1 position](https://github.com/Juanvvc/c172p-detailed/graphs/contributors) :clap: :wink:

gilbertohasnofb:
:smile: :trophy: Thanks!! But that's only because averagely 50% of my commits are fixes for the errors I introduced __myself__ in the previous ones :wink:

dg-505:
Alright. Let me just finish an instrument for the DHC-6 2D panel, then I'll start with the checklists/tutorials.

gilbertohasnofb:
:+1:

Issue #493, opened 2015-11-02
Add oil management, fuel contamination and securing aircraft to checklist and tutorial


gilbertohasnofb:
Add:
- pitot tube cover
- wheel chock
- wing tie-down strap
- oil management
- test for fuel contamination

To:

- [x] checklists
- [x] tutorials

dg-505:
I'll do it.
Shall I also write a "Preflight Check" tutorial/checklist?

gilbertohasnofb:
@dg-505 Thanks, that would be a great addition! But maybe it would be better if you wait just a bit longer to start working on this, because we have some new features concerning this coming in the next days: right now there is a PR waiting to be merged which will divide the tie-downs into two objects, then by today or tomorrow I will create a PR for the fuel inspection (water contamination), and hopefully in a couple of days at most I will have the oil management implemented. As soon as I have all this I will report here, how that sounds?

Some other things:

- when creating the tutorial, I'd recommend using the walker for pre-flight inspection
- we can adapt the 1982 POH for the checklist and tutorial, see pages 54-57 in the link: http://www.aerodynamicaviation.com/documents/cessna-172p-1982-poh.pdf
- I'd say we should also add this pre-flight checklist to our c172p wiki page, see: #437
- when all this is implemented, I plan to create a youtube video to post in our thread, as well as in the wiki page and newsletter, showing how this inspection could be done

dg-505:
Sounds good. Nevertheless I'll start working, and when you add the new things, I'll ad them to the checklists/tutorial.

I had the same idea about the walker:
Checklist item:
`Name:Walker`
`Value:STEP OUT OF AIRCRAFT`

I'll take the original POH into account when creating the checklists. When the checklists are finished, I can add them to the wiki page. About the video: That's a very good idea!

gilbertohasnofb:
Sounds perfect! Thanks a lot!

gilbertohasnofb:
Updated title and description.

dg-505:
For now I've added the first version of the Preflight Inspection checklist to my fork. I'll create a pull request when all checklists and tutorials are finished.
For those who want to test: It's enough to replace the c172-checklists.xml in your aircraft folder with the one from my fork. I already tested it, but I may have missed some things... Further testing is highly appreciated! :point_up::heavy_exclamation_mark::smile:

gilbertohasnofb:
[...]

PR #549, opened 2015-11-16
"Preflight Inspection" checklists

dg-505:
Added a checklist for preflight inspection, removed twice contained items (in ```Preflight``` and ```Before starting engine```, and expanded the Securing aircraft checklist.

### DIFF
--- a/c172-checklists.xml
+++ b/c172-checklists.xml
@@ -2641,30 +2641,122 @@
             </binding>
         </item>
         <item>
-            <name>Parking Brake</name>
-            <value>SET</value>
-            <marker>
-                <x-m>-0.3092</x-m>
-                <y-m>-0.1992</y-m>
-                <z-m>-0.3284</z-m>
-                <scale>3.4667</scale>
-            </marker>
-            <condition>
-                <equals>
-                    <property>/sim/model/c172p/brake-parking</property>
-                    <value>1</value>
-                </equals>
-            </condition>
-            <binding>
-                <command>property-assign</command>
-                <property>/sim/model/c172p/brake-parking</property>
-                <value>1</value>
-            </binding>
-        </item>
-        <item>
             <name>Control Lock</name>
             <value>INSTALL</value>
         </item>
+        <item>
+        	<name>Walker</name>
+		    <value>STEP OUT OF AIRCRAFT</value>
+		   	<condition>
+		   		<property>/sim/walker/outside</property>
+		   	</condition>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
@dg-505 First of all, thanks for this work! :smile: I had a quick look on your code and I have only two comments about it:

- you are mixing tabs and spaces for indentation, while we decided to use 4 spaces. You can probably set your text editor to use 4 spaces instead of tabs.
- what about automatically opening the dialogs when clicking in the items for fuel quantity (one for each tank), fuel contamination check and oil check?

dg-505:
1. I'm gonna remove the tabs and replace them with 4 spaces, thanks for the hint.
2. I didn't implement the "execute"-button for these items, because for other items this will *execute* the required action (setting/releasing the parking brake, switch lights on/off, etc.). After that you can jump to the next chacklist item. Just opening the dialog window, wouldn't execute the fuel sampling, the oil check, etc. Even if the "execute"-button would open the dialog, you couldn't jump to the next step, i.e. you wouldn't really *execute* this item.

You know what I mean? It's explained a bit awkward...

gilbertohasnofb:
I think I know what you mean, and I don't know what's the best solution. I just wonder if opening the dialogs wouldn't give a help, even if it doesn't change the status of the item because it's not _executing_ it.

dg-505:
IMHO the "execute"-button should set the status to green/checked. Maybe there is a way to to the fuel sampling and the oil check with this button, but this is beyond my skills, and I don't know enough about this system...

onox:
@dg-505 Does this PR fix a particular issue?

onox:
The execute action for closing the baggage door on page 2 is instant. Use:

```xml
<binding>
    <command>nasal</command>
    <script>c172p.baggageDoor.close();</script>
</binding>
```

onox:
Perhaps change "Beacon" to "Beacon light" and add a marker.

onox:
No other remarks. Nice work @dg-505.

dg-505:
@onox this fixes partially issue #493

dg-505:
Okay, thanks @onox . Corrected.

onox:
@dg-505 Does this PR closes #493 or are there things remaining to be fixed?

dg-505:
It doesn't close it, because I plan to add the preflight inspection and the securing aircraft procedure to the tutorials. When this is done, I think #493 can be closed.

Issue #533, opened 2015-11-09
Gear lights of amphibious do not depend on electrical system


onox:
Gear lights of amphibious do not depend on electrical system. Even if battery or alternator is turned off:

![fgfs-screen-320](https://cloud.githubusercontent.com/assets/3361585/11045425/a99831f6-8727-11e5-8028-e445949cd52e.png)
![fgfs-screen-321](https://cloud.githubusercontent.com/assets/3361585/11045424/a99056c0-8727-11e5-8364-ad67d8098c30.png)

I see `Model/c172p.xml` has multiple animations that need to be adjusted, so I suggest to do this:

1. Modify `Systems/als-lights.xml` to add a `<logic>` element

2. Add the output property from the `<logic>` element to the parameter list (around line 259) in `Models/c172p.xml`. Something like:

    ```xml
    <cockpit>
        <electrical>
            <gear-amphibious>/local/output/property/of/new/logic/element</gear-amphibious>
        <electrical>
    </cockpit>
```

    It seems `sim/model/material/instruments/factor` is another local property that is not yet in the parameter list.

    Note that the reason to add it to the property list (even though it is a local property) is so that it becomes very easy to modify the plane when we get to use a new free MP property in the future (+ dual-control). (It seems at the moment we cannot use int[0] .. int[3] until most 3.4 users have upgraded to 3.6)

3. Add `<property alias="/params/cockpit/electrical/gear-amphibious"/>` to the animations

onox:
@wlbragg Would you like to do this? Add the commits to branch `bug-533`.

wlbragg:
Yes, I can finish it.

onox:
I can do this issue if you're already too busy with other issues.

wlbragg:
I'd let you! :smile:

wlbragg:
I'll work on it.

wlbragg:
I already have it all in except to tie to the alt and batt. I'll finish it and push later this morning.

wlbragg:
I guess it is tied directly to master power and no switch inline?

onox:
> I guess it is tied directly to master power alt and/or batt and no switch inline?

You mean whether it depens on a circuit breaker? I don't know. @Juanvvc @tigert @wezzer Do you know this?

wlbragg:
[...]

PR #550, opened 2015-11-16
Tie amphibious landing gear panel to alternator and battery switches.

wlbragg:
Issue #533

Right now it ties to `/systems/electrical/volts`.
If the alternator is on and the engine is running or the battery is switched on, the panel gets power.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -257,6 +257,11 @@
             <mixture>/controls/engines/current-engine/mixture</mixture>
             <carb-heat>/controls/engines/current-engine/carb-heat</carb-heat>
         </controls>
+        <cockpit>
+            <electrical>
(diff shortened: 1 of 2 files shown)


wlbragg:
Just to be sure we can use either

systems/electrical/outputs/instrument-lights
or
systems/electrical/outputs/instr

on the circuit breaker
systems/electrical/outputs/instrument-lights is labeled interior-lights
systems/electrical/outputs/instr is labeled instr-breakers

I now have it tied to "instrument-lights", (interior-lights)
Would "instr",  instr-breakers be more appropriate?

onox:
@Juanvvc @dany93 @tigert What do you prefer?

wlbragg:
Merged into #551

PR #554, opened 2015-11-19
Bug 551, 533, and 555

### LINKED ISSUE #551: Create water rudder lever
Create water rudder lever for float and amphibian models

Reference: http://austingwatson.com/yahoo_site_admin1/assets/docs/Cessna172FloatplanePOH.32573617.pdf

wlbragg:
Fixes #551, #533, #555  

Wait to merge this until I get the water rudder system deploy animation and  limiting of amphibious gear retraction/deploy to power only finished. Right now the animation is only "rudders up". I am going to tie in a  "water rudder system deploy" animation to the "rudders down" position and event. Plus limit the gear from being deployed or retracted without power available.


onox:
I've added an animation to toggle the water rudders by clicking on the handle.

@wlbragg Could you please remove the spaces from the object names? The names are already very inconsistently named in the model files (CamelCase, under_score, etc.) :confused: No need to introduce a third convention :laughing:

wlbragg:
Will do, I should have noticed that it was even more inconsistent. Thanks for adding the animation toggle to the handle,  never even occured to me to do that.

wlbragg:
OK, are we done?

wlbragg:
Oops, I need to finish the renaming with your select animation.

gilbertohasnofb:
So I just tested this, the water rudder system seems to be working perfectly well and the baggages are now properly UV mapped (thanks again!). I just noticed that both the pontoon and the amphibian models have two strange polygons showing inside the cockpit, see:

fgfs-screen-002

To better view them, simply switch from the default to float in the sim and you will see those polygons popping up into existence.

onox:
> OK, are we done?

Yes.

One small comment: when you have added a local commit and you see here on Github I have added a commit as well, you can do `git pull --rebase` instead of `git pull` to put your commit on top of (in this case) mine, thereby avoiding a merge commit.

> I just noticed that both the pontoon and the amphibian models have two strange polygons showing inside the cockpit

I see those things in the master branch as well, so not caused by this branch. Shall we fix those small mistakes when we get the new model from @thevirtualfer?

gilbertohasnofb:
>>  I just noticed that both the pontoon and the amphibian models have two strange polygons showing inside the cockpit

> I see those things in the master branch as well, so not caused by this branch. Shall we fix those small mistakes when we get the new model from @thevirtualfer?

Sounds good. I will open an issue for it anyway so we don't forget about it, ok?

onox:
:+1:

gilbertohasnofb:
See: https://github.com/Juanvvc/c172p-detailed/issues/556

PR #559, opened 2015-11-21
Tutorial: Preflight Inspection

dg-505:
After a several nights with little sleep, here is a first draft of the Preflight tutorial.
:exclamation:Testing is requested:exclamation: 

Just one thing: When checking the oil level, the used dialog window depends on the used engine (160HP or 180 HP). I didn't get it right that the correct window gets opened automatically. For the fuel sample with only one dialog it works fine. I'd appreciate any help here...

The securing aircraft tutorial will follow soon, together with a slightly reworked startup tutorial.


Tutorials/preflight.xml
```diff
@@ -0,0 +1,1056 @@
...
+            <field-of-view>48.4</field-of-view>
+        </view>
+        <marker>
+            <x-m>-0.3500</x-m>
+            <y-m>-0.3967</y-m>
+            <z-m>-0.2283</z-m>
+            <scale>2.0413</scale>
+        </marker>
+        <error>
+            <message>To save energy, swithc the battery off..</message>
```
onox: I would write "To save energy, switch off the battery." here.

onox:
When the tutorial finishes the cockpit part, the FOV is quite high (zoomed out). I need to press Ctrl+x to reset it.

onox:
When the tutorial starts, you should put on the pitot tube, otherwise the user cannot remove it during the tutorial. Check whether it's necessary to have "Allow securing aircraft" (in Aircraft Options window) enabled.

Tutorials/preflight.xml
```diff
@@ -0,0 +1,1056 @@
...
+        </error>
+        <exit>
+            <condition>            
+                <property>/consumables/fuel/tank[0]/fuel-sample-taken</property>
+            </condition>
+        </exit>
+    </step>
+
+    <step>
+        <message>If the color of the fuel is light blue, the fuel isn't contaminated and you can return the sample to the tank. If it's partially translucid, it is contaminated. Then you have to discard the sample.</message>
```
onox: IIRC "translucid" is not correct English according to @gilbertohasnofb. Perhaps change it to "transparent". Also the phrase is very long and might not fit on the screen.

onox:
Enable the tie downs when the tutorial starts, otherwise the user cannot remove them during the tutorial.

onox:
When the camera moves to the tie down of the right wing, it rotates clock-wise (to the right) instead of to the left (which is quicker). Is that something that can be changed?

Tutorials/preflight.xml
```diff
@@ -0,0 +1,1056 @@
...
+        </error>
+        <exit>
+            <condition>            
+                <property>/consumables/fuel/tank[1]/fuel-sample-taken</property>
+            </condition>
+        </exit>
+    </step>
+
+    <step>
+        <message>If the color of the fuel is light blue, the fuel isn't contaminated and you can return the sample to the tank. If it's partially translucid, it is contaminated. Then you have to discard the sample.</message>
```
onox: Same comment as above.

onox:
Wheel chocks aren't enabled by default.

onox:
Perhaps you should also change the aircraft bushkit to the default one, because the seaplanes don't have wheel chocks or tie downs.

onox:
Camera rotates in wrong direction after finishing checking the oil level.

onox:
Camera is zoomed out too much when getting back in the cockpit again.

onox:
@dg-505 I have a bunch of small comments, but otherwise nice work! :smile: :+1:

Tutorials/preflight.xml
wkitty42: On 11/21/2015 06:51 AM, onox wrote:
> IIRC "translucid" is not correct English according to @gilbertohasnofb.
> Perhaps change it to "transparent".

how about something simple? something like


If the color is light blue, there is no contamination and the sample may be 
returned to the tank. If it is cloudy, it is contaminated and should be discarded.

onox: I think the phrase itself is OK, just change the word "translucid" and spread it out over two lines instead of one.

dg-505:
Thanks for your comments!

1. Why do you have to press Ctrl+X? This high FOV remains just a short time, then the tutorial continues with "Open the door". It's kind of a "cutscene" if you know what I mean. 

2,3,5. Take a look at the ```<init>```-section:
```xml
        <set>
            <property>/sim/model/c172p/chock</property>
            <value>true</value>
        </set>
        <set>
            <property>/sim/model/c172p/tiedownL</property>
            <value>true</value>
        </set>
        <set>
            <property>/sim/model/c172p/tiedownR</property>
            <value>true</value>
        </set>
        <set>
            <property>/sim/model/c172p/pitot-cover</property>
            <value>true</value>
        </set>
```

There the pitot cover, tiedowns and wheel chocks should be set. For some reason FG doesn't set these properties when initializing the tutorial. The same with ```/controls/lighting/instruments-norm```. OTOH for the Flaps, mixture, battery, yoke, door positions, etc. it works...

4+7 I don't know exactly, but if I'm right, the reason that it rotates in the wrong direction is that the cockpit view has limits enabled. I.e. you can't turn around more than +140 deg and -140 deg. So the view movement has to rotate the other way round. To fix this the property ```/sim/view/config/limits/enabled``` has to be set to false. Maybe this can be set to false during the tutorial and set back to true when the tutorial ends. I'm gonna try that.
6. That's a good idea. I'll add that.
8. The same like 1. When the message "Tutorial finished" appears, the viewpoint/FOV should return to the standard setting.



I'll change the text of the fuel sample thing.

wlbragg:
>To fix this the property /sim/view/config/limits/enabled has to be set to false.

Why do we have this limit set, because you can't rotate 360 in your seat?

dg-505:
I think so... But I did some experiments: The limits seem to be immaterial. At the relevant points the view turns always around the wrong direction.

And I still have no idea why some properties aren't set during init.

wlbragg:
> why some properties aren't set during init.

We discovered you can set some properties in a "blank first step" instead of init. We did that when the init would override setting we wanted to set. I don't know if this is the same as your issue?
See 17079037b0f355d7b86d9ba250161fe670b3265a

dg-505:
Seems to work. I'll do some further tests.

Juanvvc:
The tutorial subsystem (in FGDATA, not in our files), if the aircraft starts on the ground, sets an initial configuration for the engine *after* the presets but *before* the init. The starting procedure on the current c172p is more complex, so any specialized configuration must be set either during the init or during a dummy first step, not in the preset section.

I consider this assumption of a simple engine a bug in FGDATA and it might affect tutorials for other propeller aircraft. Let's see if we can issue a change request to FGDATA to remove the lines about setting magnetos and throttle in tutorial.nas ( http://sourceforge.net/p/flightgear/fgdata/ci/next/tree/Nasal/tutorial/tutorial.nas )

dg-505:
I took care of @onox's comments, but still the camera rotates in the wrong direction sometimes. I have no idea how to solve this. OTOH, does this really bother?

gilbertohasnofb:
@dg-505 tail tie-downs have been added to the master, so they can be included in your work :smile:

dg-505:
> tail tie-downs have been added to the master, so they can be included in your work 

Thanks. Done!

onox:
If other people are OK with it, this PR can be merged.

dg-505:
Please wait!
I'll add the securing aircraft tutorial...

dg-505:
Before merging this, please test the securing aircraft tutorial.
When this is merged, #493 can be closed.

onox:
1. When you have to reduce throttle, the camera focuses on the RPM instrument, but the throttle control isn't visible then.

2. I think mixture marker should be moved to where the mixture control currently is.

3. Pitot cover isn't removed when starting the tutorial.

dg-505:
1., 2. I'll correct that.

3., 4., 5. The tutorial is meant to be started after a flight when the engine's still running. In this case the pitot cover, tiedowns and wheel chocks are usually not placed...
But if you nevertheless consider it to be necessary, I'll remove them at the beginning.

I didn't manage it, to start the engine properly at the beginning of the tutorial because of the complex process... I'll ad a note to the tutorial description.

onox:
> In this case the pitot cover, tiedowns and wheel chocks are usually not placed...

In that case it doesn't matter to remove the pitot cover and chocks, does it? :smile: 

> I didn't manage it, to start the engine properly at the beginning of the tutorial because of the complex process...

You're right about this, starting the engine may be too difficult or impossible with the current tutorial system.

dg-505:
> In that case it doesn't matter to remove the pitot cover, tiedowns, and chocks, does it?

Point taken, I'll add that :wink:

PR #569, opened 2015-11-22
Adjust windshield rain RPM X and Z.

wlbragg:
Minor change, gets rid of momentary downward movement of rain.


gilbertohasnofb:
I don't know if I understand what was modified, how do you suggest I test this?

wlbragg:
I use Environment/Weather/Detailed Weather 
Weather conditions = Core low pressure to get rain on windshield.

Watch rain on windshield while doing the following.

Apply brake, start engine rev to full throttle then release brake and take off.

The change was minor but it did correct a momentary reversal of rain that went down instead up up and back at the moment of RPM increase from idle to full throttle.

gilbertohasnofb:
I compared it to the master and the change is clear, thanks for the work. Can we delete the branch?

PR #572, opened 2015-11-23
No preview lights and propeller.

wlbragg:
I don't know how to select a livery for preview in FGRun, but everything else is done.
I think I screwed up the naming on this PR or issue, I'm so confused. At any rate this is the fix for FGRun preview.


wlbragg:
The livery that is being used is the one that is assigned as textures in Blender. Does anyone know if you can select a "livery" to show in fgrun VS the Blender texture that was assigned?

gilbertohasnofb:
What about the following: why not move the older D-EEQA textures (interior and exterior) to the `Models/Liveries` folder, then move the textures from `Models/Liveries/4X-CHV` to `Models` and then use these to UV map the model to the chosen default 4X-CHV? A bit annoying, I know, but currently we have the D-EEQA textures in a prominent place while it isn't the default texture any longer.

If you all think this is a good idea, I can help with that (except the Blender part). @onox, do you think this is a good idea?

onox:
I guess it's the only option we have.

wlbragg:
If we have to actually redo the UV textures in Blender, no way, at least not me! I think the UV mapping of the entire aircraft is a lot of work just to get the one we want in FGRun..
If we can swap the texture file and Blender keeps the mapping and applies it to the new texture file, then OK.
Let me see if it is that easy, if so then cool!

gilbertohasnofb:
Okay, I am working on it now. Soon I will commit it.

gilbertohasnofb:
@wlbragg I think I have it done now without any need to do anything on Blender, I just reorganized the xml files and their locations. Gimme a second to test then I will push.

gilbertohasnofb:
Done. I also fixed the horrible layout of those livery xml files... :scream:

gilbertohasnofb:
And it works:

screenshot from 2015-11-23 18 54 18

wlbragg:
Great, well gone!

gilbertohasnofb:
You too!

PR #573, opened 2015-11-23
Create new tiedown straps and hardware for amphibious.

wlbragg:
I created a whole new set of straps and hardware for the amphibious.
I didn't like the angle some of the straps would have if we used the same straps on all gear configurations. Plus I really like having the hardware sticking out of the pavement even though it is just in two of the configurations now.

The new set of straps is
~~~xml
        <path>Models/Exterior/tiedowns/tiedownLL.ac</path>
        <path>Models/Exterior/tiedowns/tiedownRL.ac</path>
        <path>Models/Exterior/tiedowns/tiedownTL.ac</path>
~~~


gilbertohasnofb:
Cool! So let's not merge this yet and I will work on the xml. 

>  I really like having the hardware sticking out of the pavement even though it is just in two of the configurations now

Me as well, it's just a pity that the default plane doesn't have them :cry: Maybe we should create tie-downs for each variant? Though that would mean 6 variants x 3 tie-downs for each

Another improvement we could do on this branch is to change the angle of the tail tie-down, I think the angle should be more "acute". See for instance this image below, note how the pavement link is far behind the rudder, while ours is under the plane:

wlbragg:
>Another improvement we could do on this branch is to change the angle of the tail tie-down

I'll make you a deal,
I'll make another set for the default and then I think the only one that won't have hardware on the pavement would be the 26".
I'll make the angle "acute" on the default as in the photo but leave the rest as is (it's a real pain to change them).
Deal?

wlbragg:
I just checked and the ski uses the 36" wheels so they both have the hardware showing.Now the amphibian and the default will too. I think we could live without the 26" showing the bottom hardware? But if you really want them I can do them as well. That would only be a total of 4 sets.
I'm not as concerned about the angle only because I'm sure it is different at different airports, but I will make any new ones (default gear) with the more acute angle.

wlbragg:
If this is a yes, I'll make name the default straps 
tiedownLD

gilbertohasnofb:
That's a deal! But I would suggest better filenames (what you are using is very confusing, such as `tiedownRL.png` for _right large_). Let's use:

`tiedownL-<variant>.png`, `tiedownR-<variant>.png` and `tiedownT-<variant>.png`, such as `tiedown-amphibious.png`. The default ones can be `tiedownL.png` of course.

gilbertohasnofb:
Once you have all tie-downs I can have a look at the code, ok? Would you please let me know?

onox:
How many tiedowns do we have now? We can actually have one set of tiedowns, and then specify the coordinates of the attachment point and the rotations (yaw and pitch) per bushkit option.

onox:
To give you an idea what I mean:

fgfs-screen-347
fgfs-screen-350

I apply the same technique to the refueling boom of the 707-tt I have modified. (Although the glorious MP system ruins any chance of getting very close to the boom)

wkitty42:
On 11/23/2015 06:41 PM, onox wrote:
> How many tiedowns do we have now? We can actually have one set of tiedowns, and
> then specify the coordinates of the attachment point and the rotations (yaw and
> pitch) per bushkit option.

that's what i was thinking with seeing all the details that the sim offers that 
are unknown to many...

then this also brings up another question... can these tiedowns be made to 
stretch slightly and then droop and even break if the stretch is too much in 
certain weather?

as an example, set everything as you would from a good flight... everything is 
off, the pitot tube is covered and all the tiedowns are on... now set your 
weather to one of the two storm scenarios... change you view to an external view 
and watch the craft move in the winds...

there are times when the craft rocks back and the rear tiedown should droop... 
then if the forward motion is strong the craft should rock forward and be 
abruptly stopped by the rear tiedown... if this forward rock is past a certain 
limit (i don't know the breaking point of the tiedown cables) then the tiedown 
should break and that aspect of being tied down should be negated and the craft 
left to rock back and forth with the rotation point centered over the wing 
tiedowns...

by the same token, if the winds are from the side, then the same drooping and 
flexing of the wing tiedowns should be visible and eventually, if the winds are 
strong enough, lead to either a broken wing or tiedown... at that point, only 
the other tiedown(s) are anchoring the craft and there could be other damage...

perhaps these details are too much for the sim and this craft at this point in 
time of development of both... i do, however, sit at times and watch the craft 
move in the weather while i work on a flight plan or other sim flight related 
tasks...

if these capabilities are available, we should try to implement them if at all 
possible... the reality factor is astounding and if we can simulate ground 
actions like these in the same way that the sim simulates the craft's actions in 
the air, then this is a huge plus... especially with this craft being the 
leading craft in the sim's distribution... even moreso if/when we can show 
others how to do the same thing with the other craft available...

gilbertohasnofb:
>  How many tiedowns do we have now? We can actually have one set of tiedowns, and then specify the coordinates of the attachment point and the rotations (yaw and pitch), and scale (distance to ground) per bushkit option.

We have only one set right now (left + right + tail). If stretching and rotating is an option, let's do this way! But I am not sure I would be able to do so, would any of you two try to implement this?

>  perhaps these details are too much for the sim and this craft at this point in time of development of both...

To be frank, that's my opinion on this. Tie-downs are normally strapped very tight, I think only the loose end would flick in the wind. Still, that is a bit too much detail IMO :smile:

wlbragg:
I thought about suggesting using animations on the hardware buckle and overhang, but couldn't figure out how to deal with the length of the strap.

@onox, can you scale z and not x or y? And if so will the scaling make the textures look worse?

@gilbertohasnofb, if we go this way I can do the positioning of the pieces.

I'm not convinced that it would be any easier or better just to have one model. I mean we're almost done the other way and now have to work up all the positions.
Do we gain on overhead or something to make this worth the added effort at this point?

gilbertohasnofb:
@wlbragg I suppose the main advantage is that having only one set of tie-downs is a more elegant solution and keeps the repository less confusing. Having 18 objects with similar but distinct names is a bit messy, but it will of course work as well. I think the stretching technique will look very good as well.

One more thing that I noticed: currently the tie-downs' positions are fixed in relation to the fuselage. The wings barely move if one apply full power, but the tail can raise quite a lot in similar condition. If we would be able to do what @onox said, so that only two points are fixed, the effect would be much better I think. Makes sense?

wlbragg:
>Having 18 objects with similar but distinct names is a bit messy

This reply will sound like I am trying to talk us out of the alternate way, but I am not. I just want to point out something that might be relevant.
Each tiedown set (strap, hardware, overhang and buckle) can be joined in Blender and be one object name, such as skiRight, skiLeft, etc. once we are done creating and positioning them. Don't forget then in that case we would have a total of 4 strap configurations with only 3 names each for a total of 12 for all 5 gear configurations. This we can do easily. It requires me to make 2 more strap configurations and join the respective sets when I am finished.
Think about the existing select animation code you have already wrote for this.

Now using the "more elegant method" I won't be able to join any of it so we will have to have each piece named (unless I still join it all and then apply the stretch to that, I'm not sure how the hardware, buckle and overhang will look stretched?). If we can't stretch it joined then we will have 5 objects X the same 4 configurations and the duplicated 5th configuration.
So were talking about 20 translation animations code blocks plus the existing select conditions.

Is it still more elegant?

wlbragg:
>so that only two points are fixed, the effect would be much better I think. Makes sense?

I'm not sure I understand this, your saying that the view of the tiedowns, will or do, look bad now because they move when power is applied and wont the other way? And that is because the other way we supposedly can fix the hardware and dynamically stretch the strap real time? I'm not sure that is possible or that I could program the code for that.

wkitty42:
On 11/23/2015 08:36 PM, wlbragg wrote:
>     Having 18 objects with similar but distinct names is a bit messy
>
> This reply will sound like I am trying to talk us out of the alternate way, but
> I am not. I just want to point out something that might be relevant.
> Each tiedown set (strap, hardware, overhang and buckle) can be joined in Blender
> and be one object name, such as skiRight, skiLeft, etc. once we are done
> creating and positioning them. Don't forget then in that case we would have a
> total of 4 strap configurations with only 3 names each for a total of 12 for all
> 5 gear configurations. This we can do easily. It requires me to make 2 more
> strap configurations and join the respective sets when I am finished.
> Think about the existing select animation code you have already wrote for this.
>
> Now using the "more elegant method" I won't be able to join any of it so we will
> have to have each piece named (unless I still join it all and then apply the
> stretch to that, I'm not sure how the hardware, buckle and overhang will look
> stretched?).

the only thing that should stretch should be the strap... not any of the anchors 
to the craft or ground or the buckle and excess in the middle of the strap...

i feel the pain but i also have to wonder about the need of, if i understand 
correctly, the straps on the model in blender or ac3d... can't they be their own 
entity with their own properties (endpoints and stretching) that are ""just 
placed"" along side the model? or am i really missing something intrinsic to 
modeling several different things and joining them together in a program like a 
simulator?

[/ignorance but i really am trying to understand and have been trying to work 
with 3d cad programs to try to understand]

wlbragg:
>the only thing that should stretch should be the strap.

I agree, in a perfect world. but then I can't join them into one name if we chose to position each object individually and that is a lot of translate animations.

>can't they be their own
entity with their own properties (endpoints and stretching) that are ""just
placed"" along side the model?

Yes, and in fact I can join one entire side together into one name. I'm not sure what your asking or if I answered it.

wlbragg:
>I apply the same technique to the refueling boom of the 707-tt I have modified.

@onox can you point me at your code in the V-22 so I can take a look at it?

onox:
> @onox, can you scale z and not x or y? And if so will the scaling make the textures look worse?

The texture may get scaled a bit, but if you give the object in Blender its average that it will have then the scaling won't be much. I think for the amphibious the tiedown's length will be the largest, and for the default the smallest.

onox:
So what you would do is make sure the tiedown is horizontal and on one of the axis. Then in XML you first scale one of its dimensions. After that you rotate it.

onox:
@wlbragg I'll look up the code.

wlbragg:
:+1:

onox:
https://github.com/onox/ExpansionPack/blob/master/Nasal/fuel_truck_dialog.nas#L154-L173

onox:
To give you some information about it: `FuelTruckPositionUpdater` is a class with 4 methods: `enable`, `disable`, `reset`, and `update`. It uses the `updateloop.UpdateLoop` class (it should be part of 3.6, but the core devs were so super slow accepting the PR, that I just added it to my own ExpansionPack). 

It calls `update` continuously. The code first computes the (relative) position of the fuel truck and then the distance, pitch, and yaw of the fuel line.

onox:
XML animations are in https://github.com/onox/ExpansionPack/blob/master/Models/Airport/Fuel-Truck/fuel-line.xml

onox:
Oh, and if you want to add the fuel truck to your own aircraft (not the c172p) then you should read https://onox.github.io/ExpansionPack/module-fuel-truck-dialog.html :wink:

wlbragg:
Cool, no I want to use it on my cargo lift :bulb:. I kind of got stuck on the rope and the tiedowns of course!

onox:
Okay, talk to you tomorrow, now I need to sleep :smile:

gilbertohasnofb:
> Each tiedown set (strap, hardware, overhang and buckle) can be joined in Blender and be one object name

I don't think you can do that, as we remove each tie-down separately. If they are joined, clicking on one of them would remove ALL tie-downs at once. So even if we use a single tie-down strap rotated and scaled for all variants, we will need to define them as three separated objects, `tiedownL`, `tiedownR` and `tiedownT`.

>> so that only two points are fixed, the effect would be much better I think. Makes sense?

> I'm not sure I understand this, your saying that the view of the tiedowns, will or do, look bad now because they move when power is applied and wont the other way? And that is because the other way we supposedly can fix the hardware and dynamically stretch the strap real time?

Yes, that's what I meant. If we can fix two points in space, one the connection point in the ground which is fixed relative to the ground, and another which is in the tail of the plane, relative to the hook (regardless if the plane is trying to move or not), then we could try to stretch it in real time. I think that's possible given @onox images above, but this is stuff beyond my knowledge.

>> the only thing that should stretch should be the strap.

> I agree, in a perfect world.

Yeah, I think any stretch of the buckle will be imperceptible and I don't think focusing on it would be reasonable. 

> So what you would do is make sure the tiedown is horizontal and on one of the axis. Then in XML you first scale one of its dimensions. After that you rotate it.

We just have to keep in mind that the extra part of the strap is pointing vertically down:

fgfs-screen-001

Maybe the best approach (and correct me if I am wrong) would be to keep our three tie-down objects, make the adjustments to the tail one to create a more acute angle (@wlbragg already knows about it) and then use the techniques @onox describe to stretch and rotate them just a little bit to fit each variant. In my mind, this would be the best way to get the best looking results without going too crazy with several models. What you all think?

wlbragg:
> clicking on one of them would remove ALL tie-downs at once

You must have realized they are already joined in a sense and so taken what I said to the next logical level. I didn't mean join all the parts on all the sets together, I meant join all 5 parts on skiLeft to skilLeft and all 5 parts on skiRight to skiRight. But really, we already do that. I was not considering that we aren't using all the pieces in a side (or set) independently (calling all 5 pieces to a side separately). Because we save them together and use them as a group, always with the same exact animation (no variance), then for all intent and purpose they are already joined, we don't call them separately anyway. We already only make one call for the entire set because of the way it is used and it is saved in its own .ac together.

I tested the existing tail tiedown and see what your talking about, I didn't realize it did that.
I have two thought on that
One, can we stop that action from happening when the aircraft is tiedown? That seems like what would happen in real life.
Two, if we can do that (stop it from happening), then if the choice is one or two more set of straps (one if for default only, two if we make a set for 26"), would you still want to redo it all to use just one set with a bunch of animations to position everything.
Although in fairness, to position everything I'm not clear on how many animations we would be talking about, because the clips on the ground and attached to the wings may only require one animation for all configurations.
I guess I can't make a decision for myself until I see onox's  code and see how he is positioning and leveraging everything. I'll still need to take a look at it, hopefully I can get to it here shortly.

> keep our current three tie-down objects (so with their own angles already), make the adjustments to the tail one to create a more acute angle (@wlbragg already knows about it) and then use the techniques @onox describe to stretch and rotate them just a little bit to fit each variant

If we can't stop the tail from rising when tiedown, then that would be a third choice.
We would have...
Use one set and make positioning animations for everything.
Use 3 or 4 sets and use some stretching animations on the strap
Use 3 or 4 sets and stop the tail from moving.

I say 3 or 4 sets and not 5 because (ski and 36" share a set effectively)
3 sets gets everything except the 26" with hardware fasteners on the ground
4 sets adds the 26' hardware on the ground.

wlbragg:
Anyone know if you can stop the "tail lift"  FDM behavior with a condition?

wlbragg:
>In case we're going for this, we should also use these external forces as Thorsten said.

What's this?

onox:
What we do currently is just secretly set the parking brake. Thorsten mentioned that you can use external forces as well. I think he mentioned it on the mailing list.

wlbragg:
OK, yeah I think we tried something to that effect but had less than optimal results, we need to try it again and troubleshoot if possible.

I am going to make a total of 4 sets of tiedowns which will cover all 5 configurations. We can deal with the tail rise after that.
I like the idea of one model with positioning animations or even the nasal solution to position and stretch, but right now it is too much for me to deal with when the other solution is close at hand and less over all work for me, at this point. So long as we can fix the "tail rise' or anchoring of the aircraft.
I hope we are all OK with that?

wlbragg:
If we decide to come back to another method, then I think we should try to modularize it for use in all aircraft!

wlbragg:
There is probably a lot of this stuff we could have or should have done that to.

gilbertohasnofb:
> I am going to make a total of 4 sets of tiedowns which will cover all 5 configurations. We can deal with the tail rise after that.

That's all right with me. But since we are going back to Blender, would you please correct the angle of the tail tie-down as we spoke about before? The rise we solve some other time (I will open an issue to remind us about it).

wlbragg:
Better?
tie

gilbertohasnofb:
That's perfect!

wlbragg:
So the left and right tie downs you think need to be almost vertical as in the picture?

wlbragg:
If I don't put some slant to them the overhang wont show very well.

gilbertohasnofb:
Not at all, see for instance:



It was just the tail one, so I think it's really good now :+1:

wlbragg:
OK, good.

wlbragg:
You know what, as I do this I discovered that it isn't unreasonable, or might be possible, to use one set (5 pieces) for each position (l, r and t) and just change the angle of the entire set until you get the bottom hardware clip to be in the correct ground level elevation.
If I did this I would start with the highest profile and it would have to be near vertical, the next highest profile swings out a bit until the clip is back at ground level, the 3rd profile swings farther and the last (default) swings out the farthest.
Do you see what I am talking about?
I think without a ton of effort I can give you and example picture of what the different angles would look like.
We have much flexibility on the angle of the tiedown as we have seen both acute and close to 90deg angles for the same tiedown position. I think it depends on the airport and the aircraft parking space your putting the aircraft at.
What do you think?

gilbertohasnofb:
I see what you are talking about and that sounds very clever, the only thing is that the bottom hardware won't be parallel to the ground if we turn the model, right? Also, wouldn't it be correct to say that changing too much the angles will cause that little strap part hanging in the side to become non-perpendicular to the ground?

wlbragg:
>the bottom hardware won't be parallel to the ground if we turn the model, right?

I guess I don't understand this question. The hardware would be close to perpendicular to ground on the highest profile aircraft, and close to parallel on the lowest profile. That's exaggerated but gets my point across I think.

>that little strap part hanging in the side to become non-perpendicular to the ground?
That is true. You could argue that the short overhang we are simulating is pretty stiff and if the highest profile version showed the hangover sticking out a bit past vertical, the lowest profile might show it pretty close to vertical.

Just a thought. I don't really have any issues with multiple sets.

gilbertohasnofb:
This is what I mean, imagine this is the front view of the wing tie-down:

untitled

wlbragg:
arc

wlbragg:
The hardware stays as it is now, I would rotate everything together as one piece.

wlbragg:
Note: the two actual straps in the blend pic are the tallest and lowest profiles, def and amph.

gilbertohasnofb:
Sorry but I am still not understanding, in your image above that piece of the strap is clearly not perpendicular to the ground depending on the rotation, which is how it should be if it's just hanging there. But it's not relevant that I understand it, if you can make it look good then go for it :smile:

wlbragg:
Right, but that small piece of overhanging strap would be somewhat stiff and may not hang perfectly perpendicular to the ground. Yes, that's a bit of a stretch, but is it worth all the extra modeling not to cheat a little?
I really don't know myself. I feel like a couple extra models that are only used when the aircraft is in a static position isn't really harming anything.
So whatever you think is good with me. I just wanted to through that out there as I discovered the possibility as I was modeling the 4 sets.

wlbragg:
One thing to look at is, we either call a different model for each profile or we make some animation block for each one of them, apples and oranges I guess.

We've gone this far, we might as well finish. If we ever want a "smaller footprint" (so to speak) we can change it later.

gilbertohasnofb:
> We've gone this far, we might as well finish. If we ever want a "smaller footprint" (so to speak) we can change it later.

:+1:

gilbertohasnofb:
Or here is another option (which you may hate me for): we forget about the hardware in the ground and bury ALL 5 straps. So we just need one single set of long straps which would be used for all variants. Advantages: one single elegant xml block, only one tie-down object for each position (L,R,T), all of them will look the same if none would show the hardware in the ground.

wlbragg:
Your killing me :wink:, but it looks so much better with the hardware.
tst

gilbertohasnofb:
I know, I know, maybe I am being overly practical. I guess you should just do what you think is best, as I am sure the result will be good.

wlbragg:
OK, this newest commit adds
tiedown36SkiL.ac
tiedown36SkiR.ac
tiedown36SkiT.ac
tiedownAmphL.ac
tiedownAmphR.ac
tiedownAmphT.ac
tiedownDefL.ac
tiedownDefR.ac
tiedownDefT.ac
All that is left is tiedown26tL, R and T

I didn't touch the XML (rename anything or add anything).

gilbertohasnofb:
> All that is left is tiedown26tL, R and T

Okay, once we have all of them I will work on the xml file.

gilbertohasnofb:
And thanks for the work! :smile_cat:

wlbragg:
It's all here.

If you get to a point where it seems like it is to much. I went ahead an took the right side tiedown starting with the amphibious (the longest strap) and rotated it for each of the other configurations, leaving everything the same(buckle, overhang, hardware). Just rotated it to get it to the correct elevation for the next longest strap, and so on and so on.
Here is what it looks like.
pos
flap
a
b
c
d
What you see is the same (single set) of tiedown rotated to fit each gear configuration.

gilbertohasnofb:
@wlbragg It actually looks super good and it would keep things so much more elegant, so I vote to use the rotation techniques with a single model of tie-downs (L,R,T). But I would need some help defining their exact locations and rotations for each variant as in XML it's a matter of trial and error and updating the sim for each try (which is a pain in the neck). Would it be possible for someone to either give me these values from Blender or directly modify the XML files?

wlbragg:
OK, yes sir I can do either. I won't have any time until later this afternoon or evening.

wlbragg:
I'm not totally sure what we need. If you want a test, try this,

Start with tiedownAmphR.ac, the model should work as is.

Then try to use it with the following rotation for the 36SkiR configuration
x=135.237
y=8.027
z=191.22

If that works then I can give you the rest later this afternoon.

wlbragg:
>it's a matter of trial and error and updating the sim for each try (which is a pain in the neck)

Yes it is, and we have all those positions in Blender so we have that data and I can give it to you, but I still have to go through a process in blender (this afternoon).

gilbertohasnofb:
@wlbragg Wait, I am a bit confused:

> Start with tiedownAmphR.ac, the model should work as is with the amph configuration.

shouldn't we simply use the `tiedown.ac` object for ALL of them since we are just rotating them? If by that you mean that the `tiedownAmphR.ac` is the longest one and the one that will be used for ALL configurations, I suggest we remove all other .ac and .blend files for the tiedowns we won't use and rename all `tiedownAmph*.ac` as `tiedown*.ac`. What do you think?

> Then try to use it with the following rotation for the 36SkiR configuration
> x=135.237
> y=8.027
> z=191.22

I think I need more data, as we need a position x,y,z for the object as well as a rotation in degrees, see:

```
<offsets>
    <x-m>-0.0</x-m>
    <y-m>0.0</y-m>
    <z-m>0.0</z-m>
    <roll-deg>0.0</roll-deg>
    <pitch-deg>0.0</pitch-deg>
    <heading-deg>0.0</heading-deg>
</offsets>
```

wlbragg:
I think I found a better way to do this.

This is all you need for the right side. I will push a new tiedown.blend and tiedownR.ac, tiedownL.ac and tiedownT.ac

~~~xml
    <animation>
        <type>rotate</type>
        <object-name>tiedownR</object-name>
        <property alias="/params/bushkit/property"/>
        <interpolation>
            <entry>
                <ind>0</ind>
                <dep>24.45</dep>
            </entry>
            <entry>
                <ind>1</ind>
                <dep>22.95</dep>
            </entry>
            <entry>
                <ind>2</ind>
                <dep>22.5</dep>
            </entry>
             <entry>
                <ind>4</ind>
                <dep>0</dep>
            </entry>
            <entry>
                <ind>5</ind>
                <dep>22.5</dep>
            </entry>
        </interpolation>
        <center>
            <x-m>0.13462</x-m>
            <y-m>2.30397</y-m>
            <z-m>0.40351</z-m>
        </center>
        <axis>
            <x>1</x>
            <y>0</y>
            <z>0</z>
        </axis>
    </animation>
~~~

wlbragg:
OK, it's ready. All you should have to do is add the amphibious configuration and verify that the tiedown fit that configuration. It's the only one I didn't test.

gilbertohasnofb:
@wlbragg I finally had time to test this, it works beautifully! :+1: 

> All you should have to do is add the amphibious configuration

Okay!

gilbertohasnofb:
Done. This PR is ready for testing and merging. @onox maybe you would like to test it?

gilbertohasnofb:
By the way, is it "amphibious" or "amphibian"? Isn't the latter better?

wlbragg:
>"amphibious" or "amphibian"

I just looked it up and actually amphibious would be the correct term, amphibian is a class of animal. Whew knew! :smile:

gilbertohasnofb:
:+1:

gilbertohasnofb:
Oh, no! I completely forgot to limit the amphibian tie-downs to solid grounds.... I am restoring this branch and working on it.

gilbertohasnofb:
Actually, I will limit ALL variants to now be allowed to have tie-downs on water, so that if someone successfully crash land on water with the regular model, tie-downs won't be an option :smile: and it also makes the XML coding easier. What do you think @onox? Give me green light and I will work on this straight away.

onox:
Sounds like a good idea.

onox:
> now be allowed

Now or not? :open_mouth:

gilbertohasnofb:
"not"! Edited my previous post :smile:

gilbertohasnofb:
@dg-505 had recommended me to use the property `/gear/gear[~i~]/ground-is-solid`, but this property doesn't seem to exist in our tree. @wlbragg recommended `/fdm/jsbsim/hydro/active-norm` but this is true only if the float or amphibian are properly floating in the water.

onox:
Use `/fdm/jsbsim/gear/unit[?]/solid`.

onox:
Oh wait, use this one instead: `/fdm/jsbsim/ground/solid`.

gilbertohasnofb:
:+1:

gilbertohasnofb:
Yep, that did the trick. I will open a new PR to merge this branch again as soon as I finish this. But @onox, please help me out with a little problem here. So now I have the following `<logic>`:

```
    <logic>
        <name>Tiedowns Can Have</name>
        <input>
            <and>
                <not-equals>
                    <property>/fdm/jsbsim/bushkit</property>
                    <value>3</value>
                </not-equals>
                <property>/sim/model/c172p/allow-securing-aircraft</property>
                <property>/fdm/jsbsim/ground/solid</property>
            </and>
        </input>
        <output>
            <property>/sim/model/c172p/tiedowns-can-have</property>
        </output>
    </logic>
```

Which works great, except when one quits the sim with the tie-downs in place on a solid runway and then loads the sim in a port airport. In that case, the tie-downs won't be visible, but the property `/sim/model/c172p/tiedowns` will be set to true. So basically we need something like this:

if `/fdm/jsbsim/ground/solid` is true, then `/sim/model/c172p/tiedowns` is set to false, else do nothing. AFAIK <logic> blocks do not have an "else" behaviour, they simply force a value constantly. So what would be the best approach for this? Should I make a Nasal function for it or is there a more elegant way of handling this only using XML?

onox:
What does `/sim/model/c172p/tiedowns` do?

gilbertohasnofb:
That's the property that sets the parking brake. The others are:

`tiedowns-can-have`, which is what allows the tie-downs to be set and unset, and `tiedowns-visible` which is what makes them, well, visible. It works very well except now that we can have a same variant being allowed to have tiedowns when quitting the sim (amphibian on a runway, for instance) but then not allowed when the sim starts (amphibian on water).

gilbertohasnofb:
So now we have to basically listen to `/fdm/jsbsim/ground/solid` and when it goes false, then `/sim/model/c172p/tiedowns` should be set to false. So my question is there a way of doing so with xml or do we need something more complicated? I can do that in Nasal, but I think it would be more elegant to do it otherwise.

onox:
Let me think...

gilbertohasnofb:
Ok.

onox:
I don't see the `/sim/model/c172p/tiedowns` property in `Systems/ground-effects.xml`.

onox:
Parking brake is controlled by `/sim/model/c172p/tiedownL-visible`, etc.

gilbertohasnofb:
Actually that property is broken into three now:

`sim/model/c172p/tiedownL`
`sim/model/c172p/tiedownR`
`sim/model/c172p/tiedownT`

They are present in the `Models/c172p.xml` and `gui/dialogs/aircraft-dialog.xml`

onox:
I still don't understand what the problem is, but you can do things like this:

```xml
<logic>
    <enable>
        ...
    </enable>
    <input>
        <false/>
    </input>
    <output>
        <property>...</property>
    </output>
</logic>
```

gilbertohasnofb:
I think things got a bit messy with that code, maybe we should have only the `-can-have` and `-visible`. I will think if it's possible to clean it, okay?

And thanks for the solution above, I forgot that `<logic>` can have `<enable>`

onox:
Is it the checkbuttons in the GUI you're talking about? :open_mouth:

gilbertohasnofb:
Those three properties (`tiedownL`, `tiedownR` and `tiedownT`) are connected to the checkbuttons but are also present in the `Models/c172p.xml`. I will take a deeper look on this and see if I can clean it, ok? I am also having a hard time understanding why we are using them. My suspicion is that I implemented it in a simpler way in the beginning (it was only the default variant that could have them) and then with all these modifications and added features it got a bit out of hand. Let me check it.

onox:
So the problem is you want to explicitly uncheck the checkbuttons in the GUI when you're on water?

onox:
In that case you can use the XML I posted above. Use `<property>/sim/model/c172p/tiedowns-can-have</property>` in the `<enable>` and use multiple outputs.

gilbertohasnofb:
> So the problem is you want to explicitly uncheck the checkbuttons in the GUI when you're on water?

I am not sure that's the only problem. If so, then the solution you posted above will fix it of course, but I am unhappy with how messy things are. Let me sit on this one for a while and I will try to clean it, ok?

onox:
Ok.

gilbertohasnofb:
Okay, I think I got it. I did a complete clean-up of that code and indeed those three properties weren't necessary at all. See what you think in this PR here: https://github.com/Juanvvc/c172p-detailed/pull/580

wkitty42:
On 11/28/2015 04:12 AM, Gilberto Agostinho wrote:
> Actually, I will limit ALL variants to now be allowed to have tie-downs on
> water, so that if someone successfully crash land on water with the regular
> model, tie-downs won't be an option :smile:

well, there goes that trick out the door... the trick of, after a crash, turning 
everything off and setting the tie-downs so that when we restart we restart 
clean and tied down like there never was an accident...

or does crashing turn off the automatic saving of the settings unless we do a 
repair?? (which may not get us into a flyable condition)

wlbragg:
>does crashing turn off the automatic saving of the settings

Interesting observation, we may need to address this.

gilbertohasnofb:
That's a tricky one: after you crash, the sim does not reload automatically but you will stare at a broken aircraft. This mean that if we would reset it to the previous state (switches, tie-downs, etc.) they would appear IMMEDIATELY after a crash, which is silly. To be honest, given that you crashed a plane, there is no reasonable "previous state" to compare to :smile: Maybe having to tick a few check boxes to enable the tie-downs is the price to pay for crashing such a beautiful plane :wink:

wlbragg:
>Maybe having to tick a few check boxes to enable the tie-downs is the price to pay for crashing such a beautiful plane :wink:

You got a point, that is a reasonable penalty.

wkitty42:
what i was trying to say, knowing that the sim doesn't reset and that our repair doesn't yet flip the craft upright (or did i miss that??), is that if you crash, you exit and the state of the craft is saved... so the next time you start the sim, the craft's settings are whatever they were when you crashed... the thought, after i actually wrote it and @wlbragg made his comment, is that a crash state won't save any craft settings so you can start it again in a good/proper state with the craft properly shut down and secured... including the wheels being down on the amphibious :wink:

gilbertohasnofb:
> you exit and the state of the craft is saved... so the next time you start the sim, the craft's settings are whatever they were when you crashed...

wkitty42:
On 11/28/2015 06:06 PM, Gilberto Agostinho wrote:
>     you exit and the state of the craft is saved... so the next time you start
>     the sim, the craft's settings are whatever they were when you crashed...

right... that's what i'm talking about possibly having the crash flag 
disabling... especially if the tie-downs are not available on the water... if 
you crash on water, you won't be able to set them after turning everything else 
off... then when you exit and start the sim, the craft won't have the tie-downs 
because they weren't allowed while you were crashed on the water...

my "trick" comment above was about crashing and then setting everything as you 
would if you had landed and secured the craft properly... you won't be able to 
do that if the tie-downs are not allowed on water if you crash on water :wink:

is that clearer? :wink:

gilbertohasnofb:
What you are writing is clear to me, what I question is the usefulness of it and how to implement it properly.

> then when you exit and start the sim, the craft won't have the tie-downs because they weren't allowed while you were crashed on the water...

But are you aware that not everyone will want the tie-downs enabled like that?

> my "trick" comment above was about crashing and then setting everything as you would if you had landed and secured the craft properly... you won't be able to do that if the tie-downs are not allowed on water if you crash on water

What's the difference between checking the tie-down boxes before resetting the sim or after restarting the sim?

wkitty42:
On 11/28/2015 07:14 PM, Gilberto Agostinho wrote:
> What you are writing is clear to me, what I question is the usefulness of it and
> how to implement it properly.

how to implement it properly? there's apparently a check for exiting the sim 
which causes the craft's setting to be saved... i would extend that check to 
also see if the crash property is true... if it is, do not save the settings... 
keep the original ones...

>     then when you exit and start the sim, the craft won't have the tie-downs
>     because they weren't allowed while you were crashed on the water...
>
> But are you aware that not everyone will want the tie-downs enabled like that?

if they are operating fairly realistically and starting with a cold secure 
craft, why wouldn't they?

>     my "trick" comment above was about crashing and then setting everything as
>     you would if you had landed and secured the craft properly... you won't be
>     able to do that if the tie-downs are not allowed on water if you crash on water
>
> What's the difference between checking the tie-down boxes before resetting the
> sim or after restarting the sim?

fluidity... one would normally approach their cold secure craft and start their 
preflight checkout... they wouldn't approach and set everything off and string 
the tie-downs and then do their preflight undoing what they've just done...

look, it shouldn't be that hard to do... just one more condition to check as i 
note in my first paragraph above... crashed? no craft settings saved... no 
crash, save the settings like now...

wlbragg:
I think it wouldn't be a bad idea, but I'm not sure we can do it without a source code change.

As far as I know the only "save setting " code is in c172p-set.xml
~~~xml
<!-- Save some properties when the simulator is shut down -->
        <aircraft-data>
            <path>/sim/rendering/shadow-volume</path>
            <path>/sim/model/occupants</path>
            <path>/sim/model/c172p/allow-securing-aircraft</path>
            <path>/sim/model/c172p/pitot-cover</path>
            <path>/sim/model/c172p/chock</path>
            <path>/sim/model/c172p/tiedownL</path>
            <path>/sim/model/c172p/tiedownR</path>
            <path>/sim/model/c172p/tiedownT</path>
            <path>/sim/model/c172p/brake-parking</path>
            <path>/instrumentation/airspeed-indicator/serviceable</path>
            <path>/engines/active-engine/oil_consumption_allowed</path>
            <path>/engines/active-engine/oil-level</path>
            <path>/fdm/jsbsim/settings/damage</path>
            <path>/fdm/jsbsim/bushkit</path>
            <path>/controls/engines/active-engine</path>
            <path>/environment/aircraft-effects/cabin-heat-set</path>
            <path>/environment/aircraft-effects/cabin-air-set</path>
            <path>/fdm/jsbsim/running</path>
            <path>/fdm/jsbsim/complex</path>
            <path>/fdm/jsbsim/weather</path>
            <path>/consumables/fuel/save-fuel-state</path>
            <path>/consumables/fuel/contamination_allowed</path>
            <path>/instrumentation/save-switches-state</path>
            <path>/consumables/fuel/tank[0]/level-gal_us</path>
            <path>/consumables/fuel/tank[1]/level-gal_us</path>
        </aircraft-data>
~~~
There is no logic mechanics to put a condition on it.

wkitty42:
ok... just toying around here... maybe something like the following??
```
   <ifthen>
      <eq>
        <property>/sim/crashed</property>
        <value>'false'</value>
      </eq>
      <aircraft-data>
        blah blah
      </aircraft-data>
    </ifthen>
```
or possibly
```
   <condition>
      <not>
        <property>/sim/crashed</property>
      </not>
      <aircraft-data>
        blah blah
      </aircraft-data>
    </condition>
```
i don't know if that will remove those path values from the aircraft-data property when crash is true, though... like i said, just playing around and brainstorming... i am trying to test and see if changing /sim/crashed from false to true and back causes /sim/aircraft-data to appear and disappear but i'm not having any luck... it is either there without either attempt or it is gone with either of them... i also thought to try to put the conditional inside the aircraft-data tree but that only added the condition, not and property items to a tree inside aircraft-data :disappointed: 

there should be a way and i just don't know enough about it to know which way to go or how to do it... if a core code change is needed to enable this capability, we certainly can ask for it... i think the condition is available and we just need to be able to either do this or not...

i'm starting to think and actually hoping that the -set.xml file is a static xml file and doesn't get executed like other xml code... it doesn't make sense that it would so perhaps moving this section to another xml file which is executed in the loop would allow for this check to work? maybe something like the xml files in System with a PropertyList and logic layout would work?? that may not be desirable, though, if they are written or removed from the tree x times per second...

i think i'll go ask the "stupid" question and see what happens...

PR #576, opened 2015-11-26
Bug 462: Replace Nasal implementation for fog and frost with JSBSim system

### LINKED ISSUE #462: Cabin temperature gui.popup too often
When having cabin temperature either below 0°C or above 32°C a gui.popup-message (guess what?!) pops up. But this continously, so this blocks other simulation occurences, e.g. throttle percentage or even cabin air/cabin heat readout. This makes adjustments very difficult, especially in order to react on that specific message.

onox:
Fixes #462 

* The dew point of the temperature in the cabin is used instead of the dew point of outside temperature

* I replaced the "Disable frost and fog" setting with "Enable frost and fog"

* The maximum additional heat provided by the engine is 50.0 * EGT (which is usually between 0.4 and 0.6). I consider it to be sufficient for most area's except those were it's like -40 degrees celsius. But we can increase it if necessary.

* Instead of continuously spamming you with GUI tooltips, you now only get one tooltip when you go below 0 C or above 32 C. When the temperature gets within 0 and 32 C again, you get another tooltip (once).

I haven't yet implemented any frost point, I'm simply doing 3 * moisture, just like the Nasal implementation did. Also, how fast moisture increases/decreases (condensation/evaporation) should maybe depend on the relative humidity in the cabin, but I haven't implemented that currently.

For those who like to have properties on their screen, you could display the following:

Environment:
* `/environment/temperature-degc`
* `/environment/relative-humidity`

JSBSim:
* `/fdm/jsbsim/heat/cabin-air-temp-degc`
* `/fdm/jsbsim/heat/cabin-dewpoint-degc`
* `/fdm/jsbsim/heat/surface-temp-degc` (As long as the surface temp is lower than the dew point, you're going to get fog or frost)

* `/fdm/jsbsim/heat/surface-dewpoint-diff` (If this property is negative then fog or frost is decreasing. If it is positive then the fog or frost is increasing)
* `/fdm/jsbsim/heat/cabin-heated-air-degc` (Extra heat controlled by the cabin heat slider + EGT(!))

* `/fdm/jsbsim/heat/fog-level`
* `/fdm/jsbsim/heat/frost-level`


wlbragg:
I just pulled it and am testing it now.

wlbragg:
>how fast moisture increases/decreases (condensation/evaporation) should maybe depend on the relative humidity in the cabin

I agree.

A couple questions,

>I replaced the "Disable frost and fog" setting with "Enable frost and fog"

>Instead of continuously spamming you with GUI tooltips, you now only get one tooltip when you go below 0 C or above 32 C. When the temperature gets within 0 and 32 C again, you get another tooltip (once).

When we talked about changing "Disable frost and fog" to "Enable Auto Climate Control", (weren't the exact words , but you get the point), it was the idea to allow the user to check a box to automatically keep a stable climate when they got the first warning that the climate was adverse. I take it you decided against this idea. Why then change the default to off instead of on?

Also if we don't address some form of continuous warning about the adverse temperature in the cockpit, then the one warning won't mean much and the occupants will continue to freeze or roast.
Thus, my idea of a choice of automatically adjusting the sliders upon getting the warning.

Still testing!

wlbragg:
I'm testing at FHAW and gained altitude to get into some pretty cold air.
Outside temp is around -4.65c.

 I have the heat and air wide open and the cabin temp is still falling?

I got fog and no ice, is the ice still working as in Nasal or not at all?

Opening heat and air slider to max eventually started taking the fog away. I liked the speed of that. It is of course a very subjective and hard to measure event.

wlbragg:
I have ice now.

wlbragg:
Once the outside airtemp reached -4c+, it appeared that the heater couldn't keep up. 

I'm also getting some instances of instant fog and then none. It may just be that sweet spot, it may have always been that way, more testing required.

I have to quit for the night.

wlbragg:
By the way, looks nice. I like the idea and arrangement of the system.

onox:
> When we talked about changing "Disable frost and fog" to "Enable Auto Climate Control", (weren't the exact words , but you get the point), it was the idea to allow the user to check a box to automatically keep a stable climate when they got the first warning that the climate was adverse. I take it you decided against this idea. Why then change the default to off instead of on?

The setting is unrelated to your idea. It only changes a negative setting to a positive one. Just like how "Show shadow" and "Human models" are positive settings. The setting used to be "Disable frost and fog" with default off, and I changed it to "Enable frost and fog" with default on.

We can still add (or replace "Enable frost and fog") your "Enable Auto Climate Control" setting.

> Also if we don't address some form of continuous warning about the adverse temperature in the cockpit, then the one warning won't mean much and the occupants will continue to freeze or roast.
Thus, my idea of a choice of automatically adjusting the sliders upon getting the warning.

How about displaying red text at the top of the screen (like when you start FG example, that kind of text) instead of a GUI tooltip? The GUI tooltips are basically meant for displaying info when hovering above items in the cockpit, not to give status info like we do.

> Once the outside airtemp reached -4c+, it appeared that the heater couldn't keep up.

How is your engine's EGT and cabin heat slider? If outside temp is just -4 C, then it should be no problem getting the temperature above 0.

onox:
> I have the heat and air wide open and the cabin temp is still falling?

Make sure your engine is on.

Now I think of it, I should add support for open doors or windows.

onox:
:warning: Don't merge this PR yet. Need to improve some things. :warning:

wlbragg:
>How about displaying red text at the top of the screen

I like that idea.

>Make sure your engine is on.

I was at 8000 ft and climbing, it better be on! :smile:

onox:
Cabin heat and cabin air knobs know work like an old fashioned two-sided water sink as @tigert explained. Overhead air vents can be dragged/scrolled to open/close them. Opening doors and windows will influence the cabin temperature.

Only thing that remains now is taking the number of humans into account for the relative humidity and cabin temperature.

wlbragg:
>Overhead air vents can be dragged/scrolled to open/close them.

We have overhead air vents? Go figure! :laughing:

wlbragg:
See #582

onox:
:warning: Don't merge this PR yet. Need to improve some things. :warning:

onox:
PR is finally ready. Can some people test it please?

onox:
I've put `logger.nas` in `Nasal/generic/` to indicate that it's not c172p specific and can be re-used by other aircraft. For the tiedowns I will probably add [`math_ext.nas`](https://github.com/onox/ExpansionPack/blob/master/Nasal/math_ext.nas) in that directory as well.

gilbertohasnofb:
> PR is finally ready. Can some people test it please?

@wlbragg would you mind doing this yourself as you are much more familiar with these effects than I am?

onox:
@gilbertohasnofb You can also help test this :smile::

1. Test in very hot and very cold places that when you open overhead vents, cabin air, or doors that temperature goes up or down.

2. In very cold places, temperature should go up if you open cabin heat + have sufficient airspeed or propeller RPM (for airflow) + sufficient EGT (increase throttle + lean mixture).

3. Try to test that you're able to get rid of the fog (after a while).

* `/fdm/jsbsim/heat/cabin-air-temp-degc`
* `/fdm/jsbsim/heat/cabin-dewpoint-degc`
* `/fdm/jsbsim/heat/surface-temp-degc`
* `/fdm/jsbsim/heat/surface-dewpoint-diff` (If this property is negative then fog or frost is decreasing. If it is positive then the fog or frost is increasing)

Note that you get fog/frost if cabin air dewpoint > surface temperature. If you increase the temperature (using cabin heat) then cabin temperature + dew point goes up. So you should get some fog then. After a while, the surface temperature catches up (it changes more slowly) and becomes lower than cabin dew point, so fog disappears again.

Note that cabin air temp, dew point, and humidity starts at 0 unfortunately (because of JSBSim `<integrator>` elements) (I will probably ask on the mailing list whether they can fix it to use initial value), so open the doors to quickly synchronize the cabin air temperature.

onox:
If you add more humans, you get more heat + higher humidity in cabin.

wlbragg:
>Note that cabin air temp, dew point, and humidity starts at 0 unfortunately (because of JSBSim <integrator> elements)

I'll have to see what your referring to, because it was a nice feature on a cold startup that the ice or fog was automatically present.

Will be testing here over the next day or two.

onox:
> on a cold startup that the ice or fog was automatically present.

Yeah, the `moisture-norm` property is an integrator as well. I'll send an e-mail to the mailing list to ask if they can fix it. It can probably be fixed by routing every integrator to a summer (which sums it with a default value), but I hope to be able to avoid that.

gilbertohasnofb:
@onox sorry, I have been busy during these last days, I will try to test it tonight.

wlbragg:
Me too!

onox:
I've been thinking about the initial fog/frost issue: when I do `debug.dump(getprop("/environment/temperature-degc"));` in `Nasal/c172p.nas` in the `/sim/signals/fdm-initialized` listener at the end of the file, I got often incorrect temperatures. Even though summing the integrator output with this initial temperature works, it doesn't help at all if the initial temperature is incorrect.

For example in UEEE where it's about -32 C, I got 34.81001880000002 and when I then move to KNID where it's around 0, I get -12.54609119999998.

Also, personally I have found the full fog/frost when starting up FG to be not always correct and also annoying at times. Which is why I often just disabled the whole fog/frost system in the Aircraft Options window.

I think for now maybe 0 is the best option. People (and new users) then get a clear windshield and don't immediately have to worry about it. The best would be if we have a `/sim/signals/weather-initialized` signal to indicate to any code that the temperature, etc., can be safely read.

onox:
Besides that, I can display a short message though how to remove the fog/frost when `moisture-norm` > 0.5.

@gilbertohasnofb @wlbragg Can some of you test this PR? Before you start testing, just open the doors for 1 or 2 minutes then close them. Play with adding some humans and opening/closing windows or the overhead vents for example. See https://github.com/Juanvvc/c172p-detailed/issues/462#issuecomment-159987264

onox:
I've been flying with the branch for a while now, and it all works. It's mainly some constants that could be tuned. The fog/frost appears not too fast. I'm not sure humans add enough heat. Maybe it's better to just merge the PR, so that we can later hear if people think fog needs to be appear quicker or not.

All the integrators start at 0 for now, so don't expect any wild changes in the first few minutes after starting FG. Open doors/vents for speed up.

wlbragg:
>The best would be if we have a /sim/signals/weather-initialized signal to indicate to any code that the temperature, etc., can be safely read.

Are you saying this isn't available as of now, we need it exposed from the source? There has to be something exposed already that we can use.

I'm glad it's merged, we should get some feedback if it's got any issues. Great work onox!

onox:
The whole property doesn't exist (yet). I just thought of it on the spot. I'll send an e-mail to the mailing list to Thorsten.

PR #580, opened 2015-11-28
Cleaning up tiedown code, tiedowns not allowed when on water

gilbertohasnofb:
Some tests that can be done to ensure everything works well:
- confirm tie-downs are still working as they should, stopping the aircraft when any of them is enabled
- confirm that both the dialog checkboxes as well as the hotspots are working as they should (not on water, groundspeed < 1kts, no pontoons)
- add tie-downs to some variant and change the variant. If pontoons => tiedowns should disappear
- add tie-downs to amphibious, close the sim and load it in a port airport such as SP01, tiedowns should disappear


onox:
About the "If tiedowns-can-have is false then no tiedowns visible" `<logic>`:

1. Have you tried to use:

    ```xml
    <input>
        <false/>
    <input>
    ```

    I think that's more clear than using a property (that in this case happens to always be false).

2. For the outputs:

    ```xml
    <output>
        <property>/sim/model/c172p/tiedownL-visible</property>
        <property>/sim/model/c172p/tiedownR-visible</property>
        <property>/sim/model/c172p/tiedownT-visible</property>
    </output>
    ```

   I found this in code in my other projects. But it should work. Is a bit more concise.

onox:
In the `/sim/model/c172p/tiedown{L,R,T}-addable` properties you have:

```xml
<less-than>
    <value>1.0</value>
</less-than>
```

Is it possible to move it to `/sim/model/c172p/tiedowns-can-have`?

onox:
The `c172-checklists.xml` still uses `sim/model/c172p/tiedown{L,R,T}` properties.

onox:
For the pitot-cover there are currently 4 properties in use. Can you investigate whether to reduce to 3 or 2?

onox:
One last comment: perhaps the properties for the tiedowns, pitot-cover, and chocks can all be moved to, lets say, `/sim/model/c172p/securing/`. Then they're grouped together.

gilbertohasnofb:
@onox all your comments make sense, I will revise all this tomorrow and make a new commit, okay? Thanks for taking a look at the code!

gilbertohasnofb:
@onox I think I fixed everything you mentioned. As for the `pitot-tube` properties, I manage to reduce it to 3 instead of 4, I hope that's fine. Given that I basically touched everything related to pre-flight inspection here, I'd say we need to make a very thorough test to ensure everything still behaves as expected (of course I tried testing everything myself as well and it seems to be fine).

onox:
Code and testing OK. I'll merge master into this branch so you can fix `Tutorial/preflight.xml`.

gilbertohasnofb:
>  I'll merge master into this branch so you can fix Tutorial/preflight.xml.

Let me fix that with this PR

gilbertohasnofb:
Ah, I see now, that file isn't in this branch. Okay, merge and I will create a new branch and new PR for that.

onox:
Hang on. Pushing.

onox:
Pushed. You can fix it in this branch.

gilbertohasnofb:
:+1:

onox:
Just pull first so you get my merge commit.

gilbertohasnofb:
@onox Done. I tested the pre-flight tutorial and found the following problems, should I open an issue for all of them?:

- the tutorial starts with instrument lightning at 100% during day!
- close door pink circle in the wrong place (immediately after leaving the cockpit after checking the fuel level)
- funny cameras, sometimes they rotate in an anti-intuitive manner (like rotating 300 degrees instead of 60, thus looking away from the plane for some moments)

onox:
The funny camera's is something @dg-505 has problems with too. Not sure whether that can be solved.

gilbertohasnofb:
Well, I will open an issue for all of those and let him know about them. So if you are fine with my modifications to the pre-flight tutorial, you can merge this PR.

gilbertohasnofb:
I found another problem after the merge you did: the checklist xml is still showing the old property for the tail tie-down... I will fix that in this very branch, ok?

onox:
Ok.

PR #582, opened 2015-11-29
Re-UV map air vents, remove vent shadow effect, remodel air vents.

wlbragg:
This should correct a few issues with the vents. Animating them exposed a shadow shader issue with the left vent. Between that and the layers of glass it appeared the closed vent disappeared into the fuselage. Also the vent hole texture was not designed for animation and looked bad during animation. That has been reworked requiring remodeling and new UV mapping along with a change to the texture sheet panel_parts.png.


Models/c172p.xml
```diff
@@ -388,8 +388,6 @@
     <!-- interior effects -->
     <effect>
         <inherits-from>Effects/interior/c172p-interior</inherits-from>
-        <object-name>AirVent_left</object-name>
```
onox: Why were these removed? Objects will now appear black when using the flash light.

onox:
@wlbragg In https://github.com/Juanvvc/c172p-detailed/pull/576#issuecomment-160314200 you're saying:

> It is ALS or the effects that we are applying to the object that are mucking it up.

But what is mucked up exactly?

wlbragg:
I   can add the flashlight to them. Hang on and I'll post a picture of the problems.

onox:
That would be helpful. It appears you have removed the objects from the effect for a reason, but I don't understand what the reason is yet.

onox:
Btw, what I find very ugly about the glass (always has been), is that it ends near the door, but it should end around the leading edge of the wing. That's why the overhead vents are sticking through the glass.

wlbragg:
vent1
vent2
vent2a
vent3
vent4
vent5

wlbragg:
At 10% it appears to actually sink into the air frame.

onox:
And removing the two objects from the interior effects solves that?

wlbragg:
It solve the appearance of sinking into the frame.  In fact it is the glass that is blending with but only with the effect applied. The vent hole was another issue, that required a model edit an UV and texture change.

wlbragg:
Let me add the flashlight effect before merging.

wlbragg:
vent9
vent8
vent7
vent6

onox:
Okay, so we can have the flash light effect?

wlbragg:
Yep.

wlbragg:
venta
ventb
ventc

onox:
@wlbragg Did you see my comment about the `../../../../`?

wlbragg:
Done!

onox:
Thanks. I was a bit surprised it would still work.

wlbragg:
I know, how it reconciled that is puzzling.

PR #584, opened 2015-11-29
Some optimizations to the liveries

gilbertohasnofb:
- optimizing `Models/panel_parts.png` after @wlbragg new addition to the vent tube
- synchronizing modifications to `panel_parts.png` and `panel.png` to all liveries (new vent tube and water rudder textures for the first, float bar for the latter). This does not directly affect the aircraft as these additions  weren't livery dependent, but it makes things less messy: now all `panel_parts.png` and `panel.png` have the same objects on it.
- re-organizing the liveries in subfolders inside `Models/Liveries`, so only the xml files are in that folder


wlbragg:
Oops, sorry, I forgot this mod could effect other .png's.

gilbertohasnofb:
The vent wasn't livery dependent, so your modification wasn't doing any harm to any livery. But I think it's better to keep all these files synchronized, else things get messy very easily :smile:

PR #594, opened 2015-12-03
Bug 593: Reset pontoon leakage to 0 when undamaged

### LINKED ISSUE #593: Clicking "Repair" button should reset pontoon leakage
If the pontoons are damaged or broken, they start leaking (when on the water) and sink. Clicking the "Repair" button should reset the leakage to 0.

onox:
Fixes #593 

1. Start near the sea and switch to amphibian or sea plane

2. Get in the air and smash the aircraft to the ground near the water

3. Check that one or both of the pontoons are damaged or broken

4. Press Ctrl+u to get in the air and land on the water

5. You should see the damaged/broken pontoon start sinking a bit

6. Now click the "Repair" button. The leakage should disappear and the pontoon should fully float again.

@andgi To answer your question (in `Systems/bushkit.xml`): to reset a pid's integrator, you set its trigger to a negative value. In this case I used -1 as long as a pontoon is undamaged.

### DIFF
--- a/Systems/bushkit.xml
+++ b/Systems/bushkit.xml
@@ -282,9 +282,19 @@ Extra weight and drag due to bush wheels, floats and 180 hp engine
             </function>
         </fcs_function>
 
-        <!-- FIXME: How to reset this to 0 on repair? -->
+        <!-- No water leakage as long as pontoon is not damaged or broken -->
+        <switch name="left-pontoon/leaked-water-windup-trigger">
+            <default value="-1"/>
+
+            <test logic="AND" value="0">
+                pontoon-damage/left-pontoon GT 0
+            </test>
+        </switch>
+
+        <!-- If trigger is negative then the integrator is reset to 0 -->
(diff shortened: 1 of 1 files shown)


wlbragg:
Works great! Do you know if you can damage the pontoons by landing to hard on water?

Might not be a bad idea to have some random water leakage damage.

onox:
> Do you know if you can damage the pontoons by landing to hard on water?

Maybe. I couldn't quickly get damaged pontoons by crashing into the water yesterday, it would only break my wings that way. So I found crashing on land much easier.

> Might not be a bad idea to have some random water leakage damage.

@andgi made the amount of water that leaks per second fixed per damage state (5 lbs/s if damaged, 250 lbs/s if broken)

wlbragg:
I meant that there could be a random event that the floats sustained damage so the next time you land or start on water the floats are sunk. I guess that should be part of the random failure system, no more so than the "water in oil" though I guess. Does FG still implement "random failures"?

Issue #461, opened 2015-10-08
Clean-up ALS shadows code


wlbragg:
Because we opted to use only the 3d shadow and not the original 2d shadow, there is superfluous code in all the `*shadow.xml` files. I will eventually get to this unless someone beats me to it :smile:

Note to self, also remove `shadow.png`, extra autopilot filter in `Systems/ground-effects.xml` and maybe a  property def in `c172p-set.xml`

gilbertohasnofb:
Just being a bit pedantic here, but because this only involves cleaning up some code then it isn't a new feature and therefore we should use the release milestone rather than feature freeze (meaning also that we would have some months more to solve it if we don't do it until the feature freeze).

wlbragg:
:+1:

wlbragg:
Just starting to go through this and now I am confused (who me!). When we first added the shadow there were actually three shadows available, non-ALS, ALS 2d and ALS 3d. It looks like when we lost the 2d we may have also removed the Non-ALS, does this ring any bells?
Was there a reason we got rid of the Non-ALS shadow?

wlbragg:
This is where we loose it I think
https://github.com/Juanvvc/c172p-detailed/commit/d49163307e27c006d7b1c2fda676b0a26be3da3c
Note the last block of code. That translation is only for the NON-ALS shadow.
So I think I must have had the ALS 2D and the NON-ALS rolled up into this same bit of shadow logic.

wlbragg:
Do we want the NON-ALS shadow back? I need to know before I clean this up.

gilbertohasnofb:
> Do we want the NON-ALS shadow back?

Hmm, I would say probably not. If someone is using Rembrandt then they already have shadows. On top of that our 3D shadow is pretty good I'd say. What do you think, @onox ?

onox:
I vote for not bringing the shadow back. If people want FG to be pretty, they should use ALS.

wlbragg:
OK, just to be sure, make sure your not confusing a GUI chioce of 2D or 3D, that's not what I am talking about. This would not even be selectable to the user (other than the choice of shadow or no shadow). If they are on ALS they get the ALS shadow if chosen. If not on ALS or Rembrandt they get the non-ALS shadow (if they have chosen to display shadow at all).
I really don't care either and would just as soon not have to re-work this again, but I want to make sure this isn't taking away from some that don't have the ability to use ALS or Rembrabdt but still want shadows.

[...]

PR #602, opened 2015-12-07
bug-461

wlbragg:
Unless we decide to do any changes as described in #461 this is all that is left to the cleanup.

There is an unneeded filter in ground-effects.xml for output to `/position/gear-agl-m` that was being used for the 2d shadow. I don't think it is used for anything else so if we want to delete it we can.


onox:
`shadow.png` is used by `shadow.ac`, but that file doesn't seem to be in use (it looks like it was used by the 2D shadow).

You can delete the unneeded filter.

wlbragg:
>shadow.png is used by shadow.ac

Oops, forgot that one. I'll get to it tomorrow.

onox:
I can do it right now.

Issue #610, opened 2015-12-10
Returning to repaired aircraft during replay with crashed gear makes gear stick in the ground


onox:
It is possible to make the gear stick through the ground by doing the following:

1. Crash the gear
2. Repair the aircraft
3. Watch the replay and then exit the replay mode while the gear is crashed
4. After having exited the replay mode, the gear stick through the ground because the z-positions aren't reset

I've already fixed this here locally.

PR #611, opened 2015-12-10
Bug 608: Kill engine if propeller is broken

### LINKED ISSUE #608: Engine should be killed if propeller is broken
Engine should be killed if propeller has stopped (broken). This should also happen if you for example break the propeller by collapsing the nose gear.

Can be fixed by using `/engines/active-engine/killed`. It seems it is currently used in the animations (that is good), but not by the FDM or engine systems (because if I set full throttle, I still get RPM). Engine is still running in this screenshot:

fgfs-screen-403

onox:
Fixes #608 

Not all calls to `killengine()` can be removed (read: moved to XML) because some depend on the altitude at the moment a wing gets damaged.

In `Nasal/engine.nas` I kill the engine temporarily for 0.25 seconds when oil level is low. I have also increased the maximum random duration from 3 to 10 seconds. 3 would give too much coughing.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -253,7 +253,9 @@
         <water-rudder-norm>sim/multiplay/generic/float[5]</water-rudder-norm>
         <engine>
             <rpm>engines/engine[2]/rpm</rpm>
+            <crashed>sim/multiplay/generic/int[3]</crashed>
         </engine>
+
         <!-- TODO To be made MP properties in the future when we get passengers or dual-control -->
         <controls>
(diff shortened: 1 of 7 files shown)


onox:
After this PR gets merged I have another PR coming up that fixes #610

gilbertohasnofb:
I will test and merge this tonight if no one else does it first.

gilbertohasnofb:
:+1: 

> In Nasal/engine.nas I kill the engine temporarily for 0.25 seconds when oil level is low. I have also increased the maximum random duration from 3 to 10 seconds. 3 would give too much coughing.

Looks and sounds better

PR #615, opened 2015-12-11
Bug 610: Fix gear sticking in ground when exiting replay mode with damaged gear

### LINKED ISSUE #610: Returning to repaired aircraft during replay with crashed gear makes gear stick in the ground

onox:
Fixes #610 

Follow the instructions in that issue to see that you can no longer reproduce it (exiting replay mode should show repaired gear (not sticking through the ground) if you had it repaired before entering replay mode)


onox:
@wlbragg Unrelated to the issue above, but related to the z-positions: why do you only set the z-position for unit 5 in `bothwingcollapse`. Why not for unit 4 as well?

Also, unit 4 and 5 have a default z-position of 50, so why setting it to 50 again in `upsidedown`? It seems redundant.  Setting unit 12's to 90 in that function looks strange to me as well. It's always 90 unless you have pontoons/amphibious and are on/above the water.

onox:
Is it true gear unit 23, 24, and 25 are always -22 for ski's, and always 0 for all the other bushkits?

onox:
Another thing: it seems the listener for `contact~"unit[12]/WOW"` never gets triggered (even when it should).

In JSBSim we could do:

If `/fdm/jsbsim/wing-damage/left-wing` == 1.0 and `/fdm/jsbsim/contact/unit[12]/WOW` is true then `/fdm/jsbsim/contact/unit[4]/z-position` is 40, otherwise 50

Same for right wing. In that case we could remove `upsidedown`. Is that correct @wlbragg?

onox:
Anyone who reads this PR but whose name is not @wlbragg can ignore the comments above, and instead test the fix for default/26"/36" bushkits gear issue :smile:

wlbragg:
It's probably going to be later tonight before I can look at this and recall any of my prior actions on it. But I will try to remember why I did what I did.
I do know two things, 
As you probably already know, contacts are a bitch!
At the time I did this I didn't totally understand how the contacts worked (or more precisely, how they position), I still don't.
I could have made lots of mistakes on this. Just beware, where I left it gave me the least amount of superball bounce at the most impact angles. As far as the particulars of why and what, I'll refresh my memory tonight and try to answer your questions.

wlbragg:
Obviously the code has had multiple hands and refactoring done to it so what I did and what remains may not be the original intent.

> why do you only set the z-position for unit 5 in bothwingcollapse. Why not for unit 4 as well?

Without testing to see what actually happens I'd say that I either left it out accidentally or it got removed accidentally somewhere along the way.

>Also, unit 4 and 5 have a default z-position of 50, so why setting it to 50 again in upsidedown? It seems redundant.

Unless it is getting set somewhere that isn't obvious, it should be at 50 on reset damage and poll for water if over land, so I can't see it being needed at that location.

>Setting unit 12's to 90 in that function looks strange to me as well. It's always 90 unless you have pontoons/amphibious and are on/above the water

It may have been to account for a special circumstance when you flipped at the edge of water and land, a case where you get stuck with 160 but need 90?

>Is it true gear unit 23, 24, and 25 are always -22 for ski's, and always 0 for all the other bushkits?

Unless damaged, yes, that should be true.

>In JSBSim we could do:
If /fdm/jsbsim/wing-damage/left-wing == 1.0 and /fdm/jsbsim/contact/unit[12]/WOW is true then /fdm/jsbsim/contact/unit[4]/z-position is 40, otherwise 50In that case we could remove upsidedown. Is that correct

Technically yes, except possibly for the special circumstance at the water/land edge where we need to set it to 90 (if that was even the case, but I do recall that special circumstance) . If I remember correctly, the 160 on contact 12, was to get the right pitch on upside-down on water, 90 on land. It's been so long since I created it I really don't know all the reasoning behind what I did (shame on me for not commenting more at the time).

onox:
> Unless damaged, yes, that should be true.

What should their values be if the (ski's) gear is damaged?

What's the purpose of unit 23, 24, 25? To create additional friction?

onox:
> Unless it is getting set somewhere that isn't obvious, it should be at 50 on reset damage and poll for water if over land, so I can't see it being needed at that location.

The polling is only done for pontoons/amphibious, but with default/26"/36" gear you can crash upside down in the water as well. So the polling should be done for those bushkits too?

onox:
> What should their values be if the (ski's) gear is damaged?

I've got the same question for units 19, 20, 21, and 22 (amphibious) :smile:

onox:
> It may have been to account for a special circumstance when you flipped at the edge of water and land, a case where you get stuck with 160 but need 90?

It seems 160 is needed to put the tail high in the air for the pontoons/amphibious. Otherwise (for value 90) the aircraft is mostly horizontal.

wlbragg:
>It seems 160 is needed to put the tail high in the air for the pontoons/amphibious. Otherwise (for value 90) the aircraft is mostly horizontal.

That's right, all the videos I watched of pontoons flipping over in water left them sticking tail up at like 45 deg or so. That's what the 160 was for.

>but with default/26"/36" gear you can crash upside down in the water as well. So the polling should be done for those bushkits too?

I had no reference for anything but pontoon upside down, but now that you mention it, any version should stick up the same way (tail up). I guess it is because we "never" expected a crash over water without pontoons. It is the event of landing and taking off that increased the odds.
I'm researching the rest of the questions and will post my answers shortly.

wlbragg:
>The polling is only done for pontoons/amphibious, but with default/26"/36" gear you can crash upside down in the water as well. So the polling should be done for those bushkits too?

I still can't see any reason for that "else 50". If you haven't damaged your wings, they are already at 50, if you have upside down condition and damage, that is the only time they change to 40 (it's the only time they need it (so the damaged tips don't stick into the ground). Reset damage changes it back to 50. The value only matters if upside down, it has no influence right side up (I don't think). But it is odd I put that else in there.

wlbragg:
What should their values be if the (ski's) gear is damaged?
        setprop(gears~"unit[24]/z-position", -14);
        setprop(gears~"unit[25]/z-position", -16);

What's the purpose of unit 23, 24, 25? To create additional friction?
Not so much, more to allow the wheels attached to the skies to have contact with the ground (for spin). 23 is both small wheels on front ski and 24, 25 are rear ski wheels one on each side. They are "bogey's".

onox:
> Not so much, more to allow the wheels attached to the skies to have contact with the ground (for spin). 23 is both small wheels on front ski and 24, 25 are rear ski wheels one on each side. They are "bogey's".

Maybe I don't understand those skis yet, but we already have contacts 0, 1, and 2, so why not use those? Since the skis is basically just the 36" with some additional metal plates attached to them :smile:

wlbragg:
> but we already have contacts 0, 1, and 2, so why not use those? Since the skis is basically just the 36" with some additional metal plates attached to them

No, they also have wheels on them!

onox:
Oh, yeah, now I see those small wheels. Good I didn't touch the skis yet.

onox:
> if you have upside down condition and damage, that is the only time they change to 40 (it's the only time they need it (so the damaged tips don't stick into the ground). Reset damage changes it back to 50.

But that's only happening for the default/26"/36" bushkits. For the pontoons/amphibious the Nasal code does/did polling and sets the wing tips to -10 if you're on water. In case that old `upsidedown` function got triggered, it would be 40 only for a couple of milliseconds, before the next iteration of that polling code sets it to -10 again.

onox:
So when exactly do we want the wing tips to be -10? Old and new code just sets it to -10 if you use pontoons/bushkit and are on the water. It doesn't care about wing damage.

wlbragg:
I'm not sure I understood that.
The 40 is needed "over land" and "upside down" I guess the -10 is for "over water and "upside down", correct?

onox:
Nomatter which bushkit you use?

wlbragg:
I didn't see the -10, let me look.

onox:
https://github.com/Juanvvc/c172p-detailed/commit/ea0e96eaef213a02a15b5112c656c0e4d5d20895

onox:
You can see it in the old code as well the new code.

wlbragg:
>Nomatter which bushkit you use?

Oh, du, yes, it needs to be -10 upsidedown over water and 50 upsidedown over land unless damaged and then it is 40 upsidedown over land. I'm not sure about damaged upsidedown over water, probable still just -10.

onox:
So always -10 over water, damaged or not damaged?

onox:
> I had no reference for anything but pontoon upside down, but now that you mention it, any version should stick up the same way (tail up). I guess it is because we "never" expected a crash over water without pontoons. It is the event of landing and taking off that increased the odds.

In that case I will set the unit 12 to160 if on water, and 90 if on land, no matter which bushkit you use. Is that OK?

wlbragg:
>So always -10 over water, damaged or not damaged?

Let me go try to test that and confirm it.

>In that case I will set the unit 12 to160 if on water, and 90 if on land, no matter which bushkit you use. Is that OK?

Yes.

wlbragg:
>So always -10 over water, damaged or not damaged?

Good luck trying to test this. Just do it. It should be correct.

onox:
I don't see much/any difference between 40 and -10, but the logic is now as you said it.

onox:
```javascript
setprop(gears~"unit[24]/z-position", -14);
setprop(gears~"unit[25]/z-position", -16);
```

@wlbragg Why this difference between left and right?

wlbragg:
Slight differences in the damaged model.

wlbragg:
>I don't see much/any difference between 40 and -10, but the logic is now as you said it.

I really can't say why I had -10, there must have been a reason, but I don't know what it was.

I'll try to test this over the next day or two (regardless of when it gets merged) and see if I can discover the reason (if any) for the -10.

Nice work!

onox:
Thanks. Only contact points remaining are 19, 20, 21, and 22. If you ever need to modify some of the points, you can do that in `Systems/c172p-ground-effects.xml`.

wlbragg:
Looks like unit 23 (front ski) does need adjusting on damage, it's sticking into the ground pretty good.

I'm trying this now but need you to add it to the PR upon verification, I merge all your PR's into my master to test.
~~~xml
        <switch name="z-position-nose-ski">
            <output>gear/unit[23]/z-position</output>
            <default value="0"/>

            <test logic="AND" value="-22">
                bushkit EQ 5
                gear/unit[0]/broken EQ 0
            </test>

            <test logic="AND" value="-14">
                bushkit EQ 5
                gear/unit[0]/broken EQ 1
            </test>
        </switch>
~~~

onox:
You can just switch to the branch with `git checkout bug-610`.

wlbragg:
Changing the value is changing the point in the tree but having no effect on lifting the ski up.

onox:
You're changing `gear/unit[23]/z-position`? If I disable the whole `<switch>` and then alter between -22 and -14, there is some difference, but not much.

wlbragg:
>You're changing gear/unit[23]/z-position?

I went all the way to +14 with no change.

I'm out of here for a while, later!

onox:
Changing `gear/unit[23]/z-position` from -22 to -14 is an improvement though.

onox:
@wlbragg I have moved the remaining contact points. You might want to modify some of the `<switch>` elements for the amphibious (19, 20, 21, 22) for the case they are damaged.

Other than that, this PR is ready and can be tested.

onox:
@wlbragg One last question: what should be the value of contact points 6, 7, 8 (yellow sparks) for the skis? Currently they are 0, -14, and -16 (but those last two values were actually meant for points 24 and 25). The values for the 36" bushkit are -17.7, -16, -17.4.

wlbragg:
>what should be the value of contact points 6, 7, 8 (yellow sparks)

I'll have to test it and see what we need. Sorry I have not been available lately for much testing. I think tomorrow late I should have some free time to look at some of this stuff.

wlbragg:
I'm looking at the ski "spark" issue right now.

Here is what I think I know, we are getting smoke on (ski) gear damage because I thought we could just use the default 36" gear damage routines for the skies, so the sparks aren't there for gear damage because I modeled it to be tire burning and not metal scraping. Now with the skies we need metal scraping. Also because the front gear is below ground after damage, I don't wonder if it is also wrong on 36" bush? Because it is one and the same? I'm checking now.

wlbragg:
Do you want me to add ski spark to the same files the 36" smoke is coming from or do you have something else in mind?

onox:
The front gear of 36" and ski's are controlled by different z-positions.

onox:
For the ski's see line 110 in `Systems/c172p-ground-effects.xml`. It's currently 0 and I have added a TODO on that line.

onox:
> Do you want me to add ski spark to the same files the 36" smoke is coming from or do you have something else in mind?

When you have tire burning, what do you see on the screen? Yellow sparks still?

onox:
Let me explain: the first 3 `<fcs_function>` elements in the file are for the z-position of the gear, the 4th, 5th, and 6th are for the yellow sparks.

wlbragg:
>The front gear of 36" and ski's are controlled by different z-positions.

OK, I guess I threw that in because of the ski edge when damaged. But the gear damage is the same as the 36".

>When you have tire burning, what do you see on the screen? Yellow sparks still?

I'm checking that now.

onox:
For unit 0, 1, 2 the z-position is -22 for 36" and skis.

For unit 6 the z-position (yellow spark) is -17.7 for 36" and 0 for skis.

wlbragg:
OK, I added spark on the front gear only because the angle left the metal strut close to the ground. The back wheels are still on the rubber after damage.

The ski sparks should be on all three skies.
This will be easier if I do it and commit myself because it requires trial and error. I'll checkout bug-610 and add the sparks.and try to get that front ski out of the ground.

onox:
OK :+1:

wlbragg:
I think our misunderstanding (if any) is I never added sparks to the ski damage (I don't) think. So I still need to add it to the spark/smoke effects files.

The zpositions in Systems/c172p-ground-effects.xml are damage contact points, not spark or smoke positions as they (spark or smoke positions) technically have an xyz and are defined in a different file. Even though the zpos should be the same or similar.

wlbragg:
I have already fixed the zpos in Systems/c172p-ground-effects.xml for the skis.
contacts needed were
-25 front
-14 left
-11 right

I'm not sure how to do this and use our existing MP aliases.
I need to add different spark positions for skis (probably no smoke, only sparks) so I need to add effect files specifically for the skies (nose, left and right) (we don't have that now). I used the same effect file for the nose gear for all def, 26, 36 and ski.
Looking at Models/Effects/damage/nose_gear.xml I was looking to see if I can use the same effects file with conditions for ski VS the other three def, 26, 36.

Any suggestions?

wlbragg:
Well that was a dumb question, I guess I can use the same MP alias because it is not being used for bush or others when we are in ski config.

onox:
@wlbragg What is your plan? :smile: Why not just make yellow sparks appear for the skis. Don't think you have to duplicate anything, maybe just add/modify some conditions.

onox:
`Models/Effects/damage/nose_gear.xml` indirectly depends on `/fdm/jsbsim/contact/unit[6]/WOW`, so it's just a matter of fixing the z-position for the skis bushkit???

wlbragg:
No, the effects are called from Models/c172p.xml lines 486-503
~~~xml
   <model>
        <path>Effects/damage/nose_gear.xml</path>
        <offsets>
            <x-m> -.85 </x-m>
            <y-m>  0.0000 </y-m>
            <z-m> -.98 </z-m>
        </offsets>
    </model>
    <model>
        <path>Effects/damage/left_gear.xml</path>
        <offsets>
            <x-m> -0.7 </x-m>
            <y-m> -1.9 </y-m>
            <z-m> -0.95 </z-m>
        </offsets>
    </model>
    <model>
        <path>Effects/damage/right_gear.xml</path>
        <offsets>
            <x-m> -0.7 </x-m>
            <y-m> 1.9 </y-m>
            <z-m> -0.95 </z-m>
        </offsets>
    </model>
~~~
Those are the xyz of
nose Effects/damage/nose_gear.xml is spark only
left Effects/damage/left_gear.xml is smoke only
right Effects/damage/right_gear.xml is smoke only
using 6, 7 and 8 in c172p-ground-effects.xml

I need 
Effects/skis/nose_gear.xml is spark only
Effects/skis/left_gear.xml is spark only
Effects/skis/right_gear.xml is spark only
using 12,13 and 14  in c172p-ground-effects.xml 
none of which isn't created yet.

Although locally I already created
Effects/skis/nose_gear.xml is spark only
Effects/skis/left_gear.xml is spark only
Effects/skis/right_gear.xml is spark only

The MP problem is still an issue as we have
~~~xml
 <fcs_function name="ground/nose-gear-ground-effect-speed-kt">
            <function>
                <product>
                    <property>/fdm/jsbsim/contact/unit[6]/WOW</property>
                    <property>/velocities/groundspeed-kt</property>
                </product>
            </function>
        </fcs_function>

        <fcs_function name="ground/left-gear-ground-effect-speed-kt">
            <function>
                <product>
                    <property>/fdm/jsbsim/contact/unit[7]/WOW</property>
                    <property>/velocities/groundspeed-kt</property>
                </product>
            </function>
        </fcs_function>

        <fcs_function name="ground/right-gear-ground-effect-speed-kt">
            <function>
                <product>
                    <property>/fdm/jsbsim/contact/unit[8]/WOW</property>
                    <property>/velocities/groundspeed-kt</property>
                </product>
            </function>
        </fcs_function>
~~~
but need the same thing for 12,13 and 14

wlbragg:
>but need the same thing for 12,13 and 14

those are aliased to "actuals" for recorder.

onox:
Why not reuse `Effects/damage/nose_gear.xml` for skis as well? Then you only need something new for the left and right ski.

onox:
Unit 12 is tail?

onox:
Unit 13 is LFfloat and 14 is RFfloat.

onox:
> as you can see they have different xyz positions of the effects.

So that's why you cannot reuse  `Effects/damage/nose_gear.xml` for the nose ski?

onox:
@wlbragg What's the main problem you have right now?

wlbragg:
>So that's why you cannot reuse Effects/damage/nose_gear.xml for the nose ski?

We could get away with that, but it wouldn't be perfect. It needs to be offset a bit.

I kind of misspoke and am confusing you by interchanging contacts with aliases.

I need to look at this more. But in the meantime.

I will use only nose as an example for simplicity. I can't wrap my head around using 6, 7 and 8
for damage conditions including drag. We use it for all (wheeled) gear configurations. But then we use 23, 24 and 25 (ski wheels) also for damage, in a round about way. I'm afraid the two will conflict, am I seeing this right?

~~~xml
 <fcs_function name="z-position-effect-nose-gear">
            <output>contact/unit[6]/z-position</output>
            <function>
                <ifthen>
                    <property>gear/unit[0]/broken</property>
                    <switch>
                        <property>bushkit</property>

                        <!-- Default/26"/36"/pontoon/amphibious/skis -->
                        <value>-10.0</value>
                        <value>-18.5</value>
                        <value>-17.7</value>
                        <value>  0.0</value>
                        <value>  0.0</value>
                        <value>-25.0</value>
                    </switch>
                    <value>0.0</value>
                </ifthen>
            </function>
        </fcs_function>
~~~
and
~~~xml
 <!-- ================================================================== -->
        <!-- Contact points skis                                                -->
        <!-- ================================================================== -->

        <switch name="z-position-nose-ski">
            <output>gear/unit[23]/z-position</output>
            <default value="0"/>

            <test logic="AND" value="-22">
                bushkit EQ 5
                gear/unit[0]/broken EQ 0
            </test>

            <test logic="AND" value="-14">
                bushkit EQ 5
                gear/unit[0]/broken EQ 1
            </test>
        </switch>
~~~

Let me push what I think I needed to do so you can look at it and maybe make some sense of it. I think all that might be missing is the MP, maybe not even that.

wlbragg:
> I'm afraid the two will conflict, am I seeing this right?

I'm wondering if the damage portion of the 23, 24 and 25 switch statement shouldn't just zero them out to get them out of the way, because 6. 7 and 8 are taking over then, right?

wlbragg:
I just pushed what I have finished, I think I still need some small spark effect adjustments.

The only think else that I think might be and issue is the MP and Flight recorder, but I'm not sure.

onox:
> I'm wondering if the damage portion of the 23, 24 and 25 switch statement shouldn't just be zeroed out to get them out of the way, because 6. 7 and 8 are taking over then, right?

When you don't have skis active? Because you added 23, 24, and 25 for the skis. Default/26"/36" use 6, 7, and 8.

onox:
I'm not really sure why you added 23, 24, 25 in the first place, because skis are kinda similar to 36" variant. Am I missing something?

wlbragg:
23, 24, 25 are the wheels on the skies. They control spin. They should be moved out of the way when the skies or gears are damaged and let the damage handling contact points take control (6, 7 and 8). Which is what we're doing for the most part, except we only position 23, 24 and 25 to -14, -14  and -16 when damaged and at the same time we use 6, 7 and 8 at -25, -14 and -11 to control damage drag and effects (sparks and smoke).

wlbragg:
To add to that, we use 6, 7 and 8 "unconditionally for damage with def, 26", 36" and ski. Except now ski needs to be defined independently of the others because it has a different damage effect, all sparks and no smoke. Where as the def, 26" and 36" damage effect uses spark on the nose and smoke on the left and right.

wlbragg:
I think I have this right, we just need to make the 23, 24 and 35 switch look like this
~~~xml
       <switch name="z-position-nose-ski">
            <output>gear/unit[23]/z-position</output>
            <default value="0"/>

            <test logic="AND" value="-22">
                bushkit EQ 5
                gear/unit[0]/broken EQ 0
            </test>

            <test logic="AND" value="0">
                bushkit EQ 5
                gear/unit[0]/broken EQ 1
            </test>
        </switch>

        <switch name="z-position-left-ski">
            <output>gear/unit[24]/z-position</output>
            <default value="0"/>

            <test logic="AND" value="-22">
                bushkit EQ 5
                gear/unit[1]/broken EQ 0
            </test>

            <test logic="AND" value="0">
                bushkit EQ 5
                gear/unit[1]/broken EQ 1
            </test>
        </switch>

        <switch name="z-position-right-ski">
            <output>gear/unit[25]/z-position</output>
            <default value="0"/>

            <test logic="AND" value="-22">
                bushkit EQ 5
                gear/unit[2]/broken EQ 0
            </test>

            <test logic="AND" value="0">
                bushkit EQ 5
                gear/unit[2]/broken EQ 1
            </test>
        </switch>
~~~

wlbragg:
And then all that is left is to make sure the MP and recorder is working.

wlbragg:
@onox, I just change the above locally and it was causing a major problem I was having adjusting the 6, 7 and 8 "damage" contact points. The 23, 24 and 25 "broken" contact points were getting in the way of the ones we use for damage (6, 7 and 8), they have to be at 0 when damaged is true.
Now I have to re-adjust 6, 7 and 8 and that part is finished (other than checking MP and recorder).

onox:
@wlbragg Regarding https://github.com/Juanvvc/c172p-detailed/pull/615#issuecomment-164971349: the default value is 0 in each `<switch>`, so the second `<test>` in each switch is a bit useless :stuck_out_tongue_winking_eye:

onox:
Although you can also leave the tests in the code. Just mention in a comment then (just above each test case) that the "contact point has to be out of the way" or something.

wlbragg:
It couldn't figure out why I cleaned up the tabs

onox:
You merged `bug-610` into `bug-610`? :open_mouth:

onox:
@wlbragg I'm seeing weird things happening again: you add a commit on top of `bug-610`. Then you create a branch from the previous commit, you add a similar commit again, and then you merge the two branches:

bug-610-branches

I'm not sure how you manage to do this :stuck_out_tongue_closed_eyes: Just checkout a branch and then add commits to it. Your GUI program shows you in which branch you are, right?

onox:
@wlbragg I seems all the merge did was:

```diff
diff --git a/Models/c172p.xml b/Models/c172p.xml
index 3882df1..6749263 100644
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -559,9 +559,9 @@
     <model>
         <path>Effects/skis/left_gear.xml</path>
         <offsets>
-            <x-m> 1.012 </x-m>
-            <y-m> -0.951 </y-m>
-            <z-m> -1.291 </z-m>
+            <x-m> 2.85 </x-m>
+            <y-m> -0.51 </y-m>
+            <z-m> -2.91 </z-m>
         </offsets>
     </model>
     <model>
```

So I'm going to reset to your previous commit.

Make sure you're in the `bug-610` branch. Then do:

`git reset --hard 749f8875afb13b3e678bc1ca087373c6459af599`

This will reset your tree back to your original `Add ski damage sparks, adjust ski damage contact points.` commit.

wlbragg:
That is bizarre!
I'm sure I did everything in the normal order here. After I pushed the first time which added the new files to skis, I fixed the tabs in the existing damage effect files, along with that one point change that you reversed (that wasn't a conflict). When I went to push it, it reported a bunch of conflicts, they were all the tabs I fixed for some reason it couldn't reconcile that I fixed a bunch of tabs. I thought it was odd that I got those conflicts. Every one of them were tab spacing changes.

So why did you reverse the merge? Now the tab spacing is screwed up again :cry: and that point change was valid?

I'll push a fix and next time I get a conflict when I shouldn't be getting a conflict I'll try to reconcile locally.I just figured it couldn't evaluate the tab fix, although I don't understand why that would be.
.

wlbragg:
>So why did you reverse the merge? Now the tab spacing is screwed up again :cry: and that point change was valid?

That isn't true, I just checked it and the corrected tabs are there. So I don't have a clue what just happened. But I understand your fix and will commit the last change again.

onox:
Just make sure that you're adding a commit on top of 749f8875afb13b3e678bc1ca087373c6459af599.

onox:
If you decide to fix the formatting of a whole file, try to do it in a separate commit. That way people can safely assume that it's just formatting and not also modifications to logic (like adding/modifying a condition/expression)

wlbragg:
:+1:

onox:
@wlbragg Is this PR complete? I just upgraded simgear and flightgear and now particlesystem is dead and only half of the MP properties are transmitted, even with master branch. Problem on my end, so if you want to merge the PR, you can do that.

wlbragg:
I just verified that the ski damage and the sparks are visible over MP :smile: 

But I crashed the plane pretty hard to test it and the wings collapsed, that wasn't visible over MP anymore. :cry:

wlbragg:
After more testing, what is happening is that only one MP aircraft is seeing the others wing damage.

PR #617, opened 2015-12-12
Bug 616: Give flaps control a vertical drag direction

### LINKED ISSUE #616: Drag direction of flaps should be vertical instead of horizontal
Drag direction of flaps should be vertical instead of horizontal.

onox:
Fixes #616


gilbertohasnofb:
@onox this works much better than before, but clicking and dragging requires some good precision now as few pixels up or down move the flap. Is there a way to adjust this? Besides that, everything looks fine to me.

onox:
You mean you want dragging to be less sensitive?

gilbertohasnofb:
Yes, that's what I meant, I will check it again now.

gilbertohasnofb:
:+1:

PR #618, opened 2015-12-12
Bug 614: Color problems in the cockpit gauges & 428: Strange bug with shadows on gauges

wlbragg:
I think this may just be the fix for both of these. The interior looked really sharp I think.


gilbertohasnofb:
Looks good!

PR #619, opened 2015-12-12
Bug 604: UV map and texturize bottom panel behind pedals

wlbragg:
I think we're done here, unless @gilbertohasnofb can find some more things to fix. Pretty soon we may not need @thevirtualfer's new model. :laughing:


gilbertohasnofb:
:+1:

gilbertohasnofb:
> Pretty soon we may not need @thevirtualfer's new model.

Yeah, as much as I understand his internet situation his last commit was in March this year, and the project just can't really stop for 9 months because of him (also, it has been his decision to develop lone-wolf style instead of committing often and regularly, which we could then have helped with). 
Also, as I have been pointing out so many times, I think we really must continue our work regardless of the new model and I think we are proving this decision to be correct, the plane is already looking so much better than a couple of months before! In the end, it will be up to him to synchronize his model to the repository when he finally create a PR for it.

So you say that half-joking, but I think we are getting close to make that true :wink:

onox:
Didn't he do much work on the shape of the fuselage? :confused: 

I was hoping he would fix the engine part of the fuselage, the shape around the windows, and the glass (the glass is really botched up currently :cry:)

@gilbertohasnofb Are there issues for the oil cap and cabin air side door (on the right side of the fuselage)?

gilbertohasnofb:
> Didn't he do much work on the shape of the fuselage? 

He did, onox, but while he doesn't upload it what should we do? We can't simply stop the production line for an year until he finally uploads it. Given the situation, I think it's up to him to catch up with the modifications we are doing rather than we freezing everything because of his model that so far we haven't even seen a single commit.

> I was hoping he would fix the engine part of the fuselage, the height of the cockpit, the shape around the windows, and the glass (the glass is really botched up currently :cry:)

You have no idea how much I want that fixed as well... that fuselage border around the cockpit is so ugly and unrealistic... 

> Do we have issues for the oil cap and cabin air side door (on the right side of the fuselage)?

No we don't. But he was told about them for the new model. Anyway, I will create a new issue for those right now.

So yeah, my philosophy right now is to continue to improve the model the best I can and if we finally get an updated model, then we try to merge things the best we can (if it's worth it). Or do you see any other reasonable approach to this problem?

gilbertohasnofb:
See: https://github.com/Juanvvc/c172p-detailed/issues/621 and https://github.com/Juanvvc/c172p-detailed/issues/622

wlbragg:
Yes, that was just a joke.
@gilbertohasnofb, your right, we need to continue forward, people's priorities have a habit of changing over time and there are no guarantees, including any of us, get it while the getting is good.:wink:
I'm sure if we ever get to see the new model work well have plenty of upgrades to incorporate. I'm anxious to see some of it put into the aircraft.

gilbertohasnofb:
Agreed 100%.

onox:
I guess we just have to hope that @gilbertohasnofb gets an e-mail from @thevirtualfer with the new 3D model soon :smile: :confused: :cry:

gilbertohasnofb:
Yep... :sob:

PR #625, opened 2015-12-14
Bug 575

### LINKED ISSUE #575: Full power makes tail tie-down raise
Full power makes tail tie-down raise. Either the aircraft shouldn't move at all or the tail tie-down should be made flexible somehow.

onox:
Fixes #575 

Implements dynamic tiedowns controlled via JSBSim external forces. Under full engine power you'll see the tiedowns are a little bit flexible.

You can clearly see external forces are used if you release the tail tiedown or use just one (but not both) of the wing tiedowns. Maximum forces limited to 1000 (N?) for now. So if you crash your plane and find yourself somehow at a large distance from where you initially tied down your plane (hint: Ctrl+u) then the forces may not be sufficient to overcome friction.

Just using proportional gain (P) controllers for now. Add I term if you really want zero error. Add D for damping. In normal situations P seems to be sufficient.


gilbertohasnofb:
Looks fantastic, guys!

Issue #628, opened 2015-12-15
Problems with the start-up tutorial


gilbertohasnofb:
Problems with the start-up tutorial:

- aircraft starts with tie-downs, so I guess we must force tie-downs and chocks off as well as pitot tube cover, as well as check for oil and fuel contamination (that is, the aircraft should be in the state after pre-flight inspection). This may affect other tutorials as well but I haven't checked them all.
- I would also force the switches to the correct pre-start position: avionics and master off, all lights off, circuit breakers all on. 
- the tutorial asks me to prime the engine quite a lot, around 6 times. I thought that anything greater or equal to 3 would be fine.
- this tutorial asks me to set the throttle to "around" 10%, but several values (including 10% exactly) aren't accepted. I think anything between 10-20% should be fine right?

gilbertohasnofb:
@dg-505 would you be interested in this issue given that you have already worked with tutorials before?

dg-505:
Yes, I would be.
I also noticed that there are several problems with the other tutorials. E.g. "Landing" should start with engines running, but at least for me, it doesn't. It seems that the startup procedure of the c172p has become that complex that initializing a tutorial with engines running may be very difficult, if not impossible.

- In the first step, the tiedowns, chocks and pitot cover get removed (line 93-112). If they are nevertheless not removed, I don't know the reason. In the first step the switch positions can also be set. The breaker positions are checked in a separate step (line 193-363) so IMHO they don't need to be set at the beginning.

- The "priming"-step is from line 365-404. According to this it should move to the next step when ```/controls/engines/engine[0]/primer``` is **3 or more** and ```/controls/engines/engine[0]/primer-lever``` is **0** i.e. the Primer lever is in. To prime the engine 3 times you have to actuate the primer 6 times: 3 times pull out, 3 times push in.

- Ah, I discovered the mistake in the "throttle-setting"-step. I will correct this right now.

dg-505:
See #629

gilbertohasnofb:
> Yes, I would be.

Great! :smile: 

> In the first step, the tiedowns, chocks and pitot cover get removed (line 93-112). If they are nevertheless not removed, I don't know the reason.

Strange. I loaded the tutorial when I was tied down and it didn't remove the tie-downs by itself.

dg-505:
As #629 is now merged, can this now be closed?

gilbertohasnofb:
Yes, absolutely! I forgot to do so. Thanks again!

PR #629, opened 2015-12-15
Bug 628: Startup tutorial

dg-505:
See #628

Fix throttle setting, add lights off and master/alternator off in initial step.

### DIFF
--- a/Tutorials/preflight.xml
+++ b/Tutorials/preflight.xml
@@ -149,23 +149,23 @@ This tutorial will guide you through the Preflight inspection
         </set>
         <set>
             <property>/sim/model/c172p/securing/chock</property>
-            <value>1</value>
+            <value>true</value>
         </set>
         <set>
             <property>/sim/model/c172p/securing/tiedownL-visible</property>
-            <value>1</value>
+            <value>true</value>
         </set>
         <set>
             <property>/sim/model/c172p/securing/tiedownR-visible</property>
-            <value>1</value>
+            <value>true</value>
         </set>
         <set>
             <property>/sim/model/c172p/securing/tiedownT-visible</property>
-            <value>1</value>
+            <value>true</value>
         </set>
         <set>
             <property>/sim/model/c172p/securing/pitot-cover-visible</property>
-            <value>1</value>
+            <value>true</value>
         </set>
         <set>
             <property>sim/model/c172p/engine_flag_0</property>
--- a/Tutorials/preflight.xml
+++ b/Tutorials/preflight.xml
@@ -101,7 +101,7 @@ This tutorial will guide you through the Preflight inspection
         </set>
         <set>
             <property>/controls/lighting/instruments-norm</property>
-            <value>1</value>
+            <value>0</value>
         </set>
         <set>
             <property>/controls/engines/engine/throttle</property>
--- a/Tutorials/startup.xml
+++ b/Tutorials/startup.xml
@@ -32,7 +32,7 @@ This tutorial will take you through starting the Cessna 172p.
     <step>
         <set>
             <property>/sim/model/c172p/brake-parking</property>
-            <value>1</value>
+            <value>true</value>
         </set>
         <set>
             <property>/controls/flight/flaps</property>
@@ -80,10 +80,11 @@ This tutorial will take you through starting the Cessna 172p.
         </set>
         <set>
             <property>/controls/switches/master-bat</property>
-            <value>0</value>
+            <value>false</value>
         </set>
         <set>
             <property>/controls/switches/master-alt</property>
+            <value>false</value>
         </set>
         <set>
             <property>/controls/lighting/taxi-light</property>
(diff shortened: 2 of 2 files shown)


gilbertohasnofb:
Should we investigate the tie-downs problem a bit further before merging this, since it is also described in https://github.com/Juanvvc/c172p-detailed/issues/628?

gilbertohasnofb:
(clearing milestone as this is a PR not an issue, sorry)

onox:
The tutorial sets `/sim/model/c172p/tiedownL`, etc. to `0` instead of setting `/sim/model/c172p/securing/tiedownL-visible`, etc. to `false`.

onox:
`/sim/model/c172p/chock` has been moved to `/sim/model/c172p/securing/chock`.

`/sim/model/c172p/pitot-cover` has been moved to `/sim/model/c172p/securing/pitot-cover-visible`.

dg-505:
To my kowledge it doesn't matter if a value is set to 0/false or 1/true. In the end it does the same thing.

>  /sim/model/c172p/chock has been moved to /sim/model/c172p/securing/chock.
/sim/model/c172p/pitot-cover has been moved to /sim/model/c172p/securing/pitot-cover-visible.

Ah, that's the point.

gilbertohasnofb:
@dg-505 Some problems:

- at line 86 shouldn't you add `<value>0</value>`? My alternator isn't being set to off.
- for some reason, the pitot tube cover, tiedowns and chcoks are still not being removed, but I can't spot the problem with your code... did you test it? Is it working for you?
- strobe, beacon and nav lights, as well as pitot heat, are not being set to off to me, and I also can't find a problem in the code

Can you try to reproduce it?

gilbertohasnofb:
@dg-505 actually, if you notice, the problem happens to all properties that come after the alternator in that code; so I figured that by adding `<value>false</value>` below line 86 the problem would be solved, and voila! It is :)

Some other problems though:

- circuit breakers are not being set by the tutorial
- nor is the carburetor heat
- the primer lever starts at the position we last left it, though I think it should start in the position IN
- instrument light is also not being set to 0%

onox:
> To my kowledge it doesn't matter if a value is set to 0/false or 1/true. In the end it does the same thing.

It depends on how weakly/strongly typed the language is. For the property tree system it doesn't matter much whether you use 0/false or 1/true, so yes it does the same thing, but I still find it ugly if numbers are being used for a property that has a boolean type. Also, if you use true/false you clearly see (when you read the code) there are only two valid values: true or false. If you use numbers, it could be anything: 0, -1, 1, -2, 2, etc.

onox:
@dg-505 @gilbertohasnofb Do the problems in https://github.com/Juanvvc/c172p-detailed/pull/629#issuecomment-165099315 and https://github.com/Juanvvc/c172p-detailed/pull/629#issuecomment-165102115 still need to be fixed?

gilbertohasnofb:
@onox Yes, they do still need to be fixed. I don't think @dg-505 has made any more commits since I spotted that little problem with the code (so I don't think he committed the fix yet), and that list of other things that need to be set is still not implemented AFAIK.

gilbertohasnofb:
In the next week or so I will be travelling and I probably won't be able to collaborate much (if at all), but once I have time I will take a look at this and try to fix it myself if @dg-505 doesn't do it first. Happy new year to you all!! :beers:

dg-505:
Fixed it. Sorry guys for the delay :smiley: 
Also happy new year! :tada:

gilbertohasnofb:
@dg-505 :+1: Thanks, but I think those other things I pointed out at https://github.com/Juanvvc/c172p-detailed/pull/629#issuecomment-165102115 still need to be taken care :smile:

dg-505:
You check the circuit breakers in the tutorial step as of line 225 and the Carburetor Heat as of line 438 so I consider it unnecessary to set them when initializing the tutorial because later you make sure they're all in and the Carb heat is cold before starting the engine.

gilbertohasnofb:
> I consider it unnecessary to set them when initializing the tutorial because later you make sure they're all in and the Carb heat is cold before starting the engine.

You are completely right, it actually simulates then someone forgetting these things in the wrong position. So now that you corrected the primer lever and instrument lights, I think this is ready to be merged. I will test it right now.

gilbertohasnofb:
@dg-505 Okay, I tested and everything seems good, but I still have a bit of a problem with the last item from https://github.com/Juanvvc/c172p-detailed/issues/628 about the throttle values (sorry to be picky). The thing is the tutorial seems to want the throttle to be between 10 and 12%, and not between 10-15% as you commented in the code. Anyway, I think even that is too few and that it should be more like between 10-2x % (maybe 22-23 as the upper bound). If you think about it, our autostart function itself sets the throttle to 20% and the engine does turn on normally, so it does make sense to allow it in the tutorial as well. What do you think?

dg-505:
> sorry to be picky

No problem, it's good to have high quality standards :)

The following commit will allos the throttle to be between 15 and 20 %

dg-505:
>12 and 20 %

Oops, this is just a typo mistake in the commit message :laughing:

gilbertohasnofb:
Looks good, thanks a lot! :+1:

PR #630, opened 2015-12-15
Bug 626: Adding thermometer to the panel

gilbertohasnofb:
Adds a thermometer to the panel. Closes https://github.com/Juanvvc/c172p-detailed/issues/626


gilbertohasnofb:
:warning: :warning: :warning: still unfinished! The thermometer display isn't working yet... I am trying to adapt the code from the Comanche, which has a `Panel` subfolder and looks quite unusual to me :cry: 

Also, we currently don't have the cabin temperature in Fahrenheit so we need to create a `<filter>` for that, right?

onox:
The files are currently in `Models/Interior/Panel/Instruments/thermometer/`. What's the ideal location you actually want?

gilbertohasnofb:
I guess that location is fine, isn't it? If there is no checkbox, then it's a panel instrument methinks :smile:

onox:
You said you didn't like the `Panel` folder?

gilbertohasnofb:
No, I said the original Comanche project had a `Panel` subfolder on its root, and that's where the instrument models are loaded. This seem typical of older aircraft, and I don't really understand those xml files very well (for instance, they have tons of `<instruments>` label which we never use). So what I committed here is my first attempt of adapting what they had in that project into ours

onox:
Wait... this is looking like the old deprecated 2D panel code.

onox:
All these `<layer>` and `<chunk>` stuff is something I recognize from the old 2D panels. It's not something you normally find in regular model XML files.

Nasal/c172p.nas
```diff
@@ -194,6 +194,22 @@ var switches_save_state = func {
 };
 
 ##########################################
+# Thermometer
+##########################################
+
+var thermometer_switch = func {
+    var val = getprop("controls/switches/thermometer-switch");
+    var test = 1 + val;
```
onox: You should be able to replace these two lines (202 and 203) with:

```javascript
var test = math.mod(1 + val, 3);
```

Even better is to directly put the result in line 204 (inline it)

onox:
@gilbertohasnofb I suggest you look at `Models/Interior/Panel/Instruments/hobbs/`. And replicate its behavior.

gilbertohasnofb:
Ok, I will give it a shot. Thanks!

gilbertohasnofb:
So, the Nasal code is corrected and pushed. As for the Fahrenheit cabin temperature, where should I create the filter? I was thinking about creating a `<pure_gain>` to `c172p-heat.xml` but I don't know how to add an offset to it, would you have some recommendation @onox? 

Finally, about the Hobbs, the numbers are displayed on individual 3D objects, and this isn't the case of this thermometer. So to use something similar as the Hobbs, it would involve some work on Blender I suppose.

gilbertohasnofb:
If anyone is interested, here is the original code which I tried to adapt: https://sourceforge.net/p/flightgear/fgaddon/HEAD/tree/branches/release-3.6.0/Aircraft/pa24-250/Models/oat/

onox:
I'll write a function for you, hang on a minute.

onox:
```xml
<fcs_function name="heat/cabin-air-temp-degf">
    <function>
        <sum>
            <product>
                <property>heat/cabin-air-temp-degc</property>
                <value>1.8</value>
            </product>
            <value>32</value>
        </sum>
    </function>
</fcs_function>
```

onox:
Add it below the integrator at the end of the channel.

gilbertohasnofb:
Perfect, thanks a lot, I added it just below the `<integrator name="heat/surface-temp-degc">` block. So now I

gilbertohasnofb:
So now we "just" need to fix the display. If we go the Hobbs way, we need someone with Blender expertise to create the individual digits, and then I think I can adapt the Hobbs code myself if necessary.

onox:
Isn't it just a matter of reusing that `.png` file? You just need to alter the digits to look like LCD digits.

gilbertohasnofb:
> Isn't it just a matter of reusing that .png file? You just need to alter the digits to look like LCD digits.

No, you need to create individual "drums" for each digit on Blender. Then we must create some code which will extract each digit of the temperature for both Celsius and Fahrenheit. That's at least how the Hobbs work. 

As for the original code from the Comanche, it's strange that it's full of those 2D panel blocks since the OAT is displayed in the 3D cockpit. I was really hoping for something easier.... :cry: 

And also, once this work is done in Blender, we must remember to add the flashlight effect to it.

onox:
It uses `<panel>` it seems in http://sourceforge.net/p/flightgear/fgaddon/HEAD/tree/branches/release-3.6.0/Aircraft/pa24-250/Models/oat/oat_3d.xml

wlbragg:
I was checking this PR to determine the status and realize it is not ready.

But while I was doing that I noticed something I am concerned about. I was trying to click on the thermometer button to see if it worked yet and when I tried the "middle click" it turned the yoke to the left on every "middle click", no matter where I clicked on the screen.

By the way, I don't think it is related to this work at all. Any idea what that is all about?

gilbertohasnofb:
> Any idea what that is all about?

Not at all! I will try to reproduce this later, but today I can't unfortunately.

gilbertohasnofb:
Also, does anyone know any instrument in other aeroplanes which have a digital LCD display but are not created using those old `<panel>` blocks?

wkitty42:
On 12/18/2015 06:27 PM, Gilberto Agostinho wrote:
> Also, does anyone know any instrument in other aeroplanes which have a
> digital LCD display but are not created using those old |<panel>| blocks?

i've flown maybe a dozen of all the craft in FGAddon... the PA24-250 i only flew 
because i was trying to follow that tutorial in that PDF file... it made more 
sense to use that craft than any other due to trying to convert the instructions 
to other craft so i took the easy way out... that one is the only one that i'm 
aware of having a thermometer like that in it... i had thought that it would be 
an easy copy and add to the c172p which is why i pointed to it...

onox:
I'll close this PR for now, otherwise it shows up in the list all the time.

PR #635, opened 2015-12-30
Point controls.stepMagnetos to new property

NasalMusician:
This allows joysticks (etc.) to control it more easily.

(First time collaborating here, let me know if I need to do anything differently!)

### DIFF
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -232,6 +232,12 @@ controls.mixtureAxis = func {
     setprop("/controls/engines/current-engine/mixture", new_value);
 };
 
+controls.stepMagnetos = func {
+    var old_value = getprop("/controls/switches/magnetos");
+    var new_value = std.max(0, std.min(old_value + arg[0], 3));
+    setprop("/controls/switches/magnetos", new_value);
+};
+
 # key 's' calls to this function when it is pressed DOWN even if I overwrite the binding in the -set.xml file!
 # fun fact: the key UP event can be overwriten!
 controls.startEngine = func(v = 1) {


onox:
Thanks :smile:

Juanvvc:
This function has nothing specific to the c172p and many other aircraft may benefit from having it included in the base package. In addition, it changes the properties under "/control/switches" instead of under "/controls/engines/engine", like the original function does ( http://sourceforge.net/p/flightgear/fgdata/ci/next/tree/Nasal/controls.nas ) To me, this is pretty similar to that time when we changed the way mixture was managed, and some specialized input devices like Saitek's stopped working.

Let's keep an eye open in case this change has some side effects and bug reports. In any case, I strongly believe changes like these must be submitted to FGDATA for their consideration :) The fact that it is easier for anyone to submit bugfixes to this repository instead of FGDATA shouldn't mean the c172p has to fix the base package!

(We are already fixing the base package in some places, like walker and human models, but it was intended to be a temporal fix and I don't think this is right in the long term)

In any case, NasalMusician, please, don't let my opinion discourage you to collaborate in the project! As you can see, each of us have our own opinions and your proposal is already accepted in the project by an admin.

NasalMusician:
Thanks for merging!

@Juanvvc: Wouldn't merging it into FGData break a lot of aircraft (read: most aircraft) which use /controls/engines/engine[,]/magnetos?

I personally dislike the use of non-standard properties for controls; to me it makes most sense to define a single set of properties that most aircraft can use and share. But I see what issues you are working around with having alternative engines, and I don't know if there's a way around it that doesn't come with more assumptions. (Hm... would (optionally) copying throttle 0 to throttle 1 work?)

Anyhow, this was just a small patch to help my joystick work with this aircraft – I still have throttle and mixture to fix in my joystick now ;).

PR #639, opened 2016-01-02
Use the "y"-key to show/hide yokes

dg-505:
IMHO this is the easiest way to show/hide the yoke

Signed-off-by: Jonathan Schellhase


Juanvvc:
It would be nice if this pull request includes a modified help menu to document the use of this new key: https://github.com/Juanvvc/c172p-detailed/blob/master/c172-help.xml

dg-505:
Done :+1: Thanks, Juan

Issue #638, opened 2016-01-02
Pointmass weights question


wlbragg:
Is this /fdm/jsbsim/inertia/pointmass-weight-lbs[0] or [1]?
~~~xml
       <pointmass name="Pilot">
            <weight unit="LBS"> 180 </weight>
            <location name="POINTMASS" unit="IN">
                <x> 36 </x>
                <y> -14 </y>
                <z> 24 </z>
            </location>
        </pointmass>
~~~
I'm not sure the question matters as to there being an issue with the pointmass's.
If Pilot is [0] then
~~~xml 
<pointmass name="Left bush wheel 26 inches">
            <weight unit="LBS"> 0 </weight>
            <location name="POINTMASS" unit="IN">
                <x> 58.2 </x>
                <y>  -43 </y>
                <z> -15.5 </z>
            </location>
        </pointmass>
~~~
is correct at [5] as in
~~~xml
switch name="extra-weight-left-26-in">
            <default value="0"/>
            <test logic="AND" value="10">
                bushkit EQ 1
            </test>
            <output>/fdm/jsbsim/inertia/pointmass-weight-lbs[5]</output>
        </switch>
~~~
but then
~~~xml
       <pointmass name="Left ski">
            <weight unit="LBS"> 0 </weight>
            <location name="POINTMASS" unit="IN">
                <x> 58.2 </x>
                <y>  -43 </y>
                <z> -15.5 </z>
            </location>
        </pointmass>
        <pointmass name="Right ski">
            <weight unit="LBS"> 0 </weight>
            <location name="POINTMASS" unit="IN">
                <x> 58.2 </x>
                <y>  43 </y>
                <z> -15.5 </z>
            </location>
        </pointmass>
~~~
are wrong.

@Onox, I think this is really screwed up up. Please take a look and let me know what you think

wlbragg:
This is on the assumption that the [?] position is defined at loading position in the FDM, or is it defined somewhere else I am unaware of. We use [13] for engine when I calculate it at [15].

dany93:
If I remember well... Here is my feeling:

In c172p.xml (FDM) I think I had originally written (to help at reading)
        `
        `

but someone (you ?), when adding the skis, changed it for
        `
        `

My hypothesis is (I'm almost sure) that the number is the position in the FDM.
The issue is that you have inserted the skis "bushkit" in [13] and [14] position in the FDM, thus the engine becomes [15].

Please think well, but this should be the issue.

[...]

PR #640, opened 2016-01-03
Correct the engine pointmass positioning in the FDM.

wlbragg:
I think this is all that is required.

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -173,6 +173,18 @@
                 <z> -25 </z>
             </location>
         </pointmass>
+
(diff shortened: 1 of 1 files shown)


onox:
Oh, now I see you have already created a PR. Your PR should fix the issue (#638), but I think the skis pointmasses should remain together with the other (previous) pointmasses as is the case in your old code. You can just as well just update the indices in `Systems/bushkit.xml`.

So I did just that in PR #641. Your choice which one you want to merge (this PR or #641), just close the other one then :wink:

PR #642, opened 2016-01-03
Round front ski edges, correct right ski axle position.

wlbragg:
This is something I am doing with the Cub and thought it looks better than squared off. There was also an error in the right ski axle position x, y and z.


gilbertohasnofb:
I will check this now.

gilbertohasnofb:
:+1:

PR #645, opened 2016-01-07
FDM: Set fuel type to AVGAS and reduce max cap to 129.43 lbs

### LINKED ISSUE #644: FDM uses wrong type of fuel  --> wrong weight of fuel
According http://wiki.flightgear.org/JSBSim_Fuel_System    JSBSim calculates 6.6 lbs/ gal fuel weight as standard. This is the density of Jet Fuel. 
But the C172 uses AvGas. So it should be 6.02lbs/ gal. So fuel weight is not correct!

How to Fix:
Standard tank is 43 Gallons for the Cessna 172P , so add <type>  AVGAS</type> btween <tank></tank> tags and change each capacity to 129.43lbs.

onox:
Fixes #644

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -630,8 +630,9 @@
                 <y> -112 </y>
                 <z> 59.4 </z>
             </location>
-            <capacity unit="LBS"> 185 </capacity>
-            <contents unit="LBS"> 100 </contents>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
:+1:

PR #651, opened 2016-01-08
Bug 607

wlbragg:
I used an if statement of unit[11] or unit[12] and AGL (meters) < 5 in the physics loop. Seems to work OK.


wlbragg:
Dang it, I started with a clean master, why does it always grab the bug work I have done on other branches? :angry:

gilbertohasnofb:
Wait, why a new PR? Isn't this related to https://github.com/Juanvvc/c172p-detailed/pull/647? If not, then why the commit `parks not influenced by available light.` (did you create the new branch while on `master` or while on the sparks branch?)?

wlbragg:
On master, I don't know why this does this to me. Let me delete this and do it again.

gilbertohasnofb:
Are you sure you started with a clean master? I always do 

```
git checkout master
git pull
git status
```

and only then `git checkout -b <new-branch>`

gilbertohasnofb:
I noticed that when I create a new branch and I don't push it yet using `git push -u origin <new-branch-name>` and then I switch to master, the modifications I did get a bit messed up for some reason. This is the only thing with git that makes me confused, why creating a local branch and not pushing makes `git status` go crazy? Could be related to your problem, or not.

onox:
> Dang it, I started with a clean master, why does it always grab the bug work I have done on other branches?

If you're working on branch A (according to `git branch` or `git status`) and modify some files (according to `git status`). Then after you switch to branch B with `git checkout B`, git will try to 'replay' the modifications you did to the files in branch B. So when you do `git status` again, its output should be the same.

Don't think then when you modify some files while on branch A, that when you switch to branch B, that you're `git status` is going to be clean. All the modifications (`git diff`) are per repository, not per branch.

> I noticed that when I create a new branch and I don't push it yet using git push -u origin <new-branch-name> and then I switch to master, the modifications I did get a bit messed up for some reason. This is the only thing with git that makes me confused, why creating a local branch and not pushing makes git status go crazy? Could be related to your problem, or not.

When you have committed your changes to a new branch called A, then `git status` should become clean and you can then safely switch to another branch. If you forget to commit, then git will try to 'replay' those modifications to branch you're switching too. This isn't guaranteed to always work (if the two branches are very much different), but in that case git should tell you that you cannot switch.

So if you want to work on two branches are the same time, you either have to commit your changes before you switch to another branch, or stash it with `git stash`.

gilbertohasnofb:
> So if you want to work on two branches are the same time, you either have to commit your changes before you switch to another branch, or stash it with git stash.

Okay, so that's my problem indeed! I just find it a bit annoying because if I want to experiment in a local branch (and therefore not commit anything), I also would like to be able to check PRs and things like this. Anyway, at least now I understand why this happens, this was so puzzling for such a long time to me...

onox:
In that case you can either stash all your stuff with `git stash` (and later unstash it with `git stash pop`).

Or you commit it (`git commit -m "Work in progress"`) in your local branch A. After that you test a PR in another branch and then after that you checkout A again. After that you can "uncommit" by doing `git reset --soft HEAD~1`. What it does is leave your working tree untouched (assuming you made sure that `git status` is clean after switching back to A), but pop the last commit from your tree. The result is that everything in that last commit is visible under "Changes to be committed" in `git status`.

You should try this in a test repo on your computer.

onox:
Or instead of "uncommitting" it with `git reset --soft HEAD~1`, you can "amend" it by making more changes and then do `git commit --amend`.

gilbertohasnofb:
:+1:

onox:
So the difference between `--soft` and `--hard` for `git reset` is that `--soft` removes your commit from the tree (the chain of commits), but leaves your working copy (the files on your file system) untouched. `--hard` will also "unmodify" the files on the file system.

wlbragg:
:+1: 

Thanks!

PR #652, opened 2016-01-08
Bug 607 - Kill engine when upside down.

### LINKED ISSUE #607: Stop propeller if aircraft is upside down
Stop propeller if aircraft is upside down. Currently it keeps spinning.

fgfs-screen-402

Can probably be fixed by calling `killengine()` in `upsidedown` function in `Nasal/physics.nas`.

wlbragg:
Fixes #607 

Uses unit 11 or 12 WOW and AGL < 5 in physics loop.

### DIFF
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -60,6 +60,8 @@ var physics_loop = func
     }
     if (getprop("/fdm/jsbsim/bushkit") == 3 or getprop("/fdm/jsbsim/bushkit") == 4)
         poll_surface();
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
What `/fdm/jsbsim/contact/unit[11]/WOW` and `/fdm/jsbsim/contact/unit[12]/WOW` do exactly?

onox:
They are contact points at the top of the cockpit and fuselage.

@wlbragg Wouldn't it be better to use 9 and 10? Is the AGL < 5 check necessary?

wlbragg:
>Is the AGL < 5 check necessary

I would only because someone may want to fly upside down at a higher altitude? Although, the fuel system does eventually kill the engine when upside down long enough anyway..

>Wouldn't it be better to use 9 and 10

Yes, I think your right. That is the nose (top) and cockpit (top).

PR #655, opened 2016-01-09
Move .blend files to dev/blender/

### LINKED ISSUE #654: Move .blend files to dev/blender/
Move .blend files to dev/blender/

onox:
Fixes #654


gilbertohasnofb:
@onox there are still some `.blend` files outside the new `/dev` folder, is this on purpose? They are:

```
./Models/Exterior/chock/chock.blend
./Models/Interior/Panel/Buttons/breakers/breakers.blend
./Models/Interior/Panel/Instruments/hi/hi.blend
./Models/Interior/Panel/Instruments/hobbs/hobbs.blend
./Models/Interior/Panel/Instruments/light/light.blend
./Models/Interior/Panel/Instruments/oat/oat.blend
./Models/Interior/Panel/Instruments/tc/tc.blend
./Models/Interior/Panel/Instruments/AI/AI.blend
```

onox:
Shall we move everything in `Models/Interior/Panel/` to `dev/blender/instruments/` and move `chock.blend` to `dev/blender/c172p-yellow-chocks.blend`?

gilbertohasnofb:
I would say so, keeps everything more consistent IMO.

onox:
I have moved the last .blend files.

gilbertohasnofb:
:+1:

onox:
@gilbertohasnofb @wlbragg In case Blender cannot find the png files: go to File > External Data > Report Missing Files. Search for the image in the object tree view in Blender. Choose one of the found objects and fix its image source. Make sure that the path is relative. Repeat until Blender cannot find any missing files.

onox:
I guess it's easier to just select 1 object, switch to a different image in the "Image" tab/section, and then for reach image you select the correct image source.

Edit: I still cannot get any textures to be rendered here :frowning:

wlbragg:
Dang, I guess I should have "packed" the images in blender, I think that is supposed to keep things like this from happening?.

I kind of know what your referring to because I have been having trouble with loosing textures in the UV mapping, then you get an error when you try to export the .ac. I found pretty much the same thing, you have to re-establish the connection to them one-by-one. 

>Edit: I still cannot get any textures to be rendered here, even though Blender can find all the images 

What specifically are you referring to here?

onox:
> Dang, I guess I should have "packed" the images in blender, I think that is supposed to keep things like this from happening?.

I'm not sure if that is what we want, because it would make it more difficult to update the images. AFAIK you cannot simply replace a .png file, because you also need to pack the image in Blender. So just setting relative paths is probably the easiest solution.

> What specifically are you referring to here?

What I mean is that I cannot get Blender to display the texture on the fuselage. The fuselage remains grey, even after fixing the paths to the images. Maybe you have more luck.

wlbragg:
>remains grey

If it can't find the textures it should be purple.

I was able to get it to work by selecting the object, putting the object in edit mode, then use the texture tab to select the correct texture.

onox:
:+1:

PR #658, opened 2016-01-12
Bug 613: Mouse wheel direction of overhead air vents is inverted

### LINKED ISSUE #613: Mouse wheel direction of overhead air vents is inverted
On the control panel, scrolling the mouse wheel forward decrements the value of what we altering and pulling the mouse wheel back towards you increments the value.

However, for the overhead air vents in the top left and right corners of the windshield the direction of the mouse wheel is inverted.

We should push the mouse wheel away from us to close the vents just like we push the vent to close it.

onox:
Fixes #613 

Scrolling up moves the air vents away from you, closing them. Dragging should still be the same: in the direction of where you want to move the air vent to.

I have also reduced the increment/decrement step from 10 % to 5 %. Did the same for the cabin heat and air knobs. The tooltips sometimes show a value that is off-by-one (for example, 44 instead of 45). That's something that should be fixed in flightgear/simgear.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -1990,7 +1990,7 @@
             <binding>
                 <command>property-adjust</command>
                 <property>/environment/aircraft-effects/cabin-heat-set</property>
-                <factor>-0.1</factor>
+                <factor>-0.05</factor>
                 <min>0</min>
                 <max>1</max>
                 <wrap>0</wrap>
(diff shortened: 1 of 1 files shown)


onox:
@wkitty42 Could you perhaps test this?

wkitty42:
@onox is this code in master?

gilbertohasnofb:
> is this code in master?

No, it's on branch `bug-613`

wkitty42:
i was fixing to write that i'll pull it and give it a go but can't until the FGDATA version file is fixed... then i see that it has been merged so i'll test it in master as soon as i can use FGFS again...

Issue #633, opened 2015-12-26
Improve our lights


gilbertohasnofb:
Improve our lights using https://github.com/HHS81/c182 as reference. That plane's lights are not only smaller and more realistic, but they also fade away depending on the angle (e.g. starting at the landing lights head on makes it be seen at its brightest, and then it fades when the camera moves away).

![lights](https://cloud.githubusercontent.com/assets/5700795/12004714/2eee90b2-ab71-11e5-82c1-00adbdd1fead.png)

To be done
* [x] add new lights texture to branch `bug-633`

~~rename texture `transparent2.png` to `lights.png` and also in the `.ac` model (main and damage models)~~

* [x] rename texture `transparent2.png` to `lights.png` in the Model folder as well as in the Blender files (both main model as well as damage model)
* [ ] resize lights in Blender, particularly strobes (way too large) and the red beacon lights in the tail as well (just a bit too large)
* [x] solve the "squareness" problem of the flashing lights (see screenshots below)
* [x] add taxi lights to ~~front gear~~ ~~right~~ left wing, together with landing light (but weaker and pointing downwards)
* [x] make lights dependable on viewing angle in Blender (see c182 repository for an example of aircraft using this technique)

wkitty42:
will this also take care of the 2D(?) flashers on the wing tips and the 2D(?) landing light? i note, also, that there is no light showing on the wing or under the engine when the taxi light is on... that needs to be fixed, too...

gilbertohasnofb:
Yes, the 182 has better flashing lights as well (the red one in that image is a flashing light).

gilbertohasnofb:
I pushed the better light textures to a new branch `bug-633`, but the angle of the lights has to be done somehow in Blender (as I see that the only references to the lights in the 182 repository is in the .ac model). Can anyone have a look at it, maybe @wlbragg?

Also, I renamed our light texture from the meaningless `transparent2.png` to the meaningful `lights.png`, but I did this only to the `.ac` Model and the `.png` file and this has to be done also in the Blender file (which I don't dare to touch myself).

Finally, the sizes of the lights have to be reconsidered I think. The strobes in the wing tips are too large IMO, and also I can see some "squareness" around the flashing lights which we can try to solve, see:

![fgfs-screen-003](https://cloud.githubusercontent.com/assets/5700795/12006921/ee0c095a-abeb-11e5-9534-af0c005a4246.png)

[...]

PR #659, opened 2016-01-13
Bug #314 Correct glass external reflection.

### LINKED ISSUE #314: Add glass reflection visible from external view

wlbragg:
Fixes #314 

I think this is more of what we have in mind for the glass reflection. Please test.


onox:
The external reflection is better :+1:, but the glass is still buggy (unlike the glass in the J3Cub for example) and shows a difference depending on the viewing angle:

fgfs-screen-432
fgfs-screen-433

Another thing I find strange is that tuning parameters in `Models/Effects/exterior/exterior-glass-reflection.eff` doesn't seem to have any effect.

wlbragg:
>the glass is still buggy (unlike the glass in the J3Cub for example) and shows a difference depending on the viewing angle:

I don't think we can eliminate that issue. What it is (especially on the front glass) is that the glas object (the one that has the external reflection applied to it) is one piece including the front sides. The glas_interior  and glas_interior_sides have the interior effects applied to them. Those have the same shape but are two separate objects as opposed to one like the exterior "glas". That had to be done because the "rain on windshield" effect requires two different effects. One for the front and a different one for the sides. I verified it by hiding the interior glas_interior and glas_interior_sides and it behaved exactly like the J3Cub. The Cub has identical duplicates for exterior and interior. We would need to reshape the front and side glass that is one piece to a shape more subtle to a single rain effect, duplicate it for interior and apply the rain effect to that, using only one rain effect, the one for the front facing windshields.

>exterior-glass-reflection.eff doesn't seem to have any effect.

It may be the name used in the eff file. It had the entire path as part of the name. I removed the path part and am testing. I'll see if I can't figure out why it isn't working.

wlbragg:
I don't know onox, try disabling exterior-glass-reflection.eff from c172p.xml by commenting it out and you will see it is most definitely working, but I agree, the individual adjustments don't seem to do much.

wlbragg:
OK, try this. I actually was able to get rid of most of the funny effects caused by changing angles. What I did was create a new material in Blender that removed most of the influencing lighting effects from the glass and applied it to the interior glass. It didn't seem to affect the interior effects but reduced the collision of effects caused by having two surfaces to look through.
Let me know what you think.

onox:
I still get the buggy behavior :frowning:

wlbragg:
>I still get the buggy behavior

Please Describe!

Did you see any change?

wlbragg:
I tried doing what i said in a prior post. I hid the interior glass layers and made a new one identical to the glas object only a bit smaller so as to fit inside the other. It still had all the artifacts of the other version that is split into two pieces.
So, is this last commit better or worse that the original?
Is it better or worse than the last revision?

I can keep trying to adjust the interior glass material settings and see if there is something that will eliminate the artifacts.

I guess what puzzles me most is that we don't see them in the Cub? That being said then it probably boils down to the shape of the glass.

wlbragg:
>I guess what puzzles me most is that we don't see them in the Cub? That being said then it probably boils down to the shape of the glass.

It is doing the same thing. So I am either incompetent at glass effects or it's the nature of the technology we have available to us at the moment.

onox:
@wlbragg Shall we just close this PR? We haven't determined the root cause of the buggyness yet and the exterior reflection isn't working great. (Compare it to the [c182](https://github.com/HHS81/c182))

wlbragg:
Let me reply tomorrow after I take a look!

wlbragg:
I think there are a few issues here.

1) the shape of the glass has everything to do with the effects. Our glass shape is inferior.
2) the artifacts that we are seeing are directly related to the double panes of glass. It is consistent with every other aircraft that I looked at that is using two layers of glass. Even the f15 when rotated around at certain angles is showing some artifacts (look at the side low angle). Also the f15 loock like they were able to use two pains of glass without offset? I have tried that at least two times and we can't do it.
3) strip away the interior glass and our exterior reflection is comparable to the c182 (taking into consideration the inferiority of our glass shape in Blender.

Believe me I have put hours into looking and comparing the c172p with other glass. I just spent over and hour trying, no offset using an exact duplicate of glas without the the need for the sides, same thing only offset slightly by re-sizing it. Exact duplicate gave no interior effect, re-sizing it gave an interior effect but not as good as what we have now by splitting the sides off. Re-sizing it also caused the artifacts (the artifacts are the interior glass being visible at certain angles).

So if you think the existing glass effects are superior (or less buggy) than this PR, yes, I don't mind if you close it.

I am still determined to get this right, but it may require a new windshield.

wlbragg:
Before you close this PR I urge you to compare it one more time with what is in master. If you look closely you will notice the same artifacts are in both versions, with the difference being the effects are stronger (both interior and exterior) in the PR version .That is why you notice the artifacts more.

I think the PR has better effects, but the trade-off is stronger artifacts. However, keep in mind the artifacts are only visible when viewing from the outside and rotating the aircraft.

onox:
I'll close this PR because there are conflicts (due to the binary .blend file). @wlbragg Please do all your changes to the .blend file in one branch. That will make your life easier :smile: You could use branch `bug-633` which is still open.

> the artifacts that we are seeing are directly related to the double panes of glass. It is consistent with every other aircraft that I looked at that is using two layers of glass.

You may forward this to Thorsten and ask what he thinks of it. It seems like as it is now, ALS does not support exterior glass reflections + internal glass effects if I understood you correctly.

wlbragg:
Can we shut off, hide disable or get rid of the interior glass layer with a condition if we are viewing from the outside of the aircraft?

onox:
You could use `/sim/current-view/internal`.

wlbragg:
It works, shall we open this back up?

wlbragg:
Opps hit the wron button. I'll make a new branch without conflicts.

wlbragg:
What is the best way to do this. It required a blender and ac change. I need to use the blend in #633? Should I just add it there?

onox:
You should work with the .blend file in `bug-633`, since that is the branch in which you're modifying the blend files.

onox:
@wlbragg If it's possible to toggle the glass effect on/off with `/sim/current-view/internal`, why do you need to make changes to the .blend file? :open_mouth: I thought that you told me the only way to get actual proper reflections in external view is by disabling the glass effects?

wlbragg:
What I found was two things.
It required changing the lighting settings on the glass to get better visual rendering (stronger reflections), both interior and exterior.
It also requires, (at least in all the cases I looked at), that if you have two layers of glass you see artifacts of the interior layer at certain angles. The only way around that is to hide the interior when looking at the exterior. The strength of the glass effects determines the strength of the artifacts. We have the same artifacts on our current glass, it's just not as noticeable because the strength of the effects are weaker.
The glass effects are in #633 now if you want to test.
The interior reflection is considerably stronger, maybe too strong, let me know what you all think.

onox:
Thanks for the explanation @wlbragg. I'll delete this branch (`bug-314`) because it's useless (changes have been added to `bug-633`).

PR #660, opened 2016-01-17
adjust columns so gauges fit better on widescreen monitor

tigert:
added two columns which will later be used for radiostack (see fgdata/Phi/examples/radio.*) but that needs more styling work still so empty for now.

### DIFF
--- a/WebPanel/c172p-webpanel.html
+++ b/WebPanel/c172p-webpanel.html
@@ -41,22 +41,28 @@ body,html {
 </head>
 <body data-fgpanel="true" data-fgpanel-props="c172p-webpanel-properties.json">
   <div class="row">
-    <div id="ASI" class="instrument col-xs-3" data-fgpanel-instrument="ASI.json"></div>
-    <div id="AI" class="instrument col-xs-3" data-fgpanel-instrument="AI.json"></div>
-    <div id="ALT" class="instrument col-xs-3" data-fgpanel-instrument="ALT.json"></div>
-    <div id="VOR1" class="instrument col-xs-3" data-fgpanel-instrument="VOR1.json"></div>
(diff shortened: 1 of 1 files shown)


onox:
You moved the RPM and EGT instruments. Any particular reason for this? I prefer the layout to be identical to the actual 3D cockpit in FG.

onox:
Also, could you adjust the yellow needle of the EGT instrument to start at 0 (just like the white needle is). That should keep the rotation animation simple :smile:

tigert:
Hm, true. I must have been thinking of the late model 172 which swapped the location. I amended the pull request.

I also made a new pull req for the EGT, but since it contains changes to the already changed file, it has conflicts, so I am not sure how to do it correctly. I attached the fixed EGT file to this comment, so you can close the other pull req if that is problematic and use the file from here instead.

tigert:
[EGT.svg.zip](https://github.com/Juanvvc/c172p-detailed/files/94014/EGT.svg.zip)

onox:
The changes in PR #668 are fine, no conflicts.

onox:
Is the yellow needle in the EGT instrument moving in your browser? Here it isn't.

tigert:
Hmm.

There is something weird in the properties of engine[0] - the whole subtree is not visible in the property browser, there is only engine[1] visible, but I can manually set the url to engine[0] and it then loads it!

If I set the yellow needle to use the same source as the egt needle, it animates correctly, so the SVG and the object name of the needle are correct.

However, the properties from engine[0] do not seem to work in the web panel for some reason.

onox:
Is the same thing happening for the adf property?

onox:
How are you viewing the properties? In Phi? In can see the properties there.

onox:
> If I set the yellow needle to use the same source as the egt needle, it animates correctly

I can reproduce this too.

PR #661, opened 2016-01-18
Bug 392

### LINKED ISSUE #392: Garmin 196 GPS option
We could add the Garmin 196 GPS to our dashboard as an option (default off, to be activated via the menu). This GPS model is already located in fgdata, so it's very easy to add it to our plane. For my tests, I add the following lines to `Models/c172p.xml`:

```xml
<model>
    <path>Aircraft/Instruments-3d/garmin196/garmin196.xml</path>
    <offsets>
            <x-m>-0.4</x-m>
            <y-m>0.16</y-m>
            <z-m>0.205</z-m>
            <heading-deg>-12.0</heading-deg>
    </offsets>
</model>
```

And the following to the `c172p-set.xml` inside the `<nasal>` block:

```xml
<garmin196>
    <file>Aircraft/Instruments-3d/garmin196/garmin196.nas</file>
</garmin196>
```

This is the result, which is quite nice IMO:

image1

image2

image3

The good thing about directing to this GPS inside fgdata instead of copying it to our project is that any updates would also be automatically applied to our plane. For instance, I created a better texture for it but I couldn't make it work, and I reported it to the original creator.

onox:
Fixes #392 

The model and XML for the Canvas version is added as well, but hidden. This is so that anyone who wants to create the actual Canvas implementation (currently the screen doesn't show anything), can unhide the Canvas device and then compare the looks and behavior with the non-Canvas version.


gilbertohasnofb:
Code looks good and the GPS is working as expected.

gilbertohasnofb:
Hmm, I just realized something: the non-Canvas GPS we are using is the one from FGDATA (`Aircraft/Instruments-3d/garmin196/garmin196.xml`), but wouldn't it be better if we made it local so we could in theory make some improvements to it more easily, and then ask to merge it back tp FGDATA if we make some progress? 

For instance, we could apply the black texture and use the GPS mount model for this non-Canvas version. We can also take a look on some bugs it has, for instance: https://github.com/Juanvvc/c172p-detailed/issues/392#issuecomment-162254682

onox:
I think we should fix those things in the Canvas version.

gilbertohasnofb:
Well, that would be ideal, but in the meantime we have an ugly GPS in the plane and no prospect of having a Canvas implementation of it :cry: The texture and mount would be an easy fix, for instance. The others I don't know (I suspect that the screens shifting to a bit their position is due to the positioning of all those overlapping rectangles in the model, which could also be easy to fix). But given that I am not able to do those things myself (needs Blender for all three) I leave it up to you and the others, of course :smile:

PR #667, opened 2016-01-18
Bug 666: Set heading offset when clicking on "Autostart"

### LINKED ISSUE #666: Improve Autostart
Set heading offset (depending on magnetic variation) when clicking on "Autostart".

I also wanted to remove setting the oil-level back to 7.0. If the user has chosen not enable oil management, then the oil management will stay at 7.0 If the user DID chose to enable oil management, then the user has indicated that the oil must be managed. Thus resetting it back to the default is kinda annoying.

Same goes for the water contamination (which will be 0.0 if the user has turned it off).

I want to do the same thing for the "Allow securing aircraft" feature: whether it's enabled shows whether you want it or not. This is debatable and can be reverted.

onox:
Fixes #666 

This PR changes the behavior of "Autostart": when clicking on it, the tiedowns, chocks, and pitot tube cover are no longer removed. The reasoning is that they are only visible if the user has chosen to enable "Allow securing aircraft", and having that option enabled indicates that the user wants these things.

This change is debatable and I can revert it if too many people disagree.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -54,10 +54,10 @@ var autostart = func (msg=1) {
     var fuel_level_left  = getprop("/consumables/fuel/tank[0]/level-norm");
     var fuel_level_right = getprop("/consumables/fuel/tank[1]/level-norm");
 
-    if (fuel_level_left < 0.5)
-        logger.screen.red("Warning: fuel in left tank less than 50 %!");
-    if (fuel_level_right < 0.5)
-        logger.screen.red("Warning: fuel in right tank less than 50 %!");
+    if (fuel_level_left < 0.25)
+        logger.screen.red("Warning: fuel in left tank less than 25%!");
+    if (fuel_level_right < 0.25)
+        logger.screen.red("Warning: fuel in right tank less than 25%!");
 
     setprop("/controls/engines/engine[0]/primer-lever", 0);
     setprop("/controls/engines/engine/primer", 3);
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -40,6 +40,25 @@ var autostart = func (msg=1) {
     setprop("/sim/model/c172p/securing/tiedownR-visible", 0);
     setprop("/sim/model/c172p/securing/tiedownT-visible", 0);
 
+    setprop("/consumables/fuel/tank[0]/water-contamination", 0.0);
+    setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);
+
+    # Oil level warning for lazy pro's
+    var oil_enabled = getprop("/engines/active-engine/oil_consumption_allowed");
+    var oil_level   = getprop("/engines/active-engine/oil-level");
+
+    if (oil_enabled and oil_level < 6.0)
+        logger.screen.red("Warning: low oil level! Check level via oil cap on top of engine!");
+
+    # Fuel level warnings
+    var fuel_level_left  = getprop("/consumables/fuel/tank[0]/level-norm");
+    var fuel_level_right = getprop("/consumables/fuel/tank[1]/level-norm");
+
+    if (fuel_level_left < 0.5)
+        logger.screen.red("Warning: fuel in left tank less than 50 %!");
+    if (fuel_level_right < 0.5)
+        logger.screen.red("Warning: fuel in right tank less than 50 %!");
+
     setprop("/controls/engines/engine[0]/primer-lever", 0);
     setprop("/controls/engines/engine/primer", 3);
 
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -28,18 +28,13 @@ var autostart = func (msg=1) {
     var pressure_sea_level = getprop("/environment/pressure-sea-level-inhg");
     setprop("/instrumentation/altimeter/setting-inhg", pressure_sea_level);
 
+    # Set heading offset
+    var magnetic_variation = getprop("/environment/magnetic-variation-deg");
+    setprop("/instrumentation/heading-indicator/offset-deg", -magnetic_variation);
+
     # Pre-flight inspection
     setprop("/sim/model/c172p/brake-parking", 0);
-    setprop("/sim/model/c172p/securing/chock", 0);
-    setprop("/sim/model/c172p/securing/pitot-cover-visible", 0);
-    setprop("/sim/model/c172p/securing/tiedownL-visible", 0);
-    setprop("/sim/model/c172p/securing/tiedownR-visible", 0);
-    setprop("/sim/model/c172p/securing/tiedownT-visible", 0);
-    setprop("/engines/active-engine/oil-level", 7.0);
-    setprop("/consumables/fuel/tank[0]/water-contamination", 0.0);
-    setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);
 
-    #c172p.autoPrime();
     setprop("/controls/engines/engine[0]/primer-lever", 0);
     setprop("/controls/engines/engine/primer", 3);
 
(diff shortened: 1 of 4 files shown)


gilbertohasnofb:
> This change is debatable and I can revert it if too many people disagree.

I am one of those. Autostart is for beginners or extremely lazy people (or for people testing, as I often use it myself!) and given this focus I think it should take care of everything for the user, even if he/she selected the possibility of having fuel contamination, oil management, pitot tube cover, tie-downs, etc (I once had a conversation with @Juanvvc about it -- can't find the post -- and we both agreed that was the best route). In my view, Autostart basically means "press `s` to start the plane immediately without fussing with anything else", and that's what it should do.

gilbertohasnofb:
> Set heading offset (depending on magnetic variation) when clicking on "Autostart".

This is very good though!

onox:
I have reverted the securing behavior.

gilbertohasnofb:
That's cool, @onox! But I still think that the oil management and fuel testing should be taken care by the Autostart function as well. Look at it this way: this function is one of the few non-realistic things we have in the plane, and probably the only feature developed for beginners. If people can't fly for some reason -- even if they themselves activated those options -- it will generate a lot of headache, bug reports and complaints. I _never_ propose features for n00bs, but this case is unique and I really think we should reconsider it. I don't see a reason to bother people who want just to start the engine with a single keystroke with those other things even if their checkboxes are active.

onox:
About the oil management: it only works if you do a full manual start every single time. If you're one time lazy and forget about it, you will reset it back to its default value. Plus we don't set the fuel level either.

Can't we make a distinction based on whether the user has enabled "Complex startup procedures"?

What I'm concerned with is that we do keep noobs (full Autostart) and pro's (100 % manual) into consideration, but not the lazy pro's (me :smile:). I don't want to do a full manual start every single time I start FG, but I still want to simulate all systems and not have "Autostart" option reset things back to their default value (oil level for example).

gilbertohasnofb:
Yeah, I see that. To be frank, I am also one of these "lazy pros" myself as it has been more than 3 months since I last had a flight "for fun", and while developing I only use autostart (unless otherwise necessary). 

> Can't we make a distinction based on whether the user has enabled "Complex startup procedures"?

Sounds complicated to me, and might look cryptic to users. If a checkbox will influence these behaviours, then it must be the original checkboxes of allowing oil management and fuel contamination. 

Here is a possible solution:
- fuel contamination will be set to `0.0` by the Autostart function. Since it's random anyway, there won't be any problem of resetting it and people auto starting have decidedly not to care about pre flight inspection, be him a n00b or a pro.
- oil management: fine, let's use it's own check box for deciding that. But what if when the Autostart function notices that there oil level is low then it display a message in the scree: "check your oil!" or something like that?
- maybe same for fuel, a message displayed in the screen could help a beginner quite a lot.

What do you think?

onox:
Ok, I'll do this:

1. Set fuel contamination to 0.0.

2. Display message if oil level < 6.0 (when oil management is enabled, otherwise it's irrelevant)

3. Display message if fuel level in a tank is < 50 %

Is that ok?

gilbertohasnofb:
:+1:

onox:
@gilbertohasnofb Is it OK if I add F9 key binding to show "Fuel and Payload Settings" window?

gilbertohasnofb:
I suppose so!

onox:
@gilbertohasnofb Pushed new commits.

gilbertohasnofb:
Looks good, but just two comments:

- don't people normally fly with less than 50% fuel? Maybe we should lower this to 20%.
- check the comment in the code about the space between `50` and `%` in the warning message.

onox:
> I think there shouldn't be a space between 50 and % in the warning message. Same for two lines under this one.

That depends on whether you value English style guide or SI: https://en.wikipedia.org/wiki/Percent_sign

gilbertohasnofb:
I wasn't aware that SI recommended a space between it. Nevertheless, I would say the English style is much more reasonable here, though this is such a minimal issue that I leave it up to you to choose :wink:

onox:
> don't people normally fly with less than 50% fuel? Maybe we should lower this to 20%.

We can lower it if necessary. @wlbragg @dany93 @tigert What do you guys think?

gilbertohasnofb:
Some discussions by pilots here: http://www.cessna172club.com/forum/ubbthreads.php?ubb=showflat&Number=224266

Juanvvc:
What does the "Set heading offset (depending on magnetic variation) when clicking on "Autostart" mean? You never set the heading taking into account magnetic variation. This is something you do to calculate your headings during the planning phase, not during the actual flight. The ATC uses only magnetic headings, disregarding any local magnetic variation: http://www.askacfi.com/330/how-do-i-interpret-atc-directed-headings.htm

The HI must be initialized with whatever value the magnetic compass shows, and you must do as less maths as possible during the flight.

Please, give a second thought to this message. I fly in an area where magnetic variation is under 1º and hardly noticeable. These concepts are completely rusty to me!

About the minimum amount of fuel: enough fuel to arrive to your alternate airport. You think in terms of flying time, not fuel amount. I never ever would fly with less than an hour of fuel.

onox:
> The HI must be initialized with whatever value the magnetic compass shows

That's what the code does for you. Although it cheats a little because it uses the real magnetic heading instead of what the compass says.

When you then set the heading bug to the runway heading, it should be pointing to where your nose is after you have enabled the vacuum system (for the HI)

dany93:
> > don't people normally fly with less than 50% fuel? Maybe we should lower this to 20%.

>We can lower it if necessary. @wlbragg @dany93 @tigert What do you guys think?

It never bothered me, in fact I don't mind. How can we presume what people wish to do ? Very small detail.
People usually avoid taking off with too much fuel if not necessary, but 50% is not so excessive. There are more accidents by fuel shortage than by fuel  excess in the tanks.

gilbertohasnofb:
@dany93 the question is rather: do we want to bother an user that knows what he is doing with warning messages? If using less than 50% of fuel is uncommon, then let's just merge this like it is. Else I think finding a minimal fuel value that is more critical is a welcome addition, though indeed a small detail.

dany93:
@gilbertohasnofb 
OK, I agree.
Sorry, I hadn't understood well the context.

gilbertohasnofb:
No problems, Dany!

onox:
I read something in the POH about not doing uncoordinated flight if tank is < 25% (page 2-9), so I'll reduce it to that number.

gilbertohasnofb:
Sounds good to me. After that we can merge this IMO.

onox:
@gilbertohasnofb Although I still prefer a space between number % sign in code, I've removed it in the warning messages :smile:

gilbertohasnofb:
You are very nice :smile_cat: Okay, I will test it now and then merge it!

gilbertohasnofb:
@onox Everything works perfectly, but being super picky here, wouldn't it be better to write:
"fuel level of the right tank lower than 25%!" instead of "fuel in right tank less than 25%!"?

onox:
Sure I can write that.

gilbertohasnofb:
:+1: (I just made a small correction on my own phrase above)

onox:
You want to change the oil level warning as well?

gilbertohasnofb:
No, I think that's fine.

onox:
I think "of" is more appropriate than "in".

onox:
"fuel level of the left tank" or "amount of fuel in the left tank"

gilbertohasnofb:
Both are fine with me!

gilbertohasnofb:
Oh, sorry for not merging it but I didn't see your last commit (I don't get notifications for it).

onox:
The Github page doesn't update itself?

gilbertohasnofb:
Of course it does, but I have my thesis plus a thousand pdfs and tabs opened at the same time! :wink:

onox:
Hehe. You're working on your master's thesis?

gilbertohasnofb:
Yes, and the heat of the deadline is starting to catch up with me! :fearful:

onox:
Good luck!

gilbertohasnofb:
Thanks a lot, onox! :smile:

PR #674, opened 2016-01-20
Bug 673: Add some damping to the PID controllers of the tiedowns

### LINKED ISSUE #673: Wings may oscillate if tied down while fps is very low
If the the fps is low and the wings are tied down, the aircraft may start to oscillate a bit. Adjust the PID controllers to add some damping.

onox:
Fixes #673

### DIFF
--- a/Systems/c172p-ground-effects.xml
+++ b/Systems/c172p-ground-effects.xml
@@ -611,7 +611,7 @@
             <input>fcs/tiedown-left-mag-error</input>
             <kp>7000.0</kp>
             <ki>0.0</ki>
-            <kd>0.0</kd>
+            <kd>500.0</kd>
             <clipto>
                 <min>-50</min>
                 <max>1000</max>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
How can I test this, @onox?

wlbragg:
Really the only way is to see if the aircraft remains stable when tied down. You can apply force to it when tied down and also watch that reaction.
I can do this one if you want?

gilbertohasnofb:
> Really the only way is to see if the aircraft remains stable when tied down.

I never thought it was unstable before!

> I can do this one if you want?

That would be nice :smile:

wlbragg:
That was fun, very stable. I applied full prop and it stretched slightly. I hit it with a 110 knot wind from the front, enough to collapse the front gear and it held tight. Works great!

onox:
Haha, try to press Ctrl+u while being tied down, quite funny to see :laughing:

onox:
> I never thought it was unstable before!

I would become unstable at low fps (< 15 or something).

Issue #418, opened 2015-07-14
Seat mounts need to be 3D objects


gilbertohasnofb:
We had removed the alpha channel of the `interior.png` due to some problems with visibility through the windows, but then the seat mounts (all 4 seats) lost their transparency. See:

![fgfs-screen-002](https://cloud.githubusercontent.com/assets/5700795/8675172/b1ac4d60-2a42-11e5-8fb6-ca6672e73459.jpg)

The solution is to 3D model the seat mounts (so instead of the current rectangle, we could roughly model this surface as to fit the texture, still keeping it as a 2D surface).

gilbertohasnofb:
@Juanvvc do you think you can give me some advice here? Is there some adverse effect I should expect if I once again add an alpha channel to these textures?

Juanvvc:
There could be the usual issues with the alpha channel and textures in ALS and Rembrandt. Check this thread: http://forum.flightgear.org/viewtopic.php?f=5&t=22707

These issues do not have an easy fix, especially if we want to offer support both in ALS and Rembrandt. But these mounts looked ok in the past. Just recover the alpha channel and we'll see :) At the end of the day, we can always model these seat mounts.

gilbertohasnofb:
@Juanvvc 
> Just recover the alpha channel and we'll see :) At the end of the day, we can always model these seat mounts.

Okay! Let's try the alpha channel solution first, and if someone finds a problem we can just revert it back.

gilbertohasnofb:
@Juanvvc Okay, so I tried adding back the alpha channel (just around the seat mount area in the png file) and then I found out why we removed it in the first place: the interior of the plane isn't visible at all through the windows if there is even a single pixel with alpha channel in that file! 

So now we need solution number two: model the seat mounts in Blender. I am updating the description of the issue now.

legoboyvdlp:
Hmm... may be difficult. Will see what can be done.
I wish Blender worked like WED - point and click to add vertices...

wlbragg:
You'll pick it up quick enough. My main issue with using it was the shear amount of features and not having a clue about its capabilities. But I found a quick Google search almost always gave me the info I needed. Applying that info was another thing.

legoboyvdlp:
Right... I'll see what I can do with this! 
Looks simple enough...

legoboyvdlp:
@gilbertohasnofb @Juanvvc @tigert any ideas about the size of the item in question?

[...]

Issue #670, opened 2016-01-19
Copy 3D seat mounts from c182


gilbertohasnofb:
Copy 3D seat mounts from c182: that project has detailed mounts while we are still using a shameless cardboard rectangle :wink:

legoboyvdlp:
Thank goodness; I was trying tof igure out just how to start, and couldn't.
I'll need to be lot's more comfortable with Blender before I 'fix' (break most lkkely) the Cessna.

gilbertohasnofb:
No worries, lego :smile:

wlbragg:
OK, I'm confused, are the seat mounts backwards in my last commit? If so I can rotate them on the horizontal easy enough?

gilbertohasnofb:
I honestly don't know as the model isn't identical to the images we have. IMO the difference is small and it looks fine to me, but maybe you can ask Heiko what his intentions were.

PR #678, opened 2016-01-20
Bug 633 light effects, Bug 670,418 seat mounts, Bug 314 glass effects

### LINKED ISSUE #633: Improve our lights

wlbragg:
Fixes #633, #670, #418, #314 

:warning:  Not to merge until thoroughly tested! :warning:

I don't want to wait to long on this though as the blend touches a lot of thing and we need to make a decision and get this merged before we get too many conflicts.


wlbragg:
The conflict is in the blend and ac, I don't think it will be an issue but I need to check it out closely.

wlbragg:
@onox, I merged master into 633 locally and fixed the conflict with no ill effects. How do we do this online?

onox:
If you have verified that after merging `master` into `bug-633`, that the resulting .blend and .ac files are those from the `bug-633` branch, then you can push `bug-633` to Github. The PR should then indicate that the conflict has been resolved.

wlbragg:
:+1:

onox:
PR fixed #418 and #670 (seats)

Almost fixed #314 (the reflection color is a bit too blueish) and #633 (see comments by @gilbertohasnofb)

Lights are much more pleasant to look at :smile:

wlbragg:
The tail beacon is adjusted as good as I can get it. Landing and taxi light are set side-by-side but the effect is worse when they both are on then when they were on top of each other.
Strobe planes in Blender were reduced by about 25 percent, 50% reduction was too small.

onox:
> but the effect is worse when they both are on then when they were on top of each other

Have you tried giving the taxi light and landing light a slight difference on the x axis (nose to tail)?

> Strobe planes in Blender were reduced by about 25 percent, 50% reduction was too small.

I liked the strobes as they used to be, they're now a bit difficult too see from a distance. But maybe that's my opinion, I could be wrong.

onox:
@wlbragg I have modified the visibility arc range of the nav lights a bit. Use `git pull --rebase` instead of `git pull` in case you have already committed something locally.

onox:
@wlbragg If I look in Blender, I see the red/green lights have a cone that isn't really 180 deg. That causes the lights to become invisible too soon when viewed from the front.

gilbertohasnofb:
@onox are you sure they are supposed to be 180 degrees? I once came across this image here, but I didn't check if the values are correct:

onox:
The same goes for the tail nav light:

nav-lights-cones

@gilbertohasnofb The actual visibility range can be controlled via the "flash" animation.

onox:
@gilbertohasnofb Whatever is easier for @wlbragg. I have already adjusted the XML, but the cone object itself reduces the visibility too earlier:

wing-nav-light-cone

gilbertohasnofb:
Well, I leave this decision to you two. I just wanted to share something I had once come across, but personally I have no idea how these lights should look in RL so :no_mouth:

onox:
So I think that the actual cone object should be wider than the "flash" animation allows, because currently you can see it's a cone (You can see it's a cone when you look at the light at the edge angles of the cone).

onox:
Currently the min value of the "flash" animations are still 0.5, that could be later set to a lower value or even 0.0.

onox:
In other words: I think the cone should be a bit wider at the front, so that the "flash" animation reduces the visibility to 0.0 *before* you actually reach the edge of the cone.

wlbragg:
So the cones need to be bigger? I don't understand what the cones do. What if you angled them forwards slightly to get the forward area covered?

onox:
Let me make a FG screenshot to show the problem

onox:
When you approach the edge of the cone:

fgfs-screen-449

onox:
I could try to rotate it in XML... let me check.

onox:
Hmm... I think best solution would be to widen the cone a bit in Blender.

onox:
The front side that is, see https://github.com/Juanvvc/c172p-detailed/pull/678#issuecomment-174602619

onox:
@wlbragg Can you measure the width of the cone in Blender?

wlbragg:
I think I see. For the range of view, angle wise, I thought I noticed by increasing to the larger area or angle in the XML, that we have a little room to rotate the cone slightly forwards and still see get a view (left side) from 90 - 180, admittedly though, I don't know exactly what the cone is for.
Is it only to hide the light by its shape?
Same thing for the strobes, is it that you need a surface (or vert ice) to attach a light to?
Then you can use XML to position where that light is to appear?
If we moved the light to the convex side of the cone would it take on that range of view instead of
< 180 it would be >180? I'm not suggesting this, only trying to learn  about the lighting.

Yes I can measure it, hang on.

onox:
I think the cones are just vertices to which you attach the `lights.png` texture.

onox:
> If we moved the light to the convex side of the cone

What do you mean with this?

wlbragg:
Crude measure is 0.358 m

wlbragg:
That was left wing cone.

onox:
> that we have a little room to rotate the cone slightly forwards and still see get a view (left side) from 90 - 180

For the red/green nav lights, the range is 110 deg, from 0 deg from nose to 20 deg behind wing. So I think we could just rotate the `navlight_left` and `navlight_right` objects around the z-axis a bit so that one side is perfectly aligned with the x axis?

onox:
> Crude measure is 0.358 m

I meant in degrees :wink:

onox:
Like this:

nav-light-new-cone

onox:
@wlbragg I think `navlight_back` is OK. I measured that I can rotate it about 11 deg to align it with the y axis, which means the cone is 180 - 11*2 = 158 deg wide, wider than the required 140 deg, so tail nav light is OK.

onox:
Red/green nav lights need to be rotated about 11.3 degrees.

wlbragg:
the side cones 180 - 9.239 * 2 ?

onox:
I measured 11 degrees by selecting the object and then pressing: r z 11.0

onox:
How did you measure?

onox:
But I'm not sure if that is what we want. Currently the normal of the cone is 90 deg from the x axis (perpendicular to it).

onox:
We could set it to 35 (that's 55 deg from the nose).

onox:
Then we just need to make sure that the "flash" animation has a min value of 0.0 instead of 0.5.

onox:
@wlbragg Try this:

1. Rotate the red/green nav lights 35 around the z axis (towards the front)

2. Go to the flash animations (assuming you have done `git pull --rebase` to get my commit)

3. Change `<min>` from 0.5 to 0.0

4. Change `<offset>` from 0.0 to -0.5

onox:
These settings might give a shorter descend from 3.0 to 0.0:

power = 2.0
factor = 12.0
offset = -3.6
min = 0.0
max = 3.0

wlbragg:
Sorry, doing two things at the same time today.

cone3

onox:
If I select face angles, I see 67.955: (90 - 67.955) / 11.0225

onox:
Anyway, could you try 35 deg + the settings from https://github.com/Juanvvc/c172p-detailed/pull/678#issuecomment-174621852?

onox:
I'm interested to know whether that gives better results :smile:

wlbragg:
OK, I know nothing so definitely don't go by my GUESS!

Yes, give me a little bit and I'll see what I can do.

onox:
It seems XML changes here are OK, so I think if you just rotate red/green nav lights in Blender 35 deg, that should be enough.

I'm currently adjusting the tail light XML.

onox:
@wlbragg Have you done any commits already? XML is ready here locally.

wlbragg:
No I haven't. Will it mess up the XML if I also adjust pos x and y by a tiny bit?

wlbragg:
Rotate 35 deg cuts into the wind just a tad.

onox:
> Will it mess up the XML if I also adjust pos x and y by a tiny bit?

No.

wlbragg:
OK

wlbragg:
I'm pulling now, I haven't done anything since my last commits.

wlbragg:
You didn't touch anything in the blend?

onox:
Nope. Just the XML file. But my latest changes in the XML aren't pushed yet.

onox:
I can push those as well if you want.

wlbragg:
Yes, go ahead.

onox:
OK, hang on a second.

wlbragg:
:+1:

onox:
Pushed.

wlbragg:
:+1:

wlbragg:
I thought the strobes were fine the way they were, as you can never see the size of the circle but you need the distance. You want me to put that back or a compromise?

onox:
If you want to revert shrinking the strobes, go ahead.

onox:
Who said the strobes were too big?

gilbertohasnofb:
me... :disappointed: I didn't try the new ones yet, but I really think the original was were huge, they were like twice the landing light size!

onox:
But they blink, so don't they simulate blinding you a bit? :smile:

gilbertohasnofb:
Not really, I think there is a difference between the halo and the "ball" of light itself. It just looked outrageously large to me!

wlbragg:
But do we ever see the ball?

gilbertohasnofb:
Well, I do

onox:
@gilbertohasnofb What do you think of the new strobes when you zoom out?

I'll let @wlbragg decide what he thinks best.

gilbertohasnofb:
But really, you two are doing all the work so the decision should be up to you. I was just giving my 2 cents.

wlbragg:
@gilbertohasnofb, check them now for size and distance and give me your opinion!

gilbertohasnofb:
@onox I will give it a try, just a sec

wlbragg:
Moving the cones changes the XML x, y and z position of the lights. So I am trying to get it as close as I can by adjusting the cones and then the rest will have to be done in the XML. My experience has been lots of small corrections until its right. Also it may not be an issue cutting into the wing, still testing.

gilbertohasnofb:
Hmm, I love the new size of the strobes, here it's zoomed out:

fgfs-screen-002

For me the problem comes from two things:

- is the strobe also sensitive to angles? I had the impression they are, and I think they shouldn't be.
- aren't the strobes and nav lights too close to each other? From the same distance as in the image above, but looking from the front, I can just see a short blink and it's indeed very difficult to tell the strobes are there.

onox:
Just rotate those red/green cones 35 deg, as I said before. That should put the normal of them at 55 deg from the nose.

onox:
@gilbertohasnofb Did you try zooming out to 150 meter?

gilbertohasnofb:
That's the most zoom out I can get by pressing `x`. Does zooming out via the menu allows more? I will try again

wlbragg:
OK, try that. The nav lights may need a slight XML adjustment.

onox:
@gilbertohasnofb Don't use x. That's the FOV. Use mouse wheel.

gilbertohasnofb:
oh

onox:
Make sure you have reset the FOV with Ctrl + x. Mouse wheel only works in select views like Helicopter View.

wlbragg:
Taxi and landing light, I think I need to re-position their cones offset side-by-side (I hadn't done that, only in the XML) and see what kind of visual we get. Right now it's unacceptable. Also @onox, you think I should have the taxi light offset closer to the plane than the landing light?

onox:
Yeah, maybe that will fix your problem. Not sure.

gilbertohasnofb:
Okay, so I got it now to 150m and it looks fine to me:

fg3

@wlbragg look what happens once we get in the angle where the nav lights are visible:

fg2

Probably we should move the strobes a bit backwards so they don't superimpose each other. Or maybe make the strobes slightly "away" from the wings, so that it covers the nav lights when blinking (that may be a better option).

wlbragg:
@gilbertohasnofb, OK choose which one you want to try first, offset sideways or in front?

gilbertohasnofb:
@wlbragg sideways I think! it will probably look good from any angle, while the other may be sensible to problems. Let's see :smile:

onox:
Navigation lights are OK now :+1:

onox:
Hmm... although they intersect with the wing. @wlbragg If possible you could try an arc width of 110 deg for red/green.

onox:
But the dimming of the nav lights is OK now when viewing from the front.

wlbragg:
>Hmm... although they intersect with the wing.

Not sure what you mean, in blend they do but I don't see it in the visual.

>try an arc width of 110 deg for red/green.

Where?

onox:
Let me make another screenshot.

wlbragg:
The taxi and landing lights, the cones interfere with each other no matter what I do, any suggestions?

onox:
nav-light-110

wlbragg:
OK, got it!

onox:
Or better give it some extra margin, 120. We limit it to 110 in the "flash" animation anyway.

wlbragg:
I think I can do that, but what are you seeing and where that make you think that is necessary?

onox:
https://github.com/Juanvvc/c172p-detailed/pull/678#issuecomment-174653303 Press F5 to refresh the page.

wlbragg:
OK

onox:
Do you see the image in that comment?

wlbragg:
Yes, I see it now.

onox:
:+1:

wlbragg:
If I move it away (then the XML will need adjusted.

onox:
Yes, you can update its position if you have too. The other settings (power, offset, factor, min, max) do not need to be adjusted. They are correct for now.

wlbragg:
OK, this is going to take a little bit, I'm also adjusting the taxi and landing and strobes as I go. My refresh takes long enough that I like to get as much as I can out of each refresh of FG.

onox:
You can do Debug > Reload Aircraft Model in FG :smile:

wlbragg:
While we were debating it all I managed to get this, the artifact is where the cones overlap, We won't be able to allow any overlap, I don't think.
Thanks for the "debug" tip, I already use it, I've just worked on computers and in programming too long and have an aversion to waiting, it causes major stress anymore :smile: .
cone6

onox:
Hehe. You're welcome.

gilbertohasnofb:
@wlbragg Your image above looks very good to me! :+1:

wlbragg:
OK, it's all really close. I have to run for about 30 minutes, test and let me know if it needs any further adjustments.

wlbragg:
The strobes may need more offset, the landing and taxi are adjusted farther than in the last image and look at them from angle above the aircraft. The nav lights cannot currently be moved any closer to the plane without moving behind the cones (bad effect).
See ya all later.

onox:
You moved the nav lights? :open_mouth: I thought you were going to modify the shape of the nav light cones. The nav lights seem to be visibly distant from the wing tips :frowning:

wlbragg:
Re-positioned the taxi and landing cones and XML slightly, there was a bad effect when using just the taxi. I'm really happy with them now.

I need feedback on the strobes specifically.

wlbragg:
>You moved the nav lights?

Yes, I thought either option was a possibility.
You think the shape will work better? I think I have the ability to do that, let me try if it is not close enough the way it is.

onox:
It looks like the taxi and landing lights are good now :+1:

wlbragg:
OK, close to perfection on the nav's

wlbragg:
The only other thing is the strobe positions, I need @gilbertohasnofb to sign-off on that, I guess.

onox:
Navigations lights are better now! :+1: :smile:

gilbertohasnofb:
@wlbragg sorry that I am a bit away, but I am keeping track of the conversation here. Should I test the new strobes now or do you need to work on them still?

wlbragg:
Test them now and let me know what you think compared to the last time you saw the. That may give me an idea on how much more to move them, if any.

gilbertohasnofb:
Ok, I will do it right now :+1:

onox:
What do you guys think of the brightness of the tail nav light? I think it's a little bit too dim.

wlbragg:
Me too!

gilbertohasnofb:
I agree with that as well. But everything else is looking fantastic in my opinion, really beautiful work! :clap:

gilbertohasnofb:
What a difference from the previous lights :smile:

wlbragg:
:clap: :clap: :clap:

onox:
I'm updating the XML of the tail. Also increased the brightness of the green nav light a little bit.

gilbertohasnofb:
:+1:

onox:
Pushed. Are the brightnesses ok?

gilbertohasnofb:
Testing now.

gilbertohasnofb:
Looks much better, @onox! The only thing I would still modify before merging this PR is the tail red beacon:

fg

If we can't solve the collision, then maybe we should raise the light so it doesn't collide with the glass, what do you guys think?

onox:
> maybe we should raise the light

I'm not fond of that idea :frowning:

onox:
Perhaps split the beacon object into two parts and hide the top part when the light blinks?

wlbragg:
Me either, I tried that locally and didn't like the results. 
What I don't know or understand is the XML off-set, what is that for?
Is there a way to offset the light only from the viewers perspective?

wlbragg:
I already split it to be able to change the transparent texture. We can try hiding it.

gilbertohasnofb:
Yes, we can try that, but it sounds like a hack to me. Heiko wasn't able to figure out why are we having this problem and the c182 isn't?

wlbragg:
BeaconOff is the clear, red glass. BeaconMount is the base of it now (new  object).

onox:
I meant the top half of th red glass. But it does indeed feel like a hack.

onox:
@wlbragg The c182s doesn't have the problem. Maybe we should first check how that plane does it.

wlbragg:
>I meant the top half of th red glass. But it does indeed feel like a hack.

Oh, I agree.

I tried offsetting in XML various different ways and it always offset in one direction only and not from the viewers perspective.

I'm not prepared to change Blender versions to run the suggested Python scripts.

onox:
c182s seems to use a "blend" animation.

wlbragg:
I can take another look at the c182 and see if I can recognize the differences and the technique used.

>c182s seems to use a "blend" animation.

Makes sense!

onox:
And "alpha-test", "billboard", and "dist-scale". If I'm looking at the correct object (PlaneX)

onox:
I don't mind if we ignore that little problem for now and merge this PR :stuck_out_tongue_winking_eye:

wlbragg:
Works for me!
I will take another look at the tail beacon later on and see if I can figure it out, but I am in favor of getting what we have merged and closed.

gilbertohasnofb:
Okay, who wants to press the big green button? :yum: 

And I will open a new issue for the beacon!

onox:
First! :laughing:

gilbertohasnofb:
hehehe

(see: https://github.com/Juanvvc/c172p-detailed/issues/682 )

wlbragg:
:+1: :clap:

gilbertohasnofb:
:clap: very nice work!

onox:
Yeah, the lights look great :smiley: Nice work guys

wlbragg:
Agreed, thanks for all the help and training!

onox:
Now someone else may merge that other PR that I opened today :wink:

gilbertohasnofb:
> Now someone else may merge that other PR that I opened today 

Would you be able to check that, @wlbragg ? I still have to read two articles before going to sleep :sob:

onox:
Thanks you @wlbragg :smile:

gilbertohasnofb:
@wlbragg Oh, I see you just did, thanks!

onox:
Could someone notify on the forum in the FGPromo thread that they should download the lastest master and reshoot the scene with the c172p. Mention that we have proper external glass reflection and better lighting now.

wlbragg:
>notify on the forum in the FG promo thread

Done.

PR #681, opened 2016-01-25
Bug 680: Do not kill engine if "Damage" is disabled

### LINKED ISSUE #680: Propeller can be crashed while "Damage" is disabled
Propeller and engine can be killed while "Damage" is unchecked.

onox:
Fixes #680


wlbragg:
Good one!

Issue #592, opened 2015-12-02
Duplicate wing objects


onox:
With the current Blender files we have 10 objects for 10 states.
A lot of objects are partly duplicated because of this (the vertices).

Example 1: the middle section of `wing_1BB` and `wing_1BD` are (basically) the same.
Example 2: right outer wing part of `wing_1RGLB` and `wing_1LD` are the same.

| Left | Right | Blender |
| --- | --- | --- |
| ok | ok | (in default Blender file) |
| dmg | ok | wing_1LD |
| brk | ok | wing_1RGLB |
|  |  |  |
| ok | dmg | wing_1RD |
| dmg | dmg | wing_1BD |
| brk | dmg | wing_1RDLB |
|  |  |  |
| ok | brk | wing_1GRB |
| dmg | brk | wing_1LDRB |
| brk | brk | wing_1BB |
|  |  |  |
| col | col | wing1Collapse |

If we split the wings in two halves (left and right) we can reduce it to
6 or 7 objects for 10 states. If you want to make a change to one of the parts,
for example the left outer damaged wing, you only have to modify 1
object (`left_dmg`) instead of 3 (`wing_1LD`, `wing_1BD`, and `wing_1LDRB`).
It will make the model XML file easier and smaller.

These are the objects you get:
- `left_ok`
- `left_dmg`
- `right_ok`
- `right_dmg`
- `col_and_col` (or you can choose `left_col` + `right_col`)
- `inner_wing`

So now we only need about 5 select animations (`inner_wing` object is always
displayed) instead of 10.

The only thing the 3D modeller has to be aware of is to make sure the vertices
of the outer parts precisely align with the `inner_wing` object so that the user
doesn't see the cutting edge when a wing halve (left and/or right) is ok.

For all the 10 states here's which objects to display:

| Left | Right | Model XML file |
| --- | --- | --- |
| ok | ok | inner_wing + left_ok + right_ok |
| dmg | ok | inner_wing + left_dmg + right_ok |
| brk | ok | inner_wing + right_ok |
|  |  |  |
| ok | dmg | inner_wing + left_ok + right_dmg |
| dmg | dmg | inner_wing + left_dmg + right_dmg |
| brk | dmg | inner_wing + right_dmg |
|  |  |  |
| ok | brk | inner_wing + left_ok + right_brk |
| dmg | brk | inner_wing + left_dmg + right_brk |
| brk | brk | inner_wing |
|  |  |  |
[...]

PR #685, opened 2016-02-04
Bug 683: Use new ALS light effect for navigation lights

### LINKED ISSUE #683: External lights aren't visible from certain angles
Gijs wrote in the forum:

> The red and green lights should be visible when flying head on towards the aircraft. The table from FAR §23.1391 says that between 0° to 10°, the minimum (!) intensity should be 40 candelas. Below 0° the lights on the c172p are now completely invisible, whereas I would expect a gradual reduction. See for example the following videos:
> 
> https://www.youtube.com/watch?v=KR-Girnjcyg
> 
> https://www.youtube.com/watch?v=0kaXYQb5ko4

onox:
Fixes #683 

Basic renderer and Rembrandt are still supported. The position of the green
and red navigation lights are translated if the corresponding wing is
damaged. The navlight_{left,right,back} objects, used by the non-ALS
renderers, make use of this animation as well so that all the other
navlight_* objects can be deleted from the Blender files.

@wlbragg If the navigation lights function OK, then you can start deleting a whole bunch of `navlight_*` objects from the Blender files. See https://github.com/Juanvvc/c172p-detailed/blob/bug-683/Models/c172p.xml#L3557-L3575 These objects are no longer used.

* Have not touched the strobes, beacon, and/or taxi/landing lights (yet)
* Because I was a bit lazy, green/red navigation lights are off when the whole wing is collapsed. Is OK IMO because wing collapses when you smash your aircraft against the ground; lights probably won't survive that.
* On your own computer nav lights are moved when a wing gets damaged. I added support for MP aircraft as well, but I haven't tested it. MP is kinda broken for me since the particle system broke.


onox:
Have not touched the strobes, beacon, and/or taxi/landing lights.

onox:
I use one translate animation for the left wing, and one for the right wing. I'm assuming that damaging the opposite wing has no effect on the other wing. This assumption is currently false because the outer wing parts are not mirrors of each other. If it was true (left and right are identical) it would be functionally the same, but simplify the code and Blender model.

So the translate positions are not fully correct in all cases, as you can see in the screenshots below.

Making the outer wing parts separate objects and identical mirrors simplifies calculations and probably the Blender file as well. Current wings have very slight variations, which is nice, but I doubt most users will really notice and, and it makes the aircraft more complicated (because of all the duplicated wings).

Read #592 for more information.

fgfs-screen-460
Tiny red dot is missing when the opposite wing is damaged:
fgfs-screen-461

Right wing differs depending on state of left wing:
fgfs-screen-462
fgfs-screen-464

onox:
Also, because I was a bit lazy, green/red navigation lights are off when the whole wing is collapsed. Is OK IMO because wing collapses when you smash your aircraft against the ground; lights probably won't survive that.

**TL;DR;** If anyone doesn't understand the wing issue: wings and navigation light objects are heavily duplicated. If you look at https://github.com/Juanvvc/c172p-detailed/blob/bug-683/Models/c172p.xml#L3557-L3575 you see that you can avoid those duplicates by re-using the objects. Which is what I did for the navigation lights. The wings need to be fixed in #592.

gilbertohasnofb:
@onox I tested this branch and it looks really good, thanks for implementing this! I just wonder if there is really no fix for the beacon light colliding with the lamp glass object, it makes the effect look not so nice :cry:

wlbragg:
If you all can wait, I'll give my opinion and critique on this by tomorrow afternoon.

wlbragg:
I'm looking at this now and I forgot to pull a new fgdata so the effect of course isn't working, is there a way to condition for that so you can default to the non-ALS in that event, or is that not relevant because to get the new aircraft (with the effect) you need the new fgdata?

I see the fallback (non-ALS) is the original nav-lights, do we still want to use the "limiting code" for non-ALS? We are really close to getting it correct. All that was really left to do was flatten the angle on the cone (which I did locally and it looked much better) and change the view angle range? Limiting the cones angle to almost 180 is basically the same as the c182 using just a square plane (with the exception that it is round to better match the light texture (even though the fading at the outer edge make that unnecessary anyway).

Are we going to wait to re-positioning the actual nav light fixtures until we get a/the new c172p model? The wingtip shape of our existing model may need improvement and I still think we need to move the lights down and maybe forward, IMO.

I haven't even looked at the new work yet, doing so now.

wlbragg:
Wow, why does everything we do always present as many issues as we solve?
tip

I wish I had taken a picture of the quads before I added the new fgdata light effects. Do you think we need to angle or maybe even try a very slight angled cone in place of the quad?

Maybe it will help with the black artifact and this cutting into the wingtip?

gilbertohasnofb:
> I wish I had taken a picture of the quads before I added the new fgdata light effects. 

I think the branch in the PR by jarlarntzen still has the older lights.

wlbragg:
I mean the new quads as they are now with the new light effect. It was an error because the effect wasn't in fgdata yet, but it showed their relationship (angle) to the wing.

Basically this
tip1

wlbragg:
I'm not sure this positioning is correct, in fact I'm more sure it is not. Unfortunately testing using a cone is considerable work getting it positioned, translated and rotated correctly.

@onox, are you rotating and translating the procedural_light object for all three nav light positions, basically like I show the left one in this blend shot?
tip2

Are you planning on leaving the "bad" offset when damage to only one wing is present until we unify and mirror the damage, I'm asking if that is the plan? Or are we going to do every needed offset for each type of damage? Or are we going to do what I already did with the taxi and landing lights, (because of the amount of work involved), that they are considered "broken" when we physically damage the wing?

wlbragg:
Hum, I guess I don't understand this very well. I'm a bit frustrated because it seems we opened a whole new can of worms. My first impressions, (which may be faulty), are I am disappointed with what I am seeing. I'm not seeing much gain yet to what we already had but a whole new set of issues!
But admittedly, I really don't understand exactly how this works.

wlbragg:
At this point in time I don't think we will be ready to add these light by Tuesday, at least that is my opinion. I feel there are questions and issues than need to be addressed. Maybe they can be worked out by then, maybe I am over reacting. I just know I am feeling really apprehensive because of what I am seeing.

onox:
@wlbragg The cones in the Blender file are used by Basic + Rembrandt. These cones are NOT used by ALS.

Instead, ALS uses `Models/Effects/lights/procedural_light.ac`. If you import it, you see it's a quad and the vertex shader (`Shaders/light-ALS.vert`), if I read Thorsten's GLSL correctly, makes sure it is always pointed towards the viewer.

Maybe the vertex shader could be modified a bit to offset the whole quad. This is what we would need for the beacon light to prevent the quad from colliding with the beacon object.

onox:
> @onox, are you rotating and translating the procedural_light object for all three nav light positions, basically like I show the left one in this blend shot?

Thorsten's vertex shader is transforming the quad to make it point towards the viewer.

The Basic + Rembrandth cones are left untransformed, except that I translate them when there's damage.

> Are you planning on leaving the "bad" offset when damage to only one wing is present until we unify and mirror the damage, I'm asking if that is the plan?

Yes. I can adjust the green damaged light position a bit so that its error is a bit less if just the right wing is damaged.

> Or are we going to do every needed offset for each type of damage? Or are we going to do what I already did with the taxi and landing lights, (because of the amount of work involved), that they are considered "broken" when we physically damage the wing?

We could, but there's not much time anymore. The main benefit of using Thorsten's procedural lights is the viewing angles are now much better and you don't see the ugly cones anymores.

onox:
@wlbragg For now what I need you to do is remove all the `navlight_*` objects from `dev/blender/c172p-damage.blend` and export to `Models/Effects/damage/c172p-damage.ac`.

wlbragg:
>ALS uses Models/Effects/lights/procedural_light.ac

 I understood that, what I wasn't sure about is how it was translated and rotated, but it makes sense that the shader rotates it to always face the viewer. However the way it cuts into the model and the black artifact of the physical  nav_light is very undesirable and enough to ruin the effect IMO.

>The main benefit of using Thorsten's procedural lights is the viewing angles are now much better and you don't see the ugly cones any more.

I want to agree except the way HHS81 uses a simple quad instead of the cone in the c182 I think actually looks better than the shader. The light cutting into the wing and the dark artifact at the extreme view angles is very disappointing. I think at this point the c182 non-ALS looks better than the shader effect.

Any idea why the light is cutting into the wing at certain angles?

But I am not about to suggest not using the shader, although I don't know why, better looking is better looking. But your right about being out of time and you and Thorsten have put too much work into it to simply do away with it. However I think, even if not right now, we should ultimately make both methods available (shader or no shader) and the non-shader version should mimic the c182.

I should be able to remove all the damaged nav lights from c172p-damage.ac before Tues but it may be Monday before I do.

wlbragg:
>Maybe the vertex shader could be modified a bit to offset the whole quad. This is what we would need for the beacon light to prevent the quad from colliding with the beacon object.

I pointed Thorsten to all our conversations on the subject so he might take this need into account when he created the shader, I don't know if he read it or not and tried to account for those needs. It sounds like your pretty sure that we don't have that option as it stands now?

wlbragg:
Maybe I am over reacting, I've had a bad week. After testing these lights a bit more especially after dark, they are kind of growing on me. The effect of looking at angles is pretty cool.
 I'm sure those of you that don't care for the spikes (star effect) are a bit disappointed (I guess it never really bothered me). I didn't know the available light was going to play a part in whether the lights showed the spikes. As it gets darker all the sudden the lights spike? I was under the assumption that we could control that by the light intensity not the available light?
Anyway, I still think we should offer both and I don't mean the old default, I mean the new ones like on the c182. If there are no objections I might work on that after the merge.

onox:
The cutting in the wing might be able to be solved if the vertex shader moves the vertices of the quad a bit closer to the viewer. For a beacon light this would work (note that currently we're not using the ALS shader for the beacon yet). For the nav lights on the wing it could work 99% of the time. It won't help of course if the wing is definitely between you and the light (when you look at the right nav light from the left, for example), but in that case the nav light probably isn't even visible anyway, so even in that case you won't notice the quad.

onox:
So in other words: the quad could orbit around its origin, so that it won't intersect with any possible object that happens to be at its origin.

HHS81:
> So in other words: the quad could orbit around its origin, so that it won't intersect with any possible object that happens to be at its origin.

@onox and @wlbragg : that's possible without touching the shader itself.
The procedural-lights need the procedural-lights.ac, a simple quad made in blender. Simply Shift the quad on the x-axis in the negative area a bit. Unfortunately it won't completely stop the cutting into surfaces, but it will work most of the time. 

I really do like the definition of viewing angles, but not that much how the lights itself looks. I don't like the star effect, as it looks cartoonish. It is true that we see small and bright light sources as "stars", but not as a perfect drawn star like we see here. 

I work in the short final area of an international airport, and I can see the lights of the aircraft very good. The lights I originally created came closer to what I could see there, and that's not perfect stars. Maybe a solution like we have on the reflection-shader would be great. This shader is a lot customizable, you can let define the strength of the shader by the .ac-materials, or a texture-greymap, or/ and properties.... Letting the user decide if he wants to let the shader create the light-texture procedural, or using an own texture, would be great.

onox:
> The procedural-lights need the procedural-lights.ac, a simple quad made in blender. Simply Shift the quad on the x-axis in the negative area a bit. Unfortunately it won't completely stop the cutting into surfaces, but it will work most of the time. 

Oh right, of course. I can try that tomorrow.

> I really do like the definition of viewing angles, but not that much how the lights itself looks. I don't like the star effect, as it looks cartoonish. It is true that we see small and bright light sources as "stars", but not as a perfect drawn star like we see here.

I also prefer to have a bit more artistic control over the lights. I'll let @wlbragg do the diplomacy :wink:

HHS81:
> more artistic control over the lights

That's describes it very well.

wlbragg:
@onox, @gilbertohasnofb , before I forget this, regarding the model viewed in FGRun.
The one in fgdata is D-EEQA, the current one in this branch is 4X-CHV. We did this on purpose because 4X-CHV is to be the default?

wlbragg:
>more artistic control over the lights

I need to know specifically what we need done. Can that even be achieved with the shaders? Those light are procedural, so unless an "artist" wants to submit some math to create the style of lights we want, I'm not sure how to achieve this?
Unless Thorsten is willing to give us a hook to the math that is generating the light prior to the environment shader influence, I'm not sure that is even possible.

onox:
> We did this on purpose because 4X-CHV is to be the default?

Yes, we made sure that the pngs that the ac files point to are the ones from the 4X-CHV.

onox:
> I need to know specifically what we need done. Can that even be achieved with the shaders? Those light are procedural, so unless an "artist" wants to submit some math to create the style of lights we want, I'm not sure how to achieve this?
Unless Thorsten is willing to give us a hook to the math that is generating the light prior to the environment shader influence, I'm not sure that is even possible.
Or the argument on what the light should look like, which I already attempted to convey at the start, from the perspective of those that don't like the spikes

Being able to define how much of that star effect we want to see.

wlbragg:
This is going to be difficult today, GitHub has suddenly lost the ability to notify me of any notifications. :angry: 

>Being able to define how much of that star effect we want to see.

Sound like a simple request to me!

onox:
@wlbragg It has indeed some issues at the moment: https://status.github.com/

onox:
@wlbragg @HHS81 I tried moving the quad in the .ac file, but it dims the light. So it's not a useable solution.

wlbragg:
I just fixed https://github.com/Juanvvc/c172p-detailed/issues/624 and want to push the blend and ac to this PR, OK?

HHS81:
@onox: hmmm....works here perfect, otherwise I had never been able to create the beacon light.
Shift it only a liitle bit, only a few centimetres. It its true that with more it will dim the light, but with just a few centimetres it is hardly noticeable.

HHS81:
@onox: to prevent the dimming, use the translate-animation instead! 

```
<animation>
  <type>scale</type>
  <x-offset>0.5</x-offset>
  <y-offset>0.5</y-offset>
  <z-offset>0.5</z-offset>
 </animation>
 
<animation>
<type>translate</type>
  <object-name>beacon</object-name>
<offset-m>0.1</offset-m>
	<axis>
		<x>1</x>
		<y>0</y>
		<z>0</z>
	</axis>
</animation>
```

please notice that scale animation multiplies with the translate-animation. So here: 
scale-x-offset * translate-offset-m = in-sim-offset-m
0.5 * 0.1m = 0.05m = 5cm offset in sim

No dimming

onox:
We haven't modified the beacon light in this issue, but it is something to keep in mind.

@wlbragg If you want to merge this PR, go ahead, it fixes the visibility of the nav lights on the wings. The shift-quad-on-x-axis may work for the beacon light, but I guess it isn't going to be sufficient to prevent any wing tips sticking through the quad. Maybe Thorsten has better ideas.

onox:
@HHS81 Thanks for the XML, I'll look at it in the future. Why use scale + translation instead of just translation?

HHS81:
@onox:  Damn: 
works on the directional light like the taillight, but not on the non-directional like the beacon. I have deleted my posting. 

You can't scale the quads in Blender, you have to use the scale-animation, says Thorsten. You can use of course just the translation. I used a scale-animation since the lights were too big for my taste.

PR #689, opened 2016-02-07
Fix gauges and reconstructed ADF

tigert:
So there just was something mysteriously wrong with the ADF svg. I basically removed everything but the gauge face and compass ring (as they are identical) from the VOR gauge, and then cut and pasted the moving elements from the ADF, and saved it over the ADF.svg and this seemed to have fixed the coordinate issue in the needle rotation. No idea what the issue actually was, likely some weird size in some element, that I just couldn't find when I tried to look.

Also removed the watermarks and the blue from EGT.


onox:
Changes are good! :+1: 

One last thing: could you perhaps change the RPM instrument to have the green arc start at 1800 RPM instead of 5. Then it is equal to the one in the cockpit.

onox:
Or to 2100 RPM.

gilbertohasnofb:
Why 2100 RPM? The green arc extends from 1800 to 2700 RPMI think.

onox:
I'm looking at some photo's on google. At some they start at 2100/2200 RPM :worried:

gilbertohasnofb:
Okay, so in that case our 3D cockpit instrument would also be wrong. Let's research this.

onox:
Does the POH say anything about it?

gilbertohasnofb:
I don't think the POH would describe the look of a gauge, it will for sure talk about normal engine operation, which not necessarily reflects how a gauge looks like.

gilbertohasnofb:
OMG, every single shot of a 172 model P has the green arc starting at 2200, some at 2100 :scream: 

And by the way, Carenado's 172 model N for FSX and X-Plane uses the same gauge we have in the web panel:

onox:
See slide 8: http://www.slideshare.net/BajajSundar/cessna-c172r

It is about the R version, which came after the P version. It says that in the R version, the RPM was reduced from 2700 to 2400 RPM. So our upper limit of 2700 is correct. The photo shows the lower limit at 1900 RPM. Also, note that the instrument goes to 3500 RPM. So the web panel is correct in that regard, but the 3D model is not.

gilbertohasnofb:
I don't know, onox, most images I am finding start either at 2100 or 2200 until 2700, not up to 2400. As for the range, indeed the range in these images is much more similar as in the web panel:

onox:
> I don't know, onox, most images I am finding start either at 2100 or 2200 until 2700, not up to 2400. As for the range

As the slides explain: the RPM was reduced in the **R** version. So in the P version the RPM is still 2700.

onox:
Btw, the HD panel (shift+p) shows 2100 RPM as lower limit.

gilbertohasnofb:
Oh, I see! :smile: 

Look at this link, please: http://www.airliners.net/search/photo.search?aircraft_genericsearch=%3D%28%22cessna+172+skyhawk%2Fcutlass%2Fhawk+xp+%28t-41%2Fmescalero%29%22+OR+%22reims+f172+skyhawk%2Freims+rocket%2Fhawk+xp%22%29&airlinesearch=&countrysearch=&specialsearch=&keywords=172P+cockpit&sort_order=photo_id+desc&page_limit=15&daterange=&range=&thumbnails=&engine_version=6.0

All the images are from the cockpit of a 172P. I checked the first six and all green arcs range from 2100-2700, and the gauges range from 300 to 3500.

So should I open a ticket for this and fix the texture in this way?

onox:
Yes :smile:

onox:
@tigert The range is OK, but the green arc lower limit should start at 2100 RPM instead of 500.

gilbertohasnofb:
@onox done, see: https://github.com/Juanvvc/c172p-detailed/issues/690

onox:
@gilbertohasnofb It looks like the VS instrument could be improved as well (no lines after "10") and the numbers "5" and "15" needs to be bigger.

gilbertohasnofb:
:+1:

gilbertohasnofb:
Appended this info to that same issue, so one PR will close it.

onox:
:+1:

gilbertohasnofb:
I will also improve the artificial horizon, as the black model looks a bit different than ours in those images. Can you spot any other instrument texture that needs to be reworked?

onox:
They could be made higher resolution. They look lowres.

gilbertohasnofb:
Which ones specifically?

onox:
All of them :stuck_out_tongue_closed_eyes:

onox:
Let's continue in the new issue. This is going offtopic :stuck_out_tongue_winking_eye:

onox:
@tigert So just the green arc needs to be changed and then we can merge this PR and then open a new PR to merge `bug-339`. Then we have your web panel in the 2016.1 release :smile:

tigert:
The rpm gauge varies between engine model. The lower limit of green arc is not really that important in practice, but of course we should be accurate. The upper limits are more important in practice, redline and possible yellow arcs (avoid long operating on yellow arc for long time, possibly due to some vibration or such issues in the engine and airframe combination) though c172 doesnt have any yellow there. The POH should give the rpm instrument markings, I can try to look later today. 

The arcs are circles in the svg, so the start and end points are easy to tweak though.

onox:
Will fix start end point in `bug-339` directly.

PR #709, opened 2016-02-10
Bug 706: New pitot tube model undone in model

wlbragg:
I didn't test this yet, but the new model should be there.


gilbertohasnofb:
I will check it now.

wlbragg:
Maybe we should look at re-positioning this cover, both horizontal (farther back over the tube) and vertical angle hanging strait to ground.
pitot

gilbertohasnofb:
Yes, the pitot is there, but I noticed something that we may as well fix right now:

fgfs-screen-001

Now look at the texture, notice how the bottom half is only red: that's the back part! 

rbf

gilbertohasnofb:
> Maybe we should look at re-positioning this cover, both horizontal (farther back over the tube) and vertical angle hanging strait to ground.

Further back I agree, but different angle would be a problem, unless you modify the model: notice how the cylindrical part of the cover (in which the pitot tube enters) is in the same angle as the tube itself. If we make the hanging part vertical, then it won't match the pitot tube angle.

wlbragg:
pitot1
It would also need the tube circumference. enlarged a bit to fit on the tube better

gilbertohasnofb:
> It would also need the tube circumference. enlarged a bit to fit on the tube better

IMO it looks good! I think it has to be tight like this, right?

wlbragg:
How flexible are those covers, is it like a material that should bend right at the tube?

>IMO it looks good! I think it has to be tight like this, right?

Yes but if you move it back I bet the tube will shoe through (split it).

wlbragg:
Never mind, it is actually fully engaged with the tube already.

wlbragg:
It can't go back any farther.

gilbertohasnofb:
I think the model we use as source was plastic, see: 



Also notice how it doesn't go very much far and how it doesn't necessarily align with the vertical axis.

gilbertohasnofb:
I'd say to simply correct the text issue and we are good to go!

wlbragg:
Right!

gilbertohasnofb:
Thanks!

wlbragg:
I'm not positive but I think this requires another plane on the back side, that is how we got the old one double sided, but let me see.

wlbragg:
I'm not messing with it now, but it almost looks like the proportion of the cover is off slightly, (too small), as compared to your picture. That is why my first instinct was that it needed to be moved back.

wlbragg:
What's really funny is I noticed that back side when it was first added, but figured no one cared. I forgot about our quality control personnel, ha!

gilbertohasnofb:
> I'm not messing with it now, but it almost looks like the proportion of the cover is off slightly, (too small), as compared to your picture. That is why my first instinct was that it needed to be moved back.

Yes, the main label looks a bit smallish to me as well when compared to the image. But that's fine with me as it's a minute detail of something that isn't really the aircraft itself (there are tons of different pitot cover models out there I think)

> What's really funny is I noticed that back side when it was first added, but figured no one cared. I forgot about our quality control personnel, ha!

Hehehe

wlbragg:
I was wrong, if it is double sided it looks like you can map both sides independent of each other and just mirror the back side.
I'll push it shortly.

gilbertohasnofb:
:+1:

wlbragg:
OK

gilbertohasnofb:
I will test

gilbertohasnofb:
Is there anything else to fix or can this be merged after testing?

wlbragg:
It can be merged.

gilbertohasnofb:
:+1: just give me a second to try it in the sim!

gilbertohasnofb:
Perfect, thanks a lot for all the quick fixes!

gilbertohasnofb:
I guess that's it, right? Nothing else needs to be fix urgently before the release, I suppose. @onox I think you you can update the release branch and tell Stuart that we are now ready for the merge!

wlbragg:
Thanks for finding this stuff before it was too late!

gilbertohasnofb:
The credit goes to @onox

onox:
Commits backported to `release/2016.1`.

PR #726, opened 2016-03-02
Bug 723: Use indicated-speed-kt which uses an almost perfect pitot, but can fail

### LINKED ISSUE #723: Pitot cover should "break" pitot system, not ASI.
Hey, nice new stuff again! :)

I noticed you added the pitot cover and correctly having an effect on the airspeed indicator when you forget to remove the cover. However I found two issues when testing this:

  * Forgetting the pitot cover on should not break the airspeed indicator, but the pitot system instead: `/systems/pitot/serviceable`
  * The worse issue seems to be that the airspeed indicator always works no matter what. If I read the code correctly, I think it pulls the airspeed value directly from the FDM's velocity variable, not via the pitot/static systems modelling:

   ```xml
 <animation>
  <type>rotate</type>
  <object-name>Needle</object-name>
  <property>fdm/jsbsim/velocities/vias-kts</property>
   ```

I am a bit confused in how our indicated-airspeed.xml plays into this, as I am not very familiar how all this comes together. But it might be that jsbsim and / or flightgear model pitot failures already?

onox:
Fixes #723 

`/fdm/jsbsim/velocities/vc-kts` does not react to failure of the pitot
system. Use `/instrumentation/airspeed-indicator/indicated-speed-kt` which
does depend on the pitot system. Note that the indicated airspeed system
assumes a perfect pitot, thus it practically is a calibrated airspeed
indicator.

There's a small difference between `vc-kts` and `indicated-speed-kt` (see https://github.com/Juanvvc/c172p-detailed/issues/723#issuecomment-191188420) but it is quite small (< 1 kt usually).

### DIFF
--- a/Systems/indicated-airspeed.xml
+++ b/Systems/indicated-airspeed.xml
@@ -42,12 +42,20 @@
     output shown on the ASI.
 -->
 
+    <property>/instrumentation/airspeed-indicator/indicated-speed-kt</property>
+
     <channel name="indicated airspeed">
 
         <fcs_function name="systems/asi/indicated-airspeed">
+            <output>velocities/vias-kts</output>
             <function>
                 <table>
-                    <independentVar lookup="row">velocities/vc-kts</independentVar>
+                    <!-- The indicated-speed-kt property reacts to pitot
+                         failure, but it assumes the pitot is perfect.
+                         Therefore, it practically reacts as a calibrated
+                         airspeed indicator (KCAS).
+                    -->
+                    <independentVar lookup="row">/instrumentation/airspeed-indicator/indicated-speed-kt</independentVar>
                     <independentVar lookup="column">fcs/flap-pos-deg</independentVar>
                     <tableData>
                               0     10    30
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
My AI seem broken in this branch.

onox:
No problems with the AI here. Also, the AI does not depend on the pitot tube, but on the vacuum system.

gilbertohasnofb:
Ok, so I will try to test it again later, but I suspect I know the cause: if I had the pitot tube cover on and its state was saved before, then the property that makes the AI serviceable was set to false. Probably if I delete the c172p save state file in .fgfs the problem will go away. I will let you know later and then merge this.

gilbertohasnofb:
Indeed by some reason the property `/instrumentation/airspeed-indicator/serviceable` was set to `false`. Deleting my c172p.xml from the folder`.fgfs/aircraft-data` didn't do the trick, but manually setting that property to true did. After that, I restarted the sim and tested putting the cover on and off and it all works as expected.

Issue #704, opened 2016-02-10
ceiling looks black


gilbertohasnofb:
Ceiling looks black while the panel with the speaker doesn't. I don't think the ceiling should be pitch black during day at any circumstances.

![fgfs-screen-001](https://cloud.githubusercontent.com/assets/5700795/12934924/568bed0a-cf92-11e5-8e42-d4adfb4c85b9.png)

gilbertohasnofb:
@wlbragg sorry but suddenly I am discovering bug after bug :stuck_out_tongue_winking_eye:

wlbragg:
It's been there all along, it is the irradiance map. I tried to adjust the strength but it doesn't seem to be working. This one will take some time to fix and may require Thorsten to do some shader work.

Without irradiance map
![norad](https://cloud.githubusercontent.com/assets/11069680/12935390/fdc757fe-cf5a-11e5-83ab-686da482d655.jpg)

With irradiance map
![withrad](https://cloud.githubusercontent.com/assets/11069680/12935394/0451e4ae-cf5b-11e5-8534-4ff7d1f79397.jpg)

gilbertohasnofb:
I think I never looked up after the irradiance map was added :laughing: No problems then, let's add a later milestone to this issue.

wlbragg:
:-1:

legoboyvdlp:
Who's looking at the cieling anyway :laughing:

gilbertohasnofb:
Well, the guy who created that beautiful texture for the speaker, I suppose! :stuck_out_tongue_winking_eye:

wlbragg:
FYI: This is done with the default (built in), however we can do this with a mask and choose where we want it applied and how strong. But we will need someone good at making masks.

wlbragg:
I'm not sure why this wasn't immediately fixed when discovered. If I remember correctly it appeared the strength setting wasn't working before. Well it is now. The simple fix is to change the strength setting.
I'm pushing the fix now.

gilbertohasnofb:
Closed via https://github.com/Juanvvc/c172p-detailed/pull/730

PR #730, opened 2016-03-15
Bug 704: ceiling looks black

wlbragg:
Lets try this again.

Should close issue #704

### DIFF
--- a/Models/Effects/interior/c172p-interior-radiance.eff
+++ b/Models/Effects/interior/c172p-interior-radiance.eff
@@ -5,6 +5,6 @@
     <inherits-from>Effects/interior/c172p-interior</inherits-from>
     <parameters>
         <irradiance-map-type type="int">2</irradiance-map-type>
-        <irradiance-map-strength type="float">1</irradiance-map-strength>
+        <irradiance-map-strength type="float">.3</irradiance-map-strength>
     </parameters>
 </PropertyList>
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
I will test this now

gilbertohasnofb:
Hmm, it does look much better but the speaker in the ceiling still looks completely different. Can't it be affected by the irradiance map as well as the ceiling, so they match colours?

fgfs-screen-001

wlbragg:
Let me see what adding it to the effect does to it.

wlbragg:
I think that fixed it. We used to have the seats and the floor added before we realized it was too dark. I tried them before I pushed this and they worked much better than before, but with the textures they were already using (already pretty grainy) it didn't seem to enhance them much (lighting wise) so I left them out. We can add any of the interior objects we want to include the effect. In the shuttle all the switches on the panels have the effect applied (although it may be slightly different as its base is ra-irradiance).

gilbertohasnofb:
I think this looks better, thanks!

> We can add any of the interior objects we want to include the effect. In the shuttle all the switches on the panels have the effect applied (although it may be slightly different as its base is ra-irradiance).

Do you think adding the switches or other objects would enhance the visuals in some way? I am still a bit confused about the irradiance map and what it exactly does, so I can't visualize what applying it to the switches would cause

wlbragg:
All it is doing is adding shadow (or taking away light) from objects that are not calculated to be in the light rays supposedly being cast depending on the mode you set. Ours is using the mode that says only object faces that are perpendicular to the light coming through the windows get the full light. Any parallel objects are going to have light removed. So in effect it is adding a shading to give depth.
I guess I should try a few switches and knobs and see what it does.

gilbertohasnofb:
Sounds good to me. If you make some tests like that, please share some screenshots, I'm curious how the cockpit would look like.

Issue #634, opened 2015-12-30
improve flaps lever


gilbertohasnofb:
I think our flap lever is not the most common model found in the P model. Most of the images I found are of this model:

![](http://www.flylpas.com/N5314K%20panel.JPG)

Note how it has only two colours (light blue and white) as opposed to ours with three, also note that the numbers inside these colours show the max speed for that flap extension and not the angle. 

I will work on the new texture, but probably the model will have to be slightly adapted. Any volunteers to fix it? (should be some really simple adjustments)

gilbertohasnofb:
Okay, done. Pushed to a new branch `bug-634`. This is the new texture:

![flaps](https://cloud.githubusercontent.com/assets/5700795/12055548/825847d2-af2e-11e5-972c-cf8053ee4a9b.png)

And while the old flap had two textures, one for the panel one for the lever itself, I have everything in a single texture so the lever should be re-UV mapped.

So could anyone give me a hand with the 3D model?

gilbertohasnofb:
Also note that the flap lever and the flap position indicator (the tiny "T" that indicates the current flap position and which is electrically controlled) might need to be a bit repositioned, and also the xml for the flaps will need to be tweaked for the correct positions for each notch.

gilbertohasnofb:
@wlbragg when you have the chance, do you think you could have a look at this? I can fix the xml myself I think, but for the 3D model update I'd need help.

wlbragg:
Yes, I'll look at it.

gilbertohasnofb:
Thank you so much.

wlbragg:
I noticed the ignition switch lighting looked messed up so I fixed it while I was at it.
![ignit](https://cloud.githubusercontent.com/assets/11069680/12214814/b40c7316-b668-11e5-925d-18b52ce4a3b3.jpg)

Unrelated I also moved all the development files to dev folder, I got tired of them cluttering up the models folder.

wlbragg:
![flaps](https://cloud.githubusercontent.com/assets/11069680/12214850/ebe4a94c-b669-11e5-9b61-11dd59a0fce8.jpg)

gilbertohasnofb:
Thank you so much!

> I noticed the ignition switch lighting looked messed up so I fixed it while I was at it.

Looks much better, that switch always had some strange problems (first the green illumination, now this). Now it looks finally fixed :smile: 

[...]

PR #731, opened 2016-03-15
Add irradiance effect to multiple interior objects.

wlbragg:
Was bug-704 PR #730.

Sorry I pushed this before I realized you merged #730.

The following objects were added to irradiance effect, in case you want to look at each one.
~~~xml
        <object-name>TrimWheel</object-name>
        <object-name>ElevatorTrimPos</object-name>
        <object-name>FuelSelectorFace</object-name>
        <object-name>FuelSelectorLeft</object-name>
        <object-name>FuelSelectorRight</object-name>
        <object-name>FuelSelectorBoth</object-name>
        <object-name>FuelSelectorOff</object-name>
        <object-name>mike_mount</object-name>
        <object-name>Pedestal</object-name>
        <object-name>CabinAirMounts</object-name>
        <object-name>CabinAir</object-name>
        <object-name>CabinHeat</object-name>
        <object-name>BeaconSwitch</object-name>
        <object-name>CarbHeat</object-name>
        <object-name>CarbHeatMount</object-name>
        <object-name>GenSwitch</object-name>
        <object-name>LandingLightSwitch</object-name>
        <object-name>MasterSwitch</object-name>
        <object-name>Mixture</object-name>
        <object-name>MixtureMount</object-name>
        <object-name>NavLightSwitch</object-name>
        <object-name>PitotHeatSwitch</object-name>
        <object-name>PrimerLeverMount</object-name>
        <object-name>StrobeSwitch</object-name>
        <object-name>TaxiLightSwitch</object-name>
        <object-name>Throttle</object-name>
        <object-name>ThrottleMount</object-name>
        <object-name>mike</object-name>
        <object-name>mike_chord</object-name>
        <object-name>doorhandleint_right</object-name>
        <object-name>doorhandleint_left</object-name>
        <object-name>ParkingBrake</object-name>
        <object-name>PrimerLever</object-name>
        <object-name>windlocket_right</object-name>
        <object-name>windlocket_left</object-name>
        <object-name>windlocketbak_right</object-name>
        <object-name>windlocketbak_left</object-name>
~~~

also
objects in 
yokes.xml
~~~xml
        <object-name>Yoke</object-name>
        <object-name>YokeMount</object-name>
~~~

breakers.xml
~~~xml
        <object-name>autopilot_breaker</object-name>
	  <object-name>radio5_breaker</object-name>
	  <object-name>radio4_breaker</object-name>
	  <object-name>radio3_breaker</object-name>
	  <object-name>radio2_breaker</object-name>
	  <object-name>radio1_breaker</object-name>
	  <object-name>turn-coordinator_breaker</object-name>
	  <object-name>strobe_breaker</object-name>
	  <object-name>bcnlt_breaker</object-name>
	  <object-name>landing_breaker</object-name>
	  <object-name>navlt_breaker</object-name>
	  <object-name>intlt_breaker</object-name>
	  <object-name>instr_breaker</object-name>
	  <object-name>pitot-heat_breaker</object-name>
	  <object-name>master_breaker</object-name>
      <object-name>flaps_breaker</object-name>
      <object-name>AvionicsSwitch</object-name>
      <object-name>swithtype2_base</object-name>
      <object-name>Breakers</object-name>
~~~

peddals.xml
~~~xml
        <object-name>RightPedal</object-name>
        <object-name>LeftPedal</object-name>
~~~

flaps.xml
~~~xml
        <object-name>FlapSwitch</object-name>
        <object-name>FlapPos</object-name>
        <object-name>rect</object-name>
~~~

mag-compass.xml
~~~xml
  <object-name>Ring</object-name>
   <object-name>Casing</object-name>
~~~


gilbertohasnofb:
> Sorry I pushed this before I realized you merged #730.

No worries about it.

The results look really good, but I think we should try adding the instruments, radios and magneto switch as well to the effect, since they look very detached from the rest of the panel, see what you think:

fgfs-screen-001

wlbragg:
So your liking the effect?

This is a pretty extensive change, lots of objects and files and I had to reconfigure the effects links as I was getting some path errors.
What was done was the easier stuff. The rest requires going into each of their respective files and setting the effect. So I need to make sure I get everything and that it doesn't mess up anything. I guess we need to add the panel face also?
Let me keep working on this and lets get feedback from others (onox) while were at it.

gilbertohasnofb:
> So your liking the effect?

I think I like it, yes. It's kind of difficult for me to compare it while I am in the sim, as your mind gets used to what you are seeing. I will make some screenshots of this branch vs the master branch and compare them to see the difference in details (not today, unfortunately, as it's already late here).

> Let me keep working on this and lets get feedback from others (@onox) while were at it.

Sounds good, I'd appreciate hearing the opinions of others on this as well!

wlbragg:
> It's kind of difficult for me to compare

Same problem I had, I had a snapshot prior and after and accidentally deleted it. But from what I saw I liked it. But we definitely need to make sure everything is accounted for or it will stick out like a sore thumb, apparently.
Have a good night.

gilbertohasnofb:
I will try to compare them then, and let's see how it fares. But so far I also liked what I have seen here!

> Have a good night.

Thank you, you as well!

wlbragg:
OK, I think I have everything included that needs to be. We need to check this carefully.

wlbragg:
irr1
irr2

onox:
Does the flash light and shadow still work?

I see `AirVent_left` and `AirVent_right` still use the flashlight effect. Can they use the irradiance effect instead? Does it look good? The same for some objects in `Models/Effects/pontoon/pontoon.xml`. You can delete the flashlight effect if you can move those objects to the irradiance effect.

onox:
`wing_1L`, `wing_1R`, and `Firewall` use the old interior effect. Can they use the irradiance effect and still look good?

wlbragg:
Back up, I just realized something. All the objects that are in a separate calculated space (I don't know the words for it), the instruments and any switches that are in their own file, separate from the c172.ac, cannot have the shadow on them and that is what I just did because I basically combined all three effects into one (shadow, irradiance, flashlight and whatever else is included in model-interior).
In other words the bad shadow effect has been applied to the instruments again, oops!
So what I need to do is still have two different effect bases.
Instruments and switches not in c172.xml (or actually c172p.ac) need everything but shadow (one effect definition). Everything else that is in c172.ac gets everything, including shadow (a second effects definition).

I checked the wing_1l and 1r, Firewall and air vents. Yes we can add them with no ill effects. The firewall however I don't think is even visible so really doesn't need to have any effect applied (some might verify that it isn't visible).

Does that make sense?

Someone may want to verify that the bad shadow effect on the instruments is still bad now that it is combined with the irradiance effect (remember the shadow on the instruments didn't have the same global position as the shadow on the panels and such). It still should be off even with the irradiance through in but may be more subdued.

wlbragg:
OK, I finished adding all three effects to all the interior objects. I have to say that even though the shadow may not line up perfectly on the instruments, it still works because your eyes aren't looking for that. I think it look good. But it isn't a problem to remove the shadow from all objects that are in their own file. I'll push what I have and you all can be the judge.
Yes, the flashlight is still there, in fact I found a few things that didn't have it applied to yet.

wlbragg:
Don't merge this yet even if it's approved. I have a couple files not being used and also want to rename one, but am waiting to see if I am going to have to separate out the objects that are in their own files to a different effect config.

gilbertohasnofb:
I think this is looking really good. The effect basically makes the panel quite darker than before, particularly if not directly illuminated by sunlight, am I right? I think this creates a nice effect as in RL the amount of illumination from the environment during daytime is much higher than the cockpit, so it looks more natural to me. 

The instrument shadows are looking fine to me, and the flashlight works well here. But I noticed that the instrument needles are sometimes too bright, and are inconsistently having shadow applied to them (sometimes the whole instrument is dark but a needle is fully white). @wlbragg can you see if you can reproduce what I mean?

wlbragg:
I agree with everything you said, I even reduced the effect one more point from .3 to .2  to lighten it up a bit more.
>But I noticed that the instrument needles are sometimes too bright, and are inconsistently having shadow applied to them 

That is the problem, the shadow is calculated in the space the object resides in. Anything that has translate or maybe even offset applied to it will have the shadow calculated differently.

If necessary, we could pull any thing that doesn't look right, even just a needle but not the rest of the instrument out of the shadow effect and just apply the flashlight and irradiance. I just need to have the objects we don't want included identified. That is why I haven't deleted the extra effect file yet because I have a feeling we will need to remove a few objects from having the shadow. But I think as it is before any of these changes, we could have applied the shadow to at least some of the instrument faces without ill effect.

gilbertohasnofb:
> That is the problem, the shadow is calculated in the space the object resides in. Anything that has translate or maybe even offset applied to it will have the shadow calculated differently.

Maybe a solution is to have all needles set to a fixed amount of illumination (that is, disabling the effect on them) and we can try to find a value that looks good regardless of the cockpit being in the shadow or direct light. What do you think?

wlbragg:
Is it just the needles you notice not looking good?

gilbertohasnofb:
The needles is what was the most obvious, but I have seen some strange shadows in the AI. I guess the more an instrument moves (AI, HI), the more obvious these artifacts will be, right?

wlbragg:
Yes and no, I noticed the oil temp or oil pressure face (I think it was) was in light while a shadow was crossing the entire area once.

gilbertohasnofb:
Hmm.... that's tricky. The fixed illumination values for the needles wouldn't be bad because one would hardly notice it, but the instruments would appear brighter and darker in relation to the rest of the panel if they would be fixed, and that might not look good.

wlbragg:
I'm back to the opinion we may have to remove the shadow from all the gauges (like before) but still apply the irradiance effect. It's just too iffy on whether we will get an appropriate shadow and when.

wlbragg:
I've left it set up so I can do the two effect config relatively easy if you want to see it (no shadows on the instruments like it was before).

gilbertohasnofb:
> I've left it set up so I can do the two effect config relatively easy if you want to see it (no shadows on the instruments like it was before).

Ok, do you think you could do that? I guess that could do the trick, and the irradiance effect can take care of the overall illumination level of the instruments and needles. Let's try that :smile:

wlbragg:
OK, try that.

gilbertohasnofb:
Thanks! I just tested it and it looks much better, it's very consistent and the amount of illumination on the instruments is taken care by the irradiance effect, so it looks very natural to me:

fg-0

fg-1

fg-2

The only thing that still stands out to me is that the AI looks ultra shiny under direct light. I changed the hue of the AI when I redid all instrument textures, and since then I have been noticing that. Could that be taken care as well? I don't know what it is, but every time the sun is behind me the AI shines as if instrument illumination would be on!

wlbragg:
OK, let me see if I can figure out what's wrong with it.

gilbertohasnofb:
:+1: Thanks :smile:

wlbragg:
Try that and see if it made any difference, I can't test it now, I've got to leave for a bit. I'll check back in later and see where were at.

gilbertohasnofb:
> Try that and see if it made any difference

Not really. See:

fg

Compare it with the colour of the AI in its texture:

ai2

It's strange that the texture is just a little bit more desaturated than the previous one, but the brightness sticks out so much more!

To test it yourself, I suggest setting the weather to Core High Pressure Region, which has the strongest sunlight and therefore makes it much more obvious.

onox:
@wlbragg You made the c172p project depend on the SpaceShuttle in https://github.com/Juanvvc/c172p-detailed/pull/731/commits/901136840c728a79162335dabc65301e86ed6a82

onox:
In https://github.com/Juanvvc/c172p-detailed/pull/731/commits/6296e802554ed068e990e8fa1154f63e119c92d5 you updated ` dev/blender/c172p-pontoon.blend
`, but shouldn't you export to the .ac file as well?

wlbragg:
Oops, I'll fix the Shuttle path reference when I get back on my computer, Good thing you're paying attention. I'm away at the moment. I didn't do anything to the pontoon AC file. I just fixed the texture paths in the blend because I noticed they hadn't been fixed since we reorganized.

onox:
Okay.

gilbertohasnofb:
oops, I wanted to see what that (n) was about and apparently I gave someone a thumbs down! Sorry :smile_cat:

onox:
There's only 1 thumbs down (my comment) and I did that :stuck_out_tongue:

gilbertohasnofb:
I can see two thumbs down on that comment!

onox:
You're right, I had to F5.

wlbragg:
OK, try this. I'm really not sure how to completely change the emissive value of the AI, I also noticed the flap lever started out noticeably white (or light) so I toned it down also. Speaking of the flap switch, shouldn't that be able to be changed with the mouse? I don't think it is hot.

gilbertohasnofb:
> OK, try this. I'm really not sure how to completely change the emissive value of the AI. I changed the spec and diff and hardness values in the blend and also put it back in the effect that includes the shadow (but only the two objects that have the blue face on them "roll" and "pitch"). I also noticed the flap lever started out noticeably white (or light) so I toned it down also.

Hmm, I really don't like the new look. The AI has a complete different tone than the texture, and it looks too dark. The flaps are also looking too dark. They also don't seem to change at all depending on the light, are they still under the irradiance effect? Compare the blue colours:

ai2

fg

And note that the screenshot above was taken with the sun fully behind, so that should be the maximal of light.

>  Speaking of the flap switch, shouldn't that be able to be changed with the mouse? I don't think it is hot.

You are right, I didn't notice it either. I think this happened when we worked on the new flap switch on issue #634. I will open a new issue for it.

wlbragg:
OK, we'll try again.

wlbragg:
The flap switch is back to where it was with a "slight" change to the spec color and I added the shadow effect to it. I think it looks really good now and the shadow effect fits the scene.
The  AI is reverted to what it was with the exception of leaving the shadow effect applied. See it that helps.

onox:
@wlbragg Shouldn't it be:

```xml
<positive-x>Aircraft/c172p/Models/Effects/interior/shadow_px.png</positive-x>
<negative-x>Aircraft/c172p/Models/Effects/interior/shadow_nx.png</negative-x>
<positive-y>Aircraft/c172p/Models/Effects/interior/shadow_py.png</positive-y>
<negative-y>Aircraft/c172p/Models/Effects/interior/shadow_ny.png</negative-y>
<positive-z>Aircraft/c172p/Models/Effects/interior/shadow_pz.png</positive-z>
<negative-z>Aircraft/c172p/Models/Effects/interior/shadow_nz.png</negative-z>
```

as it used to be?

wlbragg:
No, the only time you use that is if you want shadows. We just want the irradiance and the flashlight effect applied. But the overall model-interior effect was written to default to a cube that has blue sky and green grass on it. So if you don't want shades of green and blue at times applied to the object you need to pseudo create a white cube. (it's referenced in ALS technical notes for several effects)

wlbragg:
I changed just the color instead of the color and intensity of the AI roll and pitch objects. I left the shadow applied. Here is the results at full morning light coming through the back window.
Before
ai1
After
ai2
With shadow showing on it.
ai

wlbragg:
Note: the flap selector switch has an almost horizontal face on it so as far as the light coming through the window it should be the darker side of the object, thus it is suppose to be darker. Any object faces that are perpendicular to the window with the light coming through will have the brightest faces.

wlbragg:
A couple things here I want to ask about. The hobbs panel has a lighting param that I don't think we have tied in yet. It is /controls/lighting/instrument-lights.

Also there is an effect called ra-irradiant and it is suppose to simulate the amount of panel light that lights up the cockpit. I've been experimenting with it and am just about ready to push it for your testing and comments.

wlbragg:
The only thing I still would like to change is that even when the strength is set to 0 there appears to be some residual ambience lighting. I asked Thorsten about it and am awaiting his answer.

gilbertohasnofb:
@wlbragg Sorry for my late reply, I am travelling since Thursday and I hadn't find the time to test this. So:

> I changed just the color instead of the color and intensity of the AI roll and pitch objects. I left the shadow applied. Here is the results at full morning light coming through the back window.
Before

Sorry but I still find the AI too dark now :cry: Also, I don't think it's a good idea to change colours anywhere else than directly in the textures themselves. And personally, I think that hue of blue I used in the AI texture files is really good, it's just a matter of getting a consistent result in the sim. 

As for the shadows in the instruments, I absolutely love the effect, but it's still buggy here: the shadows in the panel itself are obviously fine but it seems to be broken in the AI and other moving instruments. If we can't solve it, I am afraid that I'd vote to simple disable it on these instruments, as it actually breaks the immersion IMO.

The flap lever looks fine, maybe a tad too dark. But the background of the flaps lever is also having the shadow bug, which is strange since it never moves its position!

> Turn panel lighting all the way up and change view to look at pilot side door handle then turn off pwr to bat and alt to get a good demonstration.

I don't understand what it's supposed to happen. I did that and I see nothing unusual.

wlbragg:
>Sorry for my late reply

No worries.

The color I changed was the specular light reflection color, I honestly don't even know what that actually does, I'll put it back. The shadow on any instrument that doesn't reside in the cockpit.ac space (so to speak) won't ever have the correct shadow as compared to the main shadow so I guess until there is a fix for it we have to leave it of (the AI and the flaps). So I guess if we aren't going to try to change the emissive value of the reflecting light in the blend then the only other alternative is to change the texture (as far as I know). Anyway I reverted the AI and the flaps back to not include the shadow and reverted the AI specular.

>I don't understand what it's supposed to happen. I did that and I see nothing unusual.

Your suppose to do this in the dark (at night) this will show you what the residual ambience lighting does to the interior cockpit objects (light emitting from the instruments). Think about what light from the instrumentation should do to the inside of the cockpit, it should light it up somewhat.

gilbertohasnofb:
@wlbragg Thanks for reverting the AI and instrument shadows, I think that for now we should leave it like this indeed until we find a way of applying shadows to moving surfaces.

>>  I don't understand what it's supposed to happen. I did that and I see nothing unusual.
> 
> Your suppose to do this in the dark (at night) this will show you what the residual ambience lighting does to the interior cockpit objects (light emitting from the instruments). Think about what light from the instrumentation should do to the inside of the cockpit, it should light it up somewhat.

I am so sorry but I still don't get it. I tried exactly what you are saying but I don't know what to expect and I haven't seen anything unusual in the door handle nor any other objects around it. Maybe if you could add some picture it would make it easier to understand :smile:

wlbragg:
OK, I think part of it is because it is very subtle. So you seeing only a very slight change. But what you are seeing even before you turn on the lights, you wouldn't be seeing at all without the effect.
Also I still haven't been able to turn it off entirely when the console lights are all off so you aren't seeing the full effect from off to on. I'll take a couple of screen shots tomorrow and post them so you can see what I am talking about. It past my bed time.

gilbertohasnofb:
@wlbragg I think I finally understand it! I see some very soft red reflections on the door handle regardless of the instrument illumination, and they shouldn't be there if the lights are off, correct? The red reflections do intensify a bit if I turn the instrument illumination on, but they are always present. Is it that what you are talking about?

wlbragg:
Yes, I was just getting ready to post the images. It's subtle but can be made to be any strength we want. It should be black when the panel lighting is off. I am working on that still, I have the answer if it will work.

ra0

ra1

wlbragg:
In the final version if this. You should see the second image when the panel lights are on and the first picture when they are off (including the panel lights that you see on in that first picture, they would be off also).

wlbragg:
OK, now the residual ambience lighting is working like it's suppose to.

wlbragg:
I must be tired, two commits, two typos, sorry.
The way I set this up required the props to have values as I was using them in a function product solution. I suppose we could eliminate the props if we just used a hard value in the function.

gilbertohasnofb:
> OK, now the residual ambience lighting is working like it's suppose to.

It looks _really_ great, I love the effect! But I think you need to add a conditional for the battery charge: right now the amount of red light in the cabin depends only on the instrument lightning knob position, even if the battery is off.

wlbragg:
Oops, I got to thinking about that after I pushed it and couldn't remember if I accounted for the electrical path or not. You think the strength of the effect is OK then?

gilbertohasnofb:
Yes, the effect itself looks beautiful, I think it's just about the electrical path as you say. After that, do you still plan anything else for this PR or can we merge it?

One day if somehow we manage to add shadows to moving objects (if Thorsten implements it, for instance) we can open a new PR. I really would like to have these new effects you worked on in the master soon! :smile:

wlbragg:
OK, I'll finish it up today and then ,yes, you can merge.

gilbertohasnofb:
:+1:

wlbragg:
I just checked that and it is tied into the correct place, I think.
If you turn the bat off but still have the alt on and the engine is running then you get the power to operate the lights. with or without bat. Turn the alt off and then turn the bat off and you loose the instrument light and the effect. Unless I missed a condition I am not aware of, if so, give me the exact scenario that you think isn't working correctly.

gilbertohasnofb:
The exact scenario was: ALT OFF, BAT OFF, Instrument Lightning 100%. The whole aircraft should be dark, but the red reflection can be seen. Just play with the lightning knob while ALT and BAT are OFF and you'll see.

> If you turn the bat off but still have the alt on and the engine is running then you get the power to operate the lights

Power to operate lights is not the same as lights being on, right?

wlbragg:
I swear I just did that and it is working. If I have the instrument lighting all the way up and kill both the alt and the bat, they go off and so does the effect?

wlbragg:
Are you sure you have the most current?

gilbertohasnofb:
Yes, I am up-to-date. I will  try deleting the c172p saved states in `.fgfs` and try again, gimme a second.

gilbertohasnofb:
Hmm, the problem is still there. Please see these images (you may have to download them and see in full screen, one after the other, in order to be able to see the red reflective light in the cockpit):

fgfs-screen-001

fgfs-screen-002

On both, ALT and BAT are off and the only thing I did was to turn the panel lightning knob from 0% to 100%.

gilbertohasnofb:
@wlbragg if you can't reproduce this, I'd say we need someone else to test.

wlbragg:
I can barely see it in the images, that being said, if I can barely see it in these screen shots then I know it is fairly pronounced in the sim as I have he same issue with trying to save a screen shot with the same lighting.

wlbragg:
Check c172p\Systems\c172p-interior-lighting.xml and make sure you have 
~~~xml
<property>/controls/lighting/instruments-norm</property>
~~~
in all three functions.

wlbragg:
Wow, it suddenly appears to have quit working. OK, let me see if I can figure out what's going on.

wlbragg:
I'm using /controls/lighting/instruments-norm which I thought was a param that the batt and alt switch controlled (which id does sometimes because it was working). Now this time I loaded it it didn't change from 1 to 0 when I shut off both alt and batt?

gilbertohasnofb:
> I can barely see it in the images

if you open them in separate tabs than flick between them using Ctrl+Tab, you will be able to spot the difference.

> Check c172p\Systems\c172p-interior-lighting.xml and make sure you have 
> in all three functions.

Yep, I have.

> I'm using /controls/lighting/instruments-norm which I thought was a param that the batt and alt switch controlled (which it does sometimes because it was working). Now this time I loaded it it didn't change from 1 to 0 when I shut off both alt and batt?

It doesn't change to me either regardless of battery or alternator or any combination of circuit breakers, it only responds to the instrument lightning knob.

wlbragg:
That's why I am so confused, because it absolutely was changing on my system at one point in time. That's OK, I know what I need to do, it's a little more complex but I'll just have to do it. I think I can tie into an existing filter.

onox:
@wlbragg `/controls/lighting/instruments-norm` is the input property. Use `/sim/model/material/instruments/factor`.

wlbragg:
Oooh, oooh, thank you onox.

wlbragg:
You know what recall that param. I'll bet I was using it and somehow reverted back to just the instruments norm. I recall having a three param product at one point.

wlbragg:
That was it, I already had figured this out and somehow lost it. Thank you, I really didn't want to go down the route I was just about to do.

@gilbertohasnofb, thats why we need to keep onox happy, so he will stick around and keep us honest! :smile:

gilbertohasnofb:
> thats why we need to keep onox happy, so he will stick around and keep us honest!

Oh, absolutely! :smile: :+1:

@wlbragg so once you have this fixed, I can test again and merge this PR!

wlbragg:
Yes

wlbragg:
I'm testing it now!

wlbragg:
OK, that should do it!

gilbertohasnofb:
Ok, I will test it right away. Can I merge it immediately if everything looks good?

onox:
@wlbragg One more thing...

wlbragg:
Yes?

onox:
`Systems/c172p-interior-lighting.xml` has an odd system name and channel name.

wlbragg:
Thank you, I wasn't thrilled about that either, I'm open for suggestions!

onox:
And you might as well add `Copyright (c) 2015 Wayne Bragg` to the top of the license text.

gilbertohasnofb:
Ok, it is working now, but I have a small suggestion: at 70% of panel illumination, there is barely any red reflection in the panel. At 80% the reflection gets quite stronger, at 90% it's really strong. But the instrument lightning per se barely changes from 70% to 90%. So to make the effect more realistic, I'd make the red reflections already strong at lower panel illumination values, i.e. make the result more linear.

gilbertohasnofb:
Also, is the radio digits also contributing to the effect? Maybe they could contribute a bit at least.

onox:
Have you tried changing the instrument lighting and messing with the battery, and then see how it looks in replay mode? It's a JSBSim system, so IIRC the system isn't executed during replay mode.

onox:
Maybe it's better to use the autopilot XML system. Each function is a product, so we don't need any fancy JSBSim components.

onox:
@wlbragg Try something like this:

```xml
<filter>
    <name>RGB Red Factor</name>
    <type>gain</type>
    <input>
        <expression>
            <product>
                <property>/controls/lighting/instruments-norm</property>
                <property>/sim/model/c172p/lighting/ra-r</property>
                <property>/sim/model/material/instruments/factor</property>
            </product>
        </expression>
    </input>
    <output>
        <property>/fdm/jsbsim/lighting/rgb-ra-r-factor</property>
    </output>
</filter>
```

Maybe you should change the output property since it's no longer a JSBSim system.

onox:
Add the filters to the bottom of `Systems/als-lights.xml`. Seems like a good place.

wlbragg:
> I'd make the red reflections already strong at lower panel illumination values

I'm not sure about this, I'm making a quick change that needed to be done regardless, I'm pushing it now and tell me if that helped at all. If not then I'll have to do something more fancy.

> is the radio digits also contributing to the effect? Maybe they could contribute a bit at least. 

OK, will require some more thinking and props locating for the function, but it should be doable.

OK @onox, I'll do that. I really wasn't sure which system to use and that was the first thing I thought of.

onox:
You could use `/sim/model/c172p/rgb-ra-r-factor` as output property.

gilbertohasnofb:
@wlbragg those are just ideas, ok? You implemented it all so I think it's fair you judge it, particularly since you must consider the amount of effort for small changes is worth it.

wlbragg:
>Add the filters to the bottom of Systems/als-lights.xml. Seems like a good place.

Yes, that is where I wanted to put it and almost started out that way.
>You could use /sim/model/c172p/rgb-ra-r-factor as output property.

OK on that!

@gilbertohasnofb, let me make these changes onox is suggesting, then you can try it. Then we can go from there.

onox:
I've changed it to `/sim/model/c172p/lighting/rgb-ra-r-factor`.

wlbragg:
>I've changed it to /sim/model/c172p/lighting/rgb-ra-r-factor.

10-4

gilbertohasnofb:
:+1:

onox:
Btw.

onox:
```xml
208                     <property>/controls/lighting/instruments-norm</property>
209                     <property>/sim/model/c172p/lighting/ra-r</property>
210                     <property>/sim/model/material/instruments/factor</property>
```

Why is `instruments-norm` in that product? Shouldn't it be removed (since we already use `factor`)?

wlbragg:
Does the factor change with the instrument lighting knob, or is it a 0-1?

onox:
The `factor` gets the value of `instruments-norm` if the voltage is in the correct range, otherwise it becomes 0.0.

onox:
So having `instruments-norm` in that product seems redundant.

wlbragg:
OK.

Do you know anything about the radio stack lighting so I can incorporate it?

wlbragg:
I've never really noticed if you can even turn off the radio stack lighting?

onox:
Avionics switch?

gilbertohasnofb:
> Do you know anything about the radio stack lighting so I can incorporate it?

It's not the radio stack lightning as that is currently tied to the panel lightning (though we discussed here https://github.com/Juanvvc/c172p-detailed/issues/728 that they should be decoupled, but that's a different issue), I am simply talking about the regular radio digits. As long as the avionics switch is on, then the radios will display their digits.

onox:
@wlbragg Have you made any commits? I have some changes that I want to commit in a minute (including adding the filters above)

wlbragg:
>@wlbragg Have you made any commits? I have some changes that I want to commit in a minute (including adding the filters above)

No I haven't I was just getting ready to commit the filters, but go ahead.

onox:
Are you already using the `lighting/rgb-ra-b-factor` property somewhere? I searched for it but it doesn't seem to be in use.

onox:
Anyway, you can use `/sim/model/c172p/lighting/rgb-ra-b-factor` etc. Pushing the change in minute.

onox:
Ah, it's used in an `.eff` file. I'll change it.

wlbragg:
On thing I may have already committed is
~~~xml
<ra-irradiance-map-strength type="float"><use>/controls/lighting/instruments-norm</use></ra-irradiance-map-strength>
changed to
<ra-irradiance-map-strength type="float">1</ra-irradiance-map-strength>
~~~

>Are you already using the lighting/rgb-ra-b-factor property somewhere? I searched for it but it doesn't seem to be in use.

I don't know now, I already made the changeover to the filters so I don't know what I had exactly.

onox:
Can you push the commit?

onox:
Or does it include more stuff?

wlbragg:
I already changed everything so that one commit would be difficult.

onox:
Ok, well, I'm gonna commit then and I'll let you solve any merge conflicts :wink:

wlbragg:
If you doing the filters and removing the old function then go ahead and do that. I'll just pull the results new and go from there.

onox:
Pushed. You can do `git pull --rebase` to put your commit on top.

wlbragg:
What does the scale tag do?

wlbragg:
It's doesn't appear to be working?

onox:
It's just a gain. It should be. See http://wiki.flightgear.org/Autopilot_configuration_reference#Recursive_definition

wlbragg:
The factor is scaling but the effect isn't working.

onox:
So the output property is correctly modified if you inspect it in the property browser?

wlbragg:
Never mind, I screwed up a file.

wlbragg:
Looks good.

wlbragg:
OK, I just need to see if I can add the radio digit lighting. The Avionics switch appears to be the one to use, any idea how to tie it in to the existing filter? I'm not sure it is worth making a new filter set for?

onox:
Have you checked which conditions exactly causes the digits to appear? Is the volume dial needed?

wlbragg:
All I checked was the avionic switch which shut them off.

onox:
If I scroll the power dial, the digits disappear as well.

wlbragg:
Yes each one has a volume that shuts it off.

onox:
So the digits can be made to disappear even if avionic switch is on.

wlbragg:
Yes

onox:
Have you found out where exactly in `Models/Interior/Panel/Instruments/kx165/kx165-1.xml` they control the visibility of the digits? I can't find it.

onox:
Hmm. I think it's around line 220.

onox:
```xml
215   <animation>
216     <type>select</type>
217     <object-name>CommUse</object-name>
218     <object-name>CommStby</object-name>
219     <condition>
220       <and>
221         <property alias="../../../../params/comm-serviceable"/>
222         <property alias="../../../../params/power-comm"/>
223         <greater-than>
224           <property>systems/electrical/outputs/comm[0]</property>
225           <value>6.0</value>
226         </greater-than>
227       </and>
228     </condition>
229   </animation>
```

onox:
@wlbragg Perhaps you can move that `<and>` to a `<logic>` filter and put it in some file in `Systems/`. And then modify `kx165-1.xml` and `kx165-2.xml` to directly use the output property of the new `<logic>` elements.

4 `<logic>` elements because of (comm[0]) with `comm-serviceable` and `nav-serviceable`, and the same for (comm[1]`. Perhaps add a `<filter>` that combines and uses the 4 `<logic>` elements to outputs a float.

After that you should make use of the float to compute the extra RGB red factor.

onox:
Make that 2 `<logic>` elements: nav of comm[0] and comm[1] are controlled by the comm radio's themselves.

wlbragg:
I'm not sure I follow the logic but I did what I think will work with the exception of the following.
What type of logic do I use to (sum) multiple inputs into one output.

basically I need

~~~xml
<input>
  <sum>
    <property>/sim/model/c172p/lighting/panel-ra-b</property>
    <property>/sim/model/c172p/lighting/comm0-ra-b</property>
    <property>/sim/model/c172p/lighting/comm1-ra-b</property>
    <property>/sim/model/c172p/lighting/adf-ra-b</property>
  </sum>
</input>
<output>
  <property>/sim/model/c172p/lighting/rgb-ra-b-factor</property>
</output>
~~~

wlbragg:
Unless there is a way to add a couple additional properties to
~~~xml
    <filter>
        <name>RGB Red Factor</name>
        <type>gain</type>
        <input>
            <property>/sim/model/c172p/lighting/ra-r</property>
            <scale>
                <property>/sim/model/material/instruments/factor</property>
            </scale>    
        </input>
        <output>
            <property>/sim/model/c172p/lighting/rgb-ra-r-factor</property>
        </output>
    </filter>
~~~

wlbragg:
This is where I am at based on my understanding of you suggestion.
~~~xml
    <logic>
        <name>Comm 0 On/Off</name>
        <input>
            <and>
                <property>instrumentation/comm[0]/serviceable</property>
                <property>instrumentation/comm[0]/power-btn</property>
                <greater-than>
                    <property>systems/electrical/outputs/comm[0]</property>
                    <value>6.0</value>
                </greater-than>
            </and>
        </input>
        <output>
            <property>/sim/model/c172p/lighting/comm0</property>
        </output>
    </logic>
~~~
I used this for the comm radios and then
~~~xml
    <filter>
        <name>Comm0 Red Factor</name>
        <type>gain</type>
        <gain>1.0</gain>
        <input>
            <expression>
                <table>
                    <property>/sim/model/c172p/lighting/comm0</property>
                    <entry><ind>   0 </ind><dep> 0.0 </dep></entry>
                    <entry><ind>   1 </ind><dep> 0.0125 </dep></entry>
                </table>
            </expression>
        </input>
        <output>
            <property>/sim/model/c172p/lighting/comm0-r-norm</property>
        </output>
    </filter>
~~~
This gives me a value to add to the ra lighting section along with the panel lighting of
~~~xml
    <filter>
        <name>Panel Red Factor</name>
        <type>gain</type>
        <input>
            <property>/sim/model/c172p/lighting/ra-r</property>
            <scale>
                <property>/sim/model/material/instruments/factor</property>
            </scale>    
        </input>
        <output>
            <property>/sim/model/c172p/lighting/rgb-ra-r-panel</property>
        </output>
    </filter>
~~~
I just don't know how to sum the two values and output them.

onox:
You might nee to put `<expression>` around `<sum>`.

About the table:

```xml
                    <entry><ind>   0 </ind><dep> 0.0 </dep></entry>
                    <entry><ind>   1 </ind><dep> 0.0125 </dep></entry>
```

If it is as simple as this, a simple gain is sufficient. So remove the whole table. Use `<property>/sim/model/c172p/lighting/comm0</property>` directly as input. And then use `<gain>0.0125</gain>`.

However, I see you just use comm0 in that filter. Why not add comm1 as well?

After that, you can use the output property in an `<offset>` and add that to the `RGB Red Factor` filter.

onox:
So what you get is 2 `<logic>` elements (`comm0` and `comm1`) and 1 `<filter>` (`comm0-and-comm1-ra-r`)

onox:
And use the output of that last filter in `<offset>` in `RGB Red Factor`.

wlbragg:
>You might need to put <expression> around <sum>

I think I tried that and it didn't appear to work, but I'll try it again.

>However, I see you just use comm0 in that filter. Why not add comm1 as well?

Right, I just hadn't got there yet, I was just trying to get the one to work the way I wanted. If I put both comms, each input a value of 1, and the gain being .0125, will that then double the gain of .0125 to effectively be .0250? That is what I would want.

>After that, you can use the output property in an <offset> and add that to the RGB Red Factor filter.

What is the "offset" do, I don't understand that?

Our ra factor currently is this
~~~xml
 <filter>
        <name>Panel Red Factor</name>
        <type>gain</type>
        <input>
            <property>/sim/model/c172p/lighting/panel-ra-r</property>
            <scale>
                <property>/sim/model/material/instruments/factor</property>
            </scale>    
        </input>
        <output>
            <property>/sim/model/c172p/lighting/rgb-ra-r-factor</property>
        </output>
    </filter>
~~~
and I will need to add the new output of that "last filter" to it, that is where I need the sum, or a way to do that.

wlbragg:
Well that's why we keep you around, it worked this time. The expression then sum worked, the filter gain on the comm's worked. By gosh I think I got it.

wlbragg:
@gilbertohasnofb this is ready to test. You might let onox look at the code and get his stamp of approval prior to committing.
There could be room for adjustment or improvement of the cumulative effects of all the different pieces of equipment that contribute to the effect. But over all I like it, I think it really helps the cockpit lighting situation at night.

wlbragg:
@onox, the only thing I have reservations about is, I left all the filters in als-lights.xml. I guess technically the 2 comm filters that output for the on off condition of comm1 and comm2 should probably be in there own system file, but that seemed a bit much for just two filters, so I left them in als-lights. You can make that call.

wlbragg:
This last commit changes the direction of the effect so more of the cockpit gets the light effect. If the overall effect is to strong we can lower the values in c172p-set.xml.

onox:
I'm going to see if I can simplify it a bit.

onox:
Btw, why did you add `power-btn` to `Int Red Factor`?

onox:
And why 3 extra factors? Shouldn't the radio's only increase the red?

onox:
@wlbragg With that extra `power-btn` in the `<sum>` you get:

0 if radio's are both off
0.03 if 1 radio is on
0.045 if both radio's are on

Without the `power-btn` property, you get:

0 if radio's are both off
0.015 if 1 radio is on
0.03 if both radio's are on

Which situation do you want?

onox:
@wlbragg I pushed one commit that uses `<offset>`. Please study it if you are not familiar with it. `<scale>` and `<offset>` are general elements that you can use in `<input>` elements. See http://wiki.flightgear.org/Autopilot_configuration_reference#Linear_transformation_of_the_input_value_.3Coffset.3E

Note that in the commit I removed the use of the extra `power-btn` property. So the second situation in https://github.com/Juanvvc/c172p-detailed/pull/731#issuecomment-200557299 applies now. Maybe it's better to just increase the gain a little bit if you want to have 0.045 if both radio's are on.

onox:
> I guess technically the 2 comm filters that output for the on off condition of comm1 and comm2 should probably be in there own system file, but that seemed a bit much for just two filters, so I left them in als-lights

That's right, ideally they should be in a separate file. Maybe they be moved to `Systems/NAVandGSfilters.xml`. (That file should also get renamed to something like `Systems/instruments.xml`)

onox:
Btw, I see that the fuselage (the interior) is indeed affected by the red digits of the radio's, but the instrument panel isn't.

onox:
With the instrument panel I mean the panel around all the basic instruments (AI, ASI, VS, NAV, ADF, etc.)

wlbragg:
>Btw, why did you add power-btn to Int Red Factor?

That is a third light source (adf) that can be turned on and off.

>And why 3 extra factors? Shouldn't the radio's only increase the red?

I really don't have a definitive answer for that other than the entire effect is defined by an RGB value, that is 3 values to make the color. The panel lights default to the three values of r=.05 g=.025 b=.025 so as far as I was concerned, with that as a basis for any light, any other lights that are emitting light should also proportionally emit all 3 values.

>With that extra power-btn in the <sum> you get:

>0 if radio's are both off
0.03 if 1 radio is on
0.045 if both radio's are on

>Without the power-btn property, you get:

>0 if radio's are both off
0.015 if 1 radio is on
0.03 if both radio's are on

>Which situation do you want?

If your going to remove the ADF, or should I just say, with any of the add light sources, it's a judgment call as to how much we want to increase the effect. But it needs to be enough to make a visual difference or it's not worth doing it at all.

wlbragg:
>but the instrument panel isn't.

I didn't noticed that, if there is objects that should get the effect they have to be in the c172p-interior definition. that brings up another option or situation. We apply a different irradiance effect to the buttons and knobs (anything in the c172p-interior-radiance definition) technically most of that should have the ra-irradiance effect also. The reason for the two definitions is one included the shadow (c172p-interior) and the other does not (c172p-interior-radiance). So I suppose we need to add the ra-irradiance lines to c172p-interior-radiance and see what it does.

wlbragg:
> I suppose we need to add the ra-irradiance lines to c172p-interior-radiance and see what it does.

Just beware that also includes all the instrument faces that are emitting the light. I need to go through the objects with a fine toothed comb (so to speak) and make sure I have all the necessary objects in the right definitions. It's possible we may even need a third definition. I need to look at the object again while this is fresh in my mind.

onox:
About `<property>/instrumentation/adf[0]/power-btn</property>`. I'll add that back.

wlbragg:
:+1:

onox:
Btw, I'm not talking about the instruments themselves, I really mean the background (black if you use the default livery).

wlbragg:
That should be in the ra definition. Let me check it.

wlbragg:
Yeah, because that has the shadow applied and that is the same definition as the ra-irradiance.

onox:
Pushed the fix adding back the ADF light source (`power-btn` property)

wlbragg:
panel_1_gauges is the object around all instruments, it has the effect applied.

So your done? I can pull?

onox:
I'm done, you can pull.

wlbragg:
Looks good, you want me to separate the comm filters as per?
> Maybe they be moved to Systems/NAVandGSfilters.xml. (That file should also get renamed to something like Systems/instruments.xml)

onox:
Sure, you can move the 2 comm `<logic>` elements.

wlbragg:
OK, after that, as far as I am concerned, we're done.

gilbertohasnofb:
@wlbragg I had a quick test now and it looks really good! The only thing I would change is that I think the radio digits are contributing too much to the effect, I think the red reflections caused by them should be barely perceptible, particularly when compared with the main instruments.

onox:
@gilbertohasnofb Look in `Systems/als-lights.xml`. The gain for red is 0.015, green and blue is 0.0075.

gilbertohasnofb:
@onox I didn't mean to mess up with colour components, I simply meant that the effect (which looks red to me, which is why I used the term) seems too strong when only the avionic switch is on but the panel lightning is set to 0%. The digits in the radio should be far too weak to illuminate the cockpit by that much. But the effect caused by the instrument lightning (at 100%) is really really good. And it feels much more linear then before when going from 0% to 100%, which is great.

gilbertohasnofb:
so to make it all clear, "red reflections" == "irradiance effect" in the post above

onox:
Well the 3 gains control how much influence the radio digits have. So if you think the effect caused by the radio digits should be weaker, then you should lower those gains.

gilbertohasnofb:
@onox I see now. Diminishing those values of the `Inst <colour> Factor` filters to `0.010 and 0.006` seems to do the trick IMO. But try this:

- just radio digits on but panel lightning at 0%: very weak irradiance effect as expected
- avionics switch off and panel lightning at 100%: a lot of irradiance effect as expected
- now flick the avionics on and off: the radio digits contribute A LOT to the effect if the panel lightning is at 100%. I think it should contribute just a tiny bit, just like the difference between total darkness and the first situation described here

wlbragg:
@gilbertohasnofb if you happy with .010 and .006 on the inst filters, go ahead and change them and commit it. I agree it should have too much of an effect. But it has to have enough to be worth the code to have it contribute at all.

wlbragg:
Or I can change them and finish moving the comm filters and then it should be ready to merge.

gilbertohasnofb:
My main problem right now is that the digits contribute too much when the panel lightning is at 100%. At 0% it's fine. This modification from .015 to 0.10 and .0075 to .0060 improves the situation at 0% panel lightning, but the one at 100% is still a problem. Can you take a look and see what you think?

> Or I can change them and finish moving the comm filters and then it should be ready to merge.

That is probably easier since you are already modifying the file.

onox:
I noticed another thingy: the power dial of the ADF doesn't have any effect.

onox:
@gilbertohasnofb The effect should work as it should: inspect `/sim/model/c172p/lighting/rgb-ra-r-factor`. if you set instrument lighting to 100 % and turn off ADF (power), then the value will be 0.08. If you turn off avionics, it becomes 0.08 - 0.03 = 0.05.

If you set instrument lighting to 0 %, then the value will be 0.03 or 0.0.

onox:
@wlbragg For the ADF we should use `operable` instead of `power-btn`. That should fix it.

onox:
`/instrumentation/adf[0]/operable`.

wlbragg:
>use operable instead of power-btn. That should fix it.

:+1: 

>That is probably easier since you are already modifying the file.

:+1:

wlbragg:
At 0.10 and .006 I think it is good. I don't think the addition of the comm's to the 100% panel lighting is too much. I think it looks good. Any less and it wouldn't be worth having.

wlbragg:
I'm OK with where this is at.

onox:
@gilbertohasnofb Do you really think the effect at 100 % is a bit too much? I can add a table that depend on the instrument lighting with a value of 1.0 for 0 % and 0.8 at 100 %. Something like that.

wlbragg:
> I can add a table that depend on the instrument lighting with a value of 1.0 for 0 % and 0.8 at 100 %. Something like that.

I'm OK with that, I wouldn't be OK with lowering the comm and adf only factors (without panel) any more or they won't even be visible.

onox:
fgfs-screen-557
fgfs-screen-558

onox:
@gilbertohasnofb @wlbragg I have pushed the commit. In the diff you see I use the number 0.5. This can be increased (up to 1.0) if the glow reduction is too much.

onox:
At 0 % instrument lighting there should be no change. Only when you increase it to 100 %, you'll see that the red glow of the digits contribute less than it used to be.

wlbragg:
Looks really good. I'm good to go with it. So @gilbertohasnofb, if your satisfied you can merge it.

gilbertohasnofb:
I think it's perfect now!

gilbertohasnofb:
Really good job guys :clap: :clap: :clap: 

@wlbragg I agree with you when you write...:

> Any less and it wouldn't be worth having.

..., but sometimes having a delicate effect is really quite something. It shows the developers really care for small things. And I think the end result here is exactly those cases, it may not be obvious to someone not expecting to see them, but when he/she realises they will think "holy moley, those devs think of everything!" :smile:

gilbertohasnofb:
I posted a note about this in our forum thread: http://forum.flightgear.org/viewtopic.php?f=4&t=25157&p=280359#p280359

onox:
Nice work @wlbragg.

@gilbertohasnofb You have retired as well on the forum?

wlbragg:
:+1: Thanks to both of you for helping implement this.

gilbertohasnofb:
@onox Yes, I did. I have been really busy with RL lately and I have been working more on FG than I should (too many projects). Also, I had some couple of frustrating events in the last months which turned to be too much for me to take, so I decided to retire. 

But the bad news for you guys is that you didn't get rid of me from this project here, as nothing changed in relation to the c172p project :wink: 

@wlbragg Practical question: are you subscribed to the c172p thread in the forum?

wlbragg:
Yes

gilbertohasnofb:
Great!

wlbragg:
Let me me know if ever you want me to or think a milestone needs to be posted in the thread. In other words, If I space out something that should be said in the forum, kick me.

gilbertohasnofb:
(y)

wkitty42:
just because someone is listed as "retired" on the forum doesn't mean that they 
cannot post... "retired" is only a group, AFAIK... they can still read and post 
if they want to...

gilbertohasnofb:
@wkitty42 That's not entirely correct. Indeed I can read the forum, but so can any anonymous person as no login is required for reading, but I certainly can't post nor log in nor subscribe to a thread, which is why I asked to have my profile retired in the first place. It really isn't _just_ a forum group.

onox:
> @onox Yes, I did.

Welcome to the club :bowtie: :sunglasses:

wkitty42:
On 03/25/2016 02:15 PM, Gilberto Agostinho wrote:
> @wkitty42 That's not entirely correct. Indeed I can read the forum, but so
> can any anonymous person as no login is required for reading, but I certainly
> can't post nor log in nor subscribe to a thread, which is why I asked to have
> my profile retired in the first place. It really isn't /just/ a forum group.

whatever, dude... i know that in the last three weeks i've seen multiple folks 
who asked to be retired posting numerous times... it0uchpads is/was one of 
those... there are a couple of others, too...

in any case, this is not the proper place to discuss this... i just thought i'd 
point it out, though...

EOT

gilbertohasnofb:
@onox Thanks! :sunglasses: 

@wkitty42 That's because those two users (the other was legoboy) contacted Curt to reactivate their accounts after just a couple of days. 

> whatever, dude..
> [...]
> EOT

Chill out, please.

PR #735, opened 2016-03-25
Bug 732: Make flap lever switch object draggable and scrollable

### LINKED ISSUE #732: flap lever doesn't have hotspots
flap lever doesn't have hotspots and cannot be controlled via click or drag.

onox:
Fixes #732

### DIFF
--- a/Models/Interior/Panel/Flaps/flaps.xml
+++ b/Models/Interior/Panel/Flaps/flaps.xml
@@ -54,27 +54,8 @@
     <animation>
         <type>slider</type>
         <drag-direction>vertical</drag-direction>
-        <drag-scale-px>20</drag-scale-px>
+        <drag-scale-px>30</drag-scale-px>
 
-        <!-- Flaps moving down -->
-        <decrease>
-            <binding>
-                <command>nasal</command>
-                <script>controls.flapsDown(1);</script>
-            </binding>
-        </decrease>
-
-        <!-- Flaps moving up -->
-        <increase>
-            <binding>
-                <command>nasal</command>
-                <script>controls.flapsDown(-1);</script>
-            </binding>
-        </increase>
-    </animation>
-
-    <animation>
-        <type>translate</type>
         <property alias="/params/flaps-lever"/>
         <object-name>FlapSwitch</object-name>
         <axis>
@@ -100,6 +81,22 @@
                 <dep>0.052</dep>
             </entry>
         </interpolation>
+
+        <!-- Flaps moving down -->
+        <decrease>
+            <binding>
+                <command>nasal</command>
+                <script>controls.flapsDown(1);</script>
+            </binding>
+        </decrease>
+
+        <!-- Flaps moving up -->
+        <increase>
+            <binding>
+                <command>nasal</command>
+                <script>controls.flapsDown(-1);</script>
+            </binding>
+        </increase>
     </animation>
 
     <animation>


gilbertohasnofb:
I wish I could test this right now, but I updated my simgear using the saiacot's PPA some hours ago and since then FG crashes during startup... I need to see what is going on :sob:

gilbertohasnofb:
@onox Okay, so my simgear is fixed now. So about the flaps, it all works but there is a small thing that I'd like to know your opinion: left clicking on the lever raises the flaps while middle clicking lower them. But clicking and dragging follow the direction of the drag _except_ if the drag takes too much time after clicking. Basically, we have two systems of mouse interaction with the same lever, and I think the result is a bit confusing. Personally, I would remove the first behaviour and simply use the click and drag method, so simply middle clicking or middle clicking wouldn't do anything. What do you think?

wlbragg:
I didn't try the middle click just the "left click down and move up/down" and the middle scroll up/down. They both worked well and were smooth.
I did get the "can't find marker" error, I assume this is a branch that doesn't have that fixed yet?

gilbertohasnofb:
> I did get the "can't find marker" error, I assume this is a branch that doesn't have that fixed yet?

gilbertohasnofb:
> I didn't try the middle click just the "left click down and move up/down" and the middle scroll up/down. They both worked well and were smooth.

It all works indeed, as I said, but I find it a bit confusing. Also, do we really need to give a mouse user three options on how to use flaps?! 1) left and middle clicks 2) click and drag 3) mouse wheel. I would say to keep only click and drag, because it "feels" realistic since that's how you handle a flap lever in an airplane. Left and middle lick and mouse wheel are nice for knobs because it feels like you are turning one, same for switches.

wlbragg:
I don't have a strong opinion on this either way. But I do like the middle mouse button roll to move knobs and switches.
When I have a issue of choices where one choice doesn't really effect the other, I always opt for the most choices. I don't don't think it is confusing, just multiple choices. If one gets in the way of the other then I might feel differently about it.
But like I said, not a strong opinion.

gilbertohasnofb:
I think offering too many choices is a staple of open source projects, and though in theory it's all good when they don't get on the way of each other, in practise that's seldom the case. Look at the mess of our Advanced Weather system, which must be initiated manually and no one knows about it unless told so. Or the fact that we have three rendering approaches to FG: the default rendering (which looks horrible), Rembrandt (which is unmaintained for years) and ALS (which is the best looking by far but not activated by default). It's all there, but one must know where to click. Anyway, this isn't really relevant for the case here, just felt like stating my 2 cents :wink: 


> If one gets in the way of the other then I might feel differently about it.

The scroll certainly does not get on the way, but the left/middle click and click+drag do get on the way of each other. I would be absolutely fine with leaving the scroll and drag options only.

wlbragg:
>Look at the mess of our

"Everything", you have a good point! If you aren't a regular, you would never know this stuff. Like I said, it doesn't really matter to me if we only leave the drag and drop. You can decide as far as I am concerned.

onox:
@gilbertohasnofb If you look at the diff, you see the [slider](http://wiki.flightgear.org/Knob_/_slider_animation) animation is being used. It has separate "increase" and "decrease" actions. The documentation does not suggest that it is possible to either 1) enable/disable dragging; 2) enable/disable scrolling; and 3) choose mouse buttons. So you just get all of them. Besides, don't all the other controls have the same problem (that you can click or scroll or drag)?

gilbertohasnofb:
@onox You are right, the other controls display the same behaviour (I use the wheel button on most of others so I never realised it before). In this case, let's just merge this! :smile:

PR #744, opened 2016-04-16
better conditional for oil dialog

gilbertohasnofb:
Better conditional for oil dialogs. 

Closes https://github.com/Juanvvc/c172p-detailed/issues/743

And although discussed in that issue, I have not implemented the same conditionals for the securing objects. I attempted to do so, but the result was buggy and I couldn't narrow down the problems. Also, it looked quite inelegant to add the message "you can't add/remove tiedowns/chocks/pitot tube cover with the engine on" + the conditional to 12 hotspots in the Models.xml (that is, 6 for the tiedowns, 2 for each type of chocks and 2 for the pitot tube cover). I am not sure it's worth all this trouble since they are working just fine as of now (with groundspeed < 1.0). But in case you guys think it's worth to implement this, then do not merge this PR yet and let's discuss how to do it further.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -7922,6 +7922,9 @@
                 <condition>
                     <and>
                         <property>/engines/active-engine/oil_consumption_allowed</property>
+                        <not>
+                            <property>/engines/active-engine/running</property>
+                        </not>
                         <equals>
                             <property>/controls/engines/active-engine</property>
                             <value>0</value>
@@ -7935,6 +7938,9 @@
                 <condition>
                     <and>
                         <property>/engines/active-engine/oil_consumption_allowed</property>
+                        <not>
+                            <property>/engines/active-engine/running</property>
+                        </not>
                         <equals>
                             <property>/controls/engines/active-engine</property>
                             <value>1</value>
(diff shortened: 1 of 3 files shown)


dany93:
The condition for oil dialog can't be discussed.
For the others, that's not that important (in my opinion).

gilbertohasnofb:
> The condition for oil dialog can't be discussed.

What do you mean, Dany?

dany93:
I mean that I agree with you on preventing from checking oil when the engine is running (or on displaying some warning). Anyway, I think that checking the oil level when the engine is running is not only dangerous, but it will give a wrong indication.
IRL, nothing prevents anyone from doing this. That is just a nonsense.

gilbertohasnofb:
:+1:

gilbertohasnofb:
By the way, this PR is ready for testing and possibly merging

onox:
When I'm in the air, I cannot use the slider, but there's no red message in the dialog. (With engine off)

gilbertohasnofb:
I see, I am improving it now

gilbertohasnofb:
@onox can you try testing again?

Models/c172p.xml
```diff
@@ -7962,6 +7980,25 @@
...
+                                <property>/fdm/jsbsim/gear/unit[23]/WOW</property>
+                                <property>/fdm/jsbsim/hydro/active-norm</property>
+                            </or>
+                        </not>
+                        <property>/engines/active-engine/oil_consumption_allowed</property>
+                    </and>
+                </condition>
+                <command>nasal</command>
+                <script>
+                    gui.popupTip("You can't check the oil level while on the air!");
```
onox: "in the air".

gui/dialogs/c172p-oil-160.xml
```diff
@@ -85,17 +85,47 @@
...
+                    </or>
+                </not>
             </visible>
             <color>
                 <red>0.9</red>
                 <green>0.1</green>
                 <blue>0.1</blue>
             </color>
-            <label>Slider disabled while engine is running</label>
+            <label>Slider disabled while on the air!</label>
```
onox: "in the air"

gui/dialogs/c172p-oil-180.xml
```diff
@@ -86,17 +86,47 @@
...
+                    </or>
+                </not>
             </visible>
             <color>
                 <red>0.9</red>
                 <green>0.1</green>
                 <blue>0.1</blue>
             </color>
-            <label>Slider disabled while engine is running</label>
+            <label>Slider disabled while on the air!</label>
```
onox: "in the air"

gilbertohasnofb:
@onox oops, sorry for the bad English. Fixed now.

PR #748, opened 2016-05-09
Bug 746

### LINKED ISSUE #746: Return to a visible heading bug
Hi all,

At present only using the 'next' default c172p, version 2016.2.0.

Perhaps someone could remind me how to set up to use the latest from here. I keep the repo updated. Maybe this **Feature Request** has already been addressed, but could not find any references.

Originally the heading bug was like 2 large **orange teeth**. Very visible. As it should be.

Now it seems to be two very **faint** red lines, maybe just 1 pixel wide. Even with the illumination turned up to 100% they are still barely visible. you have to zoom in all the time.

Could we go back to the large orange markers please, if not already done.

gilbertohasnofb:
Fixes #746

- fix irradiance effect problem with the heading bug
- new heading bug for the HI


gilbertohasnofb:
oh no, my master wasn't clean when I first created the branch......

@onox heeeeeelp!

onox:
What's the code for the oil dialog doing here? :stuck_out_tongue_winking_eye:

gilbertohasnofb:
Yeah, I messed this up, sorry! :sob:

onox:
Those 2 commits are still in some other branch?

onox:
Going to force push.

gilbertohasnofb:
They are in the branch `bug-743`, see the unmerged PR https://github.com/Juanvvc/c172p-detailed/pull/744

onox:
Pushed.

gilbertohasnofb:
Should I do something locally or is this all fixed now?

onox:
You can pull if you want.

gilbertohasnofb:
ok, it worked. So this is ready for testing and possibly merging

gilbertohasnofb:
(and thanks a lot, @onox !)

onox:
The heading bug looks a bit weird.

gilbertohasnofb:
you mean the new form?

onox:
fgfs-screen-618

gilbertohasnofb:
Oh ,that's a bug

gilbertohasnofb:
lemme test it again, wait a second

gilbertohasnofb:
I can't reproduce it, @onox, it looks fine here:

fgfs-screen-001

wlbragg:
Mine look good also!

onox:
Don't zoom in. It looks OK when zoomed in, but not when not zoomed in.

gilbertohasnofb:
Default zoom looks fine to me as well:

fgfs-screen-002

wlbragg:
I can't confirm using PR, how do I get the PR?

gilbertohasnofb:
I think you need to force pull in this branch

onox:
Is there any UV map that needs to be updated?

gilbertohasnofb:
Not that I know. The heading bug texture is actually in the same size as the instrument, but mostly transparent except for the bug. So whatever you draw that will be superimposed above the instrument. I see no reason why this bug is happening now to you but not before, since all we did was tweak the texture a bit... could it be that the xml file alterations are causing this somehow?

gilbertohasnofb:
can you maybe test the new texture but with the old hi.xml file?

onox:
Why would the old `hi.xml` make any difference?

gilbertohasnofb:
We modified two things:

- the png file with the heading bug
- the xml file that handles the instrument

I just can't think why the modified png would cause any problems if the older ones were working, so to troubleshoot it I would think that we could try to see if the xml is affecting it or not at all

wlbragg:
I ended up testing the old texture with the new xml and it looked nice. Better that the newer "thick" texture. Sorry @gilbertohasnofb. My GIT system is breaking and I think I am to the point that I need to reinstall it. The git program itself is crashing and making my life miserable.

>I just can't think why the modified png would cause any problems
Me either

onox:
Old heading bug looks a bit weird:

fgfs-screen-619

Might be a OpenGL driver issue. Let's merge this.

wlbragg:
Hold on, that is a positioning issue, maybe.

onox:
Do we want the new or the old bug?

onox:
> The git program itself is crashing and making my life miserable.

On Windows or Linux?

wlbragg:
Windows.

@gilbertohasnofb , did you extend the length of the texture or move it?

gilbertohasnofb:
There are three bugs used throughout this project:

- the original one, which does not fit the current HI model and I don't want to be reused
- the thin one that onox posted a picture of just above, which doesn't look very good either
- the latest thicker one, modelled after the photo above, and which fits the model well, is quite visible and IMO is the better option



@onox if you are seeing this problem independently of the png texture, then the problem must be something else, right?

gilbertohasnofb:
@wlbragg oh yes, I did that. The new bug is a bit smaller than the original one I think, but do you think that would be affecting it?

gilbertohasnofb:
I really don't think this is a UV mapping problem. Case and point: if I substitute the bug texture for one without any transparent area, such as this one:

hi03

The result is this:

fgfs-screen-001

So as I said before, I think the heading bug isn't a tiny animated object UV-mapped, but actually a whole plane superimposed in front of the HI and which is only rotated

wlbragg:
Right, looking at the blend and the code, it would have to be the entire sheet size or possibly the needle being longer. I'm not really sure. There is no "center" coords for the rotation animation. I think onox would have to play with the position of the needle texture to see if that made a difference.

onox:
Let me check.

gilbertohasnofb:
The point is that the texture should get distorted in any way, regardless of the centre coordinates of the rotation, right? And I don't see this bug here, regardless of the zoom. 

@onox out of curiosity, could you test the very first heading bug of our project? here is the file: https://github.com/Juanvvc/c172p-detailed/blob/6c9c2893945757a831acbb42a944e9b06ffc2d84/Models/Interior/Panel/Instruments/hi/hi03.png

onox:
Are we using some mipmap filter?

gilbertohasnofb:
I think I found out the problem: you have texture compression enabled, don't you?

gilbertohasnofb:
for some reason it is affecting the bug quite heavily!

wlbragg:
I moved that sheet and could not reproduce what onox is seeing.

wlbragg:
Could it be because of the big sheet with a little object on it?

onox:
How do I know I have texture compression enabled?

gilbertohasnofb:
I enable/disable it in FGRun. In the last page, go to Advance, then Rendering then look for the fourth toggle in the left

onox:
Yep, you found it @gilbertohasnofb. It was texture compression.

gilbertohasnofb:
So the question is now why is it affecting the bug so heavily and not other textures?! People tend to use the compression as it gives better performance

wlbragg:
Any idea what the command line is. I am using it in fgrun but I doubt qt is using it unless by default. I am currently running qt and not fgrun.

wlbragg:
bump
>Could it be because of the big sheet with a little object on it?

onox:
`--prop:/sim/rendering/texture-compression=false`

gilbertohasnofb:
Once we had some problems with the ball in the turn coordinator that had some artifacts depending on the angle, does anyone remember how we solved it? Could it somehow be related?

- command line should be: `--prop:/sim/rendering/texture-compression=on`

> Could it be because of the big sheet with a little object on it?

I have no idea...

gilbertohasnofb:
I think that prop takes `on` or `off`

onox:
@gilbertohasnofb Next question :stuck_out_tongue_winking_eye: how do I enable anisotropic filtering?

gilbertohasnofb:
> how do I enable anisotropic filtering?

Oh, that I don't know! I don't use any anti aliasing nor filtering because it kills my performance. FGrun just has some generic texture filtering values in the same rendering dialog, but nothing about anisotropic filtering

wlbragg:
--texture-filtering=16
  --prop:/sim/rendering/multi-sample-buffers=1
  --prop:/sim/rendering/multi-samples=4

wkitty42:
> People tend to use the compression as it gives better performance

really?? i don't think i've ever used texture compression... what is the actual 
--command-line property to use for it?? i haven't used FGRun since the Qt 
launcher became stable for daily use and i don't even have FGRun being built on 
my system any more...

wlbragg:
Look at 4 and 5 posts above.

onox:
I use MSAA 2x, not 4x.

onox:
Anyway, which heading bug do we want?

wlbragg:
No mater what settings I use I am getting a consistent visual!

gilbertohasnofb:
I think the main question now should be: is this something we can fix on our side or is this bug caused by FG's texture compression algorithm and there is nothing to be done? Is the compression causing problems because of the use of transparency maybe? I heard before that transparency in png files can cause issues...

> Anyway, which heading bug do we want?

The latest one is the best IMO (obviously). It is easier to read than the previous, and is closer to the RL instrument. The original c172p bug should not be used with this HI as Cessna didn't use them together

onox:
I only notice any difference when using `--prop:/sim/rendering/multi-sample-buffers=0` and `--prop:/sim/rendering/multi-samples=0`.

gilbertohasnofb:
This is the only bug Cessna used with this model (I must have seen 1000 pictures of different models of HI and bugs....):



compare to:

onox:
I feel like those poor Rembrandt users with no MSAA :joy:

onox:
> closer to the RL instrument

Let's take the last one then.

wlbragg:
:+1:

gilbertohasnofb:
If you all agree with that and if we can't/won't do anything about the compression issue (anyway that should be a separated issue) then this should be ready to merge

onox:
I think the compression issue is something that OSG/FG needs to take care of.

gilbertohasnofb:
:+1:

wlbragg:
Oh my, I just turned off MSAA, yecK!

onox:
`/sim/rendering/texture-compression` can be the following: false/off/0/no/none or: arb, dxt1, dxt3, dxt5.

wlbragg:
Thanks!

wlbragg:
So it is off or on by default?

onox:
Default value is NULL. So my guess no compression then.

wkitty42:
On 05/09/2016 03:14 PM, wlbragg wrote:
> So it is off or on by default?

i looked for it before changing it... it did not exist... is this listed in the 
wiki or somewhere with explanations? "true" and "on" seem to be valid but i 
don't know if i see any difference between them...

sorry for the extended OT...

gilbertohasnofb:
Default seems to actually be texture compression on. When you deactivate texture compression in FGRun, you can see that it adds the string `--prop:/sim/rendering/texture-compression=off` to the command line. But when you enable it, FGRun simply removes that previous string and adds no other one, so it makes me think that that would be the default value.

wlbragg:
>/sim/rendering/texture-compression can be the following: false/off/0/no/none or: arb, dxt1, dxt3, dxt5.

According to @onox, this also support on by default and using a default type of compression but other capabilities or options are available.

gilbertohasnofb:
If it's on by default, then it's likely that really _a lot_ of people are using it, which makes the heading bug issue more problematic...

onox:
If I'm looking at the correct code (`src/Viewer/renderer.cxx` in flightgear and `simgear/scene/util/SGSceneFeatures.cxx` in simgear) then it doesn't support the words on/true etc. It only supports the options I mentioned above.

In the constructor of `SGSceneFeatures` you see:

```cpp
 41 SGSceneFeatures::SGSceneFeatures() :
 42   _textureCompression(UseARBCompression),
```

So it uses the ARB compression mode by default.

onox:
It could be that all those alpha pixels are causing the trouble. Perhaps try to make a small object (instead of a large rectangle) with no alpha pixels?

gilbertohasnofb:
I really think that would solve the issue, but that would mean that someone ( @wlbragg for instance :wink: :wink: :wink:) would have to model it. I could then replace the texture `hi03.png` from its 512x512 size and full of alpha pixels to a simple 16x16 or even smaller texture of full orange colour, which would be applied to the object.

So @wlbragg would you be interested in working on this? Shouldn't be too difficult I think, the model is supposedly quite simple:

onox:
It is only the HdgBug object (which is a quad) that has to be replaced, correct?

gilbertohasnofb:
Yes, so replace the current square to something similar to the orange bug in the image above (or using the current texture as base as well would work).

gilbertohasnofb:
It also would be nice if the object would be slightly thick, something like this but _very very_ slim:

bug

wlbragg:
Just saw this, let me look. I'm not sure how the animation will work as there is no center currently because of the shape of the rectangle, I guess.

gilbertohasnofb:
:+1: 

BTW, I am off for today, but tomorrow I am back

PR #750, opened 2016-05-09
Bug 746: Model hi indicator to smaller object.

wlbragg:
@onox, see if that corrects the issue.

See #746 #748


wlbragg:
fgfs-screen-016

onox:
Thanks @wlbragg It does indeed fix the problem. The only issue I can find is that the heading bug is slightly off-center in Blender (Y is currently about -0.00042).

gilbertohasnofb:
@wlbragg this looks really good, thanks for the quick fix. From my side, I would only replace the texture `hi03.png` with a tiny orange square and then re-UV-map the new object if necessary. If you give me green light I can replace the texture myself and then all you'd need to do is check if the UV-map in Blender would still work.

gilbertohasnofb:
Actually, I took the liberty of doing that right now. I tested the new texture (a 16x16 orange square without alpha channel) and the UV-mapping of your new object still seems to work perfectly.

wlbragg:
No problem, I meant to make note of needing a new texture for it before I made the PR but forgot. I'm glad you were able to do this, that's good news but be careful, you may get called on to do MUCH MORE! :smile:

<the heading bug is slightly off-center in Blender (Y is currently about -0.00042).

@gilbertohasnofb did you fix this also?

gilbertohasnofb:
> but be careful, if people know you can do these things, you may get called on to do MUCH MORE! :smile:

Oh, but I didn't touch anything on Blender! The thing is that regardless of how you had first UV-mapped it in Blender, a full orange texture will result in a full orange object. But actually, it would be more elegant if in Blender the UV mapping is also set for the whole 16x16 texture now as opposed to the tiny orange region of the old 512x512 texture. Makes sense?

> did you fix this also? 

No, Blender still frightens me :wink: (although I do plan to learn how to do UV-mapping at least, since I work so often with textures)

wlbragg:
Oh, OK I'll take care of it.

gilbertohasnofb:
Sorry for my uselessness regarding Blender, I _really_ want to address this sooner than later, but I just never found the time nor energy to learn it properly...

wlbragg:
No worries!
Should we keep the original texture in the hi folder and the original object in blender? I guess the answer in your opinion in no or you would have renamed it instead of overwriting it. I renamed the old needle "sheet" object to *OLD and simply hide it in the blend. But I can go ahead and remove it completely. If we are deleting the original texture then I guess I should delete the original object?

wlbragg:
We can always dig it out of git if we ever want it back!

gilbertohasnofb:
I think we should remove the old one completely, as nothing is being lost and a bug is being solved. Also, it may look very confusing in the future for other developers: nothing infuriates me more than opening an aircraft directory, finding 27 png files of which only 3 are in use...

gilbertohasnofb:
> We can always dig it out of git if we ever want it back!

Exactly! But I can't think of why we would ever want that :smile:

wlbragg:
OK, just about ready to push changes.

onox:
Old object can be removed. @wlbragg You probably already know all of this, but you might want to select HdgBug and then set origin to center of geometry or mass (press t to open a panel on the left). And then you see Y is -0.00042 in the panel on the right (press n).

wlbragg:
:+1: 
I am going to push this without testing because I HATE WINDOWS, it desired to give me the sql cache error.

onox:
:+1:

PR #758, opened 2016-05-17
Bug 757: better Autostart function

### LINKED ISSUE #757: Complex startup procedure off: Nasal error message
When I launch the aircraft with the "**Start with engine running**" checked (only this start option checked) I have this error message:
`Nasal runtime error: No such member: autostart
  at /home/dany/FG/github/c172p/Nasal/engine.nas, line 185
  called from: /home/dany/fgfs/install/flightgear/fgdata/Nasal/globals.nas, line 119`
I've had it for a long time.
Still, everything works well. It starts with the engine running.

I do not have this error if "**Complex Startup Procedure**" is checked, or both ("**Complex Startup Procedure**" and "**Start with engine running**") are checked.

_If I'm the only one, no big worry, but just in case..._

FG 2016.2.0 git (download and compile v 2.33). Updated 14 May 2016.
Linux Mint 17.

gilbertohasnofb:
Closes #757 

- better autostart function: sets oil level to max, sets fuel to 25% for each tank if any of them is lower than 25% (min recommended for safely operating the 172), automatically crank and starts the engine
- removed the "complex engine" option from dialogs
- small fix to fuel dialogs: the max level was wrongly set to be 28 gal instead of 21.5 gal
- in the `c172p-set.xml` file, the nasal file `c172p.nas` is declared before all others so that in the future any of its functions may be called from the ones below

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -43,14 +43,15 @@ var autostart = func (msg=1) {
     setprop("/consumables/fuel/tank[0]/water-contamination", 0.0);
     setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);        
 
-    # Oil level
-    var oil_enabled = getprop("/engines/active-engine/oil_consumption_allowed");
-    var oil_level   = getprop("/engines/active-engine/oil-level");
-
-    if (oil_enabled and oil_level < 6.0)
+    # Setting max oil level
+    if (getprop("/controls/engines/active-engine") == 0) {
         setprop("/engines/active-engine/oil-level", 7.0);
+    } 
+    else {
+        setprop("/engines/active-engine/oil-level", 8.0);
+    };
 
-    # Fuel level
+    # Checking for minimal fuel level
     var fuel_level_left  = getprop("/consumables/fuel/tank[0]/level-norm");
     var fuel_level_right = getprop("/consumables/fuel/tank[1]/level-norm");
 
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -41,23 +41,23 @@ var autostart = func (msg=1) {
     setprop("/sim/model/c172p/securing/tiedownT-visible", 0);
 
     setprop("/consumables/fuel/tank[0]/water-contamination", 0.0);
-    setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);
+    setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);        
 
-    # Oil level warning for lazy pro's
+    # Oil level
     var oil_enabled = getprop("/engines/active-engine/oil_consumption_allowed");
     var oil_level   = getprop("/engines/active-engine/oil-level");
 
     if (oil_enabled and oil_level < 6.0)
-        logger.screen.red("Warning: low oil level! Check level via oil cap on top of engine!");
+        setprop("/engines/active-engine/oil-level", 7.0);
 
-    # Fuel level warnings
+    # Fuel level
     var fuel_level_left  = getprop("/consumables/fuel/tank[0]/level-norm");
     var fuel_level_right = getprop("/consumables/fuel/tank[1]/level-norm");
 
     if (fuel_level_left < 0.25)
-        logger.screen.red("Warning: fuel level of the left tank lower than 25%!");
+        setprop("/consumables/fuel/tank[0]/level-norm", 0.25);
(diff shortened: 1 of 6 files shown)


wlbragg:
Just tested, works great. :+1:

Nasal/c172p.nas
```diff
@@ -43,14 +43,15 @@ var autostart = func (msg=1) {
     setprop("/consumables/fuel/tank[0]/water-contamination", 0.0);
     setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);        
 
-    # Oil level
-    var oil_enabled = getprop("/engines/active-engine/oil_consumption_allowed");
-    var oil_level   = getprop("/engines/active-engine/oil-level");
-
-    if (oil_enabled and oil_level < 6.0)
```
onox: Could you add `if (oil_enabled and oil_level < 6.0)` around the new if-else? You can keep the if-else, but add this old condition around it. It was added to avoid completely resetting the oil level when someone just has a (for once) lazy moment.

onox:
I have 1 comment about the resetting of the oil level. Other changes are OK.

onox:
You could also set the PANEL LT and RADIO LT after the other PR has been merged.

onox:
And set the parking brake.

onox:
And flaps.

gilbertohasnofb:
@onox Ok, I will fix the oil level. As for the parking brake and flaps, which condition do we want, take off configuration? (i.e. no parking brake + flaps at 10)? Also, setting panel and radio lights seems problematic, since it will depend if it's a night flight or not, should I use the sun position to set it (i.e. below horizon == lights on)?

onox:
Hmm, I was thinking about enabling parking brake. You can leave the PANEL LT and RADIO LT knobs.

Just set the flaps to 10 then?

gilbertohasnofb:
I think that maybe removing the parking brake is the better option, as when you autostart you are ready to go (to take off if you are on runway, or to taxi if you are on the gate). It also makes things even simpler for beginners. What do you think?

Else, oil correction and flaps 10 are coming now

gilbertohasnofb:
Wait a second, I just checked the checklists and actually we need to have flaps at 0 degrees for a normal take off. Flaps 10 is used for short field take off only.

onox:
Move it to 0 then?

gilbertohasnofb:
Yeah, flaps are now being set to 0. So the last thing is about the parking brakes. Should we set, unset or ignore it?

onox:
Let's ignore it for now. Don't really know what would be the best option.

gilbertohasnofb:
:+1:

gilbertohasnofb:
Ok, ready for review

Nasal/c172p.nas
```diff
@@ -44,12 +46,16 @@ var autostart = func (msg=1) {
...
-    if (getprop("/controls/engines/active-engine") == 0) {
-        setprop("/engines/active-engine/oil-level", 7.0);
-    } 
-    else {
-        setprop("/engines/active-engine/oil-level", 8.0);
-    };
+    var oil_enabled = getprop("/engines/active-engine/oil_consumption_allowed");
+    var oil_level   = getprop("/engines/active-engine/oil-level");
+    
+    if (oil_enabled and oil_level < 6.0)
```
onox: Please use brackets {} if it's not a single line.

gilbertohasnofb: oops

gilbertohasnofb: done

onox:
Thanks :+1: It's not for prettiness though. If you use brackets, then there's no doubt or ambiguity what code belongs to the if-condition and what not. Especially if you have multiple if and else conditions and if the code is badly formatted and mixes spaces and tabs then it can be difficult to understand what the programmer who wrote it meant.

gilbertohasnofb:
Thanks for the merge!

As for the brackets, I completely agree with you and actually I had understood and agreed immediately on why that was better! I just used the term "pretty" in the commit because of lack of better vocabulary; of course making the if's and loop statements as clear as possible is fundamental. I didn't mean it sarcastically or anything, it's just the fault of my clumsy English I suppose :disappointed:

onox:
Oh, okay. I thought you just misunderstood the reason :smile:

wlbragg:
Thanks guys, nice update!

gilbertohasnofb:
:smile:

PR #765, opened 2016-05-22
Bug 763: Property already defined bug

### LINKED ISSUE #763: Property already defined bug
I may have discovered how to eliminate these warnings in the console. Even though they are minor irritations, there is really no reason not to fix them if we can.

We need to see where the property is defined and if it is defined in an autopilot or system call, you should be able to remove it from there and instead place it in the -set file for initialization. We have done this n the Shuttle and the J3Cub.

wlbragg:
Fixes #763 

I'm guessing we still are OK with moving the creation of this property to the centralized location in the set?

### DIFF
--- a/Systems/indicated-airspeed.xml
+++ b/Systems/indicated-airspeed.xml
@@ -42,8 +42,6 @@
     output shown on the ASI.
 -->
 
-    <property>/instrumentation/airspeed-indicator/indicated-speed-kt</property>
-
     <channel name="indicated airspeed">
 
         <fcs_function name="systems/asi/indicated-airspeed">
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
Seems fine to me, is this ready to be merged? @onox what do you think?

onox:
I can't reproduce the problem that @wlbragg is trying to fix.

gilbertohasnofb:
Me neither (on FGRun)

gilbertohasnofb:
@wlbragg Is this PR still relevant? Its branch didn't receive any work in quite a while (now it even has some conflicts with the `master`).

wlbragg:
I don't know why some get these warnings and others don't, but I do know how to eliminate the warning for all.

I personally wouldn't stand for this type of warning/error in my code when there is a way to eliminate it. But I am apparently in the minority here and have no problem leaving things the way they are now. So this can also be closed without merging.

gilbertohasnofb:
If you'd like, I am of course willing to make as many tests as you can come up with regarding this issue. On the other hand, if you want to let it go then maybe we should also close https://github.com/Juanvvc/c172p-detailed/issues/763

PR #766, opened 2016-05-22
New ver2.10 Blender import/export script.

wlbragg:
This was used to import the kma20 ac file successfully.
It should be compatible with all the blends, but I don't know that for sure. I will be using it exclusively for importing and exporting ac files.
Script was obtained from https://forum.flightgear.org/viewtopic.php?f=18&t=13442


onox:
You used https://github.com/NikolaiVChr/Blender-AC3D? What's is the point of adding this .zip to the repository if you just can reference to the repository in, for example, `README.md`?

wkitty42:
i think the point is the same one as having the existing io_scene script in the repo and distributed to those who want to do 3D development for this (and other) craft... IIRC, that point was to ensure that everyone was using the same script so that things weren't left out when more than one person was working on the ac files...

wlbragg:
Either way is fine with me, I assumed because the other one was included we would want this one. A reference should be enough..

wkitty42:
what i was thinking of was also that if we're switching to the new one that the old one should be removed from the repo and replaced by this new one as well as making a prominent note and getting everyone else to switch to using this new script... that way everyone is on the same page and we won't see things happening like we did in the beginning of this project until everyone was using the io_scene script...

Juanvvc:
> that way everyone is on the same page and we won't see things happening like we did in the beginning of this project until everyone was using the io_scene script...

That's right. The import/export blender plugin was slightly modified by @thevirtualfer  and me to manage some complex groups the main .blend file was using. That was the reason to include the script in the first place. If you change this script, or create a reference to any other external script, please, check the main .ac file can still be generated from the .blend file!

wlbragg:
A reference to the new script is fine with me. All I know is the new script is required to import, edit and export some of the instruments.

>the only modification was the management of complex groups of objects

Do you know a specific example of what the old script exported that I can test and compare the new script to?

wlbragg:
>The only modification was the management of complex groups of objects

I am having issues with this version of the script managing "groups" in the J3Cub, so I think your concerns are justified.
Is there anyway one of you "experts" might take a look at the old and new import/export script and possibly update the "new" script to include the "grouping" updates that were applied to the "old" script, so we could have the best of both worlds. Or, maybe I could ask the maintainers of this new script to do the same and add the old script grouping updates?

gilbertohasnofb:
@wlbragg is this PR still relevant? If so what's missing for it to be merged?

wlbragg:
We can just close this. I'm not interested in tracking down differences in scripts and combining them. Eventually maybe I can get the necessary information to @Necolatis and they can combine it all into a single script.

PR #767, opened 2016-05-23
Bug 762: Prop spray

wlbragg:
Fixes #762 

Not ready for merge yet. Needs testing and also we need to look at Thorsten's suggestion about using emissive lighting.
I don't have any idea how that is done.

PR #768, opened 2016-05-23
Bug 762: Prop spray

### LINKED ISSUE #762: Prop mist while raining
From the forum, see
https://forum.flightgear.org/viewtopic.php?f=4&t=25157&p=285877#p285822

Consider spray from morning dew and on water in float or amphibious as well.

Posted from Android tablet.

onox:
Fixes #762 

Supersedes PR #767 

Not ready for merge yet. Needs testing and also we need to look at Thorsten's suggestion about using emissive lighting.
I don't have any idea how that is done.


onox:
It doesn't take into account the wetness of the ground. Should we use `/environment/rain-norm`?

onox:
Non-emissive:

fgfs-screen-646

Emissive:

fgfs-screen-645

wlbragg:
Thank you onox.

>It doesn't take into account the wetness of the ground. Should we use /environment/rain-norm?

Yes, what is the condition now? I don't see anything based on any weather condition.

gilbertohasnofb:
Is this ready for testing and merging or is there some work to be done?

onox:
The weather condition needs to be added as well. I'll take a look at it.

onox:
You can merge it if testing is OK.

gilbertohasnofb:
:+1:

gilbertohasnofb:
So I think it looks good, but unfortunately the spray is inside the cockpit when the throttle setting is low enough. If there is no way of keeping it only outside, maybe we should disable it from the interior view.

Also, I noticed something which I think is unrelated to this PR, but can anyone else confirm that the __back__ of the propeller disk look quite red at dusk and early evening while the front view shows normal colours? (click on the images below to see it in full resolution, else it's not possible to see it well just as small thumbnails on GitHub)

fgfs-screen-002

fgfs-screen-003

gilbertohasnofb:
About the propeller disk, I tested it with another branch and indeed it has nothing to do with this PR, sorry for the noise. I opened this issue for it https://github.com/Juanvvc/c172p-detailed/issues/771

onox:
> maybe we should disable it from the interior view.

I don't think that will work, not for MP aircraft anyway. Instead I have updated the particle emitter's position and speed. Hopefully you shouldn't see any particles anymore inside the cabin.

gilbertohasnofb:
> I don't think that will work, not for MP aircraft anyway.

Well, wouldn't that be a tiny detail if while inside the cockpit view you wouldn't see other 172's prop spray?

> Instead I have updated the particle emitter's position and speed. Hopefully you shouldn't see any particles anymore inside the cabin.

Indeed I don't see the spray inside the cabin any longer, but I think the position now is really too far behind the aircraft. Right now, at idle, the spray is already quite behind the tail of the aircraft, and unless one knows what it is supposed to represent, it looks rather strange to me.

I think something like this looks better, but has the cabin problem:

onox:
@gilbertohasnofb I've made the modifications as you suggested.

gilbertohasnofb:
Checking now.

gilbertohasnofb:
Looks much better to me, and the internal view is now free from spray :+1: Personally I would still bring the spray closer to the prop particularly when the engine is idle, I think it is still a bit too far back.

This is the current idle throttle spray position:

fgfs-screen-004

This is the full throttle spray:

fgfs-screen-003

I think that something like this could look better, if the idle would be around the green area and full throttle at around the purple area:

fg

But this is just an idea, if you think what you currently have is better then we can merge this!

onox:
@gilbertohasnofb You're free to modify the file to tune its behavior :)

gilbertohasnofb:
@onox I know that, but I thought the best approach was to let you know about my feedback and then let you decide on what is worth the trouble, given that this all has been your work. So I just can't tell if you really agree with me but wish me to change the code or if you think I am being picky and this isn't worth it. Because if it's the latter, then we can merge this straight away, as I wrote before:

> But this is just an idea, if you think what you currently have is better then we can merge this!

gilbertohasnofb:
@onox I am merging this PR right now as this it has been opened for quite some time and the discussion seems to be over. Your work is looking good, so it should go to the `master` branch, and if we can always tweak it in the future if anyone judges it necessary.

PR #775, opened 2016-05-30
removing the AI serviceable property from the set file

gilbertohasnofb:
Fixes https://github.com/Juanvvc/c172p-detailed/issues/773

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -256,7 +256,6 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <path>/sim/model/c172p/brake-parking</path>
             <path>/sim/model/c172p/enable-fog-frost</path>
             <path>/sim/model/c172p/garmin196-visible</path>
-            <path>/instrumentation/airspeed-indicator/serviceable</path>
             <path>/engines/active-engine/oil_consumption_allowed</path>
             <path>/engines/active-engine/oil-level</path>
             <path>/fdm/jsbsim/settings/damage</path>
@@ -598,9 +597,6 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
         <magnetic-compass>
             <pitch-offset-deg type="double">-2.7</pitch-offset-deg>
         </magnetic-compass>
-        <airspeed-indicator>
-            <serviceable type="bool">false</serviceable> <!-- since pitot tube is on for the first time one runs this plane -->
-        </airspeed-indicator>
         <save-switches-state type="bool">false</save-switches-state>
     </instrumentation>
 
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -597,6 +597,9 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
         <magnetic-compass>
             <pitch-offset-deg type="double">-2.7</pitch-offset-deg>
         </magnetic-compass>
+        <airspeed-indicator>
+            <serviceable type="bool">true</serviceable>
+        </airspeed-indicator>
         <save-switches-state type="bool">false</save-switches-state>
     </instrumentation>
 


gilbertohasnofb:
After this is fixed, I think we should ask for another merge into FGDATA, as we can't possibly have an aircraft without AI there... @onox how should we proceed, should we create a new release?

onox:
The bug report says it is only for 2016.3.0? That's a development version.

onox:
I'm using 2016.2.0 (from april) and AI works in our `master` branch.

gilbertohasnofb:
I am using 2016.1.1 and the AI is broken for me. It surprises me that it's working for you since the -set file clearly sets the serviceable property to `false` and nothing changes it back to `true`!

gilbertohasnofb:
@onox would you fire up our plane in the `master` branch and check the value of `instrumentation/airspeed-indicator/serviceable`

wlbragg:
It broke in our current master branch also.

wkitty42:
every time i see you guys saying "AI" i think of Artificial Intelligence instead of Airspeed Indicator :?

gilbertohasnofb:
> every time i see you guys saying "AI" i think of Artificial Intelligence instead of Airspeed Indicator :? 

And worse of it all, I use it wrongly all the time. AI is actually used for the Attitude Indicator, ASI is the Air Speed Indicator.

wkitty42:
ha! i'm all too familiar with doing that :smile_cat:

gilbertohasnofb:
@wlbragg can you take a look at this? I have forced the serviceable property to `true` in the set file as we discussed, so if all is good we can merge this.

gilbertohasnofb:
Just say this in the forum: https://forum.flightgear.org/viewtopic.php?f=25&t=29863

So this is the second or third report of broken AI in our plane. Guys, seriously, we need to make a new release as soon as possible, we simply can't have the default plane with a broken AI in FGDATA!

@wlbragg would you care to reply to that forum thread, please?

onox:
To be honest I still don't understand what the root cause of this problem is. When I start FG the property always gets set to true. It's even more strange that you have to explicitly set it to true. It should work without it. And why only for the AI? Why not for other instruments and systems? Are you sure this isn't just a PEBKAC problem? (user leaving the pitot cover on)

gilbertohasnofb:
> To be honest I still don't understand what the root cause of this problem is. 

The cause is that some line of codes from the original implementation of the pitot tube cover were left in the `c172-set.xml` file, forcing the property `instrumentation/airspeed-indicator/serviceable` to be set to `false`, and also saving this property value in the file `.fgfs/aircraft-data/c172p.xml`.

> When I start FG the property always gets set to true.

That's really strange, and I have no idea how your system is overriding the `c172-set.xml`. But whatever is happening on your side, it's a fact that the mentioned property is set to `false` in the set file, and several people (me, @wlbragg, @wkitty42 and at least 2 or 3 users in the forum) can reproduce the problem.

> It's even more strange that you have to explicitly set it to true

But that's incorrect. It's not needed to explicitly set it to `true`, but as it has been discussed before doing that has a benefit: look, because the value of the property was saved in the `.fgfs/aircraft-data/c172p.xml` file, once you load the fixed version of the plane in FG that file would then set the property back to `false`. But then if you quit the sim correctly, the new `.fgfs/aircraft-data/c172p.xml` would not contain any information about the property (since it's not saved any longer) and so the sim would run normal on the next run. This means that in the best case scenario the fixed airplane would still present the problem in the first time it runs. Now consider people that do not exit the sim properly. If they have still the old `.fgfs/aircraft-data/c172p.xml` in their system and they close the sim by clicking in the `X` button in the top right of the FG window or by using `Ctrl+F4`, then they would have the problem forever and cause lot's of headaches. A simple fix is to force the property to `true`, so whatever is saved on their `.fgfs/aircraft-data/c172p.xml` the set file will override it.

> And why only for the ASI? Why not for other instruments and systems?

Because the serviceable property affects only the ASI. It's as simple as that, if the property is set to `false` it "breaks" the instrument, if set to `true` the instrument runs normally

> Are you sure this isn't just a PEBKAC problem? (user leaving the pitot cover on)

Yes, I am. I tested it extensively, and so did others. And as I mentioned before, there is the fact that the `c172p-set.xml` file in FGDATA is setting `instrumentation/airspeed-indicator/serviceable` to be set to `false`, so there is absolutely no question of what is causing the issue.

gilbertohasnofb:
@onox and I have a guess why the problem is not affecting you: even though the property `instrumentation/airspeed-indicator/serviceable` is being set to `false` in the `c172p-set.xml` file, the stored value from the previous session (saved in `.fgfs/aircraft-data/c172p.xml`) seems to override it. So you shouldn't see any problem if you last set that value to `true`. This could have happened either manually via the property tree or by chance by having the old pitot tube cover removed just before pulling the modifications which stopped using that property for the cover. 

But anyway, this PR has been merged and I believe the problem is solved now, but we honestly should work into having a release for FG 2016.2.2

onox:
I have cherry-picked the commits to [`release/2016.2`](https://github.com/Juanvvc/c172p-detailed/tree/release/2016.2) Point the core devs to that branch if they want to do a 2016.2.2 release.

wlbragg:
@gilbertohasnofb  is correct, we have verified the behavior. There is actually a new post of the issue that just appeared. I was getting ready to reply with the fix and again verify that the user just upgraded from the previous version.

EDIT: I typed this note the other day and forgot to press "Comment" so it has just been sitting here waiting for a keypress, :cry:  Anyway, wkitty52 and I replied to the user issue on the forum and think we have them straightened out.

gilbertohasnofb:
> I have cherry-picked the commits to release/2016.2 Point the core devs to that branch if they want to do a 2016.2.2 release.

Perfect, thanks a lot @onox!

gilbertohasnofb:
Ok, I sent the message to the list asking for a new merge.

But after I asked that, I realized that James Turner has made a commit of his own to the c172p in FGDATA.... sigh... @onox do you know what is this code he commit for? See: https://sourceforge.net/p/flightgear/fgdata/ci/a949a5b641df53acfcf531f91f43cbdbc5397fc1/ Also, we once made clear that commits involving the c172p should be made to _this_ repo, but apparently we need to state that again over there.

wlbragg:
@gilbertohasnofb  see https://sourceforge.net/p/flightgear/mailman/message/35166174/

Maybe we should wait and let it flush out before suggesting to make sure we get any additions or changes?

Then again, maybe not. But at the least we need a patch or notified so we can duplicate any changes.

wlbragg:
I sent a message to the dev list requesting, at the least, some notification of any changes/additions.

wkitty42:
On 06/20/2016 03:42 PM, wlbragg wrote:
> Then again, maybe not. But at the least we need a patch or notified so we can
> duplicate any changes.

IIRC, there was a post by james on the dev list which did talk about it... i 
don't recall if it was anything specific, though... i can try to look back in my 
email archive and see, if needed...

wlbragg:
>IIRC, there was a post by james on the dev list which did talk about it..

@wkitty42 are you referring to what the commit is for or to them notifying us of any changes?

See https://sourceforge.net/p/flightgear/mailman/message/35166174/ for the commit reasoning.

gilbertohasnofb:
>  But at the least we need a patch or notified so we can duplicate any changes.

> there was a post by james on the dev list which did talk about it...

Well, I don't read every single post on the mailing list. I think it's very simple: our repo here is where the development is happening. If anyone wants to contribute to the c172p, then he/she is more then welcome!, but please do it here. It's very annoying for us to find about these commits there (second time this happens) and having to then adjust our repo ourselves. If one can send a patch or PR to FGDATA, then certainly one can do the same to our repo.

gilbertohasnofb:
>  But at the least we need a patch or notified so we can duplicate any changes.

Yeah, that would be the minimal. But the notification should be done in a visible way, not among 1000 other messages in the list.

onox:
The version in FGDATA should be considered read-only, otherwise things will get overwritten when a new version is copied from github to FGDATA. So they should do what @gilbertohasnofb said:

> our repo here is where the development is happening. If anyone wants to contribute to the c172p, then he/she is more then welcome!, but please do it here

gilbertohasnofb:
And as a matter of fact, one of the files that James has modified will be overwritten in the next merge since we also modified it in this PR here, so it all becomes very annoying to everyone...

gilbertohasnofb:
@onox I wrote the following e-mail to the list, hope that's all right with you all:

> Hi all,
> 
> One more thing concerning the c172p: I have noticed yesterday that there has been a commit by James Turner done to the c172p in FGDATA. We already mentioned this before that we would really appreciate if anyone wishing to contribute to the project would do so in our repository and to consider the c172p in FGDATA as "read only". First of all, it's all to easy for us to miss a discussion about these commits here in the dev list or also not to realize that there is a possible commit in FGDATA affecting the plane, and so this work may even get lost like this when we do a next merge. In this very case, James has modified a file which we have already modified further since then, and so it will require a manual merge if that commit is to be kept. So if possible, please do contribute directly to our project, and we would absolutely welcome anyone wishing to improve the c172p!
> 
> Thanks a lot,
> Gilberto

wkitty42:
On 06/20/2016 04:29 PM, wlbragg wrote:
>     IIRC, there was a post by james on the dev list which did talk about it..
>
> @wkitty42 <https://github.com/wkitty42> are you referring to what the commit is
> for or to them notifying us of any changes?

you responded to the one i was thinking about with your second posting over 
there... i didn't see any need in adding any more to what you had already posted...

gilbertohasnofb:
@onox @wlbragg And of course Thorsten is the one who first replied to that message of mine, telling me that I am wrong and that he sees no issue and that's our problem to solve... sigh... Sorry guys but I can't handle him at all, would anyone care to enter that discussion? I have no interest in discussing with him given our history, but I think it's important to our project to make things clear over there...

gilbertohasnofb:
Curt replied in a very nice way and was very reasonable about the situation, so let's just ignore Thorsten in this regard. It seems it's all good now.

Now concerning James's commit, I will manually add them and create a PR to our project, ok?

wkitty42:
On 06/21/2016 11:31 AM, Gilberto Agostinho wrote:
> Curt replied in a very nice way and was very reasonable about the situation, so
> let's just ignore Thorsten in this regard. It seems it's all good now.
>
> Now concerning James's commit, I will manually add them and create a PR to our
> project, ok?

sounds OK to me :)

gilbertohasnofb:
@wkitty42 @wlbragg could any of you tell the guys in the forum that the AIS bug fix has been incorporated in FGDATA git today? So if they use it then a `git pull` will solve the bug.

wlbragg:
@gilbertohasnofb have you followed this conversation? If so, what is going on here?
I am afraid to comment because I don't completely understand the issue and I thought we had it fixed.
This is acting like the old issue before it was corrected.
DO you still have a fresh view of it in your mind as to what is the issue?

wlbragg:
Oops, forgot the link
https://forum.flightgear.org/viewtopic.php?f=4&t=30493

wkitty42:
i'm starting to wonder if there should be a line in the -set file or somewhere to explicitly set this property to true...

onox:
That's only a workaround. If we go down that road, we're gonna need to initialize all kinds of properties. And such a workaround can only be added after the new version has already been released and users have reported bugs.

The real problem is:

1. FG seems to serialize and save settings when you properly click 'Exit' and then 'Yes', but not when a user clicks the 'x' button in the window decoration.

2. Any new version of an aircraft may or may not be compatible with an older version. Thus FG should throw away the serialized and saved data file, just like it creates new autosave and navdata files.

wkitty42:
On 09/23/2016 06:31 PM, onox wrote:
> That's only a workaround. If we go down that road, we're gonna need to
> initialize all kinds of properties. And such a workaround can only be added
> after the new version has already been released and users have reported bugs.
>
> The real problem is:
>
>  1.
>
>     FG seems to serialize and save settings when you properly click 'Exit' and
>     then 'Yes', but not when a user clicks the 'x' button in the window decoration.

agreed...

>  2.
>
>     Any new version of an aircraft may or may not be compatible with an older
>     version. Thus FG should throw away the serialized and saved data file, just
>     like it creates new autosave and navdata files.

delete your aircraft-data/c172p.xml file and start clean... AFAIK this triggers 
the flaw because the property inits as FALSE which is the problem...


-- 
        *Please keep mailing list traffic on the list* unless
        private contact is specifically requested and granted.

onox:
> Delete your aircraft-data/c172p.xml file and start clean. AFAIK this triggers the flaw because the property inits as FALSE which is the problem.

I just deleted `aircraft-data/c172p.xml` and `/instrumentation/airspeed-indicator/serviceable` is true when I start FG.

wkitty42:
On 09/23/2016 09:31 PM, onox wrote:
>> Delete your aircraft-data/c172p.xml file and start clean. AFAIK this
>> triggers the flaw because the property inits as FALSE which is the
>> problem.
> I just deleted |aircraft-data/c172p.xml| and
> |/instrumentation/airspeed-indicator/serviceable| is true when I start FG.

interesting... as i understand it, the topic on the forum was concerning a brand 
new install where the aircraft-data xml file did not exist... i'll ask them to 
remove the file or rename it and try again to see what the property initializes 
as...

gilbertohasnofb:
Hi all, sorry for my late reply. Indeed deleting that data xml file shouldn't cause the issue, as our -set file does force the AIS serviceable property as true. See lines 614-616 of our set file:

```
        <airspeed-indicator>
            <serviceable type="bool">true</serviceable>
        </airspeed-indicator>
```

So I am at lost. I thought this was solved and that this PR would have prevented any issues, as even if the user would not exit the sim properly then our set file would still force the property to be `true`.

gilbertohasnofb:
Maybe it would be nice to take a look at that data xml file it self, to compare a healthy one with one that is breaking the plane and try to figure out what is going on. But indeed deleting this file should absolutely clear the problem, as I just tested and confirmed what @onox wrote above. So it's either something else causing the issue or the method of reproducing it is not correct or complete enough.

gilbertohasnofb:
And here is a guess of what might be going on: our set file does override the property to `true` at every execution, but it might be that if the data file is of an older version of the airplane it might then re-override it back to `false`. Then because those users keep exiting the sim without the ESC key, their data files keep that serviceable property still saved as `false` when it actually shouldn't have it saved at all. But this does not explain why deleting the file still causes the issue for some people, though as I said this really shouldn't be even possible and given that we can't reproduce it, more information will be necessary to solve this.

wkitty42:
On 09/24/2016 03:42 PM, Gilberto Agostinho wrote:
> And here is a guess of what might be going on: our set file does override the
> property to |true| at every execution, but it might be that if the data file is
> of an older version of the airplane it might then re-override it back to
> |false|.

another possible problem is that in this recent case on the forums, the user had 
edited some of the files for this craft and was switching between the older one 
they had edited and the new one with 2016.3.1... i don't know what it was that 
they edited or why and i don't recall which FGFS release the ASI problem was 
fixed in...

gilbertohasnofb:
@wkitty42 well, so far I had been assuming the user had a single FG installation per OS, was using our aircraft unmodified and was not playing with the xml files. In that case there is a million things that could have happened, and though we could troubleshoot it, it would not bring anything to the project itself as the problem probably lies elsewhere. So if no one can reproduce the issue with the conditions I wrote above, I think this is a non issue for us.

wkitty42:
On 09/25/2016 08:01 AM, Gilberto Agostinho wrote:
> @wkitty42 well, so far I had been assuming the user had a single FG
> installation per OS, was using our aircraft unmodified and was not playing
> with the xml files. In that case there is a million things that could have
> happened, and though we could troubleshoot it, it would not bring anything to
> the project itself as the problem probably lies elsewhere. So if no one can
> reproduce the issue with the conditions I wrote above, I think this is a non
> issue for us.

yeah, what they did was to upgrade from 2016.2.1 to 2016.3.1 and they kept their 
modified c172p from 2016.2.1... i don't recall if we fixed this bug before 
2016.2.1 or not...

now that we've figured out more of what's going on and the user removed the 
aircraft-data xml file and started clean without the problem, i agree that it is 
not an issue for us to deal with... i suspect it was user induced...

gilbertohasnofb:
:+1:

PR #780, opened 2016-06-03
New propeller sound

gilbertohasnofb:
New propeller sound, longer loop than the older `wasp.wav`, and comes from a RL recording. Maybe it still need some tweaking, so I'd appreciate opinions. @onox? @wlbragg? Anyone else?

Here is how an engine in a 172P sounds like, I tried matching the pitch of the sound the best I could using this video (particularly the idle pitch and the take off @full throttle pitch): https://www.youtube.com/watch?v=p6vH7oo7dUE

Closes https://github.com/Juanvvc/c172p-detailed/issues/779


gilbertohasnofb:
:warning: so before merging, let's discuss first, ok?

wlbragg:
I really like it, but I don't have any recent experience with real life prop sounds to be able to really offer much of an opinion. 

>and comes from a RL recording

Is it a c172? Where did it come from?

gilbertohasnofb:
> Is it a c172? Where did it come from?

gilbertohasnofb:
That's why I gave the video as reference, to compare the sounds. IMO this `wav` file is quite good for our plane! :smile:

onox:
I'm not 100 % convinced. The volume is a bit soft and too distant. It sounds more like a [DHC6](https://youtu.be/V_tZCKAG8Tg?t=99) than a 172P :worried: 

Can't we let @algefaen or @tigert / @wezzer record some sounds?

gilbertohasnofb:
If you think so. But I don't think the wav file sounds like the DHC6 at all, have you heard the clip by itself or after I tried implementing it in FG? Because I think the wav file is decent, but I am really not satisfied with the xml configuration yet.

About recording a 172, that would be great but the problem is that unless some of these guys have access to a proper audio recorder, I doubt the result will be good. Given the volume of a running engine inside the cockpit, you'd need to adjust the recorder so that there is no clipping, else the sample is useless. So I guess using a mobile phone would not work very well... just listen to how most of the 172 videos on Youtube sound like.

wezzer:
Sorry guys, noticed this thread late. Unfortunately I don't have any recording equipment nor access to C172 at the moment, so can't help you with recording the sounds.

tigert:
@wezzer Miika has one pretty good recorder, so if you guys have time and energy... of course that could be a nice reason to check out CTL on floats.. :)

gilbertohasnofb:
@tigert that would be ideal, I'd love to get my hands on some good recordings of a 172 (I actually found some excellent sounds online, but not GPL friendly... sigh...)

But my point is that until we do find better sounds (or record them ourselves) we are still using the dreadful `wasp.wav` with its outrageously short half second loop and poor sound quality. I am still up to try to improve this if you all think the same, but then I'd appreciate some feedback on how to proceed. For instance, I had posted a message saying that the implementation of this sound isn't really satisfactory yet, but that I think the cause is the xml file and not the wav file. Can the interested among you listen to the new wav file and compare it to the `wasp.wav` in the `FGDATA/Sounds` folder? Because if we agree that the new wav is indeed better and can be made to sound like a 172, then it's a matter of tweaking the xml further.

gilbertohasnofb:
If you guys do not like this new sound then should we close this PR? Or is anyone willing to give me a hand to improve this work, by tweaking the xml file? Either way is fine with me.

wlbragg:
>Or is anyone willing to give me a hand to improve this work, by tweaking the xml file? 

What did you have in mind?

gilbertohasnofb:
> What did you have in mind?

So: I think this new sound is quite decent by itself (listen to the .wav file), with the potential of being far better than our current default sound. So while we don't get a proper recording of some of us inside a 172 cabin (which may take a while...) I think we could make an effort to improve things with what we have in hand. So while the sound is all right by itself, it's not sounding very good at the sim. The trick here is that the pitch and volume of the sound are affected by some properties from engine, and it's a matter of trial and error to get something sounding decent. I have tried a couple of possibilities and this PR has my best attempt, but I still think it is not sounding good. So if anyone has the interest (and patience) to take a look at `c172-sound.xml` and play with the numbers, that would be great. I still want to give more tries myself, of course.

wlbragg:
I'm not interested in doing any of it myself (it's not anything I've got any experience with), but if you think there is a better finished sound here than what we have, I'll test your changes and give you my opinion!

gilbertohasnofb:
@wlbragg Fair enough! Once I have the chance to do some more work on this I will let you know :smile:

gilbertohasnofb:
I am going to give up working on this PR, I just can't make it sound realistic... the new sound itself is much better, has higher quality and has a much longer loop than the default `wasp.wav` (if you pay attention to it, it has a mere 1 second loop, which can drive a person who cares for sound completely insane) but I simply can't make it sound good in the sim... :sob: 

Let's hope that someone can get us some good recordings of a 172 in the future.

PR #787, opened 2016-06-21
merging James' work

### LINKED ISSUE #786: In-air vs on-ground starts
Merge James Turner's commit for In-air vs on-ground starts

gilbertohasnofb:
Merges James' commit to the c172p in FGDATA

closes #786


gilbertohasnofb:
@onox I think it would be great to have this added to the 2016.2 release branch so that once we have the new merge (to fix the AIS) this work will be back in FGDATA

gilbertohasnofb:
:warning: Please do not merge yet. It seems that the devs use the c172p in FGDATA for making tests of new features and other things, and this does not necessarily mean that the tests are bringing new features to the aircraft. It seems this commit might be the case, and I am waiting for a confirmation of it. If that's the case, then we just close this PR and don't bother merging it.

Also, there is some interesting discussions about our plane going on in the list, I would suggest anyone interested to follow it and maybe give your 2 cents there as well. Some proposed to use a different aircraft for tests as not to mess up to mess up our work, which is great.

gilbertohasnofb:
@onox @wlbragg are you guys following the talks in the list? Maybe you could step in to give your 2 cents.

wlbragg:
Yeah, I would really like @onox to get his voice in this. I don't have the git skills or probably even the access to do any of what is being discussed without creating my own separate repository. I'm not really excited about having to learn a bunch new git magic. But I do think we need to be flexible and take the path that is most convenient for all concerned.

gilbertohasnofb:
Ok, so if I understood correctly I think we should merge this work, so that we would be back in sync with FGDATA (@wlbragg can you confirm you understood the same from the discussion?)

gilbertohasnofb:
Can anyone please merge this so we are in sync with FGDATA?

onox:
I'll add some commit and then merge it in a minute.

onox:
To be honest, I don't really see why this should be merged because they seem to be test files.

onox:
Or the `<overlay>` elements should be fixed by someone.

gilbertohasnofb:
> To be honest, I don't really see why this should be merged because they seem to be test files.

Yes, they are test files, but James is working on them now, and if our repo is to be in sync with FGDATA then they should be here as well. This avoids any chance of any work being lost during our constant merges (now every three months). It also allows us to keep an eye if any feature we are implementing here will not conflict when merged with any work done over there.

PR #796, opened 2016-07-17
Bug 795: External hotspots not clickable from cockpit any longer

gilbertohasnofb:
Closes: https://github.com/Juanvvc/c172p-detailed/issues/795

When pressing Ctrl+C, external hotspots do not show if they are hidden by the fuselage, but if you click in the correct position they do work (e.g. it's possible to open the oil dialog or add/remove chocks to the front wheel from the pilot's view).

This PR adds conditional for these so that they are only clickable if the view is external.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -7194,6 +7194,9 @@
             <binding>
                 <condition>
                     <property>/sim/model/c172p/securing/pitot-cover-removable</property>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
                 </condition>
                 <command>property-assign</command>
                 <property alias="/params/securing/pitot-cover"/>
@@ -7204,6 +7207,9 @@
             <binding>
                 <condition>
                     <property>/sim/model/c172p/securing/pitot-cover-removable</property>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
                 </condition>
                 <command>set-tooltip</command>
                 <tooltip-id>remove-pitot-tube-cap</tooltip-id>
@@ -7220,6 +7226,12 @@
             <binding>
                 <condition>
                     <property>/sim/model/c172p/securing/pitot-cover-addable</property>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
                 </condition>
                 <command>property-assign</command>
                 <property alias="/params/securing/pitot-cover"/>
@@ -7235,6 +7247,9 @@
             <binding>
                 <condition>
                     <property>/sim/model/c172p/securing/pitot-cover-addable</property>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
                 </condition>
                 <command>set-tooltip</command>
                 <tooltip-id>put-pitot-tube-cap</tooltip-id>
@@ -7340,6 +7355,11 @@
             <button>0</button>
             <repeatable>false</repeatable>
             <binding>
+                <condition>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
+                </condition>
                 <command>property-assign</command>
                 <property>/sim/model/c172p/securing/chock</property>
                 <value>false</value>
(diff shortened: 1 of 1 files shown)


Models/c172p.xml
```diff
@@ -7220,6 +7226,12 @@
             <binding>
                 <condition>
                     <property>/sim/model/c172p/securing/pitot-cover-addable</property>
+                    <not>
+                        <property>/sim/current-view/internal</property>
+                    </not>
+                    <not>
```
onox: This one seems to be a duplicate.

gilbertohasnofb:
@onox my bad, fixed now.

onox:
@gilbertohasnofb If "Allow securing aircraft" is false, then you should not be able to click on the tiedowns (left, right, tail) or the front wheel.

Also, pitot tube is yellow in internal view.

onox:
If the chocks are visible, you shouldn't be able to click on the wheel (it doesn't do anything anyway).

onox:
I think you can use the `-addable` properties?

onox:
Fuel contamination and oil cap seem to work OK :+1:

gilbertohasnofb:
Taking a look at it now.

gilbertohasnofb:
@onox are you sure what you are saying is possible in FG?

Please notice that the oil cap and the fuel contamination are invisible objects that are use as hot spots only, and hiding them when those conditions aren't met make them invisible. But the front wheel does not get hidden if securing aircraft is not true. This is how I implemented it:

```
    <animation>
        <type>pick</type>
        <object-name>NoseWheel</object-name>
        <object-name>NoseWheelBK</object-name>
        <action>
            <button>0</button>
            <repeatable>false</repeatable>
            <binding>
                <condition>
                    <property>/sim/model/c172p/securing/chock-addable</property>
                    <not>
                        <property>/sim/current-view/internal</property>
                    </not>
                </condition>
                <property>sim/model/c172p/securing/chock</property>
                <value>true</value>
            </binding>
        </action>
        <hovered>
            <binding>
                <condition>
                    <property>/sim/model/c172p/securing/chock-addable</property>
                    <not>
                        <property>/sim/current-view/internal</property>
                    </not>
                </condition>
                <command>set-tooltip</command>
                <tooltip-id>secure-with-wheel-chock</tooltip-id>
                <label>Secure With Wheel Chock</label>
            </binding>
        </hovered>
    </animation>
```

So you see, both the action and the hovered properties of the pick animation are using the conditional with the `-addable` property. When the "allow securing" is disabled, the hover message is disabled, but the wheel keeps behaving as if it's clickable and appears yellow with Ctrl+C. This seems to me to be a problem with FG and not with this xml code, unless I am missing something obvious here.

onox:
So FG shows a yellow outline even if there's no binding that can execute. Lets merge it then.

gilbertohasnofb:
That's what I believe, yes. And it also changes the cursor in that case.

wlbragg:
>But the front wheel does not get hidden if securing aircraft is not true.

Maybe we should not use an "actual" object as a hot spot? Only crate non-object hot spots (box around the wheel)?

gilbertohasnofb:
> Maybe we should not use an "actual" object as a hot spot? Only create non-object hot spots (box around the wheel)?

The problem with that is that the objects themselves are not boxes, and I think it leads to two problems: 1) looks extremely ugly under Ctrl+C. 2) it leads to confusion sometimes since from certain angles the boxes may be quite larger than the objects themselves, effectively making an invisible area clickable. I'd rather have the mouse cursor change than to deal with those two.

onox:
What if you add `<visible>false</visible>` to the pick/knob animation of the fake object?

gilbertohasnofb:
What is a fake object?

gilbertohasnofb:
Do you mean the HotSpots? Then it would solve 1), but not 2)

onox:
uhh, what I mean is you duplicate the wheel and attach an invisible knob animation to it.

gilbertohasnofb:
Yeah, that might work (the new invisible object would have to be slightly larger), though I believe you cannot click on invisible objects. Also, it would raise the polygon count.

onox:
That's true, ideally FG should make it configurable to enable/disable the yellow outline of particular objects.

gilbertohasnofb:
Or more obviously: if the pick animation condition is `false`, then the cursor shouldn't change and the object shouldn't show up in yellow under Ctrl+C

wlbragg:
Yes, that makes the most sense.

gilbertohasnofb:
I will report this to the devel-list then.

gilbertohasnofb:
Reported.

gilbertohasnofb:
James Turner wrote:

> This is ‘just a bug’ in the existing hovered conditional, should be fixable.
> I will try to remember to take a look this weekend, but if I forget please remind me

PR #805, opened 2016-08-07
Better yoke model

### LINKED ISSUE #606: Improve yoke model
Improve yoke model. Currently our model is very simplistic, from a frontal view it is basically a simple rectangle while the typical yoke for these old 172 models would look more like this:

Another image, by @tigert: 
aag

gilbertohasnofb:
Closes #606 

Previously our yoke was a simple square box, but in reality they are much rounder. This is what I could achieve on Blender:

fgfs-screen-002


gilbertohasnofb:
This is not perfect by any means, but I believe it's better than what you had before. @wlbragg what do you think about it?

onox:
The yoke looks better. The path to the texture in the `.blend` file is not relative.

gilbertohasnofb:
> The path to the texture in the .blend file is not relative.

Hmm, I don't know how to handle that in Blender :cry: Would you be able to explain to me or to fix it yourself?

wlbragg:
I look at this.

gilbertohasnofb:
@wlbragg thanks for taking a look at this! On `master`, do:

```
git pull
git checkout bug-606
```

that should be it :wink:

onox:
> Hmm, I don't know how to handle that in Blender :cry: Would you be able to explain to me or to fix it yourself?

I prefer to teach you how to fish instead of giving you a fish:

1. Open the user preferences and check under `File` that you have enabled `Relative paths`.

2. Click on the object `YokeMount` in the outline view (the object tree)

3. Click on the texture tab (the red/white chess board)

4. You should see various groups in the tab: `Preview`, `Colors`, `Image`, etc. (click on the other tabs first if you see very little, may be a bug in Blender 2.77)

5. Under `Image` you should see `source` with the absolute path to yoke.png. Click on the icon that opens the file browser.

6. In the file browser you should see on the left a check button `Relative Path`. Enable it if you see it.

7. Path should now be: `//../../Models/Interior/Yoke/yoke.png`.

onox:
When you enable texture in the 3D view (not solid, wireframe or material, but texture) you should see the texture instead of a pink object.

gilbertohasnofb:
@wlbragg by the way, editing older messages on GitHub is fine but we don't get notifications for it. So sometimes it's better to add a new comment to make sure someone will read it!

@onox I will give it a try. But fishing with Blender is quite a pain, you have no idea how long it took me just to change the shape of this yoke... :tired_face:

gilbertohasnofb:
@onox thanks for the point by point explanation! But just confirming the path should really be`//../../Models/Interior/Yoke/yoke.png` with the double `//` ?

And I assume I should do the same for all objects in that blend file, right?

onox:
Yes, the `//` indicates it is a relative path.

onox:
I think all the objects share the same image in that particular blend file, so they should all be fixed. Enable texture view in the 3D view to confirm.

gilbertohasnofb:
> I think all the objects share the same image in that particular blend file, so they should all be fixed. Enable texture view in the 3D view to confirm.

:+1:

Testing and pushing in a second!

gilbertohasnofb:
@onox done

onox:
Much better!

gilbertohasnofb:
Thanks!! :smile: 

Ok, so now I will just fix that splash screen and then I am done for today :sleepy:

onox:
Good work @gilbertohasnofb :)

gilbertohasnofb:
@onox sorry but I just realized that there is an issue with the yoke, look at how the handles connect to the main body:

fg

I will fix this on the same branch

gilbertohasnofb:
Strange, on Blender it looks fine:

screenshot from 2016-08-07 19 11 22

onox:
I think you can apply a boolean modifier to the handle to remove the vertices inside the yoke.

gilbertohasnofb:
Nevermind, it was just a bug while reloading the Aircraft Model without quitting the sim. It's all good:

fg-1

So now splash screen!

onox:
Ok! It looked ok here.

gilbertohasnofb:
Perfect :smile: 

Thanks for reviewing all these tons of PRs, today I am in a maniac state :wink:

gilbertohasnofb:
Hmm, indeed there is a problem... the body of the yoke is not moving... sigh:

fg-2

gilbertohasnofb:
So now it moves. I also confirmed the yokes can be hidden, get affected by the flashlight and irradiance effect. Should be good now.

PR #806, opened 2016-08-07
clock now displays local time

### LINKED ISSUE #747: Feature Request: display local time
See
https://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=2595#p284594
and the potential solution is in the next post.
https://forum.flightgear.org/viewtopic.php?f=4&t=25157&start=2595#p284596

gilbertohasnofb:
Closes #747

### DIFF
--- a/Systems/instruments.xml
+++ b/Systems/instruments.xml
@@ -359,5 +359,24 @@
             <property>/sim/model/c172p/lighting/trans</property>
         </output>
     </filter>
+    
+    <filter>
+        <name>Clock: UTF to local time conversion</name>
+        <type>gain</type>
+        <input>
+            <expression>
+                <sum>
+                    <property>sim/time/utc/day-seconds</property>
+                    <property>sim/time/local-offset</property>
+                </sum>
+            </expression>
+        </input>
+        <input>
+            <value>0.0</value>
+        </input>
+        <output>
+            <property>/sim/model/c172p/local-time-seconds</property>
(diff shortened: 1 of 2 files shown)


Systems/instruments.xml
```diff
@@ -359,5 +359,24 @@
...
+        <type>gain</type>
+        <input>
+            <expression>
+                <sum>
+                    <property>sim/time/utc/day-seconds</property>
+                    <property>sim/time/local-offset</property>
+                </sum>
+            </expression>
+        </input>
+        <input>
```
onox: This second `<input>` would only be useful if the previous one had a `<condition>`. Now it is useless.

Systems/instruments.xml
```diff
@@ -359,5 +359,24 @@
             <property>/sim/model/c172p/lighting/trans</property>
         </output>
     </filter>
+    
+    <filter>
+        <name>Clock: UTF to local time conversion</name>
+        <type>gain</type>
+        <input>
+            <expression>
```
onox: Instead of using `<expression>` you could try:

```xml
<input>
    <property>sim/time/utc/day-seconds</property>
    <offset>
        <property>sim/time/local-offset</property>
    </offset>
</input>
```

gilbertohasnofb:
@onox I am fixing those now

gilbertohasnofb:
@onox done

Issue #36, opened 2015-03-12
battery drain when alternator is off 


gilbertohasnofb:
XML15 - if an alternator is turned off (such by a failure), the battery should slowly drain all its energy over a certain period of time (minimally 30 minutes, probably more for a good working battery), causing a complete electrical failure. In case it was not a failure and the alternator is turned on again, the battery should recharge back over time

Juanvvc:
Actually, this is already modelled but the battery has (IMHO) too much capacity.

gilbertohasnofb:
By what I read, a battery minimal duration should be 30 minutes, but in reality it should last substantially longer. Do you know how much is the current capacity of the battery in minutes?

Juanvvc:
According to my instructor, less than 5 minutes or you have to use external power. Probably, the battery on my C172 was not the best :)

gilbertohasnofb:
Hmm, I don't know how different are these laws in different countries, but the FAA states that "In the event of a complete loss of the primary electrical power generating system, the battery must be capable of providing electrical power to those loads that are essential to continued safe flight and landing for [a]t least 30 minutes for airplanes that are certificated with a maximum altitude of 25,000 feet or less"

Source: http://rgl.faa.gov/Regulatory_and_Guidance_Library/rgFAR.nsf/0/B68CC9734E940D1F86257996006CB148

Juanvvc:
Ok, I must clarify my comment :)

The electrical system that needs most power is the ignition system. Once the engine is running, you only have to feed a few instruments, lights and radio system. If the alternator fails and according to the FAA provisions, the battery should be able to keep runnin "those loads that are essential to continued safe flight". This means: switch COM2, ADF and DME off, transponder OFF, all lights OFF and consider switching COM1 off after the PANPAN message (if any) This way, you save your power for instruments and flaps for half an hour and probably more. On ground, switch all lights on and use the starter 2-3 times and your batteries will be empty in less than 5 minutes.

Hence, it is all about loads for the electrical system. Currently, loads were estimated by the original author:

- Battery: 12 A h, "ideal amps" 30A, 24V.
- Master breaker: accepts loads up to 55A (breaker implemented by us, but the limit is not checked yet)
- Starter: 12A (that means: engine could run for 1 hour using only electric power. Very unrealistic!)
- Main lights: 2-5A each (beacon, navigation, strobe, landing all together ~15A, or ~360W )
- Radios use 0.5A
- Other instruments use 1-2A

[...]

Issue #38, opened 2015-03-12
better modelling of Loadmeter


gilbertohasnofb:
XML17 - better modelling of Loadmeter: when both alternator and battery are switched on it reads ~33 V (almost an electric overload, I believe), which is already beyond the yellow range. Turning lights, radios and instrument on and off currently make no difference, but I believe it should raise the load of the system

gilbertohasnofb:
With battery and alternator both on, we currently have the loadmeter reading a constant ~27 V regardless of the load, and a constant 24.5 V if the alternator is turned off, both regardless of the equipments being used (lights, avionics, etc., make no difference in these readings). 

The ammeter seems to work well, but shouldn't a negative read of the ammeter (such as when the alternator is off) cause the amount of voltage from the battery to drain?

gilbertohasnofb:
@callahanp the property `/systems/electrical/volts` will still be in use and will have the correct voltage value. The only difference is that our cockpit will no longer have a gauge displaying that value. But as long as you use that property in your cockpit you should be fine.

Issue #37, opened 2015-03-12
better modelling of Ammeter


gilbertohasnofb:
XML16 - better modelling of Ammeter: it should read zero if battery fully recharged and alternator on, or show the positive flow of current from the alternator to the battery if the latter is not fully recharged, or show the negative flow of energy being drained from the battery if alternator fails or is turned off

gilbertohasnofb:
I believe this has been implemented, can I close this issue?

onox:
Shouldn't the number of electrical devices that are active have an effect on the amps? It always goes to 7 currently (no pun intended).

gilbertohasnofb:
I don't know, but they certainly should have an effect on the voltage, see: https://github.com/Juanvvc/c172p-detailed/issues/38

gilbertohasnofb:
Related 'kind of' duplicate: https://github.com/c172p-team/c172p-detailed/issues/1017

I am copying the relevant information on the description of this issue here.

PR #810, opened 2016-08-07
Battery

### LINKED ISSUE #801: Add a recharge battery button to aircraft menu
Add a recharge battery button to aircraft menu. Take a look how this has been implemented in the ~~c182s~~ c150.

gilbertohasnofb:
This PR closes #801, and possibly #36, ~~37 and 38~~ (please confirm it)

- aircraft dialog show battery charge in percentage
- aircraft dialog has a recharge battery button
- battery charge is now persistent among sessions
- the user can disable the option above (that is, every time the sim is loaded it also sets the battery charge to 100%)
- battery voltage now drains! Before it was always stuck at 24.5 due to a mistake at the `electrical.nas`
- aircraft starter won't work if the battery is too low (to test it, turn the master ON and let the battery drain using simulation acceleration by pressing `a` several times)

### DIFF
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -67,7 +67,7 @@ BatteryClass.new = func {
                 ideal_volts : 24.0,
                 ideal_amps : 30.0,
                 amp_hours : 12.75,
-                charge_percent : 1.0,
+                charge_percent : getprop("/systems/electrical/battery-charge-percent") or 1.0,
                 charge_amps : 7.0 };
     setprop("/systems/electrical/battery-charge-percent", obj.charge_percent);
     return obj;
@@ -79,7 +79,7 @@ BatteryClass.new = func {
 #
 
 BatteryClass.apply_load = func (amps, dt) {
-    var old_charge_percent = getprop("/systems/electrical/battery-charge-percent");
+    var old_charge_percent = me.charge_percent;
 
     if (getprop("/sim/freeze/replay-state"))
         return me.amp_hours * old_charge_percent;
@@ -90,8 +90,8 @@ BatteryClass.apply_load = func (amps, dt) {
     var new_charge_percent = std.max(0.0, std.min(old_charge_percent - percent_used, 1.0));
 
     if (new_charge_percent < 0.1 and old_charge_percent >= 0.1)
-        gui.popupTip("Warning: Low battery! Enable alternator or apply external power to recharge battery!", 10);
-
+        gui.popupTip("Warning: Low battery! Enable alternator or apply external power to recharge battery!", 10);       
+    me.charge_percent = new_charge_percent;
     setprop("/systems/electrical/battery-charge-percent", new_charge_percent);
     return me.amp_hours * new_charge_percent;
 }
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -71,7 +71,7 @@ BatteryClass.new = func {
     var obj = { parents : [BatteryClass],
                 ideal_volts : 24.0,
                 ideal_amps : 30.0,
-                amp_hours : 12.75,
+                amp_hours : 25.00,
                 charge_percent : getprop("/systems/electrical/battery-charge-percent") or 1.0,
                 charge_amps : 7.0 };
     setprop("/systems/electrical/battery-charge-percent", obj.charge_percent);
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -71,7 +71,7 @@ BatteryClass.new = func {
     var obj = { parents : [BatteryClass],
                 ideal_volts : 24.0,
                 ideal_amps : 30.0,
-                amp_hours : 25.00,
+                amp_hours : 3.1875,
                 charge_percent : getprop("/systems/electrical/battery-charge-percent") or 1.0,
                 charge_amps : 7.0 };
     setprop("/systems/electrical/battery-charge-percent", obj.charge_percent);
@@ -357,7 +357,7 @@ var electrical_bus_1 = func() {
             # starter
             if ( getprop("controls/switches/starter") ) {
                 setprop("systems/electrical/outputs/starter", bus_volts);
-                load += 12;
+                load += 24;
             } else {
                 setprop("systems/electrical/outputs/starter", 0.0);
             }
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -31,6 +31,11 @@ var init_electrical = func {
     # Request that the update function be called next frame
     settimer(update_electrical, 0);
     print("Electrical system initialized");
+    
+    # checking if battery should be automatically recharged
+    if (!getprop("/systems/electrical/save-battery-charge")) {    
+        battery.reset_to_full_charge();
+    };   
 }
 
 var reset_battery_and_circuit_breakers = func {
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -84,7 +84,7 @@ BatteryClass.new = func {
 #
 
 BatteryClass.apply_load = func (amps, dt) {
-    var old_charge_percent = me.charge_percent;
+    var old_charge_percent = getprop("/systems/electrical/battery-charge-percent");
 
     if (getprop("/sim/freeze/replay-state"))
         return me.amp_hours * old_charge_percent;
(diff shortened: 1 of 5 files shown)


onox:
Could you move the text in the dialog to the left of the button? Dialog window is already getting very tall.

gilbertohasnofb:
:+1:

gilbertohasnofb:
Done: 

fg

Systems/electrical.xml
```diff
@@ -0,0 +1,17 @@
...
+
+<PropertyList>
+
+    <filter>
+        <type>gain</type>
+        <gain>100.0</gain>
+        <input>
+            <property>/systems/electrical/battery-charge-percent</property> <!-- this property ranges from 0.0 to 1.0 despite its name -->
+        </input>
+        <reference>0.0</reference>
```
onox: What's the purpose of this `<reference>` element? Since it is subtracted from the input, it doesn't do anything.

gilbertohasnofb: Oops, I will remove it. I used another filter as template and I just changed the values to output the correct thing without thinking if the statements were necessary. Sorry!

gilbertohasnofb: Done.

Nasal/electrical.nas
```diff
@@ -79,7 +84,7 @@ BatteryClass.new = func {
 #
 
 BatteryClass.apply_load = func (amps, dt) {
-    var old_charge_percent = getprop("/systems/electrical/battery-charge-percent");
+    var old_charge_percent = me.charge_percent;
```
onox: Could you elaborate on this change? The reason it reads from  a property is so that there's no hidden state in Nasal (`me.charge_percent`) and the flight recorder can record the property. (Add a `debug.dump(old_charge_percent);` statement to test)

gilbertohasnofb: I did these two modifications to `BatteryClass.apply_load` after comparing it with the same function in the c150. For some reason, our battery was constantly stuck at 24.5 V even though the percentage was draining, and even when the percentage was at 0% the systems would all work as if the battery was fully charged. I then tried changing this lines in the hope that they might fix the problem, and indeed they did.

onox:
If you add `debug.dump(old_charge_percent);` to `BatteryClass.apply_load` you'll see that during replay it will stay at the level of the battery just before you entered replay. (Because you no longer read from the property)

The value in the dialog still functions correctly it looks like.

onox:
(I hope you can read the inline comment in `Nasal/electrical.nas`)

onox:
> battery voltage now drains! Before it was always stuck at 24.5 due to a mistake at the electrical.nas

Which code causes the voltage to drop?

Nasal/electrical.nas
```diff
@@ -90,8 +95,8 @@ BatteryClass.apply_load = func (amps, dt) {
     var new_charge_percent = std.max(0.0, std.min(old_charge_percent - percent_used, 1.0));
 
     if (new_charge_percent < 0.1 and old_charge_percent >= 0.1)
-        gui.popupTip("Warning: Low battery! Enable alternator or apply external power to recharge battery!", 10);
-
+        gui.popupTip("Warning: Low battery! Enable alternator or apply external power to recharge battery!", 10);
```
onox: You have added some spaces at the end of the line.

gilbertohasnofb: I will remove that.

Nasal/electrical.nas
gilbertohasnofb: @onox my bad again :disappointed: This line didn't need to be changed, the line that makes the trick is the new line 99 below (`me.charge_percent = new_charge_percent;`). So should I revert this line and remove those spaces added below?

gilbertohasnofb:
@onox done

onox:
I think the capacity of the battery should be reduced, see #36.

gilbertohasnofb:
Do you mean reduced to 24 V instead of 24.x V that we have now (I don't remember the first decimal value)?

onox:
No, I mean we reduce`amp_hours : 12.75,` to 25 5.

About #37: the amps is still 7 when battery is at 100.0 %.

gilbertohasnofb:
> No, I mean we reduceamp_hours : 12.75, to 25 5.

I see, I can do that right now

> the amps is still 7 when battery is at 100.0 %.

What do you mean? Should the amperage be different when the battery is fully charged? As for what you wrote here https://github.com/Juanvvc/c172p-detailed/issues/37#issuecomment-158407765 I think you are probably right, but that would require quite a lot of work on the electrical system. Do we want to do it in this PR as well?

onox:
Don't think #38 has been fixed either; there's no change in the voltage. I'm not an electrical engineer, but AFAIK when something like the landing light is turned on, the voltage should be reduced shortly because of the sudden additional load. But I think this is difficult to model.

gilbertohasnofb:
The thing those issues become quite large after all the discussions, but the original thing was an issue with the gauges themselves I believe. Given that the battery now is draining properly, the gauges are displaying everything properly. The problem is that our instruments seem not to be affecting those values, but that's not exactly a problem with the loadmeter.

I would propose that we simply close all those Jurassic reports and open a new one with a clear indication: electrical systems needs X and Y and Z

onox:
> amp_hours : 25.00,

No this is wrong! Now you have doubled the capacity.

gilbertohasnofb:
> No, I mean we reduceamp_hours : 12.75, to 25 5.

?

onox:
Sorry that 5 should have been a %.

onox:
I'll revert the fix.

gilbertohasnofb:
please do, then we change it to `3.1875` right? (25% of `12.75`)

onox:
Pushed. I have doubled the amps of the starter as well.

gilbertohasnofb:
:+1:

onox:
Should we disallow recharging the battery while in the air?

gilbertohasnofb:
So what in your opinion should be done still on this PR? And what do you think about my proposal of opening a new issue for fixing whatever is left with the electrical system (e.g.  number of electrical devices that are active having an effect on the amps) and close those older issues?

Or if you are up for fixing everything here I am all for it, I can try to help the best I can :smile:

onox:
I think we can close #801 and #36, but I would leave #37 and #38 just open.

onox:
I'm not sure those latter two issues are easily fixable. Ideally FG should provide a system that properly simulates electrical networks, otherwise aircraft devs are reinventing the wheel all the time.

gilbertohasnofb:
Sounds good to me. To shall we merge this?

onox:
https://github.com/Juanvvc/c172p-detailed/pull/810#issuecomment-238092256

gilbertohasnofb:
I missed that comment.

> Should we disallow recharging the battery while in the air?

Yep! Or is it better to disable it if the engine is running? Would anyone recharge a battery with the engine running??

onox:
Perhaps add both conditions? disallow if in air or engine is running?

gilbertohasnofb:
:+1: And after that, is there anything else yet to be done here?

onox:
Don't think so.

gilbertohasnofb:
Ok!

wlbragg:
@onox, are you interested in adding Thorsten's new external runway lights?

If so I have the skinny on how to do it. I just don't want to take the time off of what I am doing with the Shuttle to work on it right now. But if you want to try to tackle it I'll create a new issue with the information.
If not I'll tackle it later.
Forgive me for hijacking this thread.

onox:
@wlbragg You could open a new issue for this with instructions how to add it. I'm personally also spending most of my time on other projects now.

wlbragg:
:+1:

gilbertohasnofb:
@onox done

wkitty42:
On 08/07/2016 12:34 PM, Gilberto Agostinho wrote:
> I missed that comment.
>
>     Should we disallow recharging the battery while in the air?
>
> Yep! Or is it better to disable it if the engine is running? Would anyone
> recharge a battery with the engine running??

isn't there some sort of charging system in the craft that brings the batter 
back up to its normal charge state? like in a car?

gilbertohasnofb:
> isn't there some sort of charging system in the craft that brings the batter back up to its normal charge state? like in a car?

Sure, that's the alternator, and it's doing it's job. In flight you should have your battery completely charged (check the % via the aircraft menu)

wkitty42:
On 08/07/2016 01:58 PM, Gilberto Agostinho wrote:
>     isn't there some sort of charging system in the craft that brings the batter
>     back up to its normal charge state? like in a car?
>
> Sure, that's the alternator, and it's doing it's job. In flight you should have
> your battery completely charged (check the % via the aircraft menu)

ok, i thought there was... i probably misunderstood the conversation being 
related only to this one small portion of the process... i've been up since 
03:30 and really need to go take a nap but that's not gonna happen today :( :( :(

gilbertohasnofb:
No problems :smile:

PR #817, opened 2016-08-07
Add Ground Equipment

### LINKED ISSUE #717: consider using static objects
I think we could consider using static objects around the aircraft, just like in https://github.com/HHS81/c182/issues/55.

See for instance:

gilbertohasnofb:
Closes #717 

- adds Ground Equipment menu
- from there you can activate/deactivate cones, ladder, ground power unit and fuel truck
- ladders can be climbed by the walker!

fg


onox:
@gilbertohasnofb I've refactored the code a bit to remove the code duplication. Changed the GUI dialog so that it looks more like the other dialogs like Aircraft Options.

gilbertohasnofb:
@onox Ok! 

> Changed the GUI dialog so that it looks more like the other dialogs like Aircraft Options.

Looks much better, thanks :+1:

PR #818, opened 2016-08-07
fixing issue with buttons

### LINKED ISSUE #814: Fix repair button in the aircraft menu
Fix repair button in the aircraft menu.

gilbertohasnofb:
Closes #814 

Repair and Recharge Battery buttons in the Aircraft Menu should now work when the aircraft has crashed

### DIFF
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -160,12 +160,6 @@
                         <not>
                             <property>/sim/freeze/replay-state</property>
                         </not>
-                        <or>
-                            <property>/fdm/jsbsim/gear/unit[1]/WOW</property>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
For some reason, when the aircraft has crashed the condition below is never true:

```
<or>
    <property>/fdm/jsbsim/gear/unit[1]/WOW</property>
    <property>/fdm/jsbsim/gear/unit[19]/WOW</property>
    <property>/fdm/jsbsim/gear/unit[23]/WOW</property>
    <property>/fdm/jsbsim/hydro/active-norm</property>
</or>
```

Anyway we don't need to detect contact with the ground, since we already have:

```
<less-than>
    <value>1.0</value>
</less-than>
```

Removing it makes the buttons available again when the crashed aircraft has come to a halt.

PR #821, opened 2016-08-10
Kr87 knob now works while the sim is paused

### LINKED ISSUE #401: ADF frequency cannot be changed while the sim is paused
Unlike the NAV and COM knobs, the ADF frequency cannot be changed while the sim is paused.

gilbertohasnofb:
Closes #401 

Now ADF frequency knobs (inner and outer) as well as the swap frequency button work while the sim is paused

### DIFF
--- a/Models/Interior/Panel/Instruments/kr87-adf/kr87.xml
+++ b/Models/Interior/Panel/Instruments/kr87-adf/kr87.xml
@@ -246,31 +246,20 @@
     </animation>
 
     <animation>
+        <type>select</type>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
@vainiovano thanks for the report, I can reproduce the bug here as well. I opened a new issue for it, see https://github.com/c172p-team/c172p-detailed/issues/885

> Could this change be reverted while searching for a better solution?

If you are in a rush for a fix, I suggest you revert it locally, but I would rather take some time to understand what is the problem and fixing it.

PR #823, opened 2016-08-11
Bug 30: Hide most instruments further than 25 m and radio stack 60 m

### LINKED ISSUE #30: Check all interior objects are NOT selected if the viewer is further than 50m

onox:
Fixes #30

Most instruments can be hidden at range higher than 25 meter. The radio stack is hidden beyond 60 meter because it shows a distinguishable gray background.

### DIFF
--- a/Models/Interior/Panel/Instruments/EGT/EGT.xml
+++ b/Models/Interior/Panel/Instruments/EGT/EGT.xml
@@ -32,12 +32,6 @@
   </params>
 
   <animation>
-    <type>range</type>
-    <min-m>0</min-m>
-    <max-m>10</max-m>
-  </animation>
-  
-  <animation>
     <type>material</type>
     <object-name>Needle</object-name>
     <object-name>Bug</object-name>
--- a/Models/Interior/Panel/Instruments/kr87-adf/ki227_228.xml
+++ b/Models/Interior/Panel/Instruments/kr87-adf/ki227_228.xml
@@ -32,13 +32,6 @@
   </animation>
 
   <animation>
-    <type>range</type>
-    <min-m>0</min-m>
-    <max-m>10</max-m>
-  </animation>
-
-  
-  <animation>
     <type>material</type>
     <object-name>CompassRose</object-name>
     <object-name>KI227.ADFNeedle</object-name>
--- a/Models/Interior/Panel/Instruments/kr87-adf/kr87.xml
+++ b/Models/Interior/Panel/Instruments/kr87-adf/kr87.xml
@@ -77,11 +77,6 @@
     <animation>
         <type>noshadow</type>
     </animation>
-    <animation>
-        <type>range</type>
-        <min-m>0</min-m>
-        <max-m>10</max-m>
-    </animation>
 
     <animation>
         <type>material</type>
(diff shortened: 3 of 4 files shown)


gilbertohasnofb:
@onox This image below is how the panel is looking like at the external view, zoomed in:

fg

Is this intentional? It seems that zooming and moving the camera are two different things, and unfortunately the default external view is already at a position which partially hides the cockpit

onox:
Are you using the mouse wheel to zoom in and out?

onox:
What you are doing (I think) is zooming in with X key.

gilbertohasnofb:
I am using the `x` key indeed, and I am aware that the mouse wheel changes distance and `x` changes the field of view (so the first is like walking closer to a subject, while the second is kind of like using a stronger zoom lens). I am just saying that the distance for the default external view is already too far away to show an incomplete cockpit, and I think this will affect a lot of screenshots which I don't like. I am just thinking if it would not be better to use the same minimal distance values that are used for the radios and AI for all other gauges and objects so that at least the default view isn't affected.

onox:
I chose 25 m so that if you want to fly with the external view, you get a bit better fps at that distance.

onox:
You can set it to a larger distance if you want.

gilbertohasnofb:
Yeah, I get the reason, but I don't like the side effect too much :cry:

gilbertohasnofb:
What's the difference in performance? How noticeable is it?

onox:
In my situation I go from 38 to 44 fps.

onox:
We could also modify the x/X key. But the mouse wheel zoom only works properly in helicopter view and model view. In other view it slightly rotates around some axis when you zoom in, so I had to disable the mouse wheel zoom for those views.

onox:
Easiest solution is to just set it to some higher distance, 35 or 50 perhaps?

gilbertohasnofb:
Well, that's more than 10% so that's a good difference I suppose.

> We could also modify the x/X key.

I don't like that, because I think it's very relevant to have the possibility to move closer/further _and_ zoom in/out. That's the way to compose great screenshots (and great photography for that matter).

> Easiest solution is to just set it to some higher distance, 35 or 50 perhaps?

That would work, so that the default distance would show everything. The downside is that the performance benefit would appear only if people scrolled out from the default view. I personally would still change it though.

onox:
:+1:

gilbertohasnofb:
Ok, so I restored this branch. Would you like to change it yourself or do you want me to do it?

onox:
You :stuck_out_tongue_winking_eye:

gilbertohasnofb:
Yeah, I kind of asked for it, didn't I? :stuck_out_tongue_closed_eyes: 

Ok, so I will do it straight away

onox:
> Yeah, I kind of asked for it, didn't I?

Yes, you did :smile:

gilbertohasnofb:
Done. I thought I would have to change the instruments manually, but now I see you grouped them in a super handy way, so this is a one liner! :beers:

PR #826, opened 2016-08-11
carburetor icing and carburetor heat 

### LINKED ISSUE #360: model carburettor heat + possibly carburettor icing
Now, I don't know how the flight modelling handles carburettor icing, or is it up to the aircraft modeller to do some "tricks" on the systems to model it?

Some food for our thoughts:
   * http://www.skybrary.aero/bookshelf/books/823.pdf - INDUCTION SYSTEM ICING ON PISTON ENGINES AS FITTED TO AEROPLANES, HELICOPTERS AND AIRSHIPS
   * http://www.skybrary.aero/bookshelf/books/659.pdf - Aircraft Icing Handbook

(these were linked from http://www.skybrary.aero/index.php/Piston_Engine_Induction_Icing)

Additional fun if we want to get a bit crazy is to model wing leading surface icing and such, but that might be a task for another issue :-)

gilbertohasnofb:
Fixes #360 

A first attempt to model carburetor icing and carburetor heat. Before merging, please test _extensively_ to make sure this doesn't break anything. Also, I think we should keep issue 360 opened since both @onox and I agree that this implementation here isn't really the best approach.

For testing: first of all, go to the aircraft menu and enable carb icing.

Then play with the weather:  go to the weather menu, select High Pressure scenario but do not apply, then select Manual and play around with the values `20/08` in the METAR string, the first being the temp in Celsius the second being the dew point in Celsius. The closer they are the more humidity there is in the air. Then compare with the chart below (in Fahrenheit!):

The accumulation of ice is given by the property `/engines/active-engine/carb_ice`. When it reaches `0.3` the engine starts coughing and at `0.32` it turns off. The icing rate is given by `/engines/active-engine/carb_icing_rate` times the EGT temp (so that a cold and not running engine does not accumulate ice like crazy at 0 RPM). The effectiveness of the carburetor heat is also proportional to the EGT temp.

Situations with temp and dew point at `07/07` cause _a lot_ of ice build up. But when the temps are outside the orange and green part of the graph above the ice should slowly melt.

:warning: :warning: :warning: 

Before merging: I would still like to make the RPM of the engine drop as the ice accumulates, so that the engine is perfectly healthy when  `/engines/active-engine/carb_ice` is at `0.0` but quite weak when  `/engines/active-engine/carb_ice` is at around `0.3`. Can anyone advice me how can I do this? I wanted to use a filter, but I can't find which property should I use on it. @onox any ideas?

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -58,7 +58,7 @@ var autostart = func (msg=1) {
         };
     };
 
-    # removig any ice from the carburetor
+    # removing any ice from the carburetor
     setprop("/engines/active-engine/carb_ice", 0.0);
     setprop("/engines/active-engine/carb_icing_rate", 0.0);
     setprop("/engines/active-engine/volumetric-efficiency-factor", 1.0);
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -58,6 +58,11 @@ var autostart = func (msg=1) {
         };
     };
 
+    # removig any ice from the carburetor
+    setprop("/engines/active-engine/carb_ice", 0.0);
+    setprop("/engines/active-engine/carb_icing_rate", 0.0);
+    setprop("/engines/active-engine/volumetric-efficiency-factor", 1.0);
+
     # Checking for minimal fuel level
     var fuel_level_left  = getprop("/consumables/fuel/tank[0]/level-norm");
     var fuel_level_right = getprop("/consumables/fuel/tank[1]/level-norm");
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -135,7 +135,7 @@ var carb_icing_function = maketimer(1.0, func {
         var airtempC = getprop("/environment/temperature-degc");        
         var factorX = 13.2 - 3.2 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
         var factorY = 7.0 - 2.0 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);  
-        var carb_icing_rate = math.exp( math.pow((0.8 * airtempC + 0.2 * dewpointC - 45.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.2 * airtempC - 0.8 * dewpointC + 28.0),2) / (-2 * math.pow(factorY,2)));                
+        var carb_icing_rate = math.exp( math.pow((0.6 * airtempC + 0.15 * dewpointC - 32.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.15 * airtempC - 0.6 * dewpointC + 18.0),2) / (-2 * math.pow(factorY,2))) - 0.2;        
         
         # if carb heat on, the rate decreses by a certain amount
         if (getprop("/engines/active-engine/running") and getprop("/controls/engines/current-engine/carb-heat"))
@@ -148,11 +148,14 @@ var carb_icing_function = maketimer(1.0, func {
             carb_ice = 0.0;
         if (carb_ice > 1.0)
             carb_ice = 1.0;
+            
         setprop("/engines/active-engine/carb_ice", carb_ice);
+        setprop("/engines/active-engine/carb_icing_rate", carb_icing_rate);
+        
     } 
     else {
-        var carb_ice = 0.0;
-        setprop("/engines/active-engine/carb_ice", carb_ice);
+        setprop("/engines/active-engine/carb_ice", 0.0);
+        setprop("/engines/active-engine/carb_icing_rate", 0.0);
     }; 
 });
 
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -149,18 +149,12 @@ var carb_icing_function = maketimer(1.0, func {
         
         # carb icing rate is multiplied by an oil temp factor so a cold engine doens't accumulate ice
         var oil_temp_factor = (oil_temp - 120) / 100;
-        if (oil_temp_factor < 0.0)
-            oil_temp_factor = 0.0;
-        if (oil_temp_factor > 1.0)
-            oil_temp_factor = 1.0;
+        oil_temp_factor = std.max(0.0, std.min(oil_temp_factor, 1.0));
         var carb_icing_rate = oil_temp_factor * (carb_icing_formula + carb_heat_rate);
 
         var carb_ice = getprop("/engines/active-engine/carb_ice");
         carb_ice = carb_ice + carb_icing_rate;
-        if (carb_ice < 0.0)
-            carb_ice = 0.0;
-        if (carb_ice > 1.0)
-            carb_ice = 1.0;
+        carb_ice = std.max(0.0, std.min(carb_ice, 1.0));
 
         # this property is used to lower the RPM of the engine as ice accumulates
         var vol_eff_factor = 1.0 - 2.218 * carb_ice;
@@ -175,6 +169,7 @@ var carb_icing_function = maketimer(1.0, func {
         setprop("/engines/active-engine/carb_ice", 0.0);
         setprop("/engines/active-engine/carb_icing_rate", 0.0);
         setprop("/engines/active-engine/volumetric-efficiency-factor", 1.0);
+        setprop("/engines/active-engine/oil_temp_factor", 0.0);
     };
 });
 
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -126,6 +126,37 @@ var oil_consumption = maketimer(1.0, func {
 
 });
 
+# ========== carburetor icing ======================
+
+var carb_icing_function = maketimer(1.0, func {
+    if (getprop("/engines/active-engine/carb_icing_allowed")) {
+        var rpm = getprop("/engines/active-engine/rpm");
+        var dewpointC = getprop("/environment/dewpoint-degc");
+        var airtempC = getprop("/environment/temperature-degc");        
+        var factorX = 13.2 - 3.2 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
+        var factorY = 7.0 - 2.0 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);  
+        var carb_icing_rate = math.exp( math.pow((0.8 * airtempC + 0.2 * dewpointC - 45.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.2 * airtempC - 0.8 * dewpointC + 28.0),2) / (-2 * math.pow(factorY,2)));                
+        
+        # if carb heat on, the rate decreses by a certain amount
+        var carb_heat = getprop("/controls/engines/current-engine/carb-heat");
+        if (carb_heat)
+            carb_icing_rate = carb_icing_rate - 0.01;            
+        
+        # changing in carb ice according to the rate calculated above
+        var carb_ice = getprop("/engines/active-engine/carb_ice");               
+        carb_ice = carb_ice + carb_icing_rate;
+        if (carb_ice < 0.0)
+            carb_ice = 0.0;
+        if (carb_ice > 1.0)
+            carb_ice = 1.0;
+        setprop("/engines/active-engine/carb_ice", carb_ice);
+    } 
+    else {
+        var carb_ice = 0.0;
+        setprop("/engines/active-engine/carb_ice", carb_ice);
+    }; 
+});
+
 # ========== engine coughing ======================
 
 var engine_coughing = maketimer(3.0, func {
@@ -262,5 +293,6 @@ setlistener("/sim/signals/fdm-initialized", func {
     var engine_timer = maketimer(UPDATE_PERIOD, func { update(); });
     engine_timer.start();
     oil_consumption.start();
+    carb_icing_function.start();
     engine_coughing.start();
 });
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -134,22 +134,29 @@ var carb_icing_function = maketimer(1.0, func {
         var dewpointC = getprop("/environment/dewpoint-degc");
         var dewpointF = dewpointC * 9.0 / 5.0 + 32;
         var airtempF = getprop("/environment/temperature-degf");
-        var egt_temp = getprop("/engines/active-engine/egt-norm");
+        var oil_temp = getprop("/engines/active-engine/oil-temperature-degf");
         
         # the formula below attempts to modle the graph found in the POH, using RPM, airtempF and dewpointF as variables
         var factorX = 13.2 - 3.2 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
         var factorY = 7.0 - 2.0 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
-        var carb_icing_rate = 0.01 * (math.exp( math.pow((0.6 * airtempF + 0.3 * dewpointF - 42.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.3 * airtempF - 0.6 * dewpointF + 14.0),2) / (-2 * math.pow(factorY,2))) - 0.2);
-
+        var carb_icing_formula = 0.01 * (math.exp( math.pow((0.6 * airtempF + 0.3 * dewpointF - 42.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.3 * airtempF - 0.6 * dewpointF + 14.0),2) / (-2 * math.pow(factorY,2))) - 0.2);
+        
         # if carb heat on, the rate decreses by a certain amount
         if (getprop("/engines/active-engine/running") and getprop("/controls/engines/current-engine/carb-heat"))
-            var carb_heat_rate = -0.03 * egt_temp;
+            var carb_heat_rate = -0.01;
         else
             var carb_heat_rate = 0.0;
+        
+        # carb icing rate is multiplied by an oil temp factor so a cold engine doens't accumulate ice
+        var oil_temp_factor = (oil_temp - 120) / 100;
+        if (oil_temp_factor < 0.0)
+            oil_temp_factor = 0.0;
+        if (oil_temp_factor > 1.0)
+            oil_temp_factor = 1.0;
+        var carb_icing_rate = oil_temp_factor * (carb_icing_formula + carb_heat_rate);
 
         var carb_ice = getprop("/engines/active-engine/carb_ice");
-        # carb icing rate is multiplied by EGT temp so a cold engine at 0 RPM doens't accumulate ice
-        carb_ice = carb_ice + carb_icing_rate * egt_temp + carb_heat_rate;
+        carb_ice = carb_ice + carb_icing_rate;
         if (carb_ice < 0.0)
             carb_ice = 0.0;
         if (carb_ice > 1.0)
@@ -161,6 +168,7 @@ var carb_icing_function = maketimer(1.0, func {
         setprop("/engines/active-engine/carb_ice", carb_ice);
         setprop("/engines/active-engine/carb_icing_rate", carb_icing_rate);
         setprop("/engines/active-engine/volumetric-efficiency-factor", vol_eff_factor);
+        setprop("/engines/active-engine/oil_temp_factor", oil_temp_factor);
 
     }
     else {
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -132,31 +132,37 @@ var carb_icing_function = maketimer(1.0, func {
     if (getprop("/engines/active-engine/carb_icing_allowed")) {
         var rpm = getprop("/engines/active-engine/rpm");
         var dewpointC = getprop("/environment/dewpoint-degc");
-        var airtempC = getprop("/environment/temperature-degc");        
-        var factorX = 13.2 - 3.2 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
-        var factorY = 7.0 - 2.0 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);  
-        var carb_icing_rate = math.exp( math.pow((0.6 * airtempC + 0.15 * dewpointC - 32.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.15 * airtempC - 0.6 * dewpointC + 18.0),2) / (-2 * math.pow(factorY,2))) - 0.2;        
+        var dewpointF = dewpointC * 9.0 / 5.0 + 32;
+        var airtempF = getprop("/environment/temperature-degf");
+        var egt_temp = getprop("/engines/active-engine/egt-norm");
         
+        # the formula below attempts to modle the graph found in the POH, using RPM, airtempF and dewpointF as variables
+        var factorX = 13.2 - 3.2 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
+        var factorY = 7.0 - 2.0 * math.atan2 ( ((rpm - 2000.0) * 0.008), 1);
+        var carb_icing_rate = 0.01 * (math.exp( math.pow((0.6 * airtempF + 0.3 * dewpointF - 42.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.3 * airtempF - 0.6 * dewpointF + 14.0),2) / (-2 * math.pow(factorY,2))) - 0.2);
+
         # if carb heat on, the rate decreses by a certain amount
         if (getprop("/engines/active-engine/running") and getprop("/controls/engines/current-engine/carb-heat"))
-            carb_icing_rate = carb_icing_rate - 0.01;
-        
-        # changing in carb ice according to the rate calculated above
-        var carb_ice = getprop("/engines/active-engine/carb_ice");               
-        carb_ice = carb_ice + carb_icing_rate;
+            var carb_heat_rate = -0.03 * egt_temp;
+        else
+            var carb_heat_rate = 0.0;
+
+        var carb_ice = getprop("/engines/active-engine/carb_ice");
+        # carb icing rate is multiplied by EGT temp so a cold engine at 0 RPM doens't accumulate ice
+        carb_ice = carb_ice + carb_icing_rate * egt_temp + carb_heat_rate;
         if (carb_ice < 0.0)
             carb_ice = 0.0;
         if (carb_ice > 1.0)
             carb_ice = 1.0;
-            
+
         setprop("/engines/active-engine/carb_ice", carb_ice);
         setprop("/engines/active-engine/carb_icing_rate", carb_icing_rate);
-        
-    } 
+
+    }
     else {
         setprop("/engines/active-engine/carb_ice", 0.0);
         setprop("/engines/active-engine/carb_icing_rate", 0.0);
-    }; 
+    };
 });
 
 # ========== engine coughing ======================
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -138,9 +138,8 @@ var carb_icing_function = maketimer(1.0, func {
         var carb_icing_rate = math.exp( math.pow((0.8 * airtempC + 0.2 * dewpointC - 45.0),2) / (-2 * math.pow(factorX,2))) * math.exp( math.pow((0.2 * airtempC - 0.8 * dewpointC + 28.0),2) / (-2 * math.pow(factorY,2)));                
         
         # if carb heat on, the rate decreses by a certain amount
-        var carb_heat = getprop("/controls/engines/current-engine/carb-heat");
-        if (carb_heat)
-            carb_icing_rate = carb_icing_rate - 0.01;            
+        if (getprop("/engines/active-engine/running") and getprop("/controls/engines/current-engine/carb-heat"))
+            carb_icing_rate = carb_icing_rate - 0.01;
         
         # changing in carb ice according to the rate calculated above
         var carb_ice = getprop("/engines/active-engine/carb_ice");               
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -155,13 +155,18 @@ var carb_icing_function = maketimer(1.0, func {
         if (carb_ice > 1.0)
             carb_ice = 1.0;
 
+        # this property is used to lower the RPM of the engine as ice accumulates
+        var vol_eff_factor = 1.0 - 2.218 * carb_ice;
+
         setprop("/engines/active-engine/carb_ice", carb_ice);
         setprop("/engines/active-engine/carb_icing_rate", carb_icing_rate);
+        setprop("/engines/active-engine/volumetric-efficiency-factor", vol_eff_factor);
 
     }
     else {
         setprop("/engines/active-engine/carb_ice", 0.0);
         setprop("/engines/active-engine/carb_icing_rate", 0.0);
+        setprop("/engines/active-engine/volumetric-efficiency-factor", 1.0);
     };
 });
 
--- a/Systems/engine.xml
+++ b/Systems/engine.xml
@@ -593,9 +593,6 @@
         <name>Engine 160 HP Carb Icing Factor</name>
         <type>gain</type>
         <input>
-            <condition>
-                <property>/engines/active-engine/carb_icing_allowed</property>
-            </condition>
             <property>/engines/active-engine/volumetric-efficiency-factor</property>
         </input>
         <output>
@@ -607,9 +604,6 @@
         <name>Engine 180 HP Carb Icing Factor</name>
         <type>gain</type>
         <input>
-            <condition>
-                <property>/engines/active-engine/carb_icing_allowed</property>
-            </condition>
             <property>/engines/active-engine/volumetric-efficiency-factor</property>
         </input>
         <output>
--- a/Systems/engine.xml
+++ b/Systems/engine.xml
@@ -506,6 +506,10 @@
                     <property>/engines/active-engine/oil-level</property>
                     <value>4.92</value>
                 </less-than>
+                <greater-than-equals>
+                    <property>/engines/active-engine/carb_ice</property>
+                    <value>0.32</value>
+                </greater-than-equals>
             </or>
         </input>
         <output>
@@ -562,6 +566,16 @@
                     <less-than>
                         <property>/engines/active-engine/oil-level</property>
                         <value>4.925</value>
+                    </less-than>
+                </and>
+                <and>
+                    <greater-than-equals>
+                        <property>/engines/active-engine/carb_ice</property>
+                        <value>0.3</value>
+                    </greater-than-equals>
+                    <less-than>
+                        <property>/engines/active-engine/carb_ice</property>
+                        <value>0.32</value>
                     </less-than>
                 </and>
             </or>
--- a/Systems/engine.xml
+++ b/Systems/engine.xml
@@ -584,5 +584,37 @@
             <property>/engines/active-engine/coughing</property>
         </output>
     </logic>
+    
+    <!-- ============================================================== -->
+    <!-- Carburetor Icing drops RPM                                     -->
+    <!-- ============================================================== -->
+
+    <filter>
+        <name>Engine 160 HP Carb Icing Factor</name>
+        <type>gain</type>
+        <input>
+            <condition>
+                <property>/engines/active-engine/carb_icing_allowed</property>
+            </condition>
+            <property>/engines/active-engine/volumetric-efficiency-factor</property>
+        </input>
+        <output>
+            <property>/fdm/jsbsim/propulsion/engine[0]/volumetric-efficiency</property>
+        </output>
+    </filter>
+
+    <filter>
+        <name>Engine 180 HP Carb Icing Factor</name>
+        <type>gain</type>
+        <input>
+            <condition>
+                <property>/engines/active-engine/carb_icing_allowed</property>
+            </condition>
+            <property>/engines/active-engine/volumetric-efficiency-factor</property>
+        </input>
+        <output>
+            <property>/fdm/jsbsim/propulsion/engine[1]/volumetric-efficiency</property>
+        </output>
+    </filter>            
 
 </PropertyList>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -631,6 +631,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <oil-level type="double">7.0</oil-level>
             <oil_consumption_allowed type="bool">false</oil_consumption_allowed>
             <carb_ice type="double">0.0</carb_ice>
+            <carb_icing_rate type="double">0.0</carb_icing_rate>
             <carb_icing_allowed type="bool">false</carb_icing_allowed>
             <auto-start type="bool">false</auto-start>
         </active-engine>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -259,6 +259,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <path>/sim/model/c172p/brake-parking</path>
             <path>/sim/model/c172p/enable-fog-frost</path>
             <path>/sim/model/c172p/garmin196-visible</path>
+            <path>/engines/active-engine/carb_icing_allowed</path>
             <path>/engines/active-engine/oil_consumption_allowed</path>
             <path>/engines/active-engine/oil-level</path>
             <path>/fdm/jsbsim/settings/damage</path>
@@ -629,6 +630,8 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <kill-engine type="bool">false</kill-engine>
             <oil-level type="double">7.0</oil-level>
             <oil_consumption_allowed type="bool">false</oil_consumption_allowed>
+            <carb_ice type="double">0.0</carb_ice>
+            <carb_icing_allowed type="bool">false</carb_icing_allowed>
             <auto-start type="bool">false</auto-start>
         </active-engine>
 
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -133,6 +133,21 @@
                     <command>dialog-apply</command>
                 </binding>
             </checkbox>
+            
+            <checkbox>
+                <halign>left</halign>
+                <label>Allow carburetor icing</label>
+                <property>/engines/active-engine/carb_icing_allowed</property>
+                <live>true</live>
+                <binding>
+                    <command>property-assign</command>
+                    <property>/engines/active-engine/oil-level</property>
+                    <value>7.0</value>
+                </binding>
+                <binding>
+                    <command>dialog-apply</command>
+                </binding>
+            </checkbox>
         </group>
 
         <hrule/>
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -186,7 +186,33 @@
                 </enable>
                 <binding>
                     <command>nasal</command>
-                    <script>c172p.repair_damage();electrical.reset_battery_and_circuit_breakers();c172p.click("engine-repair", 6.0)</script>
+                    <script>
+                        c172p.repair_damage();
+                        electrical.reset_battery_and_circuit_breakers();
+                        c172p.click("engine-repair", 6.0);
+                        c172p.dialog_battery_reload();
+                    </script>
+                </binding>
+                <binding>
+                    <condition>
+                        <property>sim/model/c172p/engine_flag_0</property>
+                    </condition>
+                    <command>property-assign</command>
+                    <property>/engines/active-engine/oil-level</property>
+                    <value>7.0</value>
+                </binding>
+                <binding>
+                    <condition>
+                        <property>sim/model/c172p/engine_flag_1</property>
+                    </condition>
+                    <command>property-assign</command>
+                    <property>/engines/active-engine/oil-level</property>
+                    <value>8.0</value>
+                </binding>
+                <binding>
+                    <command>property-assign</command>
+                    <property>/engines/active-engine/carb_ice</property>
+                    <value>0.0</value>
                 </binding>
             </button>
         </group>


gilbertohasnofb:
Also, we have to see if the amount of time it takes for the engine to freeze is reasonable. I think that on severe icing at low RPM it should take only a couple of minutes using this implementation above.

gilbertohasnofb:
And in case anyone is wondering why I selected the arbitrary value of `0.32` as the point at which the engine stops, even though that property ranges from 0.0 to 1.0, it's because I wanted ice to be able to accumulate even further after an engine stops, so that descending through warm air won't instantaneously let the engine come back to life.

onox:
> Before merging: I would still like to make the RPM of the engine drop as the ice accumulates, so that the engine is perfectly healthy when /engines/active-engine/carb_ice is at 0.0 but quite weak when /engines/active-engine/carb_ice is at around 0.3. Can anyone advice me how can I do this? I wanted to use a filter, but I can't find which property should I use on it. @onox any ideas?

You want some sort of offset? You could try this:

```xml
<input>
    <property>path/to/rpm</property>
    <offset>
        <property>/engines/active-engine/carb_ice</property>
        <scale>-some_multiplier</scale>
    </offset>
</input>
```

onox:
I'll will test this PR in the late afternoon to evening.

onox:
Since there's already a button to recharge the battery, should we recharge the battery via the "Repair" button as well?

gilbertohasnofb:
@onox Thanks for the code, I will try to implement it right now.

> Since there's already a button to recharge the battery, should we recharge the battery via the "Repair" button as well?

I don't see a problem with that. When an aircraft is repaired its battery will surely be recharged, so I think that the repair button must also do so. But having a separate button for only recharging the battery makes sense in my opinion, since that would not be categorized as "repair", and it makes it more obvious to the user what the button does. The repair is kind of a "make the aircraft fine to fly again", while recharge the battery just simply recharges it.

> I'll will test this PR in the late afternoon to evening.

Hmm, I think we will then miss the time Stuart had asked us to have the plane ready. Surely he said that he might be able to merge it also on Sunday if we don't make it today by 20:00 UTC, but I am kind of traumatized with how things went last time. 

Anyway, I am starting to have some second thoughts about merging this PR just hours before making a release. There is so much potential of things going wrong. For instance, there might be weather scenarios out there that shouldn't cause any icing but are actually doing it. Or maybe the accumulation rate is too great. Or maybe there are other bugs, some potentially aircraft-breaking, which we didn't foresee. @onox What do you think if we first make the release and then only after merge this, so that we have plenty of time to test it before it makes into FGDATA? I personally wouldn't be happy to push something like this in the last minute.

gilbertohasnofb:
> You want some sort of offset? You could try this:
> ```
> <input>
>     <property>path/to/rpm</property>
>     <offset>
>         <property>/engines/active-engine/carb_ice</property>
>         <scale>-some_multiplier</scale>
>     </offset>
> </input>
> ```

I think I do not understand well this onox, did you mean to use this in a filter or what? Also, the problem is that the input and output properties are the same (RPM), I thought you couldn't do that. And finally, I also thought that the RPM property could not be handled like this because it's one of the outputs of JSBSim.

dany93:
`I also thought that the RPM property could not be handled like this because it's one of the outputs of JSBSim.`
I agree.
I would rather look to engines properties.
http://wiki.flightgear.org/JSBSim_Engines
Like 
`<volumetric-efficiency>`
 ` <air-intake-impedance-factor>`
 ` <ram-air-factor>` 
All are artefacts, not reality compared to what you wish to do, but I think they could be a way to control the engine power.
They are in the properties, `fdm/jsbsim/propulsion/engine`. Hence they can be changed. In real time during simulation, I hope.

onox:
Ah, yeah, you can't modify the RPM. So you need to modify something else then. It was just an example.

onox:
> Anyway, I am starting to have some second thoughts about merging this PR just hours before making a release. There is so much potential of things going wrong. For instance, there might be weather scenarios out there that shouldn't cause any icing but are actually doing it. Or maybe the accumulation rate is too great. Or maybe there are other bugs, some potentially aircraft-breaking, which we didn't foresee. @onox What do you think if we first make the release and then only after merge this, so that we have plenty of time to test it before it makes into FGDATA? I personally wouldn't be happy to push something like this in the last minute.

We can merge the PR after the release.

gilbertohasnofb:
@onox in that case, let's create the release right now. I will continue to work this afterwards if that's fine.

gilbertohasnofb:
@dany93 @onox thanks for the info!

gilbertohasnofb:
@onox ok, RPM drop has been implemented. Now the question is if the icing effect is too strong or too weak. I suspect I went a little bit too conservative, as even in horrible conditions (e.g. temp/dew == `07/07`) it takes a long while for the carb to completely fill with ice without carb heat. But the effects of RPM drop can be noticed quite fast after ice starts to accumulate. Could you test it and give me your opinions?

Systems/engine.xml
```diff
@@ -584,5 +584,37 @@
...
+    
+    <!-- ============================================================== -->
+    <!-- Carburetor Icing drops RPM                                     -->
+    <!-- ============================================================== -->
+
+    <filter>
+        <name>Engine 160 HP Carb Icing Factor</name>
+        <type>gain</type>
+        <input>
+            <condition>
```
onox: Here you have a conditional `<input`> (because of the `<condition>` in it), but there's no default `<input>` (one without a `<condition>`). What should happen if the condition is false? Just leave the output value at its current value, or set it to 0? I think you want the latter, and in that case you should add

```xml
<input>
    <value>0.0</value>
</input>
```

after the first `<input>` block.

Systems/engine.xml
```diff
@@ -584,5 +584,37 @@
...
+        <output>
+            <property>/fdm/jsbsim/propulsion/engine[0]/volumetric-efficiency</property>
+        </output>
+    </filter>
+
+    <filter>
+        <name>Engine 180 HP Carb Icing Factor</name>
+        <type>gain</type>
+        <input>
+            <condition>
```
onox: Same issue here.

Nasal/c172p.nas
```diff
@@ -58,6 +58,11 @@ var autostart = func (msg=1) {
         };
     };
 
+    # removig any ice from the carburetor
```
onox: Typo, "removig" should be "removing".

onox:
I can do some functional testing later today.

gilbertohasnofb:
@onox thanks a lot for the comments! I will fix the typo, and as for the:

```
<input>
    <value>0.0</value>
</input>
```

Now I finally understand why some filters have it and some don't! But in this case we don't want that output property to be multiplied by zero. And as a matter of fact, I think what I should do is to remove the condition from those `<input>`'s, since that property inside them is handled by the Nasal function, which does take care of it in case the user has disabled icing in the carburetor. 

> I can do some functional testing later today.

:+1:

gilbertohasnofb:
@onox Ok, so I fixed those things. I also made some modifications to the Nasal code of that function. I am using now the oil temperature to infer the temperature of the engine, and it is working much better than when I was using the EGT (which gets fixed quite fast). So now ice does not accumulate like crazy when in the ground with a cold engine during taxi, but does accumulate properly when the engine gets up to the temperature. This temp also affects the effectiveness of the carburetor heat (i.e. cold engine == cold exhaust air == ineffective removal of ice).

Nasal/engine.nas
```diff
@@ -134,22 +134,29 @@ var carb_icing_function = maketimer(1.0, func {
...
+        
         # if carb heat on, the rate decreses by a certain amount
         if (getprop("/engines/active-engine/running") and getprop("/controls/engines/current-engine/carb-heat"))
-            var carb_heat_rate = -0.03 * egt_temp;
+            var carb_heat_rate = -0.01;
         else
             var carb_heat_rate = 0.0;
+        
+        # carb icing rate is multiplied by an oil temp factor so a cold engine doens't accumulate ice
+        var oil_temp_factor = (oil_temp - 120) / 100;
```
onox: You should clamp it with `oil_temp_factor = std.max(0.0, std.min(oil_temp_factor, 1.0));` here.

Nasal/engine.nas
```diff
@@ -134,22 +134,29 @@ var carb_icing_function = maketimer(1.0, func {
...
+            oil_temp_factor = 0.0;
+        if (oil_temp_factor > 1.0)
+            oil_temp_factor = 1.0;
+        var carb_icing_rate = oil_temp_factor * (carb_icing_formula + carb_heat_rate);
 
         var carb_ice = getprop("/engines/active-engine/carb_ice");
-        # carb icing rate is multiplied by EGT temp so a cold engine at 0 RPM doens't accumulate ice
-        carb_ice = carb_ice + carb_icing_rate * egt_temp + carb_heat_rate;
+        carb_ice = carb_ice + carb_icing_rate;
         if (carb_ice < 0.0)
```
onox: Use `std.min` and `std.max` here as well.

Nasal/engine.nas
gilbertohasnofb: I will do that right now. But just out of curiosity, what is the advantage of using it instead of those `if` statements? Is it because it's more compact or is there another benefit?

onox: Just more compact. Needing 4 lines for just clamping a value is a bit silly. I have send a small patch to the mailing list so you can use `std.clamp(oil_temp_factor, 0, 1)` instead. That is more readable.

gilbertohasnofb: >  I have send a small patch

That's great!

> Needing 4 lines for just clamping a value is a bit silly. 

Agreed, I pushed the changes (already tested, all seems to be correct)

gilbertohasnofb:
@onox when you have time could you please test this branch in the sim? I am curious what you will think about it :smile:

onox:
I haven't had much time to test it. Is the carb heat an on/off slider (boolean) in real life too? Or should it be like the throttle/mixture (float)? We could just merge it so that more people will beta test it.

gilbertohasnofb:
> I haven't had much time to test it. 

No problems!

> Is the carb heat an on/off slider (boolean) in real life too? Or should it be like the throttle/mixture (float)? 

I believe it's not boolean, but the procedure apparently is: that is, the checklists and manuals always write ON or OFF, never inches or percents, so maybe our modelling is fine.

> We could just merge it so that more people will beta test it.

That's an option, then if there is anything wrong we can either fix or revert this PR. But it would be good to announce it in the forum so people can keep an eye on it. Maybe @wlbragg could make a little note in the forum?

gilbertohasnofb:
@wlbragg could you please make a short post in the forum saying that we merged this carburetor icing work in the `master` and we are looking for bugs/comments? Icing has to be enabled via the aircraft dialog.

wlbragg:
Yes, I'll do it this afternoon, I'm on my way out the door at the moment.

gilbertohasnofb:
@wlbragg :+1: no rush and thanks a lot!

kentrosaur:
Hi, in 2016.3 I managed to find myself in a carb icing situation in a traffic pattern.  Applying carb heat kept my engine from completely stopping or getting worse but it never thawed the ice so I could not get full power without repairing the aircraft on the ground.  Any thoughts?

gilbertohasnofb:
@kentrosaur what was the weather, in particular the pressure, temperature and dew point? If you are in an severe icing situation, there is basically nothing you can do except a) divert if on the air b) not fly if on the ground.

Also keep in mind that the carburettor is more likely to develop ice at lower RPM and at higher, which might explain why this happened once you entered traffic pattern and is preparing to land.

kentrosaur:
@gilbertohasnofb looking back at the weather history this week, it was raining and 60°F/ 60°F and 30.03in, right on the edge of the serious zone for cruise power.

All of the icing procedures I've read seem to indicate that after a few minutes of using the heat, the carb ice should be melted in spite of the severity of the conditions.  In my case I fruitlessly fought the throttle, mixture (for EGT temp) and carb heat settings to no result.  

> At the first indication of carb ice:
> 1) Full carb heat ON (partial application could cause additional carb icing)
> 2) Leave carb heat ON until normal engine power returns
> 3) Monitor engine power and re-apply carb heat as necessary

gilbertohasnofb:
> right on the edge of the serious zone for cruise power.

So in those conditions, when lowering the RPM while preparing for landing you are in serious icing condition

> All of the icing procedures I've read seem to indicate that after a few minutes of using the heat, the carb ice should be melted in spite of the severity of the conditions.

I always thought that carb heat was rather a preventive measure and that it may not be very good at melting existing icing if the build up is severe, particularly if the engine RPM is not high. Can one of the pilots here confirm this behaviour to me? @tigert @algefaen @Juanvvc ?

So perhaps the question here would be: should we beef up the effect of carb heat, so that it can melt ice even at extreme conditions? Would that be realistic or is our current behaviour a better option?

gilbertohasnofb:
@kentrosaur by the way, thanks for the feedback on the carb icing system, I really appreciate it. Without real life data I tried to make it fit what I judged to be "reasonable" values, and I am sure there is a lot of room for improvements

wkitty42:
On 01/21/2017 07:10 PM, kentrosaur wrote:
>     At the first indication of carb ice:
>
>      1. Full carb heat ON (partial application could cause additional carb icing)
>      2. Leave carb heat ON until normal engine power returns
>      3. Monitor engine power and re-apply carb heat as necessary

i'm not a pilot but in the c172p, i thought the carb heat was either on or 
off... no middle ground... the above first step seems to indicate that carb heat 
is a variable setting maybe kinda like the mixture setting... am i missing 
something??

gilbertohasnofb:
@wkitty42 I think in RL the carb heat is a lever that can be adjusted continuously, but the course is quite small (when compared let's say to the throttle lever), and all procedures I know involve having it either ON or OFF. So considering this, I think the way we modelled it is quite practical.

kentrosaur:
@wkitty42 I'm used to the 172 fuel injected models but I might be able to look in a carb model later this week to verify it's similar to a throttle.  I think toggling between maximum and off is more than sufficient for most situations.  For example, checklists only mention 'full on' and on an aborted landing, the carb heat is one of the last things you'll remove since it's a relatively small effect at full throttle.  

I know the heat affects the RPM in the sim but does it also affect the mixture as implied from the POH?  I think the carb heat should be proportional to the engine temperature, so in some cases it's impossible to unfreeze the ice but most cases should be fine with the ice melting gradually and a subsequent increase in RPMs.  There might also be an effect from leaning to a higher EGT but that should also raise the main temp as well. 

This is a pretty informative article on carb heating in different aircraft.  
http://www.pilotfriend.com/training/flight_training/fxd_wing/carb_heat.htm

(my previous quote was a general AOPA bulletin about carb ice for any airframe)

Lastly from a 172P POH:
> A gradual loss of RPM and eventual engine roughness may result from the formation of carburetor ice.  To clear the ice, apply full throttle and pull the carburetor heat knob full out until the engine runs smoothly; then remove carburetor heat and readjust the throttle.  If conditions require the continued use of carburetor heat in cruise flight, use the minimum amount of heat necessary to prevent ice from forming and lean the mixture for smoothest engine operation.

algefaen:
> I always thought that carb heat was rather a preventive measure and that it may not be very good at melting existing icing if the build up is severe, particularly if the engine RPM is not high. Can one of the pilots here confirm this behaviour to me?

Sure, let me take our skyhawk for a spin, build some carb ice, and see how fast if it melts! (=

Not going to be able to give you any measurable data here for obvious reasons, but this is what I was taught:
 - Use carb heat to check if you have icing, and as prevention only if it's really likely (i.e. decending on a humid day or flying through clouds). Warm and humid is worse than cold weather, since warm air can hold more moisture.
 - If you apply carb heat and there's no ice, the RPM will go down and stay down. (This is what I check in my pre-landing checklist)
 - If you actually do have icing and you apply carb heat, your engine might sputter a little bit, running a bit rough, as the ice is melting. Initially the RPM decreases - but then it will increase rapidly as the ice is melting.
 - It should definitively go back to normal operations pretty quickly after carb heat has been applied and you can turn it off if you're out of the conditions that formed the ice in the first place
 - Don't ever let carb heat be left on on the ground, since it's unfiltered air and can suck up dust from the ground (this might not apply to all aircraft though).

The carb heat knob on our 172N has something like 4cm travel from fully cold to fully hot, and looks like a smaller version of the throttle knob. I only ever use it in either fully on or fully off positions, but I guess it's possible to leave it in between and get a mixture of air from each source.

I found my textbook, and it mentions:
> CAUTION: Partial use of carburettor heat may raise the temperature of the induction air into the temperature range that is most conducive to the formation of carburettor icing. Consult the POH.

gilbertohasnofb:
> I think the carb heat should be proportional to the engine temperature, so in some cases it's impossible to unfreeze the ice but most cases should be fine with the ice melting gradually and a subsequent increase in RPMs.

@kentrosaur That's a good idea, I will check if I can make this work.

> I know the heat affects the RPM in the sim but does it also affect the mixture as implied from the POH?
>> lean the mixture for smoothest engine operation

I am not sure if the POH implies that the icing affects the mixture or if simply when icing has accumulated one has to change the procedure of handling the mixture. What do you understand that has to be implemented concerning the mixture?

> Sure, let me take our skyhawk for a spin, build some carb ice, and see how fast if it melts! (=

@algefaen :rofl: let me rephrase that then: does one of the pilots here know how these things work, perhaps learned through their *theoretical* training? :wink:

> It should definitively go back to normal operations pretty quickly after carb heat has been applied and you can turn it off if you're out of the conditions that formed the ice in the first place

:+1:

Ok guys, great info from you all, I really appreciate it. So I will try the following:

- beef up the effect of the carb heat, trying to ensure it will be somwhat effective even in very severe icing situation
- use the engine temperature as a factor in the effectiveness of carb heat

algefaen:
> use the engine temperature as a factor in the effectiveness of carb heat

Make sure it's still at least somewhat effective even after a cold start. You're supposed to be able to use carb heat to clear carb ice you might have picked up during taxi. Oh, and apparently it's much more likely to get carb ice on the ground if you're maneuvering on grass - but that sounds a little bit tricky to simulate properly. (=

... I'm glad we're slowly but steadily getting rid of these carburetors. So much hassle compared to direct injection ...

gilbertohasnofb:
@algefaen :+1: 

I should have a first attempt to fix these things up today or tomorrow, and once that's done I will let you guys know here in case anyone would like to test it

gilbertohasnofb:
Ok guys, so I had a look at this and I pushed some new work on branch `carb-heat-improvements`. Here are some description of what is going on and what I added now:

The carb icing code is located in `Nasal/engine.nas`. The basic formula behind it is `carb_icing_formula`, which takes RPM, dew point and air temperature as variables. But no matter what combination of those factors, this formula will **always** output numbers in the range -0.002 to 0.008. Positive values imply that ice is accumulating, negative values that ice is melting. 

Whatever natural icing rate we have, once the carb heat is toggled on it will contribute to that ratio by -0.01, meaning that even in the worst icing scenarios the total ratio would be negative (0.008 + (-0.01) = -0.002) and thus ice would melt (albeit not as fast as in other scenarios). @kentrosaur so no matter what was your situation in the first report, the ice should have been melting for you reported. Maybe it was melting at a too slow rate? Perhaps, though I also find this difficult to gauge as even in a worst case scenario ice is melting very fast for me here.

Anyway, now I made some changes to the code and I would be happy if you guys could give it a go. The carb heat does not contributed to a fixed value of -0.01 any more but rather has its contribution proportional to the EGT. For a reasonable EGT value the carb heat should be at least 60% more efficacious in removing ice now.

Also, icing is not accumulated if the engine was too cold just like before, but the minimal values have changed. I had previously set a minimal value of 120 F of oil temperature (as a measurement of engine temp) for icing to be allowed to accumulate. I have now substantially lowered that to 60 F, so if you start up in a severe icing situation you may very well accumulate ice during the taxi. 

If anyone wants to test this, I would suggest observing the properties below:

- `/engine/active-engine/carb_ice`, which will give you the total amount of ice in the carburettor (between 0.0 and 1.0, though your engine should quit much before it becomes a completely solid ice cube :laughing: ).
- `/engine/active-engine/carb_icing_rate`, which gives you the final rate of ice accumulation of melting. This is after taking into account RPM, dew point and temperature, carb heater, EGT, engine temperature, and everything else. If this value is positive, you are accumulating ice, if negative it's melting ice. The numbers are very small because it shows the variation of `carb_ice` per second, which is never much. A very severe scenario will output something around +0.005. 
- worst case scenario should be when air temperature and dew point are both very close to each other at around 7 or 8 degrees Celsius

kentrosaur:
@gilbertohasnofb this is great, thanks!  I might have some time to re-test the ice conditions then try the branch this week.

The melting carb ice means there's water in the combustion chamber as well as just fuel and air so you need less fuel (leaner mixture) to burn completely with the remaining air.  This is why the POH says the engine will be rough or knocking after the carb heat is applied and smooth out as the ice/water is gone.  The overly rich mixture is leaving fuel in the exhaust which will fire at the wrong time in the cycle or even in the exhaust as a backfire.  The accumulation phase would be allowing a smaller volume of air into the engine  for the same amount of fuel, needing a leaner mixture again.

I think there should be a point where the carb heat can't recover.  Maye something around 80% ice which would limit the carb heat and rpms to something that can't overpower moderate icing conditions?

gilbertohasnofb:
@kentrosaur 

> I might have some time to re-test the ice conditions then try the branch this week.

Perfect, thanks a lot! I would suggest using manual input METAR's to test this, and then if you spot anything strange please let me know at which altitude you experienced the event and what was the METAR string used (temp, dew point and pressure).

> I think there should be a point where the carb heat can't recover. Maye something around 80% ice which would limit the carb heat and rpms to something that can't overpower moderate icing conditions?

Hmm, the carb heat efficacy is not programmed to be related to the amount of ice, it is related only to the exhaust gas temperature. Basically this system simply calculates the icing rate versus the melting rate, and adds or remove ice accordingly. So as long the rate is negative, ice will be melt away. But I suppose the EGT will lower as the carburettor starts filling up with ice, or at least when the engine is coughing and about to quit (effectively affecting the carb heat) but I need to check that. Also, there is no hard limit for which the engine will quit, this will depend on several factors. The ice is affecting the volumetric efficiency of the engine, and around 0.3 (or 30%) of carburettor ice is enough to make the engine quit.

> The melting carb ice means there's water in the combustion chamber as well as just fuel and air so you need less fuel (leaner mixture) to burn completely with the remaining air. This is why the POH says the engine will be rough or knocking after the carb heat is applied and smooth out as the ice/water is gone. 

Hmm, I have to think of a way of implementing this, but it might be a bit tricky. We have a function for engine coughing (which is activated when the oil level is too low), but that is sort of a hack made by very quickly turning the engine off and then back on at random intervals. But to make the engine run rough, I think we need most of all make it sound rough (and perhaps lower the RPM a tiny bit), but that's easier said than done. @onox any ideas or suggestions?

Ok guys, so I did some further modifications and improvements to that branch:

- some overall improvements to the code and added some comments to clarify the range of the functions
- ice builds up more slowly, but at the same time the carb heat also melts it slower. All values have been divided by 10, so not an accumulation rate of around 0.0008 would be very severe
- at the worst case scenario, the carb heater should still work but it might take a longer time
- carb heat efficacy does not depend on engine temperature any longer, only on the exhaust gas temperature

gilbertohasnofb:
@dany93 @onox @wlbragg I just tested and it seems the EGT temperature (given by `/engine/active-engine/egt-degf` is virtually constantly at around 1450. When I start the sim this value is already at around 50, but then it instantaneously jumps to 1450 and stays there no matter the RPM or anything else. If I then turn the engine off, this value naturally starts to go down as expected, and restarting the engine causes it to insta-jump back to 1450.

wlbragg:
Some background may help?
http://www.gami.com/articles/egt_myths.pdf

wlbragg:
Another tidbit!
http://aviation.stackexchange.com/questions/21466/what-appears-on-egt-gauge-if-the-mixture-is-rich-or-lean

wlbragg:
One could spend a lifetime modeling any of these systems!

dany93:
To come back to @gilbertohasnofb remark, I agree that an instantaneous temperature increase is weird. The Exhaust Gas Temperature probably increases very quickly at the engine start, but I would expect that the sensor temperature has some inertia (even if it is small to minimize this inertia).
However, the EGT in FG changes with fuel richness, which is a good start.

gilbertohasnofb:
Thanks for all the info, guys. @wlbragg those links are very interesting, thanks for sharing them (particularly the one about the myths). That said, I think we do have absolute exhaust temperature in FG, which is the property `/engines/active-engine/egt-degf`. We then have another property called `/engines/active-engine/egt-norm` which is what the EGT gauge uses. So I think when the article says that EGT does not give you absolute temperature, the writer is talking about the guage and that's correct, but in FG we do have access to that value.

> I agree that an instantaneous temperature increase is weird. [...] However, the EGT in FG changes with fuel richness, which is a good start.

Yes, of course! RPM should not affect the EGT, only the mixture, and that is being done properly. So I think this is all clear now and the only question concerning the EGT is if we want or not to tackle this instantaneous change or not. @dany93 if you are interested, we could open a new issue for it, but if not I don't think it's such a big deal.

dany93:
> ( @gilbertohasnofb wrote:) ...but if not I don't think it's such a big deal.

I agree. I just noticed it because of your remark.

wkitty42:
On 01/24/2017 10:12 AM, Gilberto Agostinho wrote:
> @dany93 <https://github.com/dany93> @onox <https://github.com/onox> @wlbragg
> <https://github.com/wlbragg> I just tested and it seems the EGT temperature
> (given by |/engine/active-engine/egt-degf| is virtually constantly at around
> 1450. When I start the sim this value is already at around 50, but then it
> instantaneously jumps to 1450 and stays there no matter the RPM or anything
> else. If I then turn the engine off, this value naturally starts to go down as
> expected, and restarting the engine causes it to insta-jump back to 1450.

does the EGT needle in the gauge act in the same manner? if not, what is 
controlling it? shouldn't it indicate the same as the property?

gilbertohasnofb:
No, the EGT gauge uses the property `/engine/active-engine/egt-norm` which doesn't show that behaviour.

gilbertohasnofb:
@dg-505 by the way, the carb heat is now working, so perhaps we could add it to the tutorials? See https://github.com/c172p-team/c172p-detailed/blob/master/Tutorials/runup.xml#L557

And even if this whole icing system is tweaked, it would be only parameters of how the icing occurs so it would be safe to work on the tutorials already.

kentrosaur:
METAR KXXX 030453Z AUTO 35007KT 10SM FEW060 13/08 A3023 RMK AO2 SLP238 T01280078

Just had the icing reproduce my non-improving iced state this evening, flying patterns at 1100MSL and ~1500rpm throttle.  As expected in 2016.4 this built up the ice and there was no recovery.

I tried out the carb-heat-improvemetns branch.  It didn't work but I'm not sure I have the git branch initialized correctly.  The code looks okay and I marked the plane to be unique in the list, is there anything else I need to do to use the heat tweaks from the branched aircraft?   Watching the /engines/active-engine/carb-ice wasn't showing an improving state on the ground or in the air with carb-heat.

gilbertohasnofb:
@kentrosaur Thanks a lot for the message, I really appreciate the detailed report!

But I have to say I have not being able to reproduce the issue here. Using your METAR, I flew at around 1100 MSL until I had enough ice accumulated, then I turned the carb heat on and watched the `/engines/active-engine/carb_icing_rate` and `/engines/active-engine/carb_ice` properties. The rate was negative, meaning that the ice was melting, and the carb ice quantity was indeed falling. After some short time @1500 rpm, I did manage to get rid of all ice. See the screenshot below:

fgfs-screen-001

After having accumulated how much ice did you turn on the carb heat? And was the carb rate accumulation still positive? If you would be able, a screenshot such as the one above (with those properties in the top corner) would help me a lot debugging this. To see the properties in the top corner, go to the Debug menu, then select Browse Internal Properties, then navigate to `/engines/active-engine/` then Shift+click on the properties you would like to display (I am interested in all those I listed on my screenshot).

> It didn't work but I'm not sure I have the git branch initialized correctly. The code looks okay and I marked the plane to be unique in the list, is there anything else I need to do to use the heat tweaks from the branched aircraft?

How are you loading the branch? Are you on Linux or Windows? And are you using the terminal/command line and git commands to load a branch?

kentrosaur:
@gilbertohasnofb I checked out the branch with git, then pointed that folder to the extra aircraft in the launcher from linux.

$ git status
On branch carb-heat-improvements
Your branch is up-to-date with 'origin/carb-heat-improvements'.

I have icing conditions on the ground today as well, so with the debug values I can verify if the rate slows/increases or goes negative when the heat is on.  It's kind of odd that when the heat turns on, the rate remains but no more ice accumulates.  I presume that's the old code in use.  I'm in the OH-CTL model with a 160hp engine.  Do I need to remove the original c172p from my main settings to use this branch even though I think I've marked them separately?  
 
fgfs-screen-020

I just noticed the engine property coughing, which would be good to have while 'carb-heat==true && carb_ice > 0' to simulate the extra water in the combustion chamber.

gilbertohasnofb:
@kentrosaur If you are on the `carb-heat-improvements` branch and you are loading the aircraft from that folder, then you should be in the correct branch. Did you do a `git pull` to get the latest modifications?

> It's kind of odd that when the heat turns on, the rate remains but no more ice accumulates. I presume that's the old code in use

Do you have the METAR and ICAO where this happened? I am really surprised you are seeing a positive rate (i.e. ice is accumulating, albeit slowly) even though you have carb heat on, so there are two options: a) it's indeed the older code or b) there is some problem with this implementation which we need to narrow down. Let's first make sure you are in the correct updated branch. Please do a `git pull`, and when you load the sim please make sure you are not loading the c172p which comes with FGDATA but rather the c172p from your local clone of this repository.

> I just noticed the engine property coughing, which would be good to have while 'carb-heat==true && carb_ice > 0' to simulate the extra water in the combustion chamber.

Yeah, I thought about it as well, but I think that coughing is not the same as running rough. Coughing effectively turns the engine off and on for some small interval, and I would much rather have a slightly different sound to make a "rough" property.

kentrosaur:
@gilbertohasnofb I got it to work, had to move the fgdata/c172p then link the git folder to that one.  The wires are crossed somewhere in my setup.  The git setup was all okay and with your branch I see all of the properties in the active-engine instead of some alternate locations.

fgfs-screen-022

>> I just noticed the engine property coughing, which would be good to have while 'carb-heat==true && carb_ice > 0' to simulate the extra water in the combustion chamber.
> Yeah, I thought about it as well, but I think that coughing is not the same as running rough. Coughing effectively turns the engine off and on for some small interval, and I would much rather have a slightly different sound to make a "rough" property.

Yeah good point on the rough sound.  Next up for another day, RPM and performance differences between 160 and 180hp models.

gilbertohasnofb:
@kentrosaur 

> I got it to work

gilbertohasnofb:
@kentrosaur 

> I got it to work

Great! So if you have a chance, please do some tests in icing condition and let me know if you see anything odd (and if you see something, I would appreciate to know the METAR, ICAO and the value of those properties).

> Next up for another day, RPM and performance differences between 160 and 180hp models.

:+1: Thanks a lot for all the tests!

kentrosaur:
Hi @gilbertohasnofb , Just re-tried the KLVK IFR tutorial with the carb icing. I realized after takeoff that I hadn't lowered the throttle but after the first localizer intercept, I had developed enough ice to drop me below cruise power.  13/8 are again pretty severe ice conditions but I feel like this shouldn't happen at full throttle.  

Inside the green arc on the tachometer, I feel like 2500rpm and above should neutralize (but not melt) almost all ice conditions, 2200-2400rpm  would have very slow buildup 10s of minutes, then below 2200rpm can probably follow the formula we have now.   

METAR KLVK 051953Z 32004KT 10SM CLR 13/08 A2990 RMK AO2 SLP123 T01330083

gilbertohasnofb:
@kentrosaur 

> 13/8 are again pretty severe ice conditions but I feel like this shouldn't happen at full throttle.
> [...]
> I feel like 2500rpm and above should neutralize (but not melt) almost all ice conditions

Well, that's not how I interpret the information from the POH, I really think that icing should happen at full throttle in those serious conditions. Take a look at this chart here for carb icing in the 172: 



13/08 is 55/46 in Farenheit, and it lands in the heart of the region "serious icing at cruise power". My understanding it that in those conditions you will develop ice at _any_ power setting if you don't use carb heat. Why do you think that shouldn't be the case?

But in our implementation here, even at such conditions you should be able to melt it using carb heat, did you try using it?

kentrosaur:
@gilbertohasnofb yeah, I was able to melt it and everything seemed to be okay.  

I've been trying to figure out how susceptible the O-320-D2J engine is to carb ice.  I feel like if it were as severe as we are modeling at cruise power, there would have been many more accidents in the bay area which frequently have severe icing conditions.  

[https://youtu.be/v0JbPvVNCWI](Piper supercub carb ice video)

The original '82 POH [http://ilwg.cap.gov/wp-content/uploads/2016/02/C172P_IM-82.pdf](C172P POH) has carb heat in the checklists anywhere a descent / large reduction in throttle occurs AND to use it during cruise if you detect ice OR you're flying in heavy rain (pg 4-17). 

I couldn't verify this from the '82 POH, maybe it was updated?
[https://www.aopa.org/-/media/Files/AOPA/Home/Pilot%20Resources/ASI/Safety%20Briefs/SB09.pdf](AOPA icing brief)
> For example, the pilot’s operating handbook (POH)
> for many single-engine Cessnas calls for the pilot to apply
> carb heat whenever power is reduced below the green
> arc on the tachometer.

In the end, it is mentioned in the normal cruise procedures so it's probably relatively accurate the way it's modeled.   A few lines in any of the tutorials about cruise icing and maybe even an addition to the cruise checklist are probably the best things to do here. 

The real result of all of this is I now greatly appreciate the fuel injection on the 172R and 172S models I fly but I also want to try out a carbureted version soon.

gilbertohasnofb:
@kentrosaur 

> I've been trying to figure out how susceptible the O-320-D2J engine is to carb ice.

Yes, this is the trick here. The POH and all other articles use only words like moderate or severe, but I never found any data to help quantify those (things that would help with ice build-up rate or even how many minutes it would take for an engine to stop, etc.). That said, I think those words are actually reasonably defined. For instance, [this AOPA pdf](https://encrypted.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=0ahUKEwji5_XO6v3RAhUJXRQKHX2zDvQQFggZMAA&url=https%3A%2F%2Fwww.aopa.org%2F-%2Fmedia%2Ffiles%2FAOPA%2FHome%2FPilot-Resources%2FASI%2FSafety-Advisors%2Fsa11.pdf&usg=AFQjCNH7ZViIu-m0_vFrY4hWBmWKwUggjQ&sig2=n78EvnXLwFG2hq5K4zkcWg&bvm=bv.146094739,d.d24&cad=rja) defines icing intensity as follows:

> - Trace - Ice becomes perceptible. Rate of accumulation of ice is slightly greater than the rate of loss due to sublimation.
> - Light - The rate of accumulation may create a problem for flight in this environment for one hour.
> - Moderate - The rate of accumulation is such that even short encounters become potentially hazardous.
> - Severe - The rate of accumulation is such that de-icing/anti icing equipment fails to reduce or control the hazard. 

It then says:

> It is possible for carburetor ice to form even when the skies are clear and the outside air temperature is as high as 90 degrees Fahrenheit, if the relative humidity is 50 percent or more particularly when engine rpm is low. This is why, when flying most airplanes with carbureted engines, students are drilled to turn on the carburetor heat before making a significant power reduction. Carburetors can, however, ice up at cruise power when flying in clear air and in clouds. The envelope for the most severe buildups of carburetor ice is between 60 and 100 percent relative humidity and 20 to 70 degrees Fahrenheit.

So I do think that our modelling here does a reasonable job simulating that.

gilbertohasnofb:
BTW, that video of the Cub is great!

wlbragg:
@gilbertohasnofb where did you get the information about
>carb icing rate is multiplied by an oil temp factor so a cold engine doens't accumulate ice

Can you point me to the source, I need to see those numbers?

wlbragg:
Also, do you remember if you stop the ice buildup if the engine stops running?

tigert:
That cub video was very good, thanks for that :-)

#### A few things that come to mind:

   * When in training, we always applied carburettor heat before reducing power for landing (basically when we joined downwind for the traffic pattern, carb heat was the first thing, then power reduction, then flaps etc..

   * Carb heat is a lever, so it is proportional, but you are supposed to use it either fully ON for maximum effect, or fully OFF (to not waste power with overly rich mixture due to the warm air). If you fly in those conditions that are suspectible for carb icing, the procedure is to apply full carb heat for a good amount of time (two minutes or so) so that you'll melt the ice properly if there is any. And like on that Cub video, you'll hear the cackle and coughing from the engine at first if there was ice, as it melts down and gets swallowed by the engine, whichafter the rpm is restored and the engine smoothens out.

#### Extra fun
The carburettor heat is basically a valve in a Y-pipe between the air intake in the nose, and a muffle around the exhaust pipe. More fun for you implementing stuff? :-)

   * The regular air intake has the air filter. The carburettor heat does not. While the intake for the heating  muffle is inside the cowling, you still should avoid using carburettor heat in the ground, as the propeller might pick up sand / dust and that all can get sucked into the engine unfiltered.

   * The air filter can get clogged by ice or snow when flying in icing conditions, or bird strike etc.. blocking air flow to the engine. By using the carburettor heat, the air filter can be bypassed and thus the air can flow again.

   * If your exhaust pipe gets corroded through inside the muffle, you'll end up with exhaust gas / carbon monoxide inside the cockpit which can kill you silently. This is why some aircraft have that carbon monoxide detection sticker in the panel.

gilbertohasnofb:
@wlbragg

> Can you point me to the source, I need to see those numbers?

I don't have numbers, just used some trial and error to find decent values. I read somewhere that ice does accumulate faster the hotter the engine is, and that a cold engine will not accumulate ice at a substantial rate. And since we don't have an engine temperature property, I used the oil temp to infer the engine temperature. But I can't find the source now, do you think this is incorrect?

@kentrosaur

About the engine coughing, look what I just found in [this AOPA link](https://encrypted.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwigm6Xhj4PSAhXM6xQKHdtTCy4QFggZMAA&url=https%3A%2F%2Fwww.aopa.org%2F-%2Fmedia%2Ffiles%2Faopa%2Fhome%2Fpilot-resources%2Fasi%2Fsafety-advisors%2Fsa25.pdf%3Fla%3Den&usg=AFQjCNFL95yImAVBGQ7uZw8NBAyrJCH9hw&sig2=7F-N1ydwHpUaqsFGp8luEw&bvm=bv.146496531,d.d24)

> The first time you experience carb ice can be startling,
> since application of full carb heat – as recommended –
> will likely cause brief coughing and choking in the
> engine as the ice in the carburetor melts and is ingest-
> ed by the engine.

So perhaps using the coughing function for when the ice is melting would be a solution after all.

@tigert thanks for all the information! This is really helpful.

wlbragg:
> So perhaps using the coughing function for when the ice is melting would be a solution after all.

Yes, from what I see it is a good solution.

wlbragg:
>@tigert thanks for all the information! This is really helpful.

:+1:

gilbertohasnofb:
> I'll see what I can find. 

If you find anything, please let me know. If the idea was bogus or I somehow misunderstood something, please let me know and I can remove the oil temp dependency. I must have read a trillion links while working on the carb heat here, and it is difficult to track down where I came across what... :cry: 

>> So perhaps using the coughing function for when the ice is melting would be a solution after all.
> Yes, from what I see it is a good solution.

:+1: And do you think I should remove the low oil trigger of the cough function, as we spoke about in that other issue/PR?

gilbertohasnofb:
@wlbragg I can't seem to find any source for the icing rate being correlated to engine temperature... I am considering to remove this dependence on oil temp.

wlbragg:
>And do you think I should remove the low oil trigger of the cough function, as we spoke about in that other issue/PR?

Unless you have something stating that is a symptom of low oil or someone here in our dev team can verify, I would say yes, remove it. Especially at the low oil mark. You may want to keep it and use it and even more aggressively at the "major damage" mark, which would be somewhere between no oil and not enough to be picked up by either an oil pump or splashed around. I am guessing they have an oil pump? I wish I knew a little more about an aircraft engine.

wlbragg:
>I can't seem to find any source for the icing rate being correlated to engine temperature... I am considering to remove this dependence on oil temp.

It makes some sense except we form ice in our air conditioning systems in +100 deg temp because of the humidity. Granted it is a somewhat different system, but some of the same principles apply, like air being restricted at high enough speeds. I read that the ice on the carb butterfly valve (this may also only be specific to certain engines) forms from the humid airflow volume being restricted. But the idle engine is where the icing occurs most so the heat energy has to play a part too. I think we need to search a little more before you remove it. I'm looking too.

wlbragg:
http://www.pilotfriend.com/training/flight_training/fxd_wing/carb_heat.htm
>The airflow through the carburettor is controlled by the throttle moving a circular flat rotating plate. Air passing by this butterfly valve is accelerated by a narrowing throat called a venturi....

>This narrowing reduces pressure and sucks cooling fuel into the air. The combination of a lower pressure, gasoline, and moisture in the air can so cool the metal parts of the intake that the moisture adheres to the parts as ice. The greater the air's humidity the more ice will be formed. As the airflow is restricted by the ice, the excess fuel so enriches the mixture that the engine to flood and misfire. The initial signs of this icing is a gradual decrease in rpm and an increase in engine roughness. Avoidance of carburettor ice can be improved by avoiding long, low power descents and by leaning to make the engine run hotter. Lycoming engines are less susceptible to ice than are Continental engines since the carburettor is attached so as to benefit from the heat of the engine itself.

>Carburettor ice is caused by the change in pressure as air passes through the venturi. Fuel atomized in the throat of the venturi evaporates. This uses heat from the metal of the carburettor and air. This air/fuel at lower pressure cools and takes even more heat away from the metal of the carburettor. At some point any moisture in the air will freeze on the metal of the carburettor. The ice that adds to the venturi constriction will both increase the cooling rate, the speed of airflow and lower the pressure. Thus, carburettor ice feeds on itself until the excessively rich mixture kills the engine. 

This is some confirmation
> Lycoming engines are less susceptible to ice than are Continental engines since the carburettor is attached so as to benefit from the heat of the engine itself.

gilbertohasnofb:
@wlbragg 

> Unless you have something stating that is a symptom of low oil or someone here in our dev team can verify, I would say yes, remove it. Especially at the low oil mark. You may want to keep it and use it and even more aggressively at the "major damage" mark

I was referring only to the cough function, but I believe the low oil mark is behaving according to the POH. I had a look at the [operation handbook](https://encrypted.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=0ahUKEwiYmenOyIPSAhWC0xQKHVbLCy8QFggbMAA&url=https%3A%2F%2Fwww.ing.unlp.edu.ar%2Fcatedras%2FA0020%2Fdescargar.php%3Fsecc%3D0%26id%3DA0020%26id_inc%3D17972&usg=AFQjCNG_NdZFn2-7GT9yPqhm3085efqMBw&sig2=ruskA0pTFl5Dif8e8C48Bg) of the Lycoming O-320, and according to the page 3-10, the minimum safe quantity in sump is 2 quarts. By what I was reading in some forums, this limit of 2 quarts is the minimum for safe operation, but one should keep in mind that the oil also helps to cool the engine, and also that at higher pressure and/or when performing maneauvers, the engine might get depleted from oil if the level is 2 quarts only. This is why the 172 POH states that 5 quarts is critical.

> This is some confirmation
>> Lycoming engines are less susceptible to ice than are Continental engines since the carburettor is attached so as to benefit from the heat of the engine itself.

If I understand this, it means that hotter engine equals to less ice, correct? So that the correlation should be the opposite of what I did with oil temp. I think if that's the case, then actually we should remove the dependency on oil temp since the formula that controls the icing rate actually uses RPM as a factor.

So what do you think about the following:

- set minimum oil limit to something below 2 quarts
- but make sure that the oil heats up prior to that 2 quarts value
- no coughing when low oil, simply kill the engine when value too low
- coughing when ice is melting from the carburettor

tigert:
Speaking of engine articles, this list is interesting read: http://www.avweb.com/news/pelican/182146-1.html#engine

wlbragg:
>If I understand this, it means that hotter engine equals to less ice, correct?

Yes, that is where I got confused also because the comment was opposite this.

>I think if that's the case then we should incorporate this and make warm engines less susceptible to icing

I agree.

> set minimum oil limit to something below 2 quarts

I'm not sure what is meant here, What happens at the minimum?

> no coughing when low oil, simply kill the engine when value too low

Again, if there is a built in physical kill switch on the engine then we could potentially see this condition and you would never ruin the engine because it would stop long before low enough oil for engine damage. (See the other thread about low oil)

Yes to all the below
    >coughing when ice is melting from the carburettor
    >warm engines less susceptible to icing
    > but make sure that the oil heats up prior to that 2 quarts value

wlbragg:
Some SuperCub pilot experiences with ice.
http://www.supercub.org/forum/showthread.php?26674-Carb-Heat-or-No-Cab-Heat

gilbertohasnofb:
@wlbragg 

> Yes, that is where I got confused also because the comment was opposite this.
> [...]
> Yes to all the below

:+1: 

As for the oil, I will reply on the other thread as to keep it organized

@tigert I can't open that link, do I need a login for it?

wlbragg:
@gilbertohasnofb  I did a hack of you code for the Cub to simulate a cool or hot engine. It's not tuned at all, It is the first iteration.

~~~xml
var oil_temp_factor = std.max(1.0, ((oil_temp - 120) / 100)+1);
~~~

gilbertohasnofb:
I will check it out. If not today, tomorrow I should work on these things.

wlbragg:
Oh, I guess this would be the add heat to melt carb ice as the engine warmed. Thus the next line needed the following change.
var carb_icing_rate = carb_icing_formula + carb_heat_rate + oil_temp_factor;

Yes, please do inspect closely. :smile:

wkitty42:
On 02/09/2017 03:46 PM, Gilberto Agostinho wrote:
> @tigert <https://github.com/tigert> I can't open that link, do I need a
> login for it?

i seem to have opened it just fine... it is a (narrow column) list of article 
links to engine related articles...

gilbertohasnofb:
@wkitty42 you are right, it was just one of those full screen pop ups which require you to click "thanks but no thanks" before proceeding

gilbertohasnofb:
@wlbragg thanks for those formulas, I like how you solved it! But I think I will tweak the values, if that's fine by you. The reason is that a) at high temps that formula is contributing way too much to the melting considering the other values I had and b) at low temps it makes ice accumulate faster, and I would rather simply have the oil temp help with melting, not with ice accumulation

gilbertohasnofb:
@wlbragg I will simply use `var oil_temp_factor = oil_temp / 833000;` so that for 0 F => oil_temp_factor = 0.0 and for 250 F => oil_temp_factor = 0.0003, which is one quarter of the efficacy of full carb heat.

gilbertohasnofb:
So, this is what is new:

- new minimum oil level: 1.97 quarts will kill the engine, anything above is fine
- oil temp raises and oil pressure lowers when below 3.0 quarts oil
- autostart will fill the oil if level < 5.0 quarts (which is Cessna's recommended minimum value for short flights), else leave it as it is
- low oil does not make the engine cough any longer
- but if ice is present in the carburettor and the ice is melting (either because of carb heat or because of natural melting) then engine will cough
- icing formula recalculated: hot engine melts a bit of ice, carb heat melts a lot of ice, ice accumulation rate diminished a little
- ice now does accumulate when the engine is off

@wlbragg I am having troubles thinking of an elegant way of how to lower the RPM just before the oil level reaches the minimum value, all I could come up with is a bit complicated... I will try again later

@all for those interested, please re-test the carb icing implementation! The formula changed substantially, so rates will be different.

wkitty42:
On 02/09/2017 06:17 PM, Gilberto Agostinho wrote:
> @wkitty42 <https://github.com/wkitty42> you are right, it was just one of
> those full screen pop ups which require you to click "thanks but no thanks"
> before proceeding

oh, i don't get any of that shite because of my protections that filter out 
scripts and ads ;) ;) ;)

wlbragg:
It all sounds good except
>ice now does accumulate when the engine is off

Why is that? I thought the ice accumulation was from the gas air mixture funneling through the carb's air intake. If the engine is off, why would it be sucking fuel?

>@wlbragg I am having troubles thinking of an elegant way of how to lower the RPM just before the oil level reaches the minimum value, all I could come up with is a bit complicated... I will try again later

Can you hard code the volumetric-efficiency-factor = whatever it needs to be to lower the rpm when oil level is from something like 2.0 quarts to 1.97 quarts?

wlbragg:
@gilbertohasnofb don't forget to change the comment about engine heat reducing ice. not increasing it. :smile:

wlbragg:
So far the only thing I've noticed is maybe there needs to be a clamp on the max amount of negative ice accumulation (or melting) on the property carb_ice. I get some strangely large negative numbers in the prop carb_ice, like 9.892749347e-06 which also seems to cause coughing, especially early on before the system stabilizes?

wlbragg:
Looking at the carb_ice syntax, I'm not sure you can use the same variable to store itself and another calculation in. that may be why were getting such odd numbers. 

In other words I don't think you can do
 
~~~xml
var carb_ice = getprop("/engines/active-engine/carb_ice");
carb_ice = carb_ice + carb_icing_rate * 0.00001;
carb_ice = std.max(0.0, std.min(carb_ice, 1.0));
~~~

instead it should be something like

~~~xml
var carb_ice_val = getprop("/engines/active-engine/carb_ice");
var carb_ice = std.max(0.0, std.min(carb_ice_val + carb_icing_rate * 0.00001, 1.0));
~~~

However I am still getting the strange negative numbers so I'm working through it one calculation at a time trying to trace where it seems to break.

wlbragg:
OK, maybe you meant to have the number that small, thus a slow build up. But I still am unsure of why coughing at the low number? Unless it is something else.

Any idea of a very favorable METAR string for icing?

wlbragg:
OK, I get it, genius!

What is happening is there is very small numbers of ice building up ie: 9.934984573e-6. If I add power and that number starts to drop lower (even though it is incredibly low already) ice is melting and water is introduced into the mixture and into the carb, thus coughing. Looks like it is working great. 

The main thing to analyze then is the speed of ice buildup and speed of thaw.

Using 13/08 is producing ice at a decent clip. I guess I need to time it on a long approach and see if it parallels the Cub video.

Only other thing I have a question about it the mixture, do you have that affecting the numbers? It is suppose to help slow icing to lean it out, I believe.

wkitty42:
slight diversion from the issue/topic but kinda relevant since there was some discussion about the oil level and how it aids in the cooling of the engine...
[His small plane (Cessna 172RG) spewing smoke over Addison, a young pilot has only a moment to act](http://www.dallasnews.com/life/people/2017/02/01/small-plane-billowing-smoke-addison-young-pilot-moment-act)

gilbertohasnofb:
> don't forget to change the comment about engine heat reducing ice. not increasing it.

Thanks, I totally forgot about it! :smile: Just pushed a fix

> OK, I get it, genius!

hehehe :laughing: So, as you probably realize the cough happens if `carb_ice` > 0 and if `carb_icing_rate` < 0, i.e. if there is ice and it is melting. I thought of adjusting the values so that coughing happens only at considerable melting rates or only when considerable carb ice has accumulated, but when the values are small carb heat seems to melt them in only a few seconds, barely giving any coughing at all.

> Using 13/08 is producing ice at a decent clip. I guess I need to time it on a long approach and see if it parallels the Cub video.

Oh, I would love about this kind of test! 13/08 gives some tough icing but it's worse at low RPMs, while the worst scenario is around 07/07 which gives severe icing at low as well as at high RPM. Though carb heat will melt in _all_ circumstances, at high RPM the 07/07 scenario will accumulate ice at almost double the rate as 13/08.

> Only other thing I have a question about it the mixture, do you have that affecting the numbers? It is suppose to help slow icing to lean it out, I believe.

gilbertohasnofb:
@wkitty42 very nice read, thanks for sharing it. In our implementation, low oil will slowly lead to high engine temps (starting at oil < 3 quarters). From that point until the failure point (1.97 quarts) there is _a lot_ of flying, so in theory someone can spot the low oil level by monitoring the oil temp and oil pressure gauges.

gilbertohasnofb:
@dany93 @wlbragg I just saw your conversation in the forum. Dany is not seeing the `/engines/active-engine/oil-temperature-degf` property because I updated the file `Systems/engine.xml` in this current branch only and this has not been merged in the master. After we merge this, that property will reflect the oil temperature for the active engine, regardless of which engine that would be. Makes sense?

dany93:
OK, no problem for me, it was just to answer @wlbragg . Thanks for the explanation.

gilbertohasnofb:
@wlbragg Oh, I pasted all the quotes I wanted to address then forgot about the last one :man_facepalming: 

> Only other thing I have a question about it the mixture, do you have that affecting the numbers? It is suppose to help slow icing to lean it out, I believe.

Do you have some source for that? Leaning below the ideal mixture position will lower the EGT which in turn will make the carb heat less effective. But how does the mixture affect the icing formation?

wlbragg:
I'll have to do some searching to find it again. It may not be worth the hassle to model it though. I think it is minimal but maybe something worth having and knowing about for engines without carb heat.

gilbertohasnofb:
@wlbragg Ok, let me know if you find something, but if the effect is small then probably it would not be worth adding it, since all these icing rates here are "guessed" anyway. But do let me know if you find something, I am quite curious about it.

@everyone So I guess this branch is pretty much done, we just need some testing before merging it. Would you guys like to make some tests, perhaps @kentrosaur would have the chance to give a go as well? Any feedback would be appreciated, and if it all feels reasonable then we can merge this and add some note about it in the wiki and perhaps in the forum as well.

kentrosaur:
@gilbertohasnofb just tried things out, it's not ready yet but I think it's close.  METAR by default was a nice 21/20.

- carb ice forms when engine is off: this shouldn't occur as there's no air flow to create the pressure drop which causes the ice to form in the carburetor.  
- oil temp acting as natural carb heat: I think this is working great, it would go positive/negative in the pattern at 1500 rpm but that leads to a second issue. 
- coughing: this surprised me on run-up with the carb heat off (rpm and temp changes).  It was pretty severe given the amount of ice (^-05).  Is there a way to control the rate? The coughing is from melting water, so with a large rate of melting ice, this is expected.  If the rate is lower, maybe  between -1 and 0 it should be much less frequent.  This would also eliminate a lot of the coughing in the pattern when it flips positive and negative.
- mixture and coughing: combined with the last point, a leaner mixture should be able to reduce the coughing 

Unrelated oil:
- The oil capacity should be 8 quarts, filled to a default of 6 or 7 for normal operations.  Is this saved between flights?

gilbertohasnofb:
@kentrosaur Fantastic feedback, thanks a lot. But I need some help with some of these points:

> - carb ice forms when engine is off: this shouldn't occur as there's no air flow to create the pressure drop which causes the ice to form in the carburetor.

Yes. Before ice would not form until a certain engine temperature was reached, but then @wlbragg and I argued that the hotter the temperature of the engine, the less ice it would develop, so I ended up removing this. Can you explain to me why carburettor ice does not form when the engine is off (or perhaps help me find some source)?

> - oil temp acting as natural carb heat: I think this is working great, it would go positive/negative in the pattern at 1500 rpm but that leads to a second issue.

:+1: 

> coughing: this surprised me on run-up with the carb heat off (rpm and temp changes). It was pretty severe given the amount of ice (^-05). Is there a way to control the rate? The coughing is from melting water, so with a large rate of melting ice, this is expected. If the rate is lower, maybe between -1 and 0 it should be much less frequent. This would also eliminate a lot of the coughing in the pattern when it flips positive and negative.

Good idea, I will see what is possible.

> mixture and coughing: combined with the last point, a leaner mixture should be able to reduce the coughing

Good idea again. I will add it to the coughing conditional that coughing should happen only when ice is accumulated at a certain point, melting is happening over a certain rate AND mixture is richer than a certain value.

> The oil capacity should be 8 quarts, filled to a default of 6 or 7 for normal operations. Is this saved between flights?

If you have the option to save aircraft state between flights activated in the aircraft menu, then yes oil level (just like fuel level) does save between flights. But notice that if you do use autostart and your oil is lower than 5 quarts, it will replenish it as recommended by the POH

wlbragg:
>Can you explain to me why carburettor ice does not form when the engine is off (or perhaps help me find some source)?

>this shouldn't occur as there's no air flow to create the pressure drop which causes the ice to form in the carburetor.

@kentrosaur is correct, I meant to reply with this in my feedback but somehow missed it. The engine needs to be running and sucking the air-fuel mixture for the pressure drop to be created. That is the cause of the icing. That also relates to the minor effect I pushed for of the mixture percentage having an influence. I still haven't found what I read about it and don't feel comfortable gong off of memory. But vaguely it has to do with the density of the gas in the mixture and the amount of evaporation.

gilbertohasnofb:
> The engine needs to be running and sucking the air-fuel mixture for the pressure drop to be created. That is the cause of the icing.

That makes sense, thanks to you all for the input. So I will add a conditional that ice will not form if the active engine is off.

wlbragg:
>So I will add a conditional that ice will not form if the active engine is off.

I think the reason I decided not to reply about building ice when the engine was not running was that you have RPM in the icing formula. So I assumed that engine not running meant no RPM's, thus no ice. Doesn't this have an influence?

gilbertohasnofb:
The thing is that RPM is not linear in that function, it works in the following way:

rpm factor x-axis

This means that lower RPM => larger icing rate. But notice that even at 0 RPM the function stills output a value. So I need an extra conditional about the engine being running or not.

wlbragg:
I've been testing this quit a bit (admittedly on the 150hp SuperCub) and the coughing at the smaller rates of ice melt does need to be toned down. But when your working on it, if there is a way to let it "occasionally" cough "just a bit" at the lower ice rates, enough to give the pilot feedback that there is icing and melt happening, that would be ideal.

wkitty42:
On 02/11/2017 07:00 AM, Gilberto Agostinho wrote:
> @kentrosaur <https://github.com/kentrosaur> Fantastic feedback, thanks a lot.
> But I need some help with some of these points:
>
>       * carb ice forms when engine is off: this shouldn't occur as there's no
>         air flow to create the pressure drop which causes the ice to form in the
>         carburetor.
>
> Yes. Before ice would not form until a certain engine temperature was reached,
> but then @wlbragg <https://github.com/wlbragg> and I argued that the hotter the
> temperature of the engine, the less ice it would develop, so I ended up removing
> this. Can you explain to me why carburettor ice does not form when the engine is
> off (or perhaps help me find some source)?


with a carburetor, there has to be airflow into the throat of the carb... that 
airflow, just like in the real world, will lower the temperature... in the real 
world we call it the "wind chill"... if there's no airflow, there cannot be wind 
chill in the carb throat and thus, no freezing of moisture there... there can't 
be air flow into the carb throat because there's no exit for the air to get to 
and leave the, now closed, carb throat... it is like blowing air into a cup... 
it comes right back out the opening... it might be possible for that to lead to 
some icing but i don't have any idea how much...

there may be situations with a parked unpowered craft where moisture can reach 
into the engine compartment or be blown in through the intake eg: parked and 
facing into the wind... if the air temperature is cold enough, it is possible 
for that moisture to freeze and cause restriction in the air flow... as the 
compartment warms up, that moisture can (and should) thaw to some extent which 
may introduce some moisture into the air flow and then into the engine... it may 
not be enough to cause any noticeable coughing or engine roughness... one can 
see this with their cars and trucks at times...


-- 
        *Please keep mailing list traffic on the list* unless
        private contact is specifically requested and granted.

wlbragg:
@gilbertohasnofb regarding the carb-icing-rate, while testing the ceiling in the SuperCub slowly climbed to 28,000 (about 7,000 more than what the absolute ceiling should be, although that's a separate story). After reaching that ceiling at 35% mixture, (after maybe a half hour?),  I cut the throttle to idle and allowed the aircraft to descend. This was using a manual METAR of 07/07. After many minutes of descending the carb-icing-rate was still negative but climbing the whole time. My question is, do we need to clamp both the negative and the positive rate. Because if it is allowed to accumulate over time, then it could take too long to get back to a positive, like in this case. Technically it should have started into the icing process soon after I went to idle decent, right?

gilbertohasnofb:
@wlbragg the temperature and dew point in a METAR is only for the ground level. When you go higher, there might be a dramatic difference between what you experience and what the METAR says. So in your case, at 28000 ft (on a Cub!) the temperature was very likely way too cold for any icing to happen. Very often you have only a layer in which icing happens, and you would be safe both over as well as under it. But that said, if you were at 28000 ft and descend all the way down to the ground with a METAR 07/07, at some point you should have picked up quite some ice.

wlbragg:
OK, putting that aside, is it possible that the carb-icing-rate could potentially be at a negative rate of increase for so long that it would take forever to reach the positive if a normal (average) ice conditions existed. For an extreme example, I am flying for a day in conditions that a negative icing rate is dropping lower and lower (or is that not the way the  carb-icing-rate number would work). Then I hit weather that is conducive to fast icing. How long is it going to take for that days worth of negative build number to get back to a positive. Or maybe I am not understanding how that carb-icing-rate is created, or more importantly its range might be limited.
Because back to my actual test above. As soon as I went to idle the  carb-icing-rate started to climb,but had it been dropping so long that it is going to take forever to reach a positive, even though right then, at that moment in time, ice should be building, but we have to wait until the carb-icing-rate gets back to a positive.
I hope I am making sense here and not misunderstanding how the carb-ice and  carb-icing-rate are derived.

gilbertohasnofb:
> How long is it going to take for that days worth of negative build number to get back to a positive. Or maybe I am not understanding how that carb-icing-rate is created, or more importantly its range might be limited.

The carb icing rate changes instantaneously, what takes time to build or melt is the carb ice itself. The rate depends on that sum of three components: carb heat (which only contributes negatively, i.e. melting), oil temp (only negative again) and the icing formula (which can be positive or negative). The icing formula will output positive values in icing conditions (e.g. 07/07, 13/08, etc.) and positive values for non icing conditions. Also, RPM affects the formula, so if you have light icing at a certain temp/dew point combination at low RPM, you might have melting at a higher RPM. But this is all instantaneously, so if you suddenly reach a layer of icing conditions then you should have a positive rate immediately. Makes sense?

So if you are seeing only a negative rate it's because: a) the conditions are not good for icing at all or b) the oil temp factor is contributing more than the icing formula, which can happen in very mildly icing conditions.

Does this answer what you are trying to understand or did I misunderstand you?

kentrosaur:
@wkitty42 windchill is only an apparent temperature which results in more rapid cooling to the ambient temperature and unrelated to carb icing which is from the pressure drop in the carburetor.   [experiment #4 water phase change in a vacuum chamber](https://youtu.be/Ti9C_cLSR0A)

For the second case, I think the carb would be one of the last parts of the plane experiencing icing in those conditions.

@wlbragg in your case, I think you just need to in a non icing metar (15/-10) then hit a pocket of humid air (15/8).  You'll start developing ice **immediately** as the air coming into the carburetor now has a lot more moisture.  There shouldn't be any memory on your engine temperature as that's just a mitigation of the ice, not a cause.

wlbragg:
OK, I was misunderstanding the computation of the carb-icing-rate, I'm good, thanks.

dany93:
@wkitty42 wrote:

> with a carburetor, there has to be airflow into the throat of the carb... that airflow, just like in the real world, will lower the temperature... in the real world we call it the "wind chill"

No.
In the carburetor, the freezing is not due to the simple airflow ( = the wind). It is due to the gas expansion. The air is sucked by the pistons through the carburetor throat, which makes its pressure decrease.
If it would be due to the simple airflow, the carburetor temperature could never decrease below the atmospheric temperature.

wkitty42:
On 02/11/2017 01:10 PM, kentrosaur wrote:
> @wkitty42 <https://github.com/wkitty42> windchill is only an apparent
> temperature which results in more rapid cooling to the ambient temperature and
> unrelated to carb icing which is from the pressure drop in the carburetor.
> experiment #4 water phase change in a vacuum chamber <https://youtu.be/Ti9C_cLSR0A>

ahh, yes... i was thinking thermally and not about the pressure reduction in the 
carb throat during operation... thanks for the reminder and the pointer to the 
video... we actually did that exact experiment back in junior high school but it 
took a lot longer and was much louder because our vacuum pump was a lot more 
primitive than that one in the video...

> For the second case, I think the carb would be one of the last parts of the
> plane experiencing icing in those conditions.

i'm not sure where the carb is mounted in relation to the intake port where the 
air is actually drawn in, filtered and the fed to the carb... i was also basing 
a lot off of motor vehicles as that's where i have the most experience ;)

wkitty42:
On 02/11/2017 01:36 PM, dany93 wrote:
> @wkitty42 <https://github.com/wkitty42> wrote:
>
>     with a carburetor, there has to be airflow into the throat of the carb...
>     that airflow, just like in the real world, will lower the temperature... in
>     the real world we call it the "wind chill"
>
> No.
> In the carburetor, the freezing is not due to the simple airflow ( = the wind).

true to a point but if you put your finger in the airflow, it will definitely 
feel colder just like when it is 60F outside and the wind chill has it at 50F 
like we had here the other day ;)

in any case, as just noted, i was thinking purely thermally and forgot about the 
pressure drop due to the vacuum which also contributes to the change in boiling 
point and temperature reduction...

gilbertohasnofb:
Ok, so I pushed a small new commit and now the icing formula will output 0 if engine is off. Notice that the `carb_icing_rate` could still be negative if the engine is off (if the engine is hot for instance), but it won't ever be positive.

I am working now on the coughing function and the effect that the mixture would have on it. Once I have something I will push and let you guys now.

gilbertohasnofb:
Ok guys, I had to resort to a bunch of trickery to make this work, but let's give this a shot:

- the cough function is improved and now its interval is variable

- fuel contamination and icing cause coughs. In the case of fuel contamination, the coughing interval depends on the quantity of water. Near no contamination will cause coughs every 17-20 seconds, while critical contamination will cause coughs in the interval below 3 seconds. To test this, adjust the values of `/consumables/fuel/tank[0]/water-contamination` and `/consumables/fuel/tank[1]/water-contamination`, keeping in mind that each of these properties can have a value of up to 0.2 before the engine completely quits

- icing affects cough in a similar way, the more ice in the carburettor, the more often the coughs will be once it's melting. A tiny bit of ice causes coughs every 7-10 seconds, while near critical icing accumulation will drop that interval to averagely 2 seconds

- leaning a bit the engine (relative to the max EGT you could have at that pressure altitude, of course) will make the engine stop coughing. I am using the property `/engines/active-engine/egt-norm` to judge that: when leaned for max EGT, this value is around 0.7x. The coughing will stop if this value will be 0.5 or lower (to judge this, either watch the property or watch the EGT gauge and make sure the needle doesn't cross the middle of its range (i.e. 0.5 in a scale from 0.0 to 1.0).

- since the coughing function uses a timer, and it's the refresh rate that gets affected by all these, you might still get "one last cough" after leaning or after the ice has melted completely

- a stopped engine does not accumulate (further) ice

Please, test, test, test and if possible give me some feedback.

kentrosaur:
@gilbertohasnofb looks great and I'm going to give it a quick test here but one thing stands out:

> icing affects cough in a similar way, the more ice in the carburettor, the more often the coughs will be once it's melting. A tiny bit of ice causes coughs every 7-10 seconds, while near critical icing accumulation will drop that interval to averagely 2 seconds

The sinister thing about carb ice is the only indication you have is lower power/rpms until the carburetor is almost frozen.  I think it should only cough while it is melting (lots of heat) or when you're about to have the engine cut out.  This follows the concept that the rough engine is from the extra water in the mixture from the melting ice and the fuel mixture is too rich, causing it to explode somewhere besides the piston or on the wrong stroke.

gilbertohasnofb:
@kentrosaur note that I wrote _once it's melting_. That is, the cough is not related only to ice accumulation (that only makes you lose the power), so you can ice up your carb until the engine quits without hearing a single cough, but if the ice melts (due to carb heat, engine heat or air temp/dew change) _then_ the quantity of ice is used to judge the amount of cough.

gilbertohasnofb:
> looks great and I'm going to give it a quick test here

Thanks, really appreciate it!

kentrosaur:
@gilbertohasnofb I misread that, it's working great!  Default metar, couldn't overcome the engine ice which is good.  Changed to the 07/07 built up some ice, then got it to cough eventually when it melted!  I feel like it's pretty realistic now.

A couple of more unrelated things on the 172 (should I make a separate ticket?)

- radio (metar on com2) keeps talking after switching off master or avionics switches.  
- toggle walker outside then inside, the magnetos always jump to both from off after re-entering. We can fix this, or we can add an interesting animation of the walker being hit by the propeller from an inadvertent engine start if bumped.
- 8 quart oil capacity

wlbragg:
>it's working great!

Yes, much better. It's not so aggressive coughing now. Good work. I already have adapted it to the Cub. Maybe we will get some more users testing it by it being in the Cub too.

Oh, I think you may have a typo in the comment
># a warm engine will accumulate less oil than a cold one, which is what oil temp factor is used for

Don't you want it to read "less ice" instead of "less oil"?

gilbertohasnofb:
@wlbragg 

> Yes, much better. It's not so aggressive coughing now. Good work. I already have adapted it to the Cub. Maybe we will get some more users testing it by it being in the Cub too.

Thanks a lot! I think if no one has anything else to comment here we can simply merge this and then adapt from the feedback that the users would give to us (if there would be any).

> I think you may have a typo in the comment

:laughing: :+1: I promise I will get this right this time!

@kentrosaur 

> - radio (metar on com2) keeps talking after switching off master or avionics switches.
> - toggle walker outside then inside, the magnetos always jump to both from off after re-entering. We can fix this, or we can add an interesting animation of the walker being hit by the propeller from an inadvertent engine start if bumped.

Would you please open a new issue for each of this? I can try taking a look at them after.

> - 8 quart oil capacity

I am quite sure our oil capacity is fine. As far as I know, the _total_ oil capacity of the 160 HP is indeed 8 quarts and of the 180 HP is 9 quarts, but there is always 1 quart inside the engine. This means that the 160 HP will have at most 7 quarts in the oil sump and the 180 HP will have at most 8 quarts in the oil sump, where the measurements are taken from. So this is what I simulated with that gui dialog, you can add up to 7 "measured" quarts to the engine, but in fact the aircraft will have 8 quarts. Also, the critical value of 2 quarts according to Lycoming manual should be read as "2 measured quarts from the oil sump". Makes sense?

So all: I am pushing a last fix to the comment typo and opening a PR

PR #828, opened 2016-08-12
Fixing console error message when changing engines

### LINKED ISSUE #827: Changing engines output error message
When changing between the 160 HP and 180 HP engines, I get the following error:

`Failed to execute command dialog-close`

This is because I had added the following bit of code binded to the engine change:

```
                    <binding>
                        <command>dialog-close</command>
                        <dialog-name>c172p-oil-dialog-160</dialog-name>
                    </binding>
```

That is, if you change from the 160 to 180 HP and the oil dialog of the 160 HP was opened, it was force closed, and vice-versa. This is necessary because each engine has its own dialog, and you shouldn’t be able to open the wrong one. The issue is that if the dialog is not opened, FG outputs an error. So is there a way of adding a conditional to that binding which would detect if the dialog is opened or not?

gilbertohasnofb:
Closes #827 

When changing between the 160 HP and 180 HP engines, the user would get the following error:

`Failed to execute command dialog-close`

This is because of how the oil dialogs was handled. See the issue above for the explanation of what caused it and how this fix work

### DIFF
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -436,13 +436,17 @@
                         </script>
                     </binding>
                     <binding>
-                        <command>property-assign</command>
-                        <property>/engines/active-engine/oil-level</property>
-                        <value>7.0</value>
+                        <command>dialog-show</command>
+                        <dialog-name>c172p-oil-dialog-180</dialog-name>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
@onox this is pretty safe to be merged (after reviewing of course) before the release

PR #832, opened 2016-08-12
autostart now recharges the battery

### LINKED ISSUE #831: Autostart should recharge battery
Autostart should recharge battery

gilbertohasnofb:
Closes #831

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -9,6 +9,8 @@ var autostart = func (msg=1) {
         return;
     }
 
+    electrical.reset_battery_and_circuit_breakers();
+
     setprop("/controls/switches/magnetos", 3);
     setprop("/controls/engines/current-engine/throttle", 0.2);
     setprop("/controls/engines/current-engine/mixture", 0.95);


onox:
The battery already gets charged after the autostart? Do we really want to reset the battery to a full charge during every autostart?

What I don't like about resetting all these things in autostart is that it's very easy and tempting for someone to press autostart. But doing that resets everything and then there's little that stays persistent across multiple FG sessions.

So can we maybe instead always reset circuit breakers, and set the battery charge to a certain minimum if it's below that threshold? (A charge level that is high enough to start the engine)

gilbertohasnofb:
> The battery already gets charged after the autostart?

Well, the battery gets charged by the alternator if and only if the engine is running. If the battery level is too low, autostart will fail.

> What I don't like about resetting all these things in autostart is that it's very easy and tempting for someone to press autostart. But doing that resets everything and then there's little that stays persistent across multiple FG sessions.

Well, we agreed that autostart should take care of everything since it's aimed at beginners, it really should start the engine regardless of what a n00b did (no oil, fuel contaminated, low fuel, no battery, etc.). If we are lazy to do things "properly", that's the price to pay :stuck_out_tongue_closed_eyes: 

> So can we maybe instead always reset circuit breakers, and set the battery charge to a certain minimum if it's below that threshold? (A charge level that is high enough to start the engine)

I honestly don't see the point because battery charge is different than oil or fuel level in one fundamental issue: the battery is self recharging! So even if autostart would set the battery to whatever low charge using that logic, after the flight the battery will very likely be fully charged anyway. It takes around 15 minutes to go from 30% to 100%, so I personally think this PR is the way to go.

gilbertohasnofb:
@onox if we merge this soon enough, I think I might ask Stuart to replace the `c172.nas` file at FGDATA. I am really afraid this issue here will make tons of beginners puzzled :cry:

onox:
Tell Stuart to look at the `release/2016.3` branch. Put me in the CC if you can.

gilbertohasnofb:
:+1:

PR #839, opened 2016-09-10
Autostart handles battery charge locally

### LINKED ISSUE #838: "Start with engine running" does not work: Nasal errors
When I launch the simulator with "Start with engine running" **checked**, the engine is not running and I get Nasal errors (only when checked).

```
Nasal runtime error: non-objects have no members
 at /home/dany/FG/github/c172p/Nasal/electrical.nas, line 43
  called from: /home/dany/FG/github/c172p/Nasal/c172p.nas, line 12
  called from: /home/dany/FG/github/c172p/Nasal/c172p.nas, line 325
  called from: /home/dany/FG/github/c172p/Nasal/c172p.nas, line 573
  called from: /home/dany/fgfs/install/flightgear/fgdata/Nasal/globals.nas, line 119
```

I get this with the C172P embedded in the current FG 2016.4.0 Pack version and the C172P 95260d7 github version (both from10 sept 2016).

The "Autostart" works well.

gilbertohasnofb:
Closes #838 

This PR makes changes to the autostart function:
- it now handles battery charge and circuit breakers locally instead of calling a function from electrical.nas. The issue was that when the option of starting the sim with engines on was toggled, the autostart function was being called before the electrical system was initialized. I did not manage to force it to initialize before the c172p.nas, but on top of that doing so could have caused more bugs. Doing things locally looks far more elegant in my opinion (though the original battery recharging function should stay there as it's used by one of the GUI dialogs)
- changed the order of some lines in the autostart function to make it more logical to a reader. E.g. checking for fuel level and battery charge is now done before setting the magnetos
- added more comments to make the function even clearer

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -9,8 +9,32 @@ var autostart = func (msg=1) {
         return;
     }
 
-    electrical.reset_battery_and_circuit_breakers();
+    # Filling fuel tanks
+    setprop("/consumables/fuel/tank[0]/selected", 1);
+    setprop("/consumables/fuel/tank[1]/selected", 1);
 
+    # Reset battery charge
+    setprop("/systems/electrical/battery-charge-percent", 1.0);
+    
+    # Reset circuit breakers
+    setprop("/controls/circuit-breakers/master", 1);
+    setprop("/controls/circuit-breakers/flaps", 1);
+    setprop("/controls/circuit-breakers/pitot-heat", 1);
+    setprop("/controls/circuit-breakers/instr", 1);
+    setprop("/controls/circuit-breakers/intlt", 1);
+    setprop("/controls/circuit-breakers/navlt", 1);
+    setprop("/controls/circuit-breakers/landing", 1);
+    setprop("/controls/circuit-breakers/bcnlt", 1);
+    setprop("/controls/circuit-breakers/strobe", 1);
+    setprop("/controls/circuit-breakers/turn-coordinator", 1);
+    setprop("/controls/circuit-breakers/radio1", 1);
+    setprop("/controls/circuit-breakers/radio2", 1);
+    setprop("/controls/circuit-breakers/radio3", 1);
+    setprop("/controls/circuit-breakers/radio4", 1);
+    setprop("/controls/circuit-breakers/radio5", 1);
+    setprop("/controls/circuit-breakers/autopilot", 1);
+
+    # Setting levers and switches for startup
     setprop("/controls/switches/magnetos", 3);
     setprop("/controls/engines/current-engine/throttle", 0.2);
     setprop("/controls/engines/current-engine/mixture", 0.95);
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
@dany93 this should solve the issue you reported

dany93:
Partially.
No more Nasal error messages,
but the starter makes some turns and stops,  with the message (approximately):
"The engine failed to start. You must lean the mixture and start manually"

onox:
> Doing things locally looks far more elegant in my opinion (though the original battery recharging function should stay there as it's used by one of the GUI dialogs)

I disagree with this. You have duplicated a whole chunk of code. That doesn't seem nice.

onox:
Have you tried moving

```xml
<electrical>
    <file>Nasal/electrical.nas</file>
</electrical>
```

above `<c172p>` in `c172p-set.xml`?

onox:
I see in `electrical.nas`:

```javascript
settimer(init_electrical, 0);
```

Sounds fishy.

onox:
@gilbertohasnofb Have you tried assigning `battery` and `alternator` directly where they are declared (lines 12 and 13) and replace that `settimer()` call (see comment above) with an FDM initialized callback?

dany93:
@gilbertohasnofb 
I think you have deleted your question, but I have this with only "Start with engine running" enabled. The others, not. Although I've not tested other combinations...

onox:
@gilbertohasnofb I can semi understand that you wanted to put all the setprops in one place, but since the GUI dialogs need it too, the best place is still `electrical.nas`. I have left the other changes you made untouched. I hope you don't mind.

Additionally, I have changed the system to use FG's `updateloop.UpdateLoop` so that it can be properly enable/disabled and it takes care of signals etc. `tiedowns.nas` uses it too.

gilbertohasnofb:
> Have you tried moving
> ```
> <electrical>
>     <file>Nasal/electrical.nas</file>
> </electrical>
> ```
> above <c172p> in c172p-set.xml?

I have, that didn't work

gilbertohasnofb:
@onox yeah, I completely understand that code duplicate isn't elegant from a programming point of view. I just thought that those changes made the function clearer to understand for the reason you pointed out above, but I am totally fine with reverting that

So does this last commit of yours fix the original issue or does it just improve the code concerning the things you pointed out in these last messages? I.e. can I test and potentially merge it now or is there still anything else to be done?

onox:
It should have fixed the original issue.

gilbertohasnofb:
:+1: I will test now!

gilbertohasnofb:
@dany93 

> No more Nasal error messages,
> but the starter makes some turns and stops, with the message (for a short time):
> "The autostart failed to start the engine. You must lean the mixture and start manually"

Can you test this again to see if onox's solution fix this issue? You should see that message only if the altitude is too high that the autostart can't fire up the engine with full mixture. But I suspect that whatever made you see this error should be fixed now.

@onox tested and it seems to be working perfectly fine, so let's just wait for @dany93 before merging this

dany93:
Yes, that fixes the issue.
No Nasal error messages, the engine automatically starts at sim launch.
Thank you.

PR #841, opened 2016-09-22
Bug 834

### LINKED ISSUE #834: Add fire extinguisher
Currently our plane is flying illegally as we are missing the mandatory fire extinguisher! :scream: We could also maybe add a beacon locator to the plane if those are common to the 172. Below are the models from the c150:

fg-1

fg-0

I would be happy to tackle this myself, but I have two questions about it:

- the fire extinguisher is normally located in between the seats, as in the image below. But we use that space for the amphibian water rudder cord. So what's a better solution, to add the fire extinguisher elsewhere (maybe close to the co-pilot rudders as in the c150 image above) or should I add them in between the seats for all variants except the amphibian?

- what about the beacon locator, is anyone familiar with that sort of equipment?

gilbertohasnofb:
Closes #834

- adds ELT to the back of the plane
- adds fire extinguisher to the plane, located in between the seats for gear variants and below the pilot's seat for float and amphibian
- adds flashlight effect to luggages in the back of the plane


onox:
Otherwise OK, objects are visible and flashlight effect works :+1:

gilbertohasnofb:
> Shouldn't the ELT emit some message if connected to MP and the aircraft crashes?

Yes it could, but to be honest I am not really into MP at all and so I don't feel like working on it. If you would like to implement this, I can help with testing of course!

> Otherwise OK, objects are visible and flashlight effect works :+1:

Great!

PR #843, opened 2016-09-28
Bug 811: ALS external view taxi or landing lights

wlbragg:
Please review and critique. I attempted to adjust the lights somewhat, but it is pssible they could be positioned more accurately.

I extended the lights to 4, landing, taxi, 2 x nav.

@onox, I didn't refactor the code to use the UpdateLoop you used in the tiedown code (feel free to if you think it best).
Why is the UpdateLoop a better choice?

### DIFF
--- a/Nasal/light-manager.nas
+++ b/Nasal/light-manager.nas
@@ -88,7 +88,7 @@ var light_manager = {
 
         # color values
         me.light3_r = 0.0;
-        me.light3_g = 0.1;
+        me.light3_g = 0.05;
         me.light3_b = 0.0;
 
         # spot size
@@ -100,7 +100,7 @@ var light_manager = {
         me.light4_zpos = 2.0;
 
         # color values
-        me.light4_r = 0.1;
+        me.light4_r = 0.05;
         me.light4_g = 0.0;
         me.light4_b = 0.0;
 
@@ -287,11 +287,11 @@ setlistener("/sim/signals/fdm-initialized", func {
 
     light_manager.init();
 
-    setlistener("/sim/model/c172p/lighting/landing", func (node) {
+    setlistener("/sim/rendering/als-secondary-lights/use-landing-light-ext", func (node) {
         light_manager.enable_or_disable(node.getValue(), 1);
     }, 1, 0);
 
-    setlistener("/sim/model/c172p/lighting/taxi", func (node) {
+    setlistener("/sim/rendering/als-secondary-lights/use-taxi-light-ext", func (node) {
         light_manager.enable_or_disable(node.getValue(), 2);
     }, 1, 0);
 
--- a/Nasal/light-manager.nas
+++ b/Nasal/light-manager.nas
@@ -1,0 +1,296 @@
+# provides relative vectors from eye-point to aircraft lights
+# in east/north/up coordinates the renderer uses
+
+var light_manager = {
+    run: 0,
+
+    lat_to_m: 110952.0,
+    lon_to_m: 0.0,
+
+    light1_xpos: 0.0,
+    light1_ypos: 0.0,
+    light1_zpos: 0.0,
+    light1_r: 0.0,
+    light1_g: 0.0,
+    light1_b: 0.0,
+    light1_size: 0.0,
+    light1_stretch: 0.0,
+
+    light2_xpos: 0.0,
+    light2_ypos: 0.0,
+    light2_zpos: 0.0,
+    light2_r: 0.0,
+    light2_g: 0.0,
+    light2_b: 0.0,
+    light2_size: 0.0,
+    light1_stretch: 0.0,
+
+    light3_xpos: 0.0,
+    light3_ypos: 0.0,
+    light3_zpos: 0.0,
+    light3_r: 0.0,
+    light3_g: 0.0,
+    light3_b: 0.0,
+    light3_size: 0.0,
+
+    light4_xpos: 0.0,
+    light4_ypos: 0.0,
+    light4_zpos: 0.0,
+    light4_r: 0.0,
+    light4_g: 0.0,
+    light4_b: 0.0,
+    light4_size: 0.0,
+
+    init: func {
+        # define your lights here
+
+        # light 1 ########
+        # offsets to aircraft center
+        me.light1_xpos = 23.0;
+        me.light1_ypos = 2.0;
+        me.light1_zpos = 2.0;
+
+        # color values
+        me.light1_r = 0.5;
+        me.light1_g = 0.5;
+        me.light1_b = 0.5;
+
+        # spot size
+        me.light1_size = 12.0;
+        me.light1_stretch = 1.0;
+
+        # light 2 ########
+        # offsets to aircraft center
+        me.light1_xpos = 20.0;
+        me.light2_ypos = 1.0;
+        me.light2_zpos = 2.0;
+
+        # color values
+        me.light2_r = 0.4;
+        me.light2_g = 0.4;
+        me.light2_b = 0.4;
+
+        # spot size
+         me.light2_size = 10.0;
+         me.light2_stretch = 3.0;
+
+        # light 3 ########
+        # offsets to aircraft center
+        me.light3_ypos = -6.0;
+        me.light3_zpos = 2.0;
+
+        # color values
+        me.light3_r = 0.0;
+        me.light3_g = 0.2;
+        me.light3_b = 0.0;
+
+        # spot size
+         me.light3_size = 6.0;
+
+        # light 4 ########
+        # offsets to aircraft center
+        me.light4_ypos = 6.0;
+        me.light4_zpos = 2.0;
+
+        # color values
+        me.light4_r = 0.2;
+        me.light4_g = 0.0;
+        me.light4_b = 0.0;
+
+        # spot size
+        me.light4_size = 6.0;
+
+        me.start();
+    },
+
+    start: func {
+        setprop("/sim/rendering/als-secondary-lights/num-lightspots", 4);
+
+        setprop("/sim/rendering/als-secondary-lights/lightspot/size", me.light1_size);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/size[1]", me.light2_size);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/size[2]", me.light3_size);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/size[3]", me.light4_size);
+
+        setprop("/sim/rendering/als-secondary-lights/lightspot/stretch", me.light1_stretch);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/stretch", me.light2_stretch);
+
+        me.run = 1;       
+        me.update();
+    },
+
+    stop: func {
+        me.run = 0;
+    },
+
+    update: func {
+        if (me.run == 0) {
+            return;
+        }
+
+        var apos = geo.aircraft_position();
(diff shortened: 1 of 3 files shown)


gilbertohasnofb:
I will be unable to test anything during the next week as I am travelling with my chromebook only. If 'til then nobody reviewed it, I can do it myself :smile:

wlbragg:
Only because you can't check it out, I post this for you.
https://dl.dropboxusercontent.com/u/63093453/fg/c172p/ground-lights.mp4

onox:
> @onox, I didn't refactor the code to use the UpdateLoop you used in the tiedown code (feel free to if you think it best).

Because it takes care of the loop management and signals for you.

onox:
It seems the lights do not react to the electrical voltage.

onox:
What do you think of the brightness of the nav lights on the ground? I think they are a bit too bright. Taxi/landing lights' brightness are OK.

I also wonder if the original ALS landing light is still used? It kinda conflicts with the new lights (or you would have to disable the new taxi/landing lights while in the cockpit). Do the new lights also work outside the runway/taxiways?

wlbragg:
>It seems the lights do not react to the electrical voltage.

OK, forgot that part.

>What do you think of the brightness of the nav lights on the ground? I think they are a bit too bright. 

Agreed

>I also wonder if the original ALS landing light is still used? It kinda conflicts with the new lights (or you would have to disable the new taxi/landing lights while in the cockpit).

I didn't notice, I assumed the original ALS landing light was interior view only and the new lights were exterior only. I'll check it out in more detail.

Do the new lights also work outside the runway/taxiways?

No

onox:
> with my chromebook only

https://www.youtube.com/watch?v=d_MuVwJq_XQ :smile:

onox:
and part 2: https://www.youtube.com/watch?v=OGvC0TWPk-g

wlbragg:
>I also wonder if the original ALS landing light is still used? It kinda conflicts with the new lights (or you would have to disable the new taxi/landing lights while in the cockpit).

Yeah, so what, should we remove original landing light or disable new landing light if in interior view?

wlbragg:
Hum, I am not sure what to do. I tried disabling the old landing light and taxi light, there is quite a bit of difference in perspective as to where the two different types are cast. But I don't know it it is a matter of adjusting the new ones to make it reach farther out from the interior view or if it simply isn't designed to really be as effective as the old interior view landing lights.

My first instinct is to use both and condition them on or off per view.
Need more opinions.

wlbragg:
After further review, I like the new lights better from the interior view. They cover the correct area of the runway, (right up to the leading edge of the wing) VS the old landing are only spots that are out at a distance with no light illuminating from the wing edge on.

Original ALS Interior - taxi
taxi-int-on
New ALS Exterior - taxi
taxi-new-on-i
Original ALS Interior - landing
landing-int-on
New ALS Exterior - landing
landing-new-on-i
Original ALS Interior - both landing and taxi
both-int-on
New ALS Exterior - both landing and taxi
both-new-on-i

wlbragg:
I notice the old landing lights do illuminate other objects (trees) and land covers. That is a strong case for using either both  at the same time or a view condition.

wlbragg:
After talking with Thorsten about it I guess we need to use it as intended. The new lights don't deal with the fog or anything like the original interior lights do. So I guess we need to condition the taxi and landing light off when in an interior view (I think we can still use the nav lights from the interior view).

Maybe someone can at least test the taxi light in the fog combined with the original taxi light and original landing light just to see what it looks like.

I'll do the same and give an opinion.

gilbertohasnofb:
@wlbragg thanks for the video! I agree with onox, the nav lights reflections are quite bright, I think they would look better if there is just a hint of colour on the runway. Everything else looks good to me!

@onox I am already running Linux on my Chromebook (I actually removed Chrome OS and have Mint running full time), but the computer itself is so weak and has to little storage space that I can't really have FG installed on it :smile:

onox:
> @onox (I actually removed Chrome OS and have Mint running full time),

Via this crouton script or in some other way? I thought chromebooks were kinda locked down.

Nasal/light-manager.nas
```diff
@@ -0,0 +1,303 @@
+# provides relative vectors from eye-point to aircraft lights
+# in east/north/up coordinates the renderer uses
+
+var light_manager = {
+    run: 0,
+
+    lat_to_m: 110952.0,
+    lon_to_m: 0.0,
+
+    light1_xpos: 0.0,
```
onox: There are lots of fields here getting a 0.0 value, but in the `init()` function they're initialized _again_. Can you try and see if just initializing the fields in the `init()` function is sufficient?

Nasal/light-manager.nas
```diff
@@ -0,0 +1,303 @@
...
+        delta_z = apos.alt() - vpos.alt();
+
+        setprop("/sim/rendering/als-secondary-lights/lightspot/eyerel-x-m[3]", delta_x);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/eyerel-y-m[3]", delta_y);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/eyerel-z-m[3]", delta_z);
+
+        settimer(func me.update(), 0.0);
+    },   
+
+    light1_on: func {
```
onox: When I look at all these lightx_on/off functions, they look very similar. You can simplify the code by creating a function that returns a function for you. Then you have only 1 place where you actually set the lightspot values and where you define the ALS properties:

```javascript
var lightspot_on_func = func (index) {
    return func {
        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-r", index, me.light_r[index]);
        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-g", index, me.light_g[index]);
        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-b", index, me.light_b[index]);
    }
}

var light1_on = lightspot_on_func(0);
```

Just make sure that `me.light_r`, etc. are arrays.

You can do the same thing for `lightspot_off_func` and for the other lights.

Nasal/light-manager.nas
```diff
@@ -0,0 +1,303 @@
...
+    },
+
+    light4_off: func {
+        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-r[3]", 0.0);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-g[3]", 0.0);
+        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-b[3]", 0.0);
+    },
+    
+    enable_or_disable: func (enable, light_num) {
+        if (enable) {
```
onox: Instead of generating functions like this:

```javascript
var light1_on = lightspot_on_func(0);
```

You could write `lightspot_on_func(light_num)()` instead. Imagine the number of if statements if you have 20 or 100 instead of 4 lights.

gilbertohasnofb:
@Via this crouton script or in some other way? I thought chromebooks were kinda locked down.

No, I did not use crouton, I did it the "hard" way. They are indeed kind of locked down, so I had to open my Acer C720 and remove a screw from the board which prevents any kind of write access to the BIOS. Once that was done, I then managed to completely remove Chrome OS and install a proper distro. It was relatively easy and I even found a iso of Linux Mint for this type of installation on Chromebooks, so I had no issues with drivers or whatsoever. :smiley:

wlbragg:
@onox, the UpdateLoop we're talking about using is the one in fgdata/aircraft/generic...?

I guess I didn't know that had been implemented in the c172p. Was that recently added to manage some of the newer systems like icing, oil management?

onox:
@wlbragg You don't have to include/import anything, that has already been done in the -set.xml file. Just look at `Nasal/tiedowns.nas` or `Nasal/electrical.nas` for an example how to use it.

Nasal/light-manager.nas
wlbragg: Yes, you can but I am at a loss as to why? I tested it and you can "only" create the me.var in the init section, and it works (Lua style). But why do we want to do this? It would then make it difficult at best to quickly see what a light_manager object consists of and what are we really gaining?

onox: In case we're talking about different things. What I see is this (in pseudocode):

```javascript

class X {
    foo: 0;

    function init () {
        me.foo = "bar";
    }
}
```

Why not write:

```javascript

class X {
    function init () {
        me.foo = "bar";
    }
}
```

?

> It would then make it difficult at best to quickly see what a light_manager object consists of and what are we really gaining?

If you initialize a field in just one place. Doesn't that make it easier?

wlbragg: Not necessarily, the layout is a standard c++ class layout with all the member variables and functions defined. When you init() you are basically making a "new" class x with all the vars and functions set and ready to go. Then you change any vars that need a different starting value than the default and off you go.
In c++ you have to do it like this or you will get an error. In Lua you can make the member var for the first time in a member function (init).
In this case if we don't create all the member var's where they are being initiated, we have to make sure any that are used are initialized in the init section, even those that are set to 0.
I went through and set it like that and there were a few that are not changed in the init section that are required and would have to be added to the init section and set to 0. The way it is set up now is the correct way to do it. But you are amplifying one of the reasons some experienced programmers don't care for c++ class structure. I have evolved over time to not like programming with classes. It carries more overhead than is ever needed.

Nasal/light-manager.nas
wlbragg: I considered doing something like what you suggested above but got lazy and just plugged in the function as is. I should have known yu would call me on it! :smile: 

I changed it to the following. I didn't do any arrays as per your example, again. Instead I did the following.
~~~xml
    switch_position: func(light, lightr, lightg, lightb) {
        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-r["~light~"]", lightr);
        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-g["~light~"]", lightg);
        setprop("/sim/rendering/als-secondary-lights/lightspot/lightspot-b["~light~"]", lightb);
    },

    enable_or_disable: func (enable, light_num) {
        if (enable) {
            if (light_num == 0)
                me.switch_position(light_num, me.light1_r, me.light1_g, me.light1_b);
            if (light_num == 1)
                me.switch_position(light_num, me.light2_r, me.light2_g, me.light2_b);
            if (light_num == 2)
                me.switch_position(light_num, me.light3_r, me.light3_g, me.light3_b);
            if (light_num == 3)
                me.switch_position(light_num, me.light4_r, me.light4_g, me.light4_b);
        } else {
            me.switch_position(light_num, 0.0, 0.0, 0.0);
        }
    },
~~~

wlbragg:
>You don't have to include/import anything, that has already been done in the -set.xml file. Just look at Nasal/tiedowns.nas or Nasal/electrical.nas for an example how to use it.

I changed settimer to maketimer. I really don't like using global helper function. To many chefs in the kitchen and later on down the road, someone can come along and change it and break your code.

I'm waiting on an answer from Thorsten on a different issue I have regarding the shader code. As soon as I get it resolved I'll push all the changes.

wlbragg:
It's working the way I want. The issue I have is not really affecting the visual, so I pushed it anyway.

wlbragg:
There is a patch to fix an issue with the shader I need to push to fgdata today. One that is merged we should see the following for our lights (unless something else changes between now and then).
This is the results of my latest revision.

taxi

landing

both

gilbertohasnofb:
@wlbragg Looks really good!! :+1: And I love how the nav lights are now very soft on the pavement, looks really realistic I think

PR #848, opened 2016-10-25
Fixing weights

### LINKED ISSUE #847: Error in oil consumption effect on aircraft weight
When I looked for giving a response to @wlbragg's question https://github.com/Juanvvc/c172p-detailed/issues/845 about the extra weight for 180 hp version, I was temporarily confused.
By checking the empty weight (no fuel, no pilot) in "Equipment > Fuel and Payload", I discovered that the values were different from the ones entered in the empty 160 hp FDM (1512 lbs instead of 1500 lbs in the FDM) or the FDM + extra weight due to 180 hp engine (1654 lbs instead of 1642 lbs in the FDM). The issue is that the engine oil content in FG has been added to the aircraft empty weight, in contradiction with the POHs where the "Basic Empty Weight" includes the full oil (section 6, Weight and Balance).
The issue is not really the absolute weight of the aircraft because these differences are weak and nobody will notice their effect. Moreover, the basic aircraft weight in the POHs are indicative, a pilot must know the true weight for his particular aircraft. **The issue is that, in FG, it is confusing when one searches to understand the files or to fix a bug**.
To be rigorous, the oil which has been consumed should be subtracted from the aircraft weight (instead of adding the remaining oil content).
But the difference is so weak that, in my opinion, the effect of oil consumption on the aircraft weight might be neglected. It would not be a noticeable fault and it would avoid adding useless complexity in the files.

gilbertohasnofb:
Closes #847 

- fix empty aircraft weights, as POH values include full oil tank (see issue above for full discussion)
- removes old comment about starting the 180hp engine; this comment does not seem to be actual any longer

### DIFF
--- a/Systems/bushkit.xml
+++ b/Systems/bushkit.xml
@@ -60,7 +60,15 @@ Extra weight and drag due to bush wheels, floats and 180 hp engine
 
         <switch name="extra-weight-180hp">
             <default value="0"/>
-            <test logic="AND" value="142">
+            <!--
+                The weights given by the POH include full oil level. The difference without oil can be calculated as follows:
+                160hp with oil (POH value): 1500 lbs
+                160hp without oil (7 quarts = 12 lbs): 1488 lbs
+                180hp with oil (POH value): 1642 lbs
+                180hp without oil (8 quarts = 14 lbs): 1628 lbs
+                difference without oil: 140 lbs
+            -->
+            <test logic="AND" value="140">
                 /controls/engines/active-engine EQ 1
             </test>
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
@dany93 maybe you'd like to test this

onox:
Weight is as suggested in https://github.com/Juanvvc/c172p-detailed/issues/847#issuecomment-256093774. Good idea to add an explanation about how the weight numbers are calculated :+1:

gilbertohasnofb:
Thanks, @onox! :smiley:

dany93:
Right.

PR #852, opened 2016-11-02
Fixing small bugs (carb icing and oil)

gilbertohasnofb:
Fixing small bugs:

- enabling/disabling carb icing or oil management used to reset oil level to 7 quarts
- the oil function has been improved, so how disabling oil management will automatically set the oil to the max level for that specific engine (7 quarts for 160 hp and 8 quarts for 180 hp)
- starting the sim with oil management disabled will also set the engine to the correct amount of oil now

@dany93 thanks for spotting these issues (see https://github.com/Juanvvc/c172p-detailed/issues/851), let me know if these are now fixed

### DIFF
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -89,10 +89,16 @@ var primerTimer = maketimer(5, func {
 # ========== oil consumption ======================
 
 var oil_consumption = maketimer(1.0, func {
-    if (getprop("/engines/active-engine/oil_consumption_allowed"))
-        var oil_level = getprop("/engines/active-engine/oil-level");
-    else
-        var oil_level = 7.0;
+    if (!getprop("/engines/active-engine/oil_consumption_allowed")) {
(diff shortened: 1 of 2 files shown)


dany93:
Prior to push "better oil function", have a look at the code I propose in[ consumed-oil-weight](https://github.com/dany93/c172p-detailed/tree/consumed-oil-weight).
I have completely deleted this conditional part for 7 or 8 in `engines.nas`. Managed in `aircraft-dailog.xml`, lines 122 - 157, enabling or disabling oil management automatically sets the oil to the max level ( 7 for 160 hp or 8 for 180hp). . Make sure that this part in nasal is needed, if it does not do the same thing. Needed for starting the sim ?

It's too late for me, I will look at your merge request this as soon as I can (to morrow I think).

dany93:
You've gone too fast. Please read https://github.com/Juanvvc/c172p-detailed/issues/851#issuecomment-258108677

gilbertohasnofb:
:+1:

PR #853, opened 2016-11-03
Consumed oil weight and a few bug fixes on full oil vs the engine.

dany93:
See https://github.com/Juanvvc/c172p-detailed/issues/851#issue-186835178

- Changes the way aircraft weight and CG are calculated, by _subtracting_ the _consumed_ oil from the POH Basic Empty Weight. Much more simple to understand and easier to calculate and implement.
- Fixes some bugs about the oil content (7 or 8 quarts) vs the 160 hp / 180 hp engine.

### DIFF
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -125,9 +125,31 @@
                 <property>/engines/active-engine/oil_consumption_allowed</property>
                 <live>true</live>
                 <binding>
+                    <condition>
+                        <equals>
+                            <property>/controls/engines/active-engine</property>
+                            <value type="int">0</value>
+                        </equals>
+                    </condition>
                     <command>property-assign</command>
                     <property>/engines/active-engine/oil-level</property>
                     <value>7.0</value>
+                </binding>
+                <binding>
+                    <condition>
+                        <equals>
+                            <property>/controls/engines/active-engine</property>
+                            <value type="int">1</value>
+                        </equals>
+                    </condition>
+                    <command>property-assign</command>
+                    <property>/engines/active-engine/oil-level</property>
+                    <value>8.0</value>
+                </binding>
+                <binding>
+                    <command>property-assign</command>
+                    <property>/engines/active-engine/oil-consumed</property>
+                    <value>0.0</value>
                 </binding>
                 <binding>
                     <command>dialog-apply</command>
@@ -139,11 +161,6 @@
                 <label>Allow carburetor icing</label>
                 <property>/engines/active-engine/carb_icing_allowed</property>
                 <live>true</live>
-                <binding>
-                    <command>property-assign</command>
-                    <property>/engines/active-engine/oil-level</property>
-                    <value>7.0</value>
-                </binding>
                 <binding>
                     <command>dialog-apply</command>
                 </binding>
@@ -489,6 +506,11 @@
                         <property>/engines/active-engine/oil-level</property>
                         <value>7.0</value>
                     </binding>
+                    <binding>
+                        <command>property-assign</command>
+                        <property>/engines/active-engine/oil-consumed</property>
+                        <value>0.0</value>
+                    </binding>
                 </radio>
 
                 <radio>
@@ -525,6 +547,11 @@
                         <command>property-assign</command>
                         <property>/engines/active-engine/oil-level</property>
                         <value>8.0</value>
+                    </binding>
+                    <binding>
+                        <command>property-assign</command>
+                        <property>/engines/active-engine/oil-consumed</property>
+                        <value>0.0</value>
                     </binding>
                 </radio>
             </group>
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -125,28 +125,6 @@
                 <property>/engines/active-engine/oil_consumption_allowed</property>
                 <live>true</live>
                 <binding>
-                    <condition>
-                        <equals>
-                            <property>/controls/engines/active-engine</property>
-                            <value type="int">0</value>
-                        </equals>
-                    </condition>
-                    <command>property-assign</command>
-                    <property>/engines/active-engine/oil-level</property>
-                    <value>7.0</value>
-                </binding>
-                <binding>
-                    <condition>
-                        <equals>
-                            <property>/controls/engines/active-engine</property>
-                            <value type="int">1</value>
-                        </equals>
-                    </condition>
-                    <command>property-assign</command>
-                    <property>/engines/active-engine/oil-level</property>
-                    <value>8.0</value>
-                </binding>
-                <binding>
                     <command>property-assign</command>
                     <property>/engines/active-engine/oil-consumed</property>
                     <value>0.0</value>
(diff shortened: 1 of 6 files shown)


gui/dialogs/aircraft-dialog.xml
```diff
@@ -125,11 +125,33 @@
...
                 <binding>
+                    <condition>
+                        <equals>
+                            <property>/controls/engines/active-engine</property>
+                            <value type="int">0</value>
+                        </equals>
+                    </condition>
                     <command>property-assign</command>
                     <property>/engines/active-engine/oil-level</property>
                     <value>7.0</value>
```
gilbertohasnofb: @dany93 since you already removed the binding that was setting the oil to 7 quarts when toggling the carb icing feature, you could as well remove this binding from the oil consumption toggle and so we solve everything with a single PR, what do you think? I am referring to the 3 lines above my comment

gilbertohasnofb: Actually Dany, this whole seems unecessary to me, since as far as I understand your Nasal function is taking care of the oil level by itself (so that if the oil consumption is disables, the function will reset the oil to 7 or 8 quarts according to the engine). So maybe we should remove lines 127-148, is that correct?

gilbertohasnofb:
Dany, it is all working very well with this PR and it seems to have taken care of all bugs and issues :clap: Just please take a look at my comments in the code for the `aircraft-dialog.xml`, if I am correct I think we can cut some of that code and then we would be ready to merge this.

And thanks for fixing my sloppy code by the way! :+1:

gilbertohasnofb:
One more thing, Dany: I just realized that the gross weight of the plane does not change any longer when I change the oil level, was this intentional? Try opening the fuel and payload dialog and try changing the oil level, it seems nothing changes for either engine

dany93:
I think that you are right, I even didn't notice that those lines did it twice. I have written this in so many steps...
In aircraft-dialog.xml, lines previously 127-148 (which sets oil level to 7 or 8) removed. The job is done in engine.nas and it seems enough.

gilbertohasnofb:
Great, thanks for removing those lines. But what about the weight I mentioned before? Shouldn't changing the oil level change the weight in the fuel and payload dialog?

dany93:
>I just realized that the gross weight of the plane does not change any longer when I change the oil level, was this intentional? 

Changing the oli level manually by "set?"
That is not really intentional, merely this operation is not expected. No interest. And the oil level is reset to its full value as soon as you start the engine.
The aircraft weight is calculated from the **consumed oil**. Wasn't it the target? If you want to siphon or add oil and observe the aircraft weight change on the ground, that is not simulated. Simulation has limits.

dany93:
If you wish to simulate oil siphoning, you can subtract the lacking oil from pointmass[16], as I did for the consumed oil. But for me, that's enough.

dany93:
Keep in mind that the aircraft weight reference is full oil in the engine. If you wish include siphoned oil in weight calculation, that's kind of` oil-lacking = consumed-oil + siphoned-oil` in pointmass[16].  But my opinion is that you add complexity for a negligible advantage in realism.
More  important is the oil consumption which makes the engine fail, and that is done.

dany93:
Maybe almost done. Be patient...

dany93:
Done. Can you check and test please?

gilbertohasnofb:
Hi Dany,

> Done. Can you check and test please?

Sure, my pleasure. I tested it and it works very well, thanks for implementing this.

> to include the oil level set by the slider, the best might be to take,
> instead of oil_consumed,
> oil_lacking = oil_full - oil_level
> subtracted by pointmass[16] for weight calculation.
> With oil_full = 7 or 8.
> (I only progressively discover all your subtleties, these changes can be done starting from my current code. Nr 3 seems easy enough to implement)

Yes, I had implemented a dialog for adding/removing oil a long time ago. I tested your branch right now and the dialog works very well with your algorithm, so the feature works as it should. Let's merge this PR then.

And just in case you are not completely aware of what I did in my first work with the oil system, basically there is a hot spot in the aircraft's nose on the oil cap which is only available if the oil management option is ON. Clicking on the hotspot with the engine on allows you to change the oil level.

> For me, that's enough.
> Please, do not break the code.

I have to say I have issues with these comments, Dany. First, your initial implementation had broken a feature which I had created in the past (the oil dialog), and saying that you had enough while it was not working was not very nice. Of course I do understand that you might have not being aware of this feature and also that you did more work after and now it is working, so that's fine with me. But concerning breaking code, I frankly think I broke things very few times during these years, and of course it has never on purpose so asking "please don't do it" isn't really helpful at all. Of course I do introduce some bug once in a while, but I like to think that I fix more bugs than I introduce, including bugs created by other people. Also, this is git and if code gets broken then it's always easy to either revert it or fix it. Finally, we do extensively testing before merging PRs, so the chance of me or anyone else breaking your work and getting it merged are slim.

dany93:
>  ....and saying that you had enough while it was not working was not very nice

Hi Gilberto,

Thank you for the tests and merging.

Sorry, I was tired and my meaning 

> For me, that's enough

 was not that _I had enough_ but that the functioning, as it was, was enough at my opinion (taking into account the very weak contribution of the oil consumption on the aircraft weight). My approximate English... And I thought that the only way to manually change the oil level was by "set" in the Properties... which is weird for a standard user.

I never kept an eye to control if you broke or repaired code in the past. Your contribution is important enough, I don't think like this. Only this time (no luck), I was worried after having to fix several issues in that part of code. Starting from the oil weight added to the "empty" aircraft, then the code to include it, which worked but was very intricate to use due to its form, plus one or two others of minor importance.
But I did not break the _oil dialog_, I broke (by ignorance, pardon) only its _effect on the aircraft weight_. I guess that, initially, your oil dialog was for the engine to fail by low oil level, its weight effect being a secondary, minor consequence.

But please, don't have issues for such a small incident, sorry and anyway thank you for all that you do and have done. That's the most important.

I hope that my frenglish did not fail once more for this explanation. The more I read it, the more I find mistakes. I stop.

gilbertohasnofb:
Hi Dany, thanks for your message, I really appreciate it. I think it's all clear now, thanks for explaining everything. This mixture of us being non-native speakers (myself obviously included!) together with impersonal communication by text can lead to these misunderstandings, and I am sorry I read you wrongly.

> But please, don't have issues for such a small incident

No worries about it, let's leave this subject behind right now :+1: 

Thanks again for implementing this, I think the final result is really nice and way more elegant than I had managed last year :smile: 

Cheers!

dany93:
Thank you too, Gilberto.

wlbragg:
While you're thanking each other and clearing the air, I want to thank both of you for the work. I also wanted to make clear that I injected my comment about the oil weight being negligible for all practical reasons only to help give incentive as to not having to go through leaps and bounds to implement something so minor (the weight portion), not the oil management in general. I think the oil management is a really cool feature to have for those that like the reality. I meant no offense and hope I didn't come off in a negative manner. If I did I apologize.

gilbertohasnofb:
@wlbragg 👍 Thanks for the comment and you have nothing to apologize for!

gilbertohasnofb:
This thread has become the therapy couch of the group 😆

PR #854, opened 2016-11-06
if Oil management disabled, set oil-lacking = 0

dany93:
Correction: set "oil-lacking" = 0 when "Allow oil management" is **disabled** (unchecked).
(only "oil-consumed" was reset to 0 and "full level" reset to 7 or 8)

### DIFF
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -89,13 +89,17 @@ var primerTimer = maketimer(5, func {
 # ========== oil consumption ======================
 
 var oil_consumption = maketimer(1.0, func {
+
+    var oil_level = getprop("/engines/active-engine/oil-level");
+    var oil_consumed = getprop("/engines/active-engine/oil-consumed");
+    if (getprop("/controls/engines/active-engine") == 0)
+        var oil_full = 7;
+    if (getprop("/controls/engines/active-engine") == 1)
+        var oil_full = 8;
+    var oil_lacking = oil_full - oil_level;
+    setprop("/engines/active-engine/oil-lacking", oil_lacking);
+    
     if (getprop("/engines/active-engine/oil_consumption_allowed")) {
-        var oil_level = getprop("/engines/active-engine/oil-level");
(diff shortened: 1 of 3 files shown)


dany93:
Refactoring `engine.nas` for more rational and concise.
@gilbertohasnofb if you can check and test again, thanks. There are so many combinations that I'm always afraid of missing one.
Functionning should be unchanged.

gilbertohasnofb:
@dany the code looks good, and I will test this on the sim as soon as I can (this afternoon I will be able to do that). So just give me a couple of hours and I will merge this.

> here are so many combinations that I'm always afraid of missing one.

I know the feeling! 😄

dany93:
Thanks for your response. No urgency.

gilbertohasnofb:
@dany93 I tested and it is all working well, but I don't understand what is the use of the variable `oil_consumed` in that function (and nor do I understand what is the property `/engines/active-engine/oil-consumed` used for). I searched for all other references of that property, and it seems to me it is only present in the set file (to initialize it as `0.0`) and also in the aircraft dialog file, so that when changing engines that property is reset. But other than that, it doesn't seem to affect anything in the function. Could you please check if we can cut it out as well as these other references to it?

dany93:
If I remember well, I had created the variable `oil_consumed` and` /engines/active-engine/oil-consumed` to take it into account for the weight loss, when I was not aware of the oil dialog feature. `oil-consumed` was the only contribution to the weight loss.
I have deleted all the lines referring to it.
Slightly refactored engine.nas (displaced a few lines for more clarity).
It seems good (I hope).

gilbertohasnofb:
@dany93 Thanks for the refactoring, it is looking good now. I will just test it in the sim right now to make sure no side effects are observed and then we merge this!

gilbertohasnofb:
It looks all good to me, thanks for the work Dany:+1:

dany93:
Thank you, Gilberto

PR #857, opened 2016-11-09
160 hp aircraft: empty weight correction, 1467 lbs in accordance with POH

dany93:
The initial value (already in the jan. 2012 FG version) was 1500 lbs in the FDM.
According to the **Cessna 172P Skyhawk 1982 (160hp) POH p.6-10 (112/357 pdf)** the basic empty weight (= with full oil) is **1467 lbs**.
The empty center of gravity has been recalculated, to be in accordance with this same POH (57300 lbs.ins, which gives **39.1 "**).
Probably not very noticeable on the behaviour but at least for the consistency...

The 180 hp empty weight and CG are unchanged.
The extra weight added at 160 hp to obtain the 180 hp version has been also recalculated for that.

### DIFF
--- a/Systems/bushkit.xml
+++ b/Systems/bushkit.xml
@@ -2,8 +2,8 @@
 
 <!--
 ***********************************************************************************
-c172p-detailed, May 2015 
-Extra weight and drag due to bush wheels, floats and 180 hp engine
+c172p-detailed, May 2015... Nov. 2016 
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
checking now

gilbertohasnofb:
Great :+1:

dany93:
Thank you Gilberto.

PR #858, opened 2016-11-10
Fixing issue #849 and correcting the securing properties in securing.xml

### LINKED ISSUE #849: First tutorial "Preflight" bug
See https://forum.flightgear.org/viewtopic.php?f=25&t=30813&p=297493#p297493

Basically I think the trouble is when the tutorial exits the aircraft it does so in the cockpit view and it appears you cannot interact with the external items when in cockpit view.
Verified.

dg-505:
Fix #849 

The properties for pitot cover, tiedowns and chocks in the securing tutorial were outdated. I corrected them :smirk:

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -7207,10 +7207,15 @@
             <repeatable>false</repeatable>
             <binding>
                 <condition>
-                    <property>/sim/model/c172p/securing/pitot-cover-removable</property>
-                    <not>
-                        <property>/sim/current-view/internal</property>
-                    </not>
+                    <or>
+                        <and>
+                            <property>/sim/model/c172p/securing/pitot-cover-removable</property>
+                            <not>
+                                <property>/sim/current-view/internal</property>
+                            </not>
+                        </and>
+                        <property>/nasal/tutorial/loaded</property>
+                    </or>
                 </condition>
                 <command>property-assign</command>
                 <property alias="/params/securing/pitot-cover"/>
@@ -7220,10 +7225,15 @@
         <hovered>
             <binding>
                 <condition>
-                    <property>/sim/model/c172p/securing/pitot-cover-removable</property>
-                    <not>
-                        <property>/sim/current-view/internal</property>
-                    </not>
+                    <or>
+                        <and>
+                            <property>/sim/model/c172p/securing/pitot-cover-removable</property>
+                            <not>
+                                <property>/sim/current-view/internal</property>
+                            </not>
+                        </and>
+                        <property>/nasal/tutorial/loaded</property>
+                    </or>
                 </condition>
                 <command>set-tooltip</command>
                 <tooltip-id>remove-pitot-tube-cap</tooltip-id>
@@ -7239,10 +7249,15 @@
             <repeatable>false</repeatable>
             <binding>
                 <condition>
-                    <property>/sim/model/c172p/securing/pitot-cover-addable</property>
-                    <not>
-                        <property>/sim/current-view/internal</property>
-                    </not>
+                    <or>
+                        <and>
+                            <property>/sim/model/c172p/securing/pitot-cover-addable</property>
+                            <not>
+                                <property>/sim/current-view/internal</property>
+                            </not>
+                        </and>
+                        <property>/nasal/tutorial/loaded</property>
+                    </or>
                 </condition>
                 <command>property-assign</command>
                 <property alias="/params/securing/pitot-cover"/>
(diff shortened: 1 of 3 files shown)


Models/c172p.xml
```diff
@@ -7478,10 +7521,18 @@
         <type>select</type>
         <object-name>TiedownHotSpotLeft</object-name>
         <condition>
-            <property>/sim/model/c172p/securing/tiedownL-addable</property>
-            <not>
-                <property>/sim/current-view/internal</property>
-            </not>
+            <or>
```
onox: if you have (A and B) or (A and C), then you can simplify this to this to A and (B or C). In this case A is `<property>/sim/model/c172p/securing/tiedownL-addable</property>`.

Models/c172p.xml
```diff
@@ -7571,10 +7638,18 @@
         <type>select</type>
         <object-name>TiedownHotSpotRight</object-name>
         <condition>
-            <property>/sim/model/c172p/securing/tiedownR-addable</property>
-            <not>
-                <property>/sim/current-view/internal</property>
-            </not>
+            <or>
```
onox: Same thing here :smiley:

Models/c172p.xml
```diff
@@ -7664,10 +7755,18 @@
         <type>select</type>
         <object-name>TiedownHotSpotTail</object-name>
         <condition>
-            <property>/sim/model/c172p/securing/tiedownT-addable</property>
-            <not>
-                <property>/sim/current-view/internal</property>
-            </not>
+            <or>
```
onox: And here.

gilbertohasnofb:
@onox is this ready to be merged? I see some "requested changes" above and I never seen this before on GitHub, how do we proceed from here?

dg-505:
As I didn't know how to deal with these "requested changes" I just pushed the changes in a separate commit.

onox:
> After finishing the tutorial, the "Cockpit View" is outside the airplane (looking at it from the southwest).

This last change has not been implemented yet.

dg-505:
Well, unfortunately I can't confirm this... For me it works correctly, after both the preflight and the securing tutorial the cockpit view is where it is supposed to be...

onox:
Let me try again.

onox:
I did:

1. Press v to go to external view
2. Autostart
3. Press shift+v to go to cock pit view
4. Start securing tutorial

fgfs-screen-011
fgfs-screen-012

onox:
It seems I can reproduce it when I leave the view in external view before starting the tutorial.

onox:
The code of the last step looks fine, so I'll merge this PR. Might be a bug in FG.

dg-505:
The last step of the tutorial moves the point of view a bit away, but when the tutorial finishes, it should get back into it's default position. Have no idea why it doesn't work properly for you...

onox:
Sometimes it does, but sometimes it doesn't.

gilbertohasnofb:
I will test this after and let you guys know whether I will also experiece this issue onox is reporting.

gilbertohasnofb:
Hi guys, sorry for my late reply but just now I had finally time to test this. A couple of comments:

- @onox I can't reproduce what you wrote here concerning the securing aircraft tutorial:

> 1. Press v to go to external view
> 2. Autostart
> 3. Press shift+v to go to cock pit view
> 4. Start securing tutorial

when the tutorial is finished, I am back in the cockpit view. Same thing for the preflight tutorial, I am also back in the cockpit view

- @dg-505 is it only me or the securing tutorial starts at the current airport instead of going to Barcelona? The preflight surely moves me there, but the securing tutorial didn't. Is this on purpose?

- also, with the preflight tutorial, I found a small bug. When I was asked to take the fuel sample of the right wing, I very quickly took the sample, realized it was blue and returned it, but I think I was too fast and the tutorial was not moving forwards! And because the tutorial is done in the cockpit view I wasn't able either to click under the wing to start use the sumps hotsposts myself, so I had to go to the external view and open them manually again. So I suggest a couple of things: I would say we should delay this dialog a bit if the tutorial is not listening to what the user is doing, but even better would be to make the user click under the wing to open the dialog! This would teach the user where those hotspots are, and would be more consistent. What do you guys think?

dg-505:
> Is this on purpose?
Yes, indeed this is intended. My line of thought was: When a newbie pilot has landed somewhere and wants to secure the aircraft using the tutorial, I thougt it would be better when the tutorial doesn't suddenly teleport you to another airport but keeps you at your current location

gilbertohasnofb:
> Yes, indeed this is intended. My line of thought was: When a newbie pilot has landed somewhere and wants to secure the aircraft using the tutorial

That's a valid thought and I am fine with that, though I am not a big fan of the inconsistency with the other tutorials.

> About your second point, I agree and I will change the behavior of this segmentö

Great, thanks! Maybe we should wait until the new work for preflight control surfaces check to be merged, so that you can then work on all issues with the tutorials in a new single PR, what do you think?

PR #862, opened 2016-11-14
implementing control surfaces check

### LINKED ISSUE #861: Preflight control surfaces check
I would like to implement control surfaces check for the preflight. My idea is quite simple: during the preflight, if one click on the ailerons, elevators and rudder they would move up and down a single time, as if the pilot is testing them. This is just eye candy of course, but it's quite a neat feature I have seen in other simulators. The question is how can this be done. I ran into two problems:

- the control surface properties cannot be overridden (e.g. `/surface-positions/elevator-pos-norm`). Is there another property I could use for this?
- how to smoothly change a property in Nasal? How can I make a property go from `0` to `1` in 2 seconds and back to `0` in another 2 seconds, then stop?

Does anyone have ideas or suggestions on how to tackle this?

gilbertohasnofb:
Closes #861

This PR implements control surfaces check. In the external view or using the walker, simply click on the ailerons, rudder and elevator to move them as if the pilot was testing them. Animation only works if the engine is not running and if the groundspeed is < 1.0 knots (same condition as tie-downs, chocks, etc.). The animations last for only 1-2 seconds and will override any joystick input during that time. The aileron animation temporarily disable auto-coordination if that's on, so that the ailerons move alone without moving the rudder.

Here is a video of this work (note that the video was done before the auto-coordination toggling I mentioned above):

https://youtu.be/ieVGiAv0ntI

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -195,17 +195,11 @@ var return_fuel_sample = func(index) {
 ##########################################
 var control_surface_check_left_aileron = func {
     var auto_coordination = getprop("/controls/flight/auto-coordination");
-    setprop("/controls/flight/auto-coordination", "false");
-    interpolate("/controls/flight/aileron", 1.0, 0.3);
-    settimer(func(){
-        interpolate("/controls/flight/aileron", -1.0, 0.6);
-    }, 0.3);
-    settimer(func(){
-        interpolate("/controls/flight/aileron", 0.0, 0.3);
-    }, 0.9);
+    setprop("/controls/flight/auto-coordination", 0);
+    interpolate("/controls/flight/aileron", 1.0, 0.5, -1.0, 1.0, 0.0, 0.5);
     settimer(func(){
         setprop("/controls/flight/auto-coordination", auto_coordination);
-    }, 1.2);
+    }, 2.0);
 };
 
 ##########################################
@@ -213,43 +207,25 @@ var control_surface_check_left_aileron = func {
 ##########################################
 var control_surface_check_right_aileron = func {
     var auto_coordination = getprop("/controls/flight/auto-coordination");
-    setprop("/controls/flight/auto-coordination", "false");
-    interpolate("/controls/flight/aileron", -1.0, 0.3);
-    settimer(func(){
-        interpolate("/controls/flight/aileron", 1.0, 0.6);
-    }, 0.3);
-    settimer(func(){
-        interpolate("/controls/flight/aileron", 0.0, 0.3);
-    }, 0.9);
+    setprop("/controls/flight/auto-coordination", 0);
+    interpolate("/controls/flight/aileron", -1.0, 0.5, 1.0, 1.0, 0.0, 0.5);
     settimer(func(){
         setprop("/controls/flight/auto-coordination", auto_coordination);
-    }, 1.2);
+    }, 2.0);
 };
 
 ##########################################
 # Preflight control surface check: elevator
 ##########################################
 var control_surface_check_elevator = func {
-    interpolate("/controls/flight/elevator", 0.7, 0.5);
-    settimer(func(){
-        interpolate("/controls/flight/elevator", -0.7, 1.0);
-    }, 0.5);
-    settimer(func(){
-        interpolate("/controls/flight/elevator", 0.0, 0.5);
-    }, 1.5);
+    interpolate("/controls/flight/elevator", 1.0, 0.8, -1.0, 1.6, 0.0, 0.8);
 };
 
 ##########################################
 # Preflight control surface check: rudder
 ##########################################
 var control_surface_check_rudder = func {
-    interpolate("/controls/flight/rudder", -0.7, 0.5);
-    settimer(func(){
-        interpolate("/controls/flight/rudder", 0.7, 1.0);
-    }, 0.5);
-    settimer(func(){
-        interpolate("/controls/flight/rudder", 0.0, 0.5);
-    }, 1.5);
+    interpolate("/controls/flight/rudder", -1.0, 0.8, 1.0, 1.6, 0.0, 0.8);
 };
 
 ##########################################
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -191,6 +191,68 @@ var return_fuel_sample = func(index) {
 };
 
 ##########################################
+# Preflight control surface check: left aileron
+##########################################
+var control_surface_check_left_aileron = func {
+    var auto_coordination = getprop("/controls/flight/auto-coordination");
+    setprop("/controls/flight/auto-coordination", "false");
+    interpolate("/controls/flight/aileron", 1.0, 0.3);
+    settimer(func(){
+        interpolate("/controls/flight/aileron", -1.0, 0.6);
+    }, 0.3);
+    settimer(func(){
+        interpolate("/controls/flight/aileron", 0.0, 0.3);
+    }, 0.9);
+    settimer(func(){
+        setprop("/controls/flight/auto-coordination", auto_coordination);
+    }, 1.2);
+};
+
+##########################################
+# Preflight control surface check: right aileron
+##########################################
+var control_surface_check_right_aileron = func {
+    var auto_coordination = getprop("/controls/flight/auto-coordination");
+    setprop("/controls/flight/auto-coordination", "false");
+    interpolate("/controls/flight/aileron", -1.0, 0.3);
+    settimer(func(){
+        interpolate("/controls/flight/aileron", 1.0, 0.6);
+    }, 0.3);
+    settimer(func(){
+        interpolate("/controls/flight/aileron", 0.0, 0.3);
+    }, 0.9);
+    settimer(func(){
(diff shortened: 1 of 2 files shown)


gilbertohasnofb:
@dg-505 Once this is merged, we should add new steps to the preflight tutorial so that the user has to click once on each of these surfaces (left aileron, elevator, rudder, right aileron)

dg-505:
Well, I already thought about making these steps interactive in the tutorial.

Although this scripted flight surfaces check is a really nice feature it's just kind of a fake/cheat.

At least for the tutorial I would prefer a realistic approach:
Step: Move ailerons left
Step: Move ailerons right
Step: Center ailerons
And the same for rudder and elevator

We do it like this when we're checking the sailplanes in my flying club

dg-505:
I just read your discussion about not having the yoke available during preflight check and needing to move the flight surfaces by hand. Agreed, so just forget what I just wrote.....
Another suggestion: Make the surfaces movable with the mouse (like e.g. the throttle or mixture, a slider animation would be a good choice).
Together with this, disable yoke and mouse and  keyboard input for the flight surfaces when on the ground, groundspeed <0.1 and in an outside view.

IMO this would simulate realistically that you can't move the yoke when being outside the aircraft.

What do you think about this?

onox:
> Together with this, disable yoke and mouse and keyboard input for the flight surfaces when on the ground, groundspeed <0.1 and in an outside view.

I prefer to be able to control the surface/yoke when in outside view. I often use an external view when I start taxiing or taking off. Also need it to center the controls (5 key). It is not very intuitive when surfaces work while you're moving, but suddenly no longer work when you have stopped.

Also, the yoke and surfaces are mechanically linked, so doesn't that mean that you could consider you're moving the surfaces (even when you see the yoke moving) when you use the mouse/joystick?

onox:
You could however disable controlling the yoke/surfaces via mouse/joystick while in walker view.

Nasal/c172p.nas
```diff
@@ -191,6 +191,68 @@ var return_fuel_sample = func(index) {
 };
 
 ##########################################
+# Preflight control surface check: left aileron
+##########################################
+var control_surface_check_left_aileron = func {
+    var auto_coordination = getprop("/controls/flight/auto-coordination");
+    setprop("/controls/flight/auto-coordination", "false");
+    interpolate("/controls/flight/aileron", 1.0, 0.3);
```
onox: According to the [documentation](http://wiki.flightgear.org/Nasal_library#interpolate.28.29) you should be able to replace this with:

```javascript
interpolate("/controls/flight/aileron", 1.0, 0.3, -1.0, 0.6, 0.0, 0.3);
```

Nasal/c172p.nas
```diff
@@ -191,6 +191,68 @@ var return_fuel_sample = func(index) {
 };
 
 ##########################################
+# Preflight control surface check: left aileron
+##########################################
+var control_surface_check_left_aileron = func {
+    var auto_coordination = getprop("/controls/flight/auto-coordination");
+    setprop("/controls/flight/auto-coordination", "false");
```
onox: AFAIK `"false"` is interpreted by `setprop()` as just a string. Usually you would write `0` instead (since Nasal doesn't support the `true`/`false` literals).

Nasal/c172p.nas
```diff
@@ -191,6 +191,68 @@ var return_fuel_sample = func(index) {
...
+        setprop("/controls/flight/auto-coordination", auto_coordination);
+    }, 1.2);
+};
+
+##########################################
+# Preflight control surface check: right aileron
+##########################################
+var control_surface_check_right_aileron = func {
+    var auto_coordination = getprop("/controls/flight/auto-coordination");
+    setprop("/controls/flight/auto-coordination", "false");
```
onox: See comments above.

Nasal/c172p.nas
```diff
@@ -191,6 +191,68 @@ var return_fuel_sample = func(index) {
...
+    settimer(func(){
+        setprop("/controls/flight/auto-coordination", auto_coordination);
+    }, 1.2);
+};
+
+##########################################
+# Preflight control surface check: elevator
+##########################################
+var control_surface_check_elevator = func {
+    interpolate("/controls/flight/elevator", 0.7, 0.5);
```
onox: See comments above.

Nasal/c172p.nas
```diff
@@ -191,6 +191,68 @@ var return_fuel_sample = func(index) {
...
+    settimer(func(){
+        interpolate("/controls/flight/elevator", 0.0, 0.5);
+    }, 1.5);
+};
+
+##########################################
+# Preflight control surface check: rudder
+##########################################
+var control_surface_check_rudder = func {
+    interpolate("/controls/flight/rudder", -0.7, 0.5);
```
onox: See comments above.

gilbertohasnofb:
I am also against disabling the yoke outside. This work I did with the Nasal scripting works regardless of your yoke position, since it interpolates between the extreme points. Also, most of the time during preflight the yoke will be centred anyway, and even if not this would work.

As for being able to move the surfaces up and down, that's of course doable but I personally don't find it necessary. Since inspecting the surfaces doesn't really do anything regardless of using script or clicking to move them, it's all eye candy, I would just stick to what I did previously. Also, the scripting I use is able to disable auto coordination for its duration, and click and dragging won't, so ailerons will move the pedals, nose gear and rudder if auto-coordination is on.

> Also, the yoke and surfaces are mechanically linked, so doesn't that mean that you could consider you're moving the surfaces (pretending you're standing near a surface) (even when you see the yoke moving) when you use the mouse/joystick?

Yes, which is why the 3D model of the yoke does move.

Nasal/c172p.nas
gilbertohasnofb: Nice to know that! Fixing it now

Nasal/c172p.nas
gilbertohasnofb: E.g. line 479 of this very file (`c172p.nas`)

gilbertohasnofb:
@onox Okay, so I did these changes:

- using just one `interpolation` for a series of movements
- slowed down the movements a bit
- replaced the `"false"` for `0` which indeed works as well (but please notice that using `"false"` did work for me here, and that on line 479 of this very file you will find another `"false"`)

dany93:
@dg-505 wrote
> Together with this, disable yoke and mouse and keyboard input for the flight surfaces when on the ground, groundspeed <0.1 and in an outside view.

I do not agree. In reality, they can be moved, even when stopped on the ground. Mechanical links.
Also, I (and many people I think) use the external view to control (to _understand_ for the beginners) the JS functioning. It would be the single one aircraft which wouldn't allow this.
Aside from the eye-candy thing.

gilbertohasnofb:
:+1:

dg-505:
> I do not agree


Okay, then I think I'm outvoted :wink: 


@gilbertohasnofb  If I shall include the check in the preflight tutorial, you should add something like `setprop("/sim/model/c172p/left-aileron-checked", 1);` to the Nasal script.
Reason: The tutorial checks for a property to meed a given condition

gilbertohasnofb:
@dg-505 what about if the tutorial would check if the absolute value of the surface position is larger then let's say 0.5? This way whenever the surface would cross that threshold, the condition would be established. It might be necessary to add a small delay before going to the next step so that the full animation of the surface check is concluded, though the tutorial steps often take some time to move on for me. What do you think, wouldn't that be more elegant than creating this property you propose? Or does anyone else have any suggestions on how to do this?

dany93:
@dg-505 
> Okay, then I think I'm outvoted

Sorry, Jonathan. :-)

onox:
> replaced the `"false"` for `0` which indeed works as well (but please notice that using `"false"` did work for me here, and that on line 479 of this very file you will find another `"false"`)

You're right about it, @gilbertohasnofb. It seems that `"true"` and any non-zero number (as an integer or as a string) evaluates to `true`, while `""`, `"0"`, 0, `"false"`, or any other string evaluates to `false`.

I don't like what Nasal is doing here. If the language supported booleans, `setprop()` could raise an exception because the type of the property is boolean. Most languages in which you can cast a value to a boolean type, what is considered 'empty' is cast to `false` (`""`, `[]`, `0`) and anything else (including `"false"`) to `true`. `setprop()` does seem to raise an exception if the value is not a string or number, so it might just as well look at the type of the property.

It is unfortunate that the property tree and the XML supports booleans, but Nasal doesn't.

onox:
> In the external view or using the walker, simply click on the ailerons, rudder and elevator to move them as if the pilot was testing them.

One more thing: if I I'm in the cockpit with a door open, I can still click on the aileron. Is that intentional?

gilbertohasnofb:
> One more thing: if I I'm in the cockpit with a door open, I can still click on the aileron. Is that intentional?

I didn't think about that, I should disable it just like I did for the pitot tube cover, tiedowns, chocks, etc. I will work on it and push it soon

gilbertohasnofb:
@onox done!

onox:
Nice! :smile:

gilbertohasnofb:
Thanks! :smile: :+1:

Issue #740, opened 2016-03-29
Fuselage shadow cast in wrong position at wrong time


wlbragg:
See http://forum.flightgear.org/viewtopic.php?f=4&t=25157&p=280944#p280922

I verified that we get a shadow when we shouldn't https://dl.dropboxusercontent.com/u/63093453/fg/c172p/shadowi.jpg but min is still dark. However, I am pretty sure I have seen the white one before.under the perfect condition.

Unless Thorsten is willing to look at the shader, about all we can do is put a time condition on the shadow.

wlbragg:
What would be the best condition to use for this, sun angle, time?

wkitty42:
i don't know if you know it or not but there is also work being done, slowly, on having the moon as a light source which should lead to shadows being able to be cast by it... i don't think any of this code is in fgfs yet, though...

wlbragg:
Yes I was aware.

wlbragg:
BUMP

>What would be the best condition to use for this, sun angle, time?

Any suggestions?

wkitty42:
On 05/08/2016 12:30 AM, wlbragg wrote:
> BUMP
>
>     What would be the best condition to use for this, sun angle, time?
>
> Any suggestions?

time and sun angle so as to account for the different angles at the same time 
over the seasons... remember that 20:00 is still pretty light out in the summer 
but is pretty dark in the winter... though i may be partially mixing that up 
with daylight saving time...

Issue #124, opened 2015-04-21
Implement engine damage when flying at high RPM for long time


gilbertohasnofb:
Implement engine damage when flying at a RPM higher than the red mark (2700 RPM) for long time.

gilbertohasnofb:
@onox do you want to work on this? I had proposed to help with a dialog, similar to what I did for the oil and water contamination. What do you think?

onox:
I have been rather busy with university and other things, so can't promise to work on it. You could start by searching for some sound that is heard when the engine is starting to get damaged. (Lowered RPM?)

What we could do is create some interpolation table in XML that tells how much points you get for a certain RPM.

Furthermore, I think we should lower the RPM a certain amount depending on how many points you have above a certain threshold. (another interpolation table giving a RPM reduction offset depending on collected points).

About the Aircraft Options dialog: I think it should be simplified a bit. We're starting to have a bit too many "Enable feature A", "Enable feature B", etc. checkbuttons. Personally I would combine "Save fuel level between sessions", "Save battery charge between sessions", and "Save positions of all switches between sessions". Combine them to "Save state of aircraft between sessions", because a user either want to resume its state when they restart FG or they don't. "Allow fuel contamination", "Allow oil management", and "Allow carburetor icing" basically say whether the user wants bad stuff to happen to the engine, so they could be combined to "Allow engines failures".

Just my 2 cents.

onox:
Btw, I notice that I can no longer enable the tiedowns by clicking them in external view. I can still enable them via the Aircraft Options dialog.

onox:
It seems @dg-505 accidentally removed `<not>` around `<property>/sim/current-view/internal</property>` when he simplified the expressions as I suggested (see 85a42173f0db40282f653ada481f16379ce4a107).

onox:
I'm fixing it in the release branch.

onox:
Fixed in https://github.com/Juanvvc/c172p-detailed/tree/release/2016.4

onox:
Also cherry-picked to `master`.

dg-505:
> It seems @dg-505 accidentally removed `<not>` around 


Oups, sorryy  😞

gilbertohasnofb:
> I have been rather busy with university and other things, so can't promise to work on it. 

No problems at all

> You could start by searching for some sound that is heard when the engine is starting to get damaged. (Lowered RPM?)

:+1:


[...]

PR #869, opened 2016-11-16
Simplifying aircraft dialog

### LINKED ISSUE #866: Improve aircraft dialog
Improve aircraft dialog by removing unnecessary options or by grouping similar features into a single toggle:

- "save state of aircraft between sessions" (this would take care of fuel, battery, switches, etc.)
- "enable complex engine procedures and failures" (oil management, carb icing, engine damage, fuel contamination)
- possibly remove "enable shadow" option as shadow is automatically activated when in ALS and there is no reason to remove it, the 2D shadow for default rendering has long been gone and Rembrandt has its own shadow

gilbertohasnofb:
Closes #866 

Simplifying aircraft dialog:
- "save state of aircraft between sessions" checkbox now takes care of saving fuel quantity, battery charge and switches positions between sessions
- "enable complex engine procedures and failures" checkbox enables oil management, carb icing fuel contamination (and in the future should be used for engine damage as well)
- I also removed the "enable shadow" checkbox as the shadow is automatically activated when using ALS and there is no reason to hide it, also the 2D shadow for default rendering has long been gone and Rembrandt has its own shadow, so the option is just consuming precious space there
- reworded "Human models" label to "Show human models"

### DIFF
--- a/Systems/electrical.xml
+++ b/Systems/electrical.xml
@@ -12,5 +12,19 @@
             <property>/systems/electrical/battery-charge-percent-100</property>
         </output>
     </filter>
-
+    
+    <!-- ============================================================== -->
+    <!-- Save state of aircraft between sessions                        -->
+    <!-- ============================================================== -->
+    
+    <logic>
+        <name>Save state of aircraft between sessions => save battery charge</name>
+        <input>
+            <property>/sim/model/c172p/save-state</property>
+        </input>
+        <output>
+            <property>/systems/electrical/save-battery-charge</property>
+        </output>
+    </logic>
+    
 </PropertyList>
--- a/Systems/electrical.xml
+++ b/Systems/electrical.xml
@@ -12,19 +12,5 @@
             <property>/systems/electrical/battery-charge-percent-100</property>
         </output>
     </filter>
-    
-    <!-- ============================================================== -->
-    <!-- Save state of aircraft between sessions                        -->
-    <!-- ============================================================== -->
-    
-    <logic>
-        <name>Save state of aircraft between sessions => save battery charge</name>
-        <input>
-            <property>/sim/model/c172p/save-state</property>
-        </input>
-        <output>
-            <property>/systems/electrical/save-battery-charge</property>
-        </output>
-    </logic>
-    
+
 </PropertyList>
--- a/Systems/engine.xml
+++ b/Systems/engine.xml
@@ -611,4 +611,52 @@
         </output>
     </filter>            
 
+    <!-- ============================================================== -->
+    <!-- Complex engine procedures                                      -->
+    <!-- ============================================================== -->
+
+    <logic>
+        <name>Complex engine procedures => oil consumption allowed</name>
+        <input>
+            <property>/engines/active-engine/complex-engine-procedures</property>
+        </input>
+        <output>                         
+            <property>/engines/active-engine/oil_consumption_allowed</property>
+        </output>
+    </logic>
+
+    <logic>
+        <name>Complex engine procedures => carb icing allowed</name>
+        <input>
+            <property>/engines/active-engine/complex-engine-procedures</property>
+        </input>
+        <output>
+            <property>/engines/active-engine/carb_icing_allowed</property>                
+        </output>
+    </logic>
+    
+    <logic>
+        <name>Complex engine procedures => fuel contamination allowed</name>
+        <input>
+            <property>/engines/active-engine/complex-engine-procedures</property>
+        </input>
+        <output>
+            <property>/consumables/fuel/contamination_allowed</property>
+        </output>
+    </logic>
+    
+    <!-- ============================================================== -->
+    <!-- Save state of aircraft between sessions                        -->
+    <!-- ============================================================== -->
+    
+    <logic>
+        <name>Save state of aircraft between sessions => save fuel state</name>
+        <input>
+            <property>/sim/model/c172p/save-state</property>
+        </input>
+        <output>                         
+            <property>/consumables/fuel/save-fuel-state</property>
+        </output>
+    </logic>
+
 </PropertyList>
--- a/Systems/engine.xml
+++ b/Systems/engine.xml
@@ -431,7 +431,7 @@
             <property>/controls/engines/engine[1]/mixture</property>
         </output>
     </filter>
-    
+
     <!-- ============================================================== -->
     <!-- Carburetor Heat                                                -->
     <!-- ============================================================== -->
@@ -475,11 +475,11 @@
             <property>/controls/anti-ice/engine[1]/carb-heat</property>
         </output>
     </filter>
-    
+
     <!-- ============================================================== -->
     <!-- Fuel contamination and low oil level                           -->
     <!-- ============================================================== -->
-    
+
     <logic>
         <name>Engine Killed</name>
         <input>
@@ -516,7 +516,7 @@
             <property>/engines/active-engine/killed</property>
         </output>
     </logic>
-    
+
     <!-- Low level of fuel contamination or just above oil minimal level makes engine cough-->
     <logic>
         <name>Engine Coughing</name>
@@ -584,7 +584,7 @@
             <property>/engines/active-engine/coughing</property>
         </output>
     </logic>
-    
+
     <!-- ============================================================== -->
     <!-- Carburetor Icing drops RPM                                     -->
     <!-- ============================================================== -->
@@ -609,7 +609,7 @@
         <output>
             <property>/fdm/jsbsim/propulsion/engine[1]/volumetric-efficiency</property>
         </output>
-    </filter>            
+    </filter>
 
     <!-- ============================================================== -->
     <!-- Complex engine procedures                                      -->
@@ -620,42 +620,40 @@
         <input>
             <property>/engines/active-engine/complex-engine-procedures</property>
         </input>
-        <output>                         
+        <output>
+            <name>oil</name>
             <property>/engines/active-engine/oil_consumption_allowed</property>
         </output>
-    </logic>
-
-    <logic>
-        <name>Complex engine procedures => carb icing allowed</name>
-        <input>
-            <property>/engines/active-engine/complex-engine-procedures</property>
-        </input>
         <output>
-            <property>/engines/active-engine/carb_icing_allowed</property>                
+            <name>carb-icing</name>
+            <property>/engines/active-engine/carb_icing_allowed</property>
         </output>
-    </logic>
-    
-    <logic>
-        <name>Complex engine procedures => fuel contamination allowed</name>
-        <input>
-            <property>/engines/active-engine/complex-engine-procedures</property>
-        </input>
         <output>
+            <name>fuel-contamination</name>
             <property>/consumables/fuel/contamination_allowed</property>
         </output>
     </logic>
-    
+
     <!-- ============================================================== -->
     <!-- Save state of aircraft between sessions                        -->
     <!-- ============================================================== -->
-    
+
     <logic>
-        <name>Save state of aircraft between sessions => save fuel state</name>
+        <name>Save state of aircraft between sessions</name>
         <input>
             <property>/sim/model/c172p/save-state</property>
         </input>
-        <output>                         
+        <output>
+            <name>fuel</name>
             <property>/consumables/fuel/save-fuel-state</property>
+        </output>
+        <output>
+            <name>battery</name>
+            <property>/systems/electrical/save-battery-charge</property>
+        </output>
+        <output>
+            <name>instruments</name>
+            <property>/instrumentation/save-switches-state</property>
         </output>
     </logic>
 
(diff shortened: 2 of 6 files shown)


gilbertohasnofb:
More commits:
- removed registration on panel checkbox
- registration is now persistent between sessions

Ready to be tested and possibly merged!

Systems/electrical.xml
```diff
@@ -12,5 +12,19 @@
...
+    <!-- ============================================================== -->
+    <!-- Save state of aircraft between sessions                        -->
+    <!-- ============================================================== -->
+    
+    <logic>
+        <name>Save state of aircraft between sessions => save battery charge</name>
+        <input>
+            <property>/sim/model/c172p/save-state</property>
+        </input>
+        <output>
```
onox: IIRC you can specify <property> multiple times like this:

```xml
<output>
    <property>...
    <property>...
    ...
</output>
```

gilbertohasnofb: I see. I tried several `<output>`'s before but that didn't work, I will try a single `<output` with multiple properties and see if it works

gilbertohasnofb:
@onox do you want to add that callsign randomizer to this PR since it ended up influencing the registration number as well? If so, see what I wrote https://github.com/Juanvvc/c172p-detailed/issues/124#issuecomment-261027776

onox:
Another thing: the three "Enable left/right/tail tie-down" checkboxes, could be replaced with just one "Enable tie-downs" checkbox. What do you think?

gilbertohasnofb:
> Another thing: the three "Enable left/right/tail tie-down" checkboxes, could be replaced with just one "Enable tie-downs" checkbox. What do you think?

Sounds better as well! If one wants to do the tie-downs one by one then he can use the hotspots

Systems/electrical.xml
gilbertohasnofb: hmm, that did not work onox, are you sure you can have multiple properties in the filters? It is outputting only the first property and ignoring the others

onox:
> do you want to add that callsign randomizer to this PR since it ended up influencing the registration number as well? If so, see what I wrote #124 (comment)

Should I apply the patch to this branch or do you want to do it?

gilbertohasnofb:
> Should I apply the patch to this branch or do you want to do it?

You can apply the patch to this branch.

gilbertohasnofb:
@onox About the tie-downs, I am struggling to come up with a code to handle it. The problem is that the tiedowns won't all be linked to a single condition when the user clicks on them outside, and that complicates things a lot. So let's say he added a tie-down to the tail but not to the wings, what will the checkbox in the aircraft dialog show? But it's not only that, that checkbox will output a single property that if used in a `<filter>` such as I did with those other ones would constantly force all tie-downs either to be on or off. An option is of course to just leave it as it is now...

onox:
> An option is of course to just leave it as it is now...

You can leave it as it is for now. We can deal with it later.

I have pushed the commit, pull with `git pull --rebase` if you had already added more commits locally.

gilbertohasnofb:
> You can leave it as it is for now. We can deal with it later.

:+1: 

> I have pushed the commit, pull with git pull --rebase if you had already added more commits locally.

no new local commits

gilbertohasnofb:
@onox have you seen what I wrote about not being possible to have multiple properties in a single `<output></output>` block? If that's correct, then I guess this is ready to be merged

Systems/electrical.xml
onox: Let me check, requiring multiple `<logic>` would not be nice.

gilbertohasnofb: Okay, please let me know, but I did test it and only the very first property was affected...

onox:
It seems it is indeed not possible. I'll take a last look in the C++ code and then merge.

gilbertohasnofb:
No problems, so feel free to merge this whenever you feel like, or let me know if you find a way of solving the output block issue and I can rework those

onox:
It should work :confused: In `src/Autopilot/logic.cxx`:

```c++
46   // set all outputs to the given value
47   for( OutputMap::iterator it = _output.begin(); it != _output.end(); ++it )
48     (*it).second->setValue( value );
```

gilbertohasnofb:
Hmm, would you mind making a test in xml and see if you also run into troubles like I did?

onox:
I think I have a possible solution.

onox:
The trick is to add a `<name>` to the `<output>` element.

gilbertohasnofb:
> The trick is to add a <name> to the <output> element.

Aha! Nice to know. Should I test it here and if all is well merge it or is there still anything left to be done?

onox:
You can test my changes. I think we should initialize `/sim/model/c172p/immat-on-panel` to false.

Systems/engine.xml
```diff
@@ -609,6 +609,52 @@
...
         </output>
-    </filter>            
+    </filter>
+
+    <!-- ============================================================== -->
+    <!-- Complex engine procedures                                      -->
+    <!-- ============================================================== -->
+
+    <logic>
+        <name>Complex engine procedures => oil consumption allowed</name>
```
gilbertohasnofb: I think this should just be `<name>Complex engine procedures</name>`

gilbertohasnofb:
Testing now.

> I think we should initialize /sim/model/c172p/immat-on-panel to false.

But it is, see line 100 of the set file:
`                <immat-on-panel type="bool">false</immat-on-panel>`

gilbertohasnofb:
:+1: all good, thanks for the help and review!

wlbragg:
I know this is a little late, but why did we remove the option to disable the shadow?
I am against that idea, I often don't like the shadow on for reasons of capturing images as well as not liking the looks of it under certain conditions. I don't want to have to turn off all of ALS to remove the shadow.

gilbertohasnofb:
> why did we remove the option to disable the shadow?

I (or we?) thought that toggle was a relic from when we had two shadows, and it never crossed my mind anyone was disabling the ALS shadow!

> I often don't like the shadow on for reasons of capturing images as well as not liking the looks of it under certain conditions. I don't want to have to turn off all of ALS to remove the shadow.

Well, you can also disable it via the property tree (`/sim/rendering/shadow-volume`), it's just a little bit more hassle. I personally think it's worth it if (almost) no one else uses that feature, though of course I can only guess on what people do. What do you think about this, if you would still strongly want that toggle I can negotiate 😆 (though I can't speak for @onox and the others!)

wlbragg:
>I (or we?) thought that that toggle was a relic from when we had two shadows

That was never the intent of that option. We did have an option on which shadow you wanted in case you didn't use ALS.

>Well, you can also disable it via the property tree (/sim/rendering/shadow-volume)

Yes, I could also add the GUI back in to my local copy. That really isn't my point.

If you look at it like we don't have gui options for disabling other shader features I could potentially agree. But this is a little different. The shadow has issues, it doesn't always behave nicely or look good. That is my main argument for leaving a gui option for the casual user. If it was prefect and always behaved well and looked nice then I might not be so adamant.

The reason to remove it being for GUI simplification doesn't mean much to me. It's one line item. We could remove many GUI options and force the user to allow the option to simplify the GUI. This one item is not "simplifying" anything and we are removing an easy way to switch off the shadow which may be an issue for users. 

If there is still a consensus to remove it, even though I know we aren't in the habit of asking what users want, in this case I think we should and see if it is being used. Or more precisely, how many users don't like the shadow on. Or just leave it available.

Anyway, that's my opinion and I think it needs to be left in the GUI at least until we have a perfect shadow that looks good and behaves in all situations.

I think you know by now that my opinion on these types of issues is always in favor of giving options and not taking them away, especially if the feature might be contentions.

gilbertohasnofb:
> That was never the intent of that option.

This is my mistake, absolutely. I really thought the reason that toggle was there was something else. Had I realized it I don't think I would have proposed to remove it.

> The shadow has issues, it doesn't always behave nicely or look good. That is my main argument for leaving a gui option for the casual user. If it was prefect and always behaved well and looked nice then I might not be so adamant.

I understand your point and, as I said, I am up for bringing that checkbox back. @onox are you also fine with this? If so, I can make a new PR later today.

> The reason to remove it being for GUI simplification doesn't mean much to me. It's one line item. 

Well, each of the items removed or merged together were a single item, but then again they made several options together. 

> I know we aren't in the habit of asking what users want, in this case I think we should and see if it is being used. Or more precisely, how many users don't like the shadow on.

There is this anecdote -- wrongly attributed to Henry Ford -- that says that if he had asked people what they really wanted, they would have told him what they really wanted was a faster horse. This here is also a hilarious classic:



> Or just leave it available.
> Anyway, that's my opinion and I think it needs to be left in the GUI at least until we have a perfect shadow that looks good and behaves in all situations.

👍 Fine with me! 

> I think you know by now that my opinion on these types of issues is always in favor of giving options and not taking them away

Yes, I know your point of view here and yes I do disagree with it, but let's just solve somehow solve this issue in question and move on, is that fine?

gilbertohasnofb:
@wlbragg by the way, do you know if Thorsten is planning to make improvements to the shadows? For me the two main issues are: 

- the shadow is not uniform, as some parts of the aircraft makes the shadow darker in some places, as if the shadow would have structure. E.g.:


- while the shadow does get less intense as the sun goes down (i.e. the colour starts to get closer to the surroundings), the edges do not get softer/blurred.

onox:
> The shadow has issues, it doesn't always behave nicely or look good. That is my main argument for leaving a gui option for the casual user. If it was prefect and always behaved well and looked nice then I might not be so adamant.

Can we first give Thorsten the opportunity to fix the issues with the shadow before making a decision?

> I know we aren't in the habit of asking what users want, in this case I think we should and see if it is being used. Or more precisely, how many users don't like the shadow on. Or just leave it available.
>
> Anyway, that's my opinion and I think it needs to be left in the GUI at least until we have a perfect shadow that looks good and behaves in all situations.

It has only been removed in `master` (not in `release/2016.4`) so regular users still have the option for now. We can see if there are any loud complaints on the forum, but I prefer to see Thorsten just fix the any remaining issues with the shadow.

Systems/engine.xml
onox: I didn't see your comment (seems I need to press F5 for these things), so it should be fixed in a next PR.

gilbertohasnofb:
Would anyone talk to Thorsten about it then? 

@wlbragg Also, what are the issues you see with the shadow besides the two points I raised above? Is there anything else that makes it look ugly in your opinion?

Systems/engine.xml
gilbertohasnofb: No problems. I will work on the About dialog today, so when we are reviewing it we can sneak a little fix for this there.

wlbragg:
>Can we first give Thorsten the opportunity to fix the issues with the shadow before making a decision?
>Would anyone talk to Thorsten about it then? 

I think it would be low on his priority list.

>what are the issues you see with the shadow besides the two points I raised above?

The alignment to the wheelbase gets screwed up as the sun angle gets more extreme.
Reset/Re-init pretty much breaks it in a really bad way.
The elevation issues with the terrain (clipping). 

I'm sure there are others, but what you posted and these are the most obvious and noticeable.

gilbertohasnofb:
@wlbragg I think you do make a point on this whole issue

onox:
I've e-mailed Thorsten so we can first see if he's willing to fix some or all of these issues before we add back the checkbox.

wlbragg:
I was trying the scenario of controlled landing without yoke and started the c172p using the on approach at startup, Here is another reason not to eliminate the ability to remove the shadow without removing ALS in its entirety.
fgfs-screen-002
Also issues like this https://github.com/c172p-team/c172p-detailed/issues/740

wlbragg:
Until the shadow is close to perfect, it needs to remain an option.

wlbragg:
When starting on final using the Internal Launcher the aircraft starts with engine off. One more thing, have we missed something?

onox:
> I was trying the scenario of controlled landing without yoke and started the c172p using the on approach at startup, Here is another reason not to eliminate the ability to remove the shadow without removing ALS in its entirety.

That is just a bug. Are you using the Qt launcher? The c172p doesn't have proper states support right now. Just two dummy files for "approach" and "cruise". Seems to be some problem with the `Effects/shadow-vol.eff` effect?

wkitty42:
i agree that it is a bug... one that might be fixed with the states and allowing the sim and processing to catch up with the state the craft is in... that should put the shadow on the ground but that processing has to happen first...

wlbragg:
OK, it turns out that the shadow is there upon startup in that situation because the aircraft is started up paused. As soon as you un-pause it the shadow disappears.

However I discovered something else that could be an issue.

How is the shadow param /sim/rendering/volume-shadow set to true initially now? 
Also, was the saved state parameter taken into account. I ask because I had my shadow turned off in the version that still had the GUI choice and when I checked out the new version that has the GUI choice removed I have no shadow by default, I had to go into debug and set /sim/rendering/volume-shadow to true. That will be an issue for any user that has the shadow off in the version with the GUI when they upgrade to the version without it.

onox:
I'll add back the "Enable shadow" checkbox.

PR #871, opened 2016-11-17
About dialog

### LINKED ISSUE #870: Create "About" dialog in the C172P menu
Create "About" dialog in the C172P menu with information about our project, list of collaborator, etc.

gilbertohasnofb:
Closes #870 

- new About dialog
- set file altered a bit, since the description of the plane contained many spaces and line breaks which looked ugly in the sim


onox:
Cool idea to add an image :+1:

Some comments:

- Perhaps you could add a dashed horizontal line like the "Help" menu has
- The window is rather tall. Does it look better if you put the text to the right of the image?

gilbertohasnofb:
@onox 

> Cool idea to add an image :+1:

Thanks! But the idea is actually by @Harald67 from the c150 project.

> Some comments:
> Perhaps you could add a dashed horizontal line like the "Help" menu has

I will take a look at that now.

> The window is rather tall. Does it look better if you put the text to the right of the image?

The problem with that is that the image size is hard coded, so depending on the monitor resolution it might look funny on the left side. I think the vertical layout is more guaranteed, albeit tall.

gilbertohasnofb:
@onox first suggestion is done. If you agree with in leaving the dialog tall, this is ready to merge

onox:
Why would it look funny on the left side? Don't most screens have a larger width than height?

gilbertohasnofb:
@onox you are right, nothing funny about it, it looks better:

fgfs-screen-001

onox:
Nice. Would it be better to rename "Repository Link" to something like "Show repository in browser"?

I'm wondering if that button should be shown at all for regular users. We could change it to be only visible if the version ends with "-git".

gilbertohasnofb:
> Nice. Would it be better to rename "Repository Link" to something like "Show repository in browser"?

Wilco.

> I'm wondering if that button should be shown at all for regular users. We could change it to be only visible if the version ends with "-git".

I don't see the harm in having that link over there even for regular users. And it could be an advantage for them to come across the repo: they could report bugs directly here, see how development work and maybe get the idea of collaborating to a project themselves, directly download the most up-to-date version of the project, etc. Let's just leave it there.

gilbertohasnofb:
@onox done. Would you like to add a fix for https://github.com/Juanvvc/c172p-detailed/pull/869#discussion_r88320113 before merging this?

onox:
I'm adding a small amount of padding around it for consistency with the other dialogs.

onox:
I'll sneak in the fix for that other issue too.

gilbertohasnofb:
:+1:

onox:
Pushed. I've added a `<empty><stretch>true</stretch></empty>` to push the two browser buttons to the left. I've added such an element also between the description and authors to push authors a little bit more to the bottom. You can move it to below authors to push authors against description.

onox:
Maybe "Wiki Page" should be renamed as well. To "Show Wiki Page in Browser".

gilbertohasnofb:
Looks good to me, so that was all, right?

> Maybe "Wiki Page" should be renamed as well. To "Show Wiki Page in Browser".

I think this is a bit verbose. While "Open Repository" can be a bit confusing for a beginner, what else can someone expect when clicking "Wiki Page" than opening it in the browser? :smile:

onox:
Yeah, okay, let's merge it then :smile:

gilbertohasnofb:
Great, thanks a lot for the hand! :+1:

Issue #872, opened 2016-11-17
Fire extinguisher not affected by shadows:


gilbertohasnofb:
Fire extinguisher not affected by shadows:

![fgfs-screen-001](https://cloud.githubusercontent.com/assets/5700795/20410464/728ba9dc-ad1d-11e6-9927-d81a82afd38c.png)

![fgfs-screen-002](https://cloud.githubusercontent.com/assets/5700795/20410465/7290e7d0-ad1d-11e6-91b4-099d3f289e8d.png)

gilbertohasnofb:
I suspect the same is true for the other fire extinguisher (used by some of the bush variations), the two baggages and the ELT in the back of the cabin.

@wlbragg do you think you could take a look at this, or tell me how to fix this? I thought the shadow cube would be enabled similarly to the irradiance effect by using ...

````
    <effect>
        <inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>
        ...
````

... but that is not the case apparently.

wlbragg:
>@wlbragg do you think you could take a look at this, 

I don't mind fixing it (I actually did and created a bug-872 branch) except I can't seem to push. My credentials keep failing, yet I can change them in my account?

>or tell me how to fix this?

So. I think all it needs is to change
./../../../Effects/interior/c172p-interior-radiance
to 
./../../../Effects/interior/c172p-interior
The first uses the white cube and no shadow, the latter uses the shadow cube. Beware, testing and the final effect may not match the shadow in the cockpit because it is in a different space, same reason we don't use the shadow on most of the knobs and gauge faces.

We've talked about eventually trying to line up the sub-model .ac to match the local coordinates to illuminate this undesirable effect.

wlbragg:
I finally got it to push....

gilbertohasnofb:
> We've talked about eventually trying to line up the sub-model .ac to match the local coordinates to illuminate this undesirable effect.

gilbertohasnofb:
> We've talked about eventually trying to line up the sub-model .ac to match the local coordinates to illuminate this undesirable effect.

That would be great. Do we have an open issue for that?

wlbragg:
Yes, absolute position for instruments in individual .ac files #790

onox:
Corresponding PR was merged. Closing issue.

PR #873, opened 2016-11-18
Add interior shadow effect to suitcases and fe-elt.

wlbragg:
Bug #872

Pleas beware when testing, the shadow may not play nice with the existing interior cockpit shadow as they are in different coordinate space.

### DIFF
--- a/Models/Interior/Baggages/suitcase1.xml
+++ b/Models/Interior/Baggages/suitcase1.xml
@@ -4,7 +4,7 @@
 
     <path>suitcase1.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
+        <inherits-from>../../Effects/interior/c172p-interior</inherits-from>
         <object-name>suitcase1</object-name>
     </effect>
     
--- a/Models/Interior/Baggages/suitcase1.xml
+++ b/Models/Interior/Baggages/suitcase1.xml
@@ -4,7 +4,7 @@
 
     <path>suitcase1.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>
+        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
         <object-name>suitcase1</object-name>
     </effect>
     
--- a/Models/Interior/Baggages/suitcase2.xml
+++ b/Models/Interior/Baggages/suitcase2.xml
@@ -4,7 +4,7 @@
 
     <path>suitcase2.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
+        <inherits-from>../../Effects/interior/c172p-interior</inherits-from>
         <object-name>suitcase2</object-name>
     </effect>
     
--- a/Models/Interior/Baggages/suitcase2.xml
+++ b/Models/Interior/Baggages/suitcase2.xml
@@ -4,7 +4,7 @@
 
     <path>suitcase2.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>
+        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
         <object-name>suitcase2</object-name>
     </effect>
     
--- a/Models/Interior/FE-ELT/elt.xml
+++ b/Models/Interior/FE-ELT/elt.xml
@@ -4,7 +4,7 @@
 
     <path>elt.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
+        <inherits-from>../../Effects/interior/c172p-interior</inherits-from>
         <object-name>elt</object-name>
     </effect>
 
--- a/Models/Interior/FE-ELT/elt.xml
+++ b/Models/Interior/FE-ELT/elt.xml
@@ -4,7 +4,7 @@
 
     <path>elt.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>
+        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
         <object-name>elt</object-name>
     </effect>
 
--- a/Models/Interior/FE-ELT/fe.xml
+++ b/Models/Interior/FE-ELT/fe.xml
@@ -4,7 +4,7 @@
 
     <path>fe.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
+        <inherits-from>../../Effects/interior/c172p-interior</inherits-from>
         <object-name>extinguisher</object-name>
     </effect>
 
--- a/Models/Interior/FE-ELT/fe.xml
+++ b/Models/Interior/FE-ELT/fe.xml
@@ -4,7 +4,7 @@
 
     <path>fe.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>
+        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
         <object-name>extinguisher</object-name>
     </effect>
 
--- a/Models/Interior/FE-ELT/fe2.xml
+++ b/Models/Interior/FE-ELT/fe2.xml
@@ -4,7 +4,7 @@
 
     <path>fe2.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>
+        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
         <object-name>extinguisher2</object-name>
     </effect>
 
--- a/Models/Interior/FE-ELT/fe2.xml
+++ b/Models/Interior/FE-ELT/fe2.xml
@@ -4,7 +4,7 @@
 
     <path>fe2.ac</path>
     <effect>
-        <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
+        <inherits-from>../../Effects/interior/c172p-interior</inherits-from>
         <object-name>extinguisher2</object-name>
     </effect>
 


gilbertohasnofb:
@wlbragg Hmm, I am having some issues here:

- I get the following error in the console:
```
can't find "../../../../Effects/interior/c172p-interior.eff"
can't find base effect ../../../../Effects/interior/c172p-interior
```

- those objects become untexturized:
fgfs-screen-002
fgfs-screen-003

- all other xml for these type of objects (e.g. yoke, rudder pedals, flaps, breakers) refer to `<inherits-from>../../../../Effects/interior/c172p-interior-radiance</inherits-from>` in their xml files, and not to `<inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>`

wlbragg:
The difference between the two c172p-interior-radiance and c172p-interior is the first has no shadow cubemap, the second does. They both also activate flashlight effect and irradiate light from the instruments. If the yoke.. etc.. don't use c172p-interior then they don't have a shadow effect applied to them.

OK, not sure what happened to the wing as I didn't touch it. the other is probably the path. I don't think I got that outcome, I did test it enough to know that it didn't get the results one might expect because of the different coordinate space. Sorry, I will check it out.

wlbragg:
Sorry @gilbertohasnofb for missing the path. I really did test it and it didn't err like that as I had textures. But I really didn't think I had a very defined shadow during the test (but it can be difficult to get a shadow to cross the face on some of the objects) so I chocked it up to coordinate space. Now my test shows a well defined shadow across all the object.
I verified the yokes and pedals don't have the shadow applied, again, it was because of the difference in coordinate space. Do you want me to also add the shadow to either of them?

>again, it was because of the difference in coordinate space. Do you want me to also add the shadow to either of them?

FYI: When and if we ever re-factor all the submodels to the same coordinate space as the main cabin it will require duplicating some objects. For example, the yokes, 
old school = one yoke model duplicated using xml and positioning animation. At the time there was no shadow effect requiring a central point coordinate position.
new school = two separate yoke models positioned either in the main cabin .ac or matching the main cabin positions in its own submodel .ac.

onox:
> FYI: When and if we ever re-factor all the submodels to the same coordinate space as the main cabin it will require duplicating some objects. For example, the yokes,
old school = one yoke model duplicated using xml and positioning animation. At the time there was no shadow effect requiring a central point coordinate position.
new school = two separate yoke models positioned either in the main cabin .ac or matching the main cabin positions in its own submodel .ac.

When you import a model, you can specify a static offset via `<offsets>`. Can't the C++/shader use that?

gilbertohasnofb:
> BTW, how do you get the console to show up using Linux, I use the internal QT launcher but I launch the sim from the generated .sh file that download_and_compile.sh creates.

I don't know how to do it with the QT launcher as I still use FGRun.

> Sorry @gilbertohasnofb for missing the path

No problems!

> I verified the yokes and pedals don't have the shadow applied, again, it was because of the difference in coordinate space. Do you want me to also add the shadow to either of them?

Would the fact that they are not static be a problem? Would the shadows compensate for their current position or would they affect them as if they would be in the default position? Because if they can't compensate for movement, it's better to leave them as they are

> When you import a model, you can specify a static offset via <offsets>. Can't the C++/shader use that?

:+1:

zakalawe:
Unrelated to this bug, if you launch FGFS directly the console always exists (well we use whatever console you used to run it, nothing else needed). What does the .sh file do? It's presumably things you can do some other way.

gilbertohasnofb:
@wlbragg It's all looking good now. The shadows are not perfect in the fire extinguisher because I used some offset and rotations to position it, but the effect is really plausible and you can hardly perceive the imperfection. That said, I would be very interested in what @onox wrote about making the shader take any offset and rotations into consideration.

As for the yoke and rudder, let me know about what I wrote in the previous message. If there is nothing to do about them being moveable, then we just merge this. If moving isn't a problem, then I think we should add shadows to yokes, rudder, flaps, buttons and switches, etc.

wkitty42:
On 11/18/2016 01:55 PM, James Turner wrote:
> Unrelated to this bug, if you launch FGFS directly the console always exists
> (well we use whatever console you used to run it, nothing else needed). What
> does the .sh file do? It's presumably things you can do some other way.

i suspect they've (i don't know who wrote about the problem) made some sort of 
clickey thing instead of simply opening a terminal window and typing the 
necessary command...

wlbragg:
>What does the .sh file do?

All it does is run fgfs.exe --launcher  so the QT launcher opens. But I use a Steam controller to control the aircraft. I launch FG through Big Picture mode in Steam, thus no console or terminal appears. I guess when I develop I need to execute the .sh through a terminal, then I will get the debug.

>Would the fact that they are not static be a problem?

I don't know. I guess we need to try it and see what it does. Hold on with the merge and I'll see what the shadow looks like on the non static objects. I can push it for you to test if I question the aesthetics of it.

>When you import a model, you can specify a static offset via . Can't the C++/shader use that?

That is the question! Possibly it could be made to work or maybe not, and at what cost to send verts to the shader. 
I don't know anything about order of operations for loading the effect VS loading the mesh.
Right now the sub-model shader is applied in the submodel.xml, ie:
load - path to model.xml, model.xml has applied effect
then execute the offset
I think if it was easy, or maybe even possible with the current c++ operations, Thorsten would have already done for it.
For sure though, by not using offset and instead aligning the submodel mesh with the main model and not dynamically duplicating a submodel, we can get most everything to behave with the effects.
Personally, I never subscribed to the idea of sharing resources across multiple aircraft. This being one reason. I don't think the preservation of space is a good enough reason for the trade-off of losing better effects or the ability to customize a resource. Not to mention the future limitation and breakage that can happen if someone alters or removes a shared resource.

gilbertohasnofb:
> I think if it was easy, or maybe even possible with the current c++ operations, Thorsten would have already done for it.

Hmm, I don't want us to be spamming Thorsten with a million things, but would anyone maybe ask him? Could this have been a small oversight instead of a technical problem?

wlbragg:
>>I think if it was easy, or maybe even possible with the current c++ operations, Thorsten would have already done for it.
>Could this have been a small oversight instead of a technical problem? 
I've had the conversation with him previously and it is a "technical problem", i just don't know the details. I or we can ask that question of him though. Find out if it is a bigger issue than simply the math?

gilbertohasnofb:
> I've had the conversation with him previously and it is a "technical problem", i just don't know the details. 

I see, thanks for the info!

> I or we can ask that question of him though. Find out if it is a bigger issue than simply the math?

That would be nice.

onox:
If any shader cannot be adjusted to add the offset, and we're going to put the instruments/others at their absolute positions, we should combine them (at least all the cockpit stuff) in one .blend file and make sure they use the same texture (texture atlas) where possible (unless OSG performs this optimization for us automatically, but I am not sure of that).

onox:
> Personally, I never subscribed to the idea of sharing resources across multiple aircraft. This being one reason. I don't think the preservation of space is a good enough reason for the trade-off of losing better effects or the ability to customize a resource. Not to mention the future limitation and breakage that can happen if someone alters or removes a shared resource.

Instruments that are shared are put in `Aircraft/Instruments-3d/`. I doubt that there's any aircraft using `Aircraft/c172p/Models/Interior/Panel/Instruments/`. Usually, people just copy and paste it into their own aircraft.

wkitty42:
On 11/19/2016 10:24 AM, onox wrote:
>     Personally, I never subscribed to the idea of sharing resources across
>     multiple aircraft. This being one reason. I don't think the preservation of
>     space is a good enough reason for the trade-off of losing better effects or
>     the ability to customize a resource. Not to mention the future limitation
>     and breakage that can happen if someone alters or removes a shared resource.
>
> Instruments that are shared are put in |Aircraft/Instruments-3d/|. I doubt that
> there's any aircraft using |Aircraft/c172p/Models/Interior/Panel/Instruments/|.
> Usually, people just copy and paste it into their own aircraft.

i think he's talking more about stuff like craftA using parts from craftB 
instead of copying them over to craftA's directory tree... it is akin to "dll 
hell" in today's OSes...

onox:
I noticed btw that the `PitotTube` and various `_interior` objects in the `c172p.blend` file are double sided. Is it supposed to be like that? A while ago I was modifying some canopy and had to make things single sided to remove artifacts.

> I think he's talking more about stuff like aircraf A using parts from aircraft B instead of copying them over to aircraft A's directory tree.

Yeah, that's what I meant when I wrote "I doubt that there's any aircraft using `Aircraft/c172p/Models/Interior/Panel/Instruments/`".

wlbragg:
>I noticed btw that the PitotTube and various _interior objects in the c172p.blend file are double sided.

I don't know the ramification of double sided objects when only one side is ever visible. Common sense dictates that it is a waist of resource, but that is a guess on my part. We should find out for sure and that could be another issue under "Optimizing".

>we should combine them (at least all the cockpit stuff) in one .blend file and make sure they use the same texture (texture atlas) where possible

Absolutely, my understanding from a few conversations on the dev list is this is a resource hog, to do with creation of multiple Mip Maps for each texture, at the least.

gilbertohasnofb:
@wlbragg I just suggest that if we are really going to raise the cockpit as we spoke by e-mail, we do that first before merging all instruments in a single file, else we might have to redo this same work later.

wlbragg:
:+1:

gilbertohasnofb:
@wlbragg should we merge this and if Thorsten ever finds a way of handling any offset and rotation automatically then we see if any modification will need to be done in our end? Or is there anything else that needs to be done in this PR?

wlbragg:
Yes, fine with me.

>You're not authorized to merge this pull request. 

This is the first I noticed this, has something changed?

onox:
Who are you quoting?

onox:
And you need to accept your invitation to the c172p-team 😄

wlbragg:
>Who are you quoting?

Sorry, it was the immediate post preceding mine.

>@wlbragg should we merge this

Yes, fine with me.

>And you need to accept your invitation to the c172p-team :smile:

I hadn't noticed it in my mail, thanks!

gilbertohasnofb:
I will make just one more test (it's long time since I last checked this) and then merge it

> You're not authorized to merge this pull request.

Did you manage to solve this by accepting the invitation?

gilbertohasnofb:
All good, thanks @wlbragg :+1:

wlbragg:
>Did you manage to solve this by accepting the invitation?

Yes, I had missed the invite in my email.

Issue #865, opened 2016-11-16
Create control wheel lock


gilbertohasnofb:
Create and implement control wheel lock:

![](http://www.cessna150.net/Assets/Photos/inside_photos/controllock.JPG)

gilbertohasnofb:
Volunteers for the 3D model? I can work on texture + XML.

gilbertohasnofb:
@dg-505 let's continue our talk about the lock over here

gilbertohasnofb:
I created the texture, what do you think?

![control-lock](https://cloud.githubusercontent.com/assets/5700795/20351871/35238afa-ac15-11e6-9e97-a304340683ad.png)

dg-505:
I think when it comes to textures (and screenshots) you're really a magician  😆

gilbertohasnofb:
Thanks a lot, Jonathan! I really appreciate it :smiley:

gilbertohasnofb:
Here is one more idea: let's have two identical models of the lock, one positioned at the yoke, another in the pocket on the left side of the cockpit where the checklist is. So when the lock is on the yoke, clicking on it will remove it and put it on the pocket, and clicking on the lock on the pocket will put it back to the yoke. This was done by A2A, see the following video at 6:06 https://www.youtube.com/watch?v=CtFXPuklLB8

Also, when implementing the code we can't forget to deal with hide/show yokes and the lock. I would say that if one hides the yoke, the lock should be set to removed (and so go to the pocket). This is to avoid the problem of someone hiding the yoke while locked, then not being able to fly any longer and not knowing why.

What do you think?

wlbragg:
I suppose I can create the model in Blender. Has anyone else volunteered yet?

>What do you think?

Great ideas!

gilbertohasnofb:
@wlbragg if I understood correctly @dg-505 had volunteered to model this, am I right Jonathan?

dg-505:
At least can give it a try. I think such a small piece can be a good way to learn Blender...

wlbragg:
:+1:

onox:
Here is a bunch of keys I find useful in Blender:

`t` to show panel on the left (it contains a combobox called "Set Origin")
`n` to show panel on the right (it contains 3D cursor)
`shift`+`c` to set 3D cursor to 0,0,0
`shift`+`s` -> `Cursor to Selected`
`1`, `3`, `7` (+ `shift` optionally) to show front/right/top side
[...]

Issue #859, opened 2016-11-12
Gather a list of features and bugs we fixed in 2017.1


onox:
We should gather a list of all the new features, major changes, and bugs we fixed. Please create a list (use \* to create bullet points) here.

* Control surfaces can be checked for free movement during preflight inspection
* The opening of the baggage dialog was delayed as to wait for the animation of the baggage door
* Cleaning up the Aircraft Options dialog
* Improvements to the Ground Equipment dialog
* Callsigns of new users are automatically randomized to avoid them blocking other new users on multiplayer
* New About This Aircraft dialog, with information about the project as well as relevant links
* The pilot's yoke can now be locked with a control lock (lock can be found in the bag on the left of the pilot's seat)
* Fix bug with ADF timer in the radio stack
* Added view for IFR training
* Improvements to the carburetor system (icing, carb heat)
* Improvements to the engine coughs (due to fuel contamination or carb ice)
* Lowered limit of critical oil level according to Lycoming's manual
* Critical oil will not cause the engine to cough, it will simply quit when oil reaches the critical value of 2 quarts
* Improved fuel contamination system (fixed some bugs)
* More realistic bush wheels and tires
* Fuselage shows fresnel effect

Do not forget to update `NEWS.md` before creating the release. See https://github.com/c172p-team/c172p-detailed/wiki/Release-process-and-versioning

dg-505:
* Add a control lock

gilbertohasnofb:
@dg-505 you can now edit posts as well, and to keep things easier we always edit @onox 's post above instead of writing features in a bunch of messages. But no worries, I edited the first post and added this feature there :+1:

gilbertohasnofb:
@onox I totally forgot about the date of the new release; Torsten just wrote in the dev list:

> Hi all,
> 
> I am planning to let the release builder scripts run on Friday 17th sometimes during daylight hours (European timezone). I can't promise exactly when this will happen.
> 
> Everything that''s checked into next of the gits, trunk in fgaddon and the scenemodels database by Friday, Feb. 17th 07.00 UTC should make it into the release.
> 
> Regards
> Torsten

Should we contact Stuart as soon as possible for a merge? If so, let's try to solve that carb icing PR as fast as we can, as I would be happy to have that merged in the `master`

onox:
Yeah, we should merge soon.

gilbertohasnofb:
@onox I just wrote Stuart a message, let's hope he can merge a release before Friday.

[...]

PR #874, opened 2016-11-18
3D models for the control lock

dg-505:
Here's the first step to approach #865 

Since I don't have any knowledge about UV mapping, materials and textures in Blender I'd rather leave that to someone else (maybe @onox?) :wink: 

When the 3D model is finished I would volunteer to proceed with the animation and implementation.


onox:
> Since I don't have any knowledge about UV mapping, materials and textures in Blender I'd rather leave that to someone else (maybe @onox?)

I don't really know much about UV mapping actualy, so maybe ask @wlbragg?

wlbragg:
>I don't really know much about UV mapping actually, so maybe ask @wlbragg?

Do you have a specific question or do you want me to go ahead and do the UV mapping?

If I just volunteered, let me know when the texture is pushed and we're ready to map it.

dg-505:
I would be very grateful if you do the UV mapping.

Here's the texture @gilbertohasnofb created:
control-lock

gilbertohasnofb:
@wlbragg about the texture, please use the bottom half (the purely metallic one) for the back of the sign as well as for the "cable" that enters the yoke mount.

wlbragg:
:+1:

wlbragg:
OK, I know I have ask this question before, but I have to ask again, how do I pull  a local copy of this pr?

onox:
```
git checkout -b dg-505-master master
git pull https://github.com/dg-505/c172p-detailed.git master
```

onox:
There's a blue link called "command line instructions" btw :smiley: (to the right of the green "Merge pull request" button)

wlbragg:
Thank you. While I was waiting for instruction I checked out the model and texture. To make this work easily and better, they need to match in size and shape. Right now the model face is to long and not the same shape as the texture. I am only referring to the text portion of the texture.

dg-505:
Okay, what do you suggest: Do the Mapping first, and after that I adjust the size of the label-thing, or should I first try to get it to a correct proportion and you do the UV mapping after that?
What do you think?

Btw, many thanks for volunteering on this :)

wlbragg:
If you can get the proportions correct or even really close, then I'll map it. If you need a template to size the mesh to I can push a mapped mesh you can then size the object to. If I try mapping the texture to the existing object then when you try to resize it you loose the proportions on the mapping.

dg-505:
I think I have an idea how I can get it roughly to the correct proportions.
So I'll push that soon.

dg-505:
So now the proportions of the plate/label should be roughly correct :+1:

wlbragg:
I'm a dummy when it comes to git. I pushed the UV mapped control-lock to dg-505:master I think.

dg-505:
Hmm, doesn't look like:
https://github.com/dg-505/c172p-detailed/commits/master

wlbragg:
https://github.com/Juanvvc/c172p-detailed/tree/dg-505-master

wlbragg:
OK, now it is!

dg-505:
Ah, okay :+1: 

@gilbertohasnofb What do you think, can we keep it as it is?

gilbertohasnofb:
@dg-505 would it be too much trouble to post a couple of sreenshots? I am terrible at Blender...

dg-505:
bildschirmfoto vom 2016-11-26 12-57-14

bildschirmfoto vom 2016-11-26 12-59-28

gilbertohasnofb:
This looks gorgeous :smile: Just a small comment: it seems to me that there is a tiny bit of blackness in the bottom of the lock (below the red part), is that due to the UV mapping? The texture does have a black region around that area. If so, what would be easier: that @wlbragg fix this in the UV mapping or that I remove the black part from the texture? I don't mind doing it, and I'd probably have some free time today in a couple of hours.

@dg-505 one more question, how is the back of the sign looking, was it also UV mapped using the grey metal?

dg-505:
bildschirmfoto vom 2016-11-26 14-45-10
bildschirmfoto vom 2016-11-26 14-45-36

gilbertohasnofb:
@dg-505 Looks perfect, thank you! So then how should we proceed with the idea of putting the control lock in the pocket (together with the checklists):

- should we have two models in the same .ac file, groupped into labels such as `control-lock-yoke` and `control-lock-pocket`, so that we can implement in the XML that clicking in the visible one toggles the visibility of both? This has the advantage they would also be in their absolute position, which allows all effects (flashlight, shadows) to work well on them
- or should we have a single lock in a single .ac file, and then add it twice to the model xml, using offset and rotate to position it in the pocket?

@wlbragg I suppose the first option is better, right?

wlbragg:
>@wlbragg I suppose the first option is better, right

For shader effects, yes, the first.

However, If you were to use the second option and make the absolute position the one that is in the pocket, it would have a good shadow effect and then rotate it for the locked position. When in the locked position the aircraft is in a fixed (unmoving) orientation anyway and thus should minimize the Ill effect of any shadow as you probably won't ever notice that it may not be correct as the shadow really isn't moving when the aircraft is locked and stationary.

gilbertohasnofb:
Let's go with the first option then, as there is nothing to lose with it (and I suppose it's much easier to position something in Blender than the pain in the neck method of trial and error with XML). @dg-505 would you be able to do that in Blender?

dg-505:
Any comments/reviews about the current state? :smile:

gilbertohasnofb:
@dg-505 Ok, so I tested it and it is looking _really_ good, but I do have a couple of comments:

- please do not _ever_ use capitalization in the file names and folders, as this can get messy very easily. Windows do not care for capitalization, but Linux and OS X do, so `Control-lock.png` and `control-lock.png` are not the same file for the last two systems. When I tested, the lock was completely without texture, and this is because the `.ac` model was looking for a texture named `control-lock.png` but the folder had a capitalized one. So remove the capitalization of the folders and files

- the size of the plate is too small. Compare it with the the image I posted at https://github.com/Juanvvc/c172p-detailed/issues/865 , note how the sign is around 1.3 times the size of the gauges, while ours is slightly smaller. Could you please make it larger?

- probably this would also improve with a larger model, but my impression is that there is too little of the lock sticking out of the pocket, I would make it a bit more obvious

- do you plan to make the hole in the yoke mount? I think that would look good  :smile:

Models/Interior/Control-lock/Control-lock.xml
```diff
@@ -0,0 +1,91 @@
...
+                        <value>0.05</value>
+                    </less-than>
+                    <greater-than>
+                        <property>sim/model/c172p/cockpit/yoke-elevator</property>
+                        <value>-0.05</value>
+                    </greater-than>
+                    <less-than>
+                        <property>sim/model/c172p/cockpit/yoke-elevator</property>
+                        <value>0.05</value>
+                    </less-than>
```
gilbertohasnofb: Perhaps these conditions above are not necessary. I get that what you implement might be more "realistic" from a certain point of view, but the problem is that it might look cryptic to the user: "hmm, why the lock can't be placed right now but was placed before?" Some people fly with mouse or keyboard and might not centralize the yoke back. IMO we should remove this

Models/c172p.xml
```diff
@@ -1020,6 +1020,15 @@
         </offsets>
     </model>
     <model>
+        <name>ControlLock</name>
+        <path>Interior/Control-lock/Control-lock.xml</path>
```
gilbertohasnofb: When you change the capitalization of the folder, please remember to change here as well

gui/dialogs/aircraft-dialog.xml
```diff
@@ -297,6 +302,38 @@
...
+                                <value>0.05</value>
+                            </less-than>
+                            <greater-than>
+                                <property>sim/model/c172p/cockpit/yoke-elevator</property>
+                                <value>-0.05</value>
+                            </greater-than>
+                            <less-than>
+                                <property>sim/model/c172p/cockpit/yoke-elevator</property>
+                                <value>0.05</value>
+                            </less-than>
```
gilbertohasnofb: again, these conditionals above concerning aileron and elevator position maybe should be removed

gilbertohasnofb:
Sorry, just a couple of more comments concerning the programming side of this feature: it is working really great, but I think we can improve the code a bit:

- the autostart function has to be updated to automatically remove the lock
- those conditionals concerning the position of aileron and elevator should be removed in my opinion, see my comments in your code
- default lock state should be `false` and not `true`, see comments again

gilbertohasnofb:
@dg-505 one more thing: I believe the tutorial should also be updated to include this

dg-505:
> please do not ever use capitalization

okay, removed
(but nevertheless, I used the capital letters because all folders of the interior models have capital letters..... well, doesn't really matter)

> the size of the plate is too small

I made it a bit bigger

> there is too little of the lock sticking out of the pocket

with the bigger plate, it is more obvious

> do you plan to make the hole in the yoke mount?

it is already there, but for now just in the .blend file. Hold on, it isn't perfect yet.... :wink: 

> these conditions above are not necessary

Well, I can't agree here, because it's indispensable that the yoke is centered when placing the lock. I've added a notice that the aileron and elevator have to be centered first (control-lock.xml, lines 87-113)

> Maybe better to set the lock as to false as the default

Roger, as the other securing components are also off by default, the lock is removed by default, too

> autostart function has to be updated

Done :+1: 

> I believe the tutorial should also be updated

I'm planning to do this in the next tutorial update, which is my next project for this aircraft :smile:

gilbertohasnofb:
@dg-505 thanks for the fast work! :smile: :+1: 

> (but nevertheless, I used the capital letters because all folders of the interior models have capital letters..... well, doesn't really matter)

You are absolutely right, I didn't realize it. But notice that the filenames are virtually all in lower case letters (and I personally always keep my folders in lower case as well in my repos, but nevermind). Anyway, thanks for changing that, I think it's fine as it is now, and most importantly there are no conflicts of upper/lower cases.

> I made it a bit bigger
>
>> there is too little of the lock sticking out of the pocket
>
> with the bigger plate, it is more obvious

Looks _much_ better, IMO the model is perfect right now!

>> do you plan to make the hole in the yoke mount?
>
> it is already there, but for now just in the .blend file. Hold on, it isn't perfect yet.... :wink:

Ok, so I will wait for that before merging this, is that fine with you? A single PR for this would be good

>> these conditions above are not necessary
>
> Well, I can't agree here, because it's indispensable that the yoke is centered when placing the lock. I've added a notice that the aileron and elevator have to be centered first (control-lock.xml, lines 87-113)

I completely understand your approach, and while I agree that the yoke must be centred, I don't agree it's up to the user because it's hard to do that in the sim and easy in RL. I just tried doing it with the mouse and I found it nearly impossible to center the elevator without looking at the external view. What I am saying is that when you click on the lock, the yoke will be _automatically_ centred. You know, in RL it is a bit like putting a key in a door, you will "feel" the hole and get it done, but in a sim it's much harder to judge if a key is aligned to a hole, so same story here. Try it with the mouse and see what you think.

> Roger, as the other securing components are also off by default, the lock is removed by default, too

:+1: 

>> autostart function has to be updated
> 
> Done :+1:

Perfect :+1: 

> I'm planning to do this in the next tutorial update, which is my next project for this aircraft :smile:

Not a problem, we can work on that later in a different PR

TL;DR version of what could still be worked out:

- please consider again the conditionals, test it with mouse
- yoke mount

wkitty42:
on the thing about centering the controls and using a mouse, is hitting the 5 on the keypad not the easiest (and best?) method of centering the yoke? even with my joystick i'm still using the keypad for some things sometimes...

gilbertohasnofb:
> on the thing about centering the controls and using a mouse, is hitting the 5 on the keypad not the easiest (and best?) method of centering the yoke?

A couple of reasons:

- most notebooks do not have a keypad
- a lot of people might not even know about this shortcut (I just learned it from you)
- why do we need to press a key then click on the lock instead of simply clicking on the lock? It's not like we are centring the yoke for flight control reasons. It's the same story as the flight surface check I implemented: why make someone wiggle it with a mouse instead of using a script, given that both accomplish nothing except "test the surface".

wkitty42:
@gilbertohasnofb i like your last item... sure, if we're putting the lock in, go ahead and center automatically...
on notebooks with keypads, they've always been embedded in the keyboard and required a special key to switch them on... yeah, i know...
aside from that, i use a keyboard for most everything... i guess i'm really old school... i used to touch type about 120 words per minute... i'm no where near that fast any more :wink:  but being a keyboardist is why i used the keyboard when i first started with FGFS and learned that way of doing things...

gilbertohasnofb:
@wkitty42 

>  i like your last item... sure, if we're putting the lock in, go ahead and center automatically...

You know, the point is that airplanes do not have keypads nor mouse nor screens with limited resolution either. I think the way we simulate realistic procedures have to be well thought, else it becomes just too pedantic (often you see people saying that certain things are harder to achieve in the sim than in real life!).

> on notebooks with keypads, they've always been embedded in the keyboard and required a special key to switch them on... yeah, i know...

I just bought an USB keypad as it makes life much easier on a software I use for freelancing, but never gave it a try in FG. Lately all my free time goes into developing rather than flying, so mouse suits me just fine at the moment :smile:

wkitty42:
:+1: :+1: :+1:

onox:
You can also press the 5 key on the row with all the numbers (above the R and the T), it doesn't have to be the keypad.

dg-505:
Would it be okay for you if I add a note that the yoke can be centered with key 5?

onox:
Why not automatically center the yoke like @gilbertohasnofb suggested?

dg-505:
Well, because IRL the yoke doesn't get centered automatically if you stick in the control lock. You have to center it *before* you place the lock...

onox:
What about making it a 2 steps process: first click centers the yoke, second click applies the control lock?

dg-505:
That would also be an option.

@gilberthasnofb what do you think about that?

gilbertohasnofb:
I really don't want to be _that_ guy dragging it but why two clicks? Also, why clicking on the lock would centre the yoke? That's bad UI in my opinion. If you really think that having the yoke must be centred _before_ adding the lock and that automatically and instantaneously centring it is not enough, what about using a script like what I did for the control surfaces? That is, when you click _once_ in the lock in the pocket, the yoke would be smoothly interpolated to position 0,0 for let's say half a second, and then the lock would be added. The downside is that if the yoke is already centred (which is the case for all joystick users) this would mean a lag of 0.5 seconds that won't be understood as a feature.

Still about RL vs simulation, keep in mind that while 99% of all joysticks and yokes are self centring, a Cessna yoke does not come back to the centre position automatically (there are no springs like in a Saitek or CH yoke). But adding a lock is trivial for the reasons I wrote above (the example of putting a key in a hole).

dg-505:
Well, nevermind, than I'll do it in that way that placing the lock centers the yoke automatically.....

dg-505:
So now I've pushed the hole in the yoke. :+1: 

If there are no more objections I think this is ready to be merged.... Or did I miss something? :tipping_hand_man:

wkitty42:
On 11/28/2016 06:44 AM, Gilberto Agostinho wrote:
> I really don't want to be /that/ guy dragging it but why two clicks? Also, why
> clicking on the lock would centre the yoke? That's bad UI in my opinion. If you
> really think that having the yoke must be centred /before/ adding the lock and
> that automatically and instantaneously centring it is not enough, what about
> using a script like what I did for the control surfaces? That is, when you click
> /once/ in the lock in the pocket, the yoke would be smoothly interpolated to
> position 0,0 for let's say half a second, and then the lock would be added. The
> downside is that if the yoke is already centred (which is the case for all
> joystick users) this would mean a lag of 0.5 seconds that won't be understood as
> a feature.

can't it be checked if the yoke is already centered? if so, skip the 
centering... the real question then comes of how close to centered it centered 
and if there's just a tiny bit more to go do that when the lock is applied...


that said, i'm tossed between the options... they're real decent ideas but it 
is, as someone pointed out, a simulator... clicking the lock to apply it and 
automatically centering the yoke doesn't seem like such a bad idea... pop up a 
tip that says "centering the yoke" during which the yoke centers and then 
popping the lock into place should be sufficient in most cases... shouldn't it?

gilbertohasnofb:
@dg-505 

> If there are no more objections I think this is ready to be merged.... Or did I miss something? :tipping_hand_man:

Thanks a lot, I will check it right now! 

> can't it be checked if the yoke is already centered? if so, skip the centering... the real question then comes of how close to centered it centered and if there's just a tiny bit more to go do that when the lock is applied...

Sure, that's possible. But what if a joystick calibration is 1% off? What if it's 2% off? Where do we stop? We could make the time of interpolation proportional to the difference, but then again do we really want to make such a complex Nasal function for such a tiny thing as adding a lock? I really feel this is really overcomplicating something that should be trivial (user wants to lock the yoke -> user clicks on lock -> yoke locked)

> pop up a tip that says "centering the yoke" during which the yoke centers and then popping the lock into place should be sufficient in most cases... shouldn't it?

Well, I gave a bunch of arguments before, but here is yet another one: there are people using our sim which do not speak English, and relying on messages can be problematic. As always, the clearer and more intuitive the user experience is, the better. But I think it's all solved now, particularly since the last changes by @dg-505

gilbertohasnofb:
@dg-505 very good job, this looks really good! @onox do you want to take a look at the code to make sure there is nothing to be changed? I had a look at it and it seems all good to me, but you have an eagle eye for that :wink: Else, let's merge this!

Models/Interior/control-lock/control-lock.xml
```diff
@@ -59,57 +59,29 @@
...
-                        </greater-than-equals>
-                    </or>
-                    <not>
-                        <property>sim/model/hide-yoke</property>
-                    </not>
-                </condition>
-                <command>nasal</command>
-                <script>screen.log.write("You have to center aileron and elevator before placing the control lock!", 1, 0.5, 0);</script>
+                <command>property-assign</command>
+                <property>sim/model/c172p/cockpit/yoke-aileron</property>
```
onox: What is `/sim/model/c172p/cockpit/{yoke-aileron,yoke-elevator}` exactly? If you set the `/controls/flight/{aileron,elevator}` to 0 (like you do below), the yoke and control surfaces will both go to the center position. So I don't understand why you (additionally) set `/sim/model/c172p/cockpit/{yoke-aileron,yoke-elevator}`?

onox:
I have added a review comment (press F5 if necessary).

Nasal/c172p.nas
```diff
@@ -623,3 +624,16 @@ setlistener("/sim/signals/fdm-initialized", func {
...
+var ControlLock = func {
+
+    if (getprop("/sim/model/c172p/cockpit/control-lock-placed")) {
+        setprop("/sim/model/c172p/cockpit/yoke-aileron", 0);
+        setprop("/sim/model/c172p/cockpit/yoke-elevator", 0);
+        setprop("/controls/flight/aileron", 0);
+        setprop("/controls/flight/elevator", 0);
+    }
+
+    settimer(ControlLock,0);
```
onox: I don't like this `settimer()` call here. At the very least you should be using `maketimer()` for periodic tasks. But I would suggest to use a `<filter>` in an `<autopilot>` file since that is executed at FDM rate. Currently I can still move the control surfaces a bit until Nasal wakes up, drinks some coffee, and then resets the surfaces to 0.

gilbertohasnofb:
@onox This is a bit messy of GitHub, but to read your comment I need to click on "Show outdated", and clicking on the button "View changes" show the diff of all files and show only a previous comment of mine, but not yours. Just a heads up in case @dg-505 has troubles finding it (as I had)

gui/dialogs/aircraft-dialog.xml
onox: Isn't @gilbertohasnofb's comment still relevant?

onox:
Do you guys prefer to see the yoke move more slowly to the center position? It's currently based (guessed) on mechanical limits of the surfaces in `<kinematic>` elements in the FDM file. From -1 to 1 takes 0,15 seconds. That value is low enough to keep a fast response, but high enough to not look instant (speed of light).

JSBSim Reference Manual suggests it needs to be a number (not a property or number), so if you want a slower movement of the yoke + adding the control lock _after_ the yoke has been centered. Let me know. In that case, @dg-505 should add the `<filter>` elements for resetting the yoke (see my review comment above) to a new XML file. I will then add 2 noise-spike filters and a state-machine.

onox:
> JSBSim Reference Manual suggests it needs to be a number (not a property or number), so if you want a slower movement of the yoke + adding the control lock after the yoke has been centered. Let me know.

So we cannot use a property to that specifies a higher number during setting the control lock.

gilbertohasnofb:
> Do you guys prefer to see the yoke move more slowly to the center position? 
> [...]
> So we cannot use a property to that specifies a higher number during setting the control lock.

I think the current situation is fine, but if you guys want to do something different than go for it.

dg-505:
> Do you guys prefer to see the yoke move more slowly to the center position? It's currently based (guessed) on mechanical limits of the surfaces in <kinematic> elements in the FDM file. From -1 to 1 takes 0,15 seconds. That value is low enough for a fast response, but high enough to not look instant (speed of light).
JSBSim Reference Manual suggests it needs to be a number (not a property or number), so if you want a slower movement of the yoke + adding the control lock after the yoke has been centered. Let me know. In that case, @dg-505 should add the <filter> elements for resetting the yoke (see my review comment above) to a new XML file. I will then add 2 noise-spike filters and a state-machine.

Well, unfortunately I have *no* idea what you're talking about :joy: 

I also consider the current state to be fine but if the changes you @onox propose are a better solution, I suggest that you make these changes and add your commit to this PR, because I have no clue how to implement these. It was hard enough to get this small nasal script working..... And replacing `settimer` with `maketimer` doesn't seem to work (or at least it isn't that easy...)

gilbertohasnofb:
@dg-505 you could start by taking a look at the easier ones. For instance, @onox pointed out that some of those conditionals of 5% range for centring the yoke still seem to be there (in the aircraft dialog xml). Also, what about the comment on using the properties `/sim/model/c172p/cockpit/yoke-aileron` and `/sim/model/c172p/cockpit/yoke-elevator`, which seems to be unnecessary? Then we can tackle the other ones together.

I know your code already works, but keep in mind it's important to have a clean and compact code. And we all go through this when we open PRs here, just take a look at _any_ of my work :stuck_out_tongue_winking_eye:

onox:
@dg-505 If you fix those two things that @gilbertohasnofb mentioned in the previous comment, then I'll do the `<autopilot>` thing.

dg-505:
Fixed :wink:

gilbertohasnofb:
@dg-505 Tested, it's all working fine and the code is looking good, thanks for the improvements! I also noticed something else, if the securing aircraft is disabled, the lock is still available to the user:

fgfs-screen-001

I think it should work like the chocks, which can be toggled only when the securing elements are available. On the other hand, perhaps we should scratch that option altogether and let the securing elements be on all the time. @onox what do you say? If so, then we can leave this as it is now, and after we merge this PR I could probably take a look at that

@onox about using `settimer`, I noticed that we use it in several other functions in our Nasal code. That comment of yours suggesting that `maketimer` would be better applies only it in this particular case or is it always better to avoid using `settimer` and therefore we should replace those other cases as well? Anyway, the xml solution is better of course.

dg-505:
Oups, my mistake, I'll correct it immediately

gilbertohasnofb:
@dg-505 Thanks, but that only disables the toggle! Which means that if one leaves it on and disables allow securing, the yoke will be locked _forever_ :stuck_out_tongue_closed_eyes: 

If you really want to fix it, you would have to do something similar to what I did for the chocks in the file `Systems/ground-effects.xml`, using `<logic>`. But we can just wait and see what the others think about simply removing that option of toggling the securing elements altogether.

dg-505:
> Which means that if one leaves it on and disables allow securing, the yoke will be locked forever

Really? See lines 260-262. Switching "Allow securing aircraft" off should remove the yoke....

gilbertohasnofb:
> Really? See lines 260-262. Switching "Allow securing aircraft" off should remove the yoke....

Let me test that in the sim, maybe I am missing something

dg-505:
Okay :smile:

gilbertohasnofb:
@dg-505 so I tested and as I expected, it does have the problem I mentioned. So while enabling/disabling securing elements does put the lock away, the user can click on the lock model on the pocket and add it back even if the securing toggle is off, e.g.:

fgfs-screen-002

dg-505:
Ah now I now what you mean.

gilbertohasnofb:
@dg-505 a possible solution you can try is to add a condition to the binding when clicking on the model in the pocket, i.e. it's only clickable if the enable securing is on. Then try testing combinations of activating that toggle and clicking on the models and see if it works as expected :smile:

dg-505:
Now it should be fixed...

gilbertohasnofb:
@dg-505 Almost there! What about using this condition for the lock model (line 25 of `control-lock.xml`):

```
        <condition>
            <and>
                <not>
                    <property>sim/model/c172p/cockpit/control-lock-placed</property>
                </not>
                <property>sim/model/c172p/securing/allow-securing-aircraft</property>
            </and>
        </condition>
```

This way, the lock in the pocket only shows when securing is toggled on. This helps for two things: first, it does not show something that is not available, but also when you click on the disable lock you end up opening the checklist (since it is behind it). What do you think?

onox:
@dg-505 Did you edit that file in your 3rd last commit in the Github editor?

onox:
If I take a look at https://github.com/dg-505/c172p-detailed/network, I see that you pulled a commit (from Github) that wasn't already in your local git repository (which is why you're getting a merge commit). In those cases you can do `git pull --rebase`. It will pull in commits from Github and put any new local ones on top of those. If git somehow fails to rebase, you can do `git rebase --abort` (`git status` will tell you that).

onox:
Anyway, I'm working on the state machine. Should be ready in some minutes.

dg-505:
> Did you edit that file in your 3rd last commit in the Github editor?

Eehm... I don't think so, I did `git commit --amend`, then git asked me to pull first, and then again push to master.... Hope that isn't a problem that this merge commit is included?

dg-505:
> What about using this condition for the lock model (line 25 of control-lock.xml):

Already thought about it, but I wasn't sure if I should do it. Anyway, maybe it's better to hide it. To avoid confusion...

onox:
Ah, you did `--amend`. If you do that, git will modify your last commit (which will then get a new commit ID). Since you had already pushed your unamended commit (with the old commit ID), git will see that the commit ID of your last commit in your local repository and the one on Github are different, thus it will perform a merge.

So, in this case you should either (1) not use `--amend` or (2) do use it but push with `git push --force`.

dg-505:
Oh okay I think I understood it now, thanx :+1:

dg-505:
Well, is it a big problem that there's a merge commit in the history?

onox:
No, you just get distracting merge commits that could have been avoided.

dg-505:
Okay, I'll try it in the future, but I can't promise :wink:
Git can be so f***ing complicated :roll_eyes:

onox:
I learned git via https://www.atlassian.com/git. But `git help` is indeed written by kernel developers (read: unreadable)

dg-505:
Looks like a good page, maybe it helps me to understand git better

onox:
Argh, `<state-machine>` works, but `property-interpolate` command doesn't.

onox:
Okay, I got it working now.

onox:
@dg-505 You can pull in my commit from the `dg-505-master` branch again.

gilbertohasnofb:
@onox is this ready for final testing and merging?

dg-505:
Well, I found some behaviour:
https://youtu.be/kORhEUZEJUU

gilbertohasnofb:
@dg-505 Do you mean that the interpolation time differs if one uses the checkbox and if one clicks on the lock on the pocket? If that's the case, then @onox could maybe take a look at that?

onox:
I'll take a look at it.

onox:
@dg-505 I cannot reproduce it. Have you properly merged the branch?

dg-505:
Hmmm.... seems like this problem solved itself....

Anyway, I think this is now ready to be merged?

onox:
@gilbertohasnofb Can you test it too? `dg-505-master` branch in this repository.

gilbertohasnofb:
> Can you test it too? dg-505-master branch in this repository.

Sure, doing it right now.

gilbertohasnofb:
I don't see any difference between the time of interpolation as in that video.

Thad said, I did spot a problem: that improvement I suggested here https://github.com/c172p-team/c172p-detailed/pull/874#issuecomment-263607632 and which @dg-505 had implemented seems to have been lost somehow for some reason...

dg-505:
Actually it is there in this PR, but for some reason it didn't make it into branch c172p-team/dg-505-master......

onox:
Isn't the problem that you removed the `<and>`?

onox:
Hang on...

dg-505:
No the problem is that in branch dg-505-master the `<property>sim/model/c172p/securing/allow-securing-aircraft</property>` doesn't exist

onox:
Try this:

```xml
<and>
    <property>sim/model/c172p/cockpit/control-lock-placed</property>
    <not>
        <property>sim/model/c172p/cockpit/control-lock-visible</property>
    </not>
    <property>sim/model/c172p/securing/allow-securing-aircraft</property>
</and>
```

dg-505:
AFAIK if there is no `or` FG handles the conditionals like `and` even if there is no `and`

onox:
Oh forget my comment. I'm too tired.

dg-505:
Mee too :smirk:

onox:
So is there still a problem in your master branch, @dg-505, or can we merge this PR?

dg-505:
AFAICT this should be ready to be merged.....

Oh wait, should I update NEWS.md??

onox:
No.

dg-505:
Okay then I think we're done here

dg-505:
So now my next project is updating the tutorials where I will include this feature :+1:

onox:
You can update #859 if you want 😄

onox:
@gilbertohasnofb Has @dg-505 been invited to the team already?

dg-505:
I don't think so.... I show up as "Contributor", not as "Collaborator" like you and Gilberto

onox:
I have send you an invite.

dg-505:
Looking forward :smile: :+1:

gilbertohasnofb:
> Has @dg-505 been invited to the team already?

I just sent him an e-mail a couple of minutes ago asking if he would like that, but I guess the message above answers that :smile: 

So in that case, @dg-505 officially welcome on board :wink: 

> I have send you an invite.

I don't see that invite, so I just resent it in case Jonathan doesn't see it

gilbertohasnofb:
And by the way, good work for both of you, very nice feature to have this lock implemented :+1:

dg-505:
Thank you :smile: 

> You’ll receive notifications for all issues, pull requests, and comments that happen inside the repository

Wow, that will blow my email inbox :joy:

onox:
Haha, yeah, GH can be quite spammy sometimes.

gilbertohasnofb:
@dg-505 if you use an e-mail client (and who doesn't use one in 2016?) you can add a filter that to automatically redirect any messages containing the string `[c172p-team/c172p-detailed]` in the subject to a specific folder, so that _all_ messages from this repo will be in that one place.

PR #877, opened 2016-11-23
New Thubmnails added in FGDATA by JMTurner

FGDATA:
Hi Guys
@gilbertohasnofb  @Juanvvc  @onox 

Please evaluate

James Turner added this 3 files to the C172p in FGDATA and these are still missing in the upstream repository

Happy Dev.

@IAHM-COL


gilbertohasnofb:
I don't like these images, to be honest, they look like snapshots and not carefully taken screenshots. Also, some problems: the first one shows a plane with engine off on a runway, and the second does the same but with choks! The third seems to use default weather (uglier than AW) and shows a plane dark and cold on a runway once again.

I don't want to sound too harsh but (personally) I don't think these are on the same level as let's say our splash screens. While we can merge this in order to be in sync with FGDATA, I would suggest we immediately open a new issue to create better thumbnails, which I would volunteer to take care of. @onox what do you think?

FGDATA:
fine by me

zakalawe:
Don't worry about this for now, these were test data. I will actually need to replace your splash screens shortly as well. If you would prefer to make them yourselves, I need the following:
- an exterior shot for each variant
- a decent panel shot
- a few 'feature' shots as you desire (showing what features of the 3D model you guys are most proud of)

These should be 16:9 or 4:3 aspect and probably 800x600 or 1024x768 (or equivalent sizes for 16:9 aspect).

The important thing: there must be /no/ text or borders on the image, because the new splash screen will overlay that info (i.e aircraft name, description, maybe authors) from the -set.xml automatically and dynamically.

The launcher will also make these images visible when browsing aircraft to be installed. It will help me considerable if the C172 is a good example of what is possible!

zakalawe:
Oh and JPEG format is probably better for file size, to avoid the package getting huge.

gilbertohasnofb:
@zakalawe sounds good to me! 👍 In that case, I will take care of creating these jpg files with the same names in our repo. @onox So should we close this and I will work on these soon, possibly until tomorrow?

zakalawe:
No rush from my side, but I'm delighted to have real assets to work with. Once the code works we will need new generic splash screens for fgdata too, I am sure the C172 can feature in those too, they can be more 'artistic' / showing scenery too.

gilbertohasnofb:
Ok, so I am closing this and I will work on it soon. I opened the issue #878 as a reminder. Thanks, James!

wkitty42:
On 11/23/2016 01:36 PM, Gilberto Agostinho wrote:

> I don't want to sound too harsh but (personally) I don't think these are on the
> same level as let's say our splash screens. While we can merge this in order to

ummm... i don't know where james got these from but they are part of the PoC for 
the new launcher, AIUI... he needed something to work with for this new 
feature... we can easily provide other snapshots to be used... we should only 
need the necessary dimensions he wants for these alternative pics...

plus it isn't like we can't contact him in this forum thing ;)

wkitty42:
On 11/23/2016 01:41 PM, James Turner wrote:
> Oh and JPEG format is probably better for file size, to avoid the package
> getting huge.

ummm... what about the variable splash screen capability? is only one per 
variant or several per base and/or variant?

wkitty42:
On 11/23/2016 02:59 PM, Gilberto Agostinho wrote:
> Ok, so I am closing this and I will work on it soon.


oops! sorry i'm so late to the party... ignore me for now, please ;)

Issue #885, opened 2016-11-30
Display of the ADF timer is broken


gilbertohasnofb:
Display of the ADF timer is broken, as reported @vainiovano :

> This commit breaks the display of the ADF timer while the simulation is running. The timer display is supposed to be multiplexed with the standby frequency display. When I replace the file with the previous version, the instruments seem to work as expected. Here is what I did:
>
> 1. Select Autostart from the Cessna C172P menu.
> 2. Press the FLT/ET button of the KR87 TSO panel. The right-hand-side 7-segment display should display a clock whose reading is changing approximately once per second, and the label FLT.
> 3. Press the FLT/ET button again. The display should show another time and the label ET.
>
> When this change is present, the right-hand-side display will not reflect the clock, but it will instead respond to the ADF knob. This should only happen when the FRQ label is being displayed (the FLT/ET button has not been pressed after pressing the FRQ button).
>
> It appears that pausing the simulation will pause or stop the script Nasal/kr87.nas.

gilbertohasnofb:
As @vainiovano correctly pointed out, the pausing of the display while the sim is paused is due to the fact that it is controlled by a Nasal script. On line 177, the update function uses the following to update itself:

`settimer( func { me.update() }, 0.1 );`

...but the problem is settimer also pauses when the sim pauses. This behaviour is present in virtually all other things which are controlled by Nasal (for instance, changing between distance display and time display in the DME also depends on the sim not being paused).

I will then just manually edit that original file since there has been other modifications afterwards and so we just keep living with inoperative ADF knobs while paused.

gilbertohasnofb:
Closed via https://github.com/c172p-team/c172p-detailed/pull/886

PR #886, opened 2016-11-30
Fixing display of the ADF timer

gilbertohasnofb:
Solves #885 

Basically, I broke that display in the commit https://github.com/c172p-team/c172p-detailed/commit/5bb4bb618a9467cc720a2fc7adb582a11c49fc8e

I am creating this PR rather than reverting it because 1) @onox has made other modifications to this same file after that commit and 2) I also had removed some unnecessary lines of code in that fateful commit, so this had to be done manually. Please see the issue description for more information.


dg-505:
Seems to be okay for me... IMHO this can be merged without further modifications

gilbertohasnofb:
@dg-505 Thanks for testing this, I also believe so. But I would like to wait for @onox because he might know some Nasal trick to solve what I wrote here https://github.com/c172p-team/c172p-detailed/issues/885#issuecomment-263925167 so let's keep this opened until he's around

onox:
Not aware of any trick. It's rather strange though that the FDM and Nasal are paused, but the pick/knob animations still work.

PR #903, opened 2017-02-14
bug-902

### LINKED ISSUE #902: Review fuel contamination system
It seems that the fuel contamination system has stopped to work and this needs to be investigated. E.g.: if you manually set the property `/consumables/fuel/tank[0]/water-contamination` to a value > `0.0` then open the fuel sump GUI dialog under the window and take a sample, that property is reset to `0.0` and the sample will be blue (i.e. no water contamination).

gilbertohasnofb:
Closes #902 

The fuel contamination system has been vastly improved with this PR:
- after taking a sample, the colour was shown according to the water _left_ in the tank instead of the water _in the sample itself_. This has been fixed
- so now in addition to `/consumables/fuel/tank[n]/water-contamination` we also have two properties `/consumables/fuel/tank[n]/sample-water-contamination` (for n = 0, 1). When water is present in the fuel and a sample is taken, the `water-contamination` diminishes by at most 0.2 and `sample-water-contamination` increases by that same amount. Then the colour is given according to the sample. Returning the sample reverses this process, discarding it resets the `sample-water-contamination`.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -52,8 +52,10 @@ var autostart = func (msg=1) {
 
     # Removing any contamination from water
     setprop("/consumables/fuel/tank[0]/water-contamination", 0.0);
-    setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);        
-
+    setprop("/consumables/fuel/tank[1]/water-contamination", 0.0);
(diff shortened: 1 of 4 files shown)


gilbertohasnofb:
@onox @wlbragg this is ready to be reviewed and possibly merged.

PR #904, opened 2017-02-14
Improve method of removing static models a bit

onox:
Changed the way to remove a static model a bit. Simply test if you can still remove the ground equipment.


gilbertohasnofb:
@onox Looks fine to me, I can add and remove the objects normally.

PR #905, opened 2017-02-14
Bug 587 first step: Add control lock to preflight and securing tutorials and checklists

dg-505:


gilbertohasnofb:
checklists: 

- pre flight inspection checklist: you added an item for the Control Lock as "remove (if existing)". Can't this be handled by a conditional? I.e. the [>] becomes active only if the "securing aircraft" option is allowed. If "securing aircraft" is not allowed, the button becomes inactivated.
- securing aircraft checklist: similar to above. Also, all the securing elements can be forced even if "securing aircraft" is false, which is a bit messy (e.g. if your "securing aircraft" is false, you can still add a pitot tube cover via the checklist but then can't remove it by clicking on it. similar stuff for tie-downs, chocks, etc). So better is to deactivated the [>] buttons if that property is false. For instance, try using a `<condition>` block for that, e.g.:

```
            <binding>
                <property>sim/model/c172p/securing/pitot-cover-visible</property>
                <value type="bool">true</value>
            </binding>
```

becomes:

```
            <binding>
                <condition>
                    <property>/sim/model/c172p/securing/allow-securing-aircraft</property>
                </condition>
                <property>sim/model/c172p/securing/pitot-cover-visible</property>
                <value type="bool">true</value>
            </binding>
```

It's a pity that `<enable>` does not seem to work with a checklist item, as it not only disables a button but as well make it look disabled.

- What are the tutorials called "Checklist: <something>" such as "Checklist: Securing Airplane"? They seem to be doubled from the regular tutorials. Is the idea that they just go through the checklist at the current airport? See:

fgfs-screen-001

- If so, then there is an issue with "Checklist: Securing Airplane" for instance. It does not check if the "securing aircraft" option is `true` or not, and when it comes the moment of using tie-downs the tutorial gets stuck (the user can't click there). I would suggest that for ALL tutorials you can override that property to `true` (some of them already do, others don't, please check all).

- Also, in the tutorial "Checklist: Securing Airplane", you can for some reason add a pitot tube cover even if the securing aircraft is `false`, which is not possible outside the tutorial. This seems potentially buggy so forcing that option to `true` is really important in all these

- I tried loading the tutorial "Securing Aircraft" and it didn't take me to Barcelona, is that intended? I think all tutorials should take place in Barcelona, except for the "Checklist: <something>" types.

dg-505:
- As far as I know unfortunately this is not possible by design. The checklist just doesn't support the feature to make checklist items "inactive". I also had the idea but something like this is mentioned nowhere. In addition to this, bindings aren't supported by checklists. It's just the determined syntax that gets interpreted by nasal (I think it's [FGData/Nasal/checklist.nas](https://sourceforge.net/p/flightgear/fgdata/ci/next/tree/Nasal/checklist.nas)). The <binding> tag in the checklists also doesn't seem to support conditonals: I tried the code you posted but even if "allow-securing-aircraft" is false, this [>] button still places the control lock. It's indeed a little bit annoying to struggle with this but there doesn't come any short-time solution to my mind. I think this requires an update of the whole checklist/tutorial system

- Same thing as above

- The tutorials "Checklist:XXXXXX" are in fact taken from the aircraft checklists. You will see that you have all checklist which are available in the Help -> Checklists dialog, are also in the tutorial dialog. If you start a tutorial called "Checklist:Whatevertheris", FG will just start to "read out" the individual checklist items one by one, and proceed to the next item only if the condition of the present one is fulfilled. This just simulates that your copilot reads out the checklist while you execute the items. It's just another way to work through the checklist.

- Because these tutorials "Checklist:Checksomelists" only go through the aircraft checklists they cannot set any property to any value by themselve. Only the pilot can do so. As a result it is not possible to set "allow-securing-aircraft" to true when starting the tutorial "Checklist:Securing airplane" ore another of these checklist-tutorials

- See this video. While doing this, I'm not in the tutorials:
https://youtu.be/BtrqxHmICk0 (upload should be finished in a few minutes)


- I think I can remember that we talked about you last point before: My intention was that the user who just landed somewhere and wants to use the tutorial to shutdown the engine, doesn't gets suddenly teleported to another airport. You know what I mean?

Did you understand my elucidations? Don't hesitate to ask if anyhing is unclear!!

gilbertohasnofb:
@dg-505 

Are you sure you pasted the code I did correctly? It really works here, I just tested it again. The button [>] doesn't get the disabled look, but clicking on it while "securing aircraft" is `false` does avoid the pitot tube cover to be added. I tested it in the "Securing Airplane" checklist, but it should work on all of them.

> Because these tutorials "Checklist:Checksomelists" only go through the aircraft checklists they cannot set any property to any value by themselve.

Ok. But that is possible with the other tutorials, right?

> See this video. While doing this, I'm not in the tutorials:

That's strange, I can't reproduce it. I launch FG, confirm that "securing aircraft" is `false` and I cannot add the pitot tube cover by clicking on the pitot. Something seems to be wrong over there, are you on Linux? If so, would you mind deleting your aircraft save state file located at `~/.fgfs/aircraft-data/c172p.xml` (on Windows I don't know where that would be)

> I think I can remember that we talked about you last point before: My intention was that the user who just landed somewhere and wants to use the tutorial to shutdown the engine, doesn't gets suddenly teleported to another airport. You know what I mean?

But isn't that exactly what the "Checklist: Securing Airplane" would do?

gilbertohasnofb:
Those improvements that didn't make into this PR can be found in this issue here: https://github.com/c172p-team/c172p-detailed/issues/906

PR #911, opened 2017-02-21
volumetric efficiency should be 0.85 and not 1.0

### LINKED ISSUE #909: Engine possibly too powerful
Stuart wrote in a private e-mail about a potential bug with the aircraft's engine:

> [...] I've spotted a possible bug in the aircraft.  It doesn't seem to be a regression, but might be worth looking at.
> 
> Using the Fair Weather scenario (which has 20C temperatures), taking off at full throttle and puting the autopilot into a 500fpm climb, I'm seeing 145kts airspeed and the prop at 3200rpm.  This is with the 160hp variant.  Those figures seem very high, particularly in a climb.

I can confirm I can reproduce these, and actually this bug does not involve the autopilot at all: simply take off with fair weather at full throttle and manually hold a climb rate of around 500 fpm and the speed will increase _very_ fast.

gilbertohasnofb:
Fixes #909 

`fdm/jsbsim/propulsion/engine[n]/volumetric-efficiency` should be `0.85` for either engine, not `1.0`. A filter introduced by the carburettor system was overwriting that value into `1.0`, thus making the engine more powerful than it should. The following PR fixes this by correcting the input property of that filter, which is generated in the carburettor icing Nasal implementation.

For more discussion on this, please see the issue above.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -72,7 +72,7 @@ var autostart = func (msg=1) {
     # removing any ice from the carburetor
     setprop("/engines/active-engine/carb_ice", 0.0);
     setprop("/engines/active-engine/carb_icing_rate", 0.0);
-    setprop("/engines/active-engine/volumetric-efficiency-factor", 1.0);
+    setprop("/engines/active-engine/volumetric-efficiency-factor", 0.85);
 
     # Checking for minimal fuel level
     var fuel_level_left  = getprop("/consumables/fuel/tank[0]/level-norm");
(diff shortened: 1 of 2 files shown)


wlbragg:
@gilbertohasnofb I am intimately familiar with this fix as I already made these changes to the Cub and pushed to fgaddon . Do you have any problem with me going ahead and merging this, or do you want @onox to see it first?

gilbertohasnofb:
Code-wise there is no difference, it is just a parameter change. I am 100% sure it was a mistake of mine and I am 100% sure this will fix what I introduce. What let to be tested is if the engine is now behaving as it should or if there were other things making it overpowered (though I quite frankly doubt, it would be too much coincidence to have introduced two bugs like this at once). But the problem is that I don't have enough knowledge about how _exactly_ the engine should behave and what airspeeds and rates should be observed. 

TL;DR we could merge this now but before declaring everything fine I think @dany93 or @stuartbuchanan should take a look if everything is really fine.

wlbragg:
I merged it to expedite possibly getting it include in the new release. It is for sure correcting a mistake regardless of other consideration.

gilbertohasnofb:
:+1: I will write Torsten straight away

PR #912, opened 2017-03-06
Bug 878

### LINKED ISSUE #878: Create new thumbnails
Create three thumbnails with filenams `thumbnail-exterior.jpg`, `thumbnail-ground.jpg` and `thumbnail-panel.jpg` using the recommendations by @zakalawe 

>  If you would prefer to make them yourselves, I need the following:
> 
> * an exterior shot for each variant
> * a decent panel shot
> * a few 'feature' shots as you desire (showing what features of the 3D model you guys are most proud of)
>
>These should be 16:9 or 4:3 aspect and probably 800x600 or 1024x768 (or equivalent sizes for 16:9 aspect).
>
>The important thing: there must be /no/ text or borders on the image, because the new splash screen will overlay that info (i.e aircraft name, description, maybe authors) from the -set.xml automatically and dynamically.
>
>The launcher will also make these images visible when browsing aircraft to be installed. It will help me considerable if the C172 is a good example of what is possible!
>
>Oh and JPEG format is probably better for file size, to avoid the package getting huge.

gilbertohasnofb:
Closes #878 

Implementing @zakalawe new thumbnail and splash systems. Removing text and categorizing our previous splash screens.


gilbertohasnofb:
Please test this with 2017.2.0, as these features haven't been implemented as of 2017.1.2.

onox:
If I enable full screen, the images fill the screen, but they look rather low res. Would it be better if we generate new splash screens that are Full HD?

gilbertohasnofb:
I think these are already larger than @zakalawe had recommended, but personally I would much prefer larger images than blurry ones that occupies less disk space. But why are the splash screens being stretched and thus losing resolution? Is this an intended thing with the new system? These images are in the same resolution as they have always been, and in the older system they are not stretched to fill the screen and so they look quite fine to me. @onox could you please post a screen shot for me?

zakalawe:
Splash screens have always been stretched to fit the window size, but it's more obvious that this happens, with the new layout. 

My suggestion of 1024x768 for previews is it gives a good balance of detail when browsing the hangar, and reasonable quality when scaled, even full screen on my 30" panel. Going larger sizes is possible but if you do that for 8-12 preview images, it's going to be 75% of the total aircraft size, and I don't think it's justified

Of course going to say 1280x1024 and JPEG is another option. 

James

gilbertohasnofb:
> Splash screens have always been stretched to fit the window size

@zakalawe Are you sure about that? These screen shots below show what I have always seen. The splash screens were not stretched but rather the background was filled with solid black. If I compare the size of the screen shots of the aircraft with the splash screen files, they are exactly 1:1

screenshot from 2017-03-07 09 13 33

screenshot from 2017-03-07 09 13 52

@onox could you post some screenshots of what you see so I can have an idea of what has been changed?

> Of course going to say 1280x1024 and JPEG is another option.

Yes, JPEG is a good option, I can convert all splash screens once everything else is ready. Currently our png splash screens are ~5.5 Mb large in total, and in JPG this can be lowered to ~1 Mb.

zakalawe:
Yep I'm quite sure about the scaling, but there were some rules which made it less obvious before: the old splash had to be square, we positioned that square in the centre of the screen, and sized to 80% of the shorter of the two window dimensions. So unless you have a very tall window, it's common previously to have scaling down, not up.

onox:
This is how it looks?

fgfs-screen-028

What's the problem with 2017.2 you're having, @gilbertohasnofb?

onox:
If we drop some of the old splash screens (perhaps keep 1, 3, 5, 6, and 9?) then we can recreate them in Full HD. It's only a couple of MB in total then and not much compared to the total size of certain aircraft models (from can range from 50 MB to 1.6 GB).

onox:
So I would prefer a couple of high quality over many blurred images.

gilbertohasnofb:
@onox 

> What's the problem with 2017.2 you're having

I just don't know how to compile FG myself no matter how many times I read that wiki page (the script for Debian systems for some reason does not work here, maybe there are some unmet dependencies). I tried doing it a couple of times but I really don't have the patience any longer. So I use that PPA by Saiarcot and I tend to stick to stable releases.

> So I would prefer a couple of high quality over many blurred images.

Me too. And the good news is that 1920 x 1080 JPEG files are actually quite small, I tested it with 5 screen shots and they have in total 1.6 Mb. Would this image size work with different resolutions and aspect ratios, @zakalawe ?

> If we drop some of the old splash screens (perhaps keep 1, 3, 5, 6, and 9?) then we can recreate them in Full HD.

Sure we could do that. The problem is that most of those are exactly the ones I don't have the original screen shots any longer (some were not taken by me). Here is what I got in 1920 x 1080 out of all splash screens:

c172p-0

c172p-1

c172p-2

c172p-3

c172p-4

zakalawe:
1.6MB in total seems fine to me, in terms of previews I'd like a panel shot in there, and maybe one showing the various bits of ground-equipment which is a pretty cool feature? But those should be excluded as splash screens of course.

gilbertohasnofb:
@onox @zakalawe I will some new shots for the cockpit, ski, float and ground equipment, I will try to get this done still today.

gilbertohasnofb:
@onox Ok, so I have pushed some new commits. I have recreated a couple of shots and used those ones above as well. They are all in 1920 x 1080 and the total size is 3 MB, which for me looks quite fine (it's half of what we had before with our tiny png splashes). Could you please take a look how they look in the sim and give me some feedback if anything has to be changed? And if all is good, we could possibly merge this.

gilbertohasnofb:
Anyone?

wlbragg:
Sounds to me like your opinion is good ! :smile:

wlbragg:
I just looked at them and they look good to me. I really like the one with the city behind it.

PR #917, opened 2017-04-09
fixing oil temp and pressure gauges when complex procedures is false

### LINKED ISSUE #916: oil pressure and temperature gauges are not working
There has been a report of a possible bug with the oil pressure and temperature gauges [in the forum](https://forum.flightgear.org/viewtopic.php?f=66&t=31961):

> But I've had the T&P gauge at zero for quite a while. Is this a known problem? Could I have some odd setting which is doing this?

@wlbragg I can reproduce this issue: apparently, starting the sim with "complex engine procedures and failers" set to false causes the problem. See these two images, taken some 2 minutes after starting the engine (I restarted the sim in between them):

fgfs-screen-002

fgfs-screen-003

So yeah, there is a bug to be addressed, could you please inform that user?

Also, is it a normal behaviour that the oil pressure does not rise to the green arc if the throttle remains at idle after the engine startup? (see first screenshot above, this was taken minutes after the startup and I imagine that the engine should have stabilized by that time. Once more throttle is applied the pressure does come to the green arc and stabilizes there).

gilbertohasnofb:
Closes #916 

Now starting the sim with complex engine procedures set to `false` will not break the oil temp and pressure gauges. For testing purposes: if complex engine procedures is set to `true`, then toggle it to `false` and _restart  the sim_ to confirm everything works as fine, as the bug was present only when starting the sim with this property set to `false`)

### DIFF
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -133,10 +133,13 @@ var oil_consumption = maketimer(1.0, func {
     }
 
     else {
+        # if oil consumption is not allowed, the oil level is set to full and pressure and temp factors are set to 1.0
         if (getprop("/controls/engines/active-engine") == 0)
             setprop("/engines/active-engine/oil-level", 7);
         if (getprop("/controls/engines/active-engine") == 1)
             setprop("/engines/active-engine/oil-level", 8);
+        setprop("/engines/active-engine/low-oil-pressure-factor", 1.0);
+        setprop("/engines/active-engine/low-oil-temperature-factor", 1.0);
     }
 });
 


geoffmcl:
Applied the patch, and can **confirm** the temp and press needles now move ;=))

Even an excessive run-up, with full power and breaks, will quite quickly drive the temp off scale... but drops back into the green in normal flight...

Don't know how long they have been `dead`, and always meant to file an issue, but thanks for catching it...

gilbertohasnofb:
@geoffmcl thanks for testing it. 

> Even an excessive run-up, with full power and breaks, will quite quickly drive the temp off scale... but drops back into the green in normal flight...

Yes, the question is if that's the realistic behaviour or not. I suspect it is, but I am really not familiar with the expected behaviour of oil temp and pressure in relation to the engine RPM. Same thing for the other behaviour I described the issue #916 concerning the oil pressure which does not raise to the green arc region while on idle. Perhaps @dany93 would be the person to tell if these are bugs or not.

But anyway, this PR simply fix an oversight from my side which resulted in temp and pressure being set at `0.0`, so any tweaks to be done in the engine would be another story.

dany93:
> concerning the oil pressure which does not raise to the green arc region while on idle.

A bug ? Probably not.
But this oil pressure vs RPM is probably the standard behaviour in FG (unless it would have been changed by some nasal for the C172).
I don't know if the FG default behaviour has to be tweaked for the C172 at idle, I even don't know what the real behavior at idle is.

gilbertohasnofb:
> I even don't know what the real behavior at idle is.

Me neither, and this is what I meant by bug (sorry, bad terminology). I just wonder if the oil pressure does indeed stay at around 30 PSI at idle after startup even though the green arc in the gauge starts at around 50 PSI

dany93:
Difficult to get reliable information on this, but I think that you might be right. I'd rather think that the oil pressure should rise to about 50 PSI (start of the green arc) at idle (about 600 - 630 RPM, throttle at 0%). Same or a bit higher at start of the engine (825 RPM, with throttle at 20%).

Real C172 pilots can easily confirm or deny.

In any case, that's not a big issue.

gilbertohasnofb:
> I'd rather think that the oil pressure should rise to about 50 PSI (start of the green arc) at idle (about 600 - 630 RPM, throttle at 0%).

That would be my guess as well. In the checklist I have, it says that the oil pressure should be in the green arc in 30 seconds (or 60 seconds in cold weather).

> Real C172 pilots can easily confirm or deny.

@tigert @Juanvvc @stuartbuchanan would you know this?

> In any case, that's not a big issue.

I think it might be an issue, since reaching the green arc is part of the checklist.

* * *

@dany93 @wlbragg @onox By the way, this PR is ready to be merged

onox:
Let's wait to see what they have to say about the idle oil pressure and open another issue if it needs to be changed.

gilbertohasnofb:
:+1:

gilbertohasnofb:
@tigert @algefaen @stuartbuchanan @Juanvvc Does any of you know something about what has been discussed concerning the initial oil pressure after startup?

stuartbuchanan:
Hi Gilberto,

I don't know enough about the Cessna, but certainly on my aircraft the oil
pressure comes up within a couple of seconds when the engine starts,
and stays in the green arc irrespective of the throttle position.

On the ground, the oil pressure falling below the green arc would be a
reason to immediately stop the engine to avoid engine damage.

-Stuart

gilbertohasnofb:
@stuartbuchanan thanks for the input, Stuart. I am 99.9% sure this should be the case in the 172 as well. I will open an issue for this and if any of the 172 RL pilots has anything to add, we would be happy to hear it.

wlbragg:
Has anyone checked to make sure the number on the gauge match the actual property, may be a case of bad interpolation?

PR #924, opened 2017-05-21
wing damage conditions, minor changes

dany93:
I changed a few lines (logic criteria) in the wing damage conditions. Unless I didn't understand the initial logic behind, I thought that some conditions seemed inappropriate.

I think so but I hope that it works as expected, because checking the over-g values in flight is rather hard. Moreover there are many possibilities / combinations (velocities, g-overload, roll moment).

No fundamental change, only (I hope) a refinement.

### DIFF
--- a/Systems/c172p-damage.xml
+++ b/Systems/c172p-damage.xml
@@ -243,22 +243,20 @@
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-high
             </test>
-            
+
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-medium
-                damage/roll-moment GT 4000.0
-                wing-damage/right-wing LT 1.0
+                damage/roll-moment GT 12000.0
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-high
             </test>
-            
+
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-medium
-                damage/roll-moment GT 4000.0
-                wing-damage/right-wing LT 1.0
+                damage/roll-moment GT 8000.0
             </test>
 
             <!-- ================================================================== -->
@@ -268,16 +266,16 @@
             <!-- Overspeed -->
             <test logic="AND" value="0.3">
                 /velocities/airspeed-kt GT damage/limits-vne-medium
-                damage/roll-moment LT 4000.0
-                damage/roll-moment GT -4000.0
+                damage/roll-moment LT 12000.0
+                damage/roll-moment GT -12000.0
                 wing-damage/left-wing LT 0.3
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="0.3">
                 damage/force-aero-lbs GT damage/limits-lift-medium
-                damage/roll-moment LT 4000.0
-                damage/roll-moment GT -4000.0
+                damage/roll-moment LT 8000.0
+                damage/roll-moment GT -8000.0
                 wing-damage/left-wing LT 0.3
             </test>
 
@@ -288,14 +286,14 @@
             <!-- Overspeed -->
             <test logic="AND" value="0.12">
                 /velocities/airspeed-kt GT damage/limits-vne-low
-                damage/roll-moment GT 4000.0
+                damage/roll-moment GT 8000.0
                 wing-damage/left-wing LT 0.12
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="0.12">
                 damage/force-aero-lbs GT damage/limits-lift-low
-                damage/roll-moment GT 4000.0
+                damage/roll-moment GT 8000.0
                 wing-damage/left-wing LT 0.12
             </test>
         </switch>
@@ -326,22 +324,20 @@
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-high
             </test>
-            
+
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-medium
-                damage/roll-moment LT -4000.0
-                wing-damage/left-wing LT 1.0
+                damage/roll-moment LT -12000.0
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-high
             </test>
-            
+
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-medium
-                damage/roll-moment LT -4000.0
-                wing-damage/left-wing LT 1.0
+                damage/roll-moment LT -8000.0
             </test>
 
             <!-- ================================================================== -->
@@ -351,16 +347,16 @@
             <!-- Overspeed -->
             <test logic="AND" value="0.3">
                 /velocities/airspeed-kt GT damage/limits-vne-medium
-                damage/roll-moment LT 4000.0
-                damage/roll-moment GT -4000.0
+                damage/roll-moment LT 12000.0
+                damage/roll-moment GT -12000.0
                 wing-damage/right-wing LT 0.3
             </test>
             
             <!-- Over-g -->
             <test logic="AND" value="0.3">
                 damage/force-aero-lbs GT damage/limits-lift-medium
-                damage/roll-moment LT 4000.0
-                damage/roll-moment GT -4000.0
+                damage/roll-moment LT 8000.0
+                damage/roll-moment GT -8000.0
                 wing-damage/right-wing LT 0.3
             </test>
 
@@ -371,14 +367,14 @@
             <!-- Overspeed -->
             <test logic="AND" value="0.12">
                 /velocities/airspeed-kt GT damage/limits-vne-low
-                damage/roll-moment LT -4000.0
+                damage/roll-moment LT -8000.0
                 wing-damage/right-wing LT 0.12
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="0.12">
                 damage/force-aero-lbs GT damage/limits-lift-low
-                damage/roll-moment LT -4000.0
+                damage/roll-moment LT -8000.0
                 wing-damage/right-wing LT 0.12
             </test>
         </switch>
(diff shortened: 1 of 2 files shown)


Systems/c172p-damage.xml
```diff
@@ -288,20 +289,14 @@
             <test logic="AND" value="0.12">
                 /velocities/airspeed-kt GT damage/limits-vne-low
                 damage/roll-moment GT 4000.0
-                wing-damage/left-wing EQ 0.0
+                wing-damage/left-wing LT 0.12
```
onox: It would keep this at `EQ 0.0`. It means the left wing must be OK (not damaged or broken).

Systems/c172p-damage.xml
```diff
@@ -288,20 +289,14 @@
...
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="0.12">
-                damage/force-aero-lbs GT damage/limits-lift-medium
```
onox: Here you removed a whole block it seems. What the original code says is that the wing gets light damage if the force > low limit and you roll to the right very fast (`roll-moment GT 4000`). Or if the force > medium limit and and the roll moment is between -4000.0 and 4000.0.

So in other words, if the force > low limit, then you must roll very fast, but if force > medium limit, then you can get damage even if you roll slowly.

c172p-set.xml
```diff
@@ -881,9 +881,6 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 
         <vne>158</vne>
 
-        <!-- 2550*3.8 -->
-        <max-lift-force>9690</max-lift-force>
```
onox: This one can indeed be removed. It was used in older revisions of `Systems/c172p-damage.xml`.

Systems/c172p-damage.xml
```diff
@@ -243,38 +243,39 @@
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-high
             </test>
+            
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-medium
                 damage/roll-moment GT 4000.0
-                wing-damage/right-wing GT 0.0
+                wing-damage/right-wing LT 1.0
```
dany93: I see no reason to set a condition `wing-damage/right-wing GT 0.0` to the other wing. We can break the left one even if the right one is OK.

Farther...
Initially, I added `wing-damage/right-wing LT 1.0` to break the left wing. This was in order to avoid breaking the right wing as soon as the left one was broken (by a roll moment). But it seems useless, the right wing holds as long as the high velocity is not reached. I think that this condition can be deleted too.

Systems/c172p-damage.xml
```diff
@@ -243,38 +243,39 @@
...
             </test>
 
             <!-- Over-g -->
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-high
             </test>
+
```
dany93: It was to make it easier to read. In a first step, I didn't see the two `< test>` blocks.
But that's formal, not important.
Just tell me your preference.

Systems/c172p-damage.xml
```diff
@@ -243,38 +243,39 @@
...
             <!-- Over-g -->
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-high
             </test>
+            
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-medium
                 damage/roll-moment GT 4000.0
                 wing-damage/right-wing LT 1.0
-                wing-damage/right-wing GT 0.0
```
dany93: Same response as above for lines 249 - 250.

Systems/c172p-damage.xml
```diff
@@ -243,38 +243,39 @@
...
             </test>
 
             <!-- ================================================================== -->
             <!-- Heavy damage                                                       -->
             <!-- ================================================================== -->
 
-            <!-- Over-g -->
```
dany93: Right. There is no change, it was just to make it consistent with the other blocks, which are
`<!-- Overspeed -->`
then
`<!-- Over-g -->`
Purely formal too, not important. Tell me.

Systems/c172p-damage.xml
dany93: Both give the same effect.
Also to be consistent with the rest of the file where the equivalent conditions are 'if damage is lower than this value, increase it to this value".

Here too, not important

Systems/c172p-damage.xml
dany93: The block that I removed does the same effect as the `Over-g `block under `Heavy damage` (line 267), except for 0.12 instead of 0.3. My initial idea has been damage 0.12 in asymmetric conditions (break a single wing, the aircraft is still controllable in roll, although uneasy). And 0.3 for symmetric damage, the aircraft can still fly but with a serious loss of lift.

c172p-set.xml
dany93: No comment, I obviously agree.

dany93:
Maybe I should have added a comment here for you to get a mail notice.
The subjects on which we have to discuss:
```
             <test logic="AND" value="1.0">
                 /velocities/airspeed-kt GT damage/limits-vne-medium
                 damage/roll-moment GT 4000.0
-                wing-damage/right-wing GT 0.0
+                wing-damage/right-wing LT 1.0
```
```
             <test logic="AND" value="1.0">
                 damage/force-aero-lbs GT damage/limits-lift-medium
                 damage/roll-moment GT 4000.0
                 wing-damage/right-wing LT 1.0
-                wing-damage/right-wing GT 0.0
```
Plus possibly to remove _any_ condition above, even `LT 1.0`, on the other wing damage (still to test more, but I think).

This block (near your initial lines 296 - ...) that I removed
```
             <!-- Over-g -->
             <test logic="AND" value="0.12">
-                damage/force-aero-lbs GT damage/limits-lift-medium

```
The other ones are purely formal.

dany93:
Hi onox,

I've been waiting for your response for a week.
What should I do? Should I push a new commit but based _only_ on my opinion? (will include what is expressed above).

Or is it a sign that you completely reject this PR?

Systems/c172p-damage.xml
onox: > We can break the left one even if the right one is OK.

But now you have changed it to`wing-damage/right-wing LT 1.0`. So that means that when the right wing is broken, you cannot break the left wing anymore unless you go faster than `limits-vne-high`. Is that on purpose?

Systems/c172p-damage.xml
onox: If damage to the left wing should not depend on the amount of damage of the other wing, shouldn't the `wing-damage/right-wing LT 1.0` condition be removed as well then?

Systems/c172p-damage.xml
onox: These conditions exist in order to force the pilot to exceed `limits-lift-high` or `limits-vne-high`, so I think they should stay.

Systems/c172p-damage.xml
onox: > except for 0.12 instead of 0.3. And I think that 0.3 will have priority.

I see it now, the 0.3 block should indeed have priority over it.

Systems/c172p-damage.xml
onox: hmm.. these conditions might actually be redundant. But they're sort of a documentation aid, so that you do not have to take the other blocks above it into account when reasoning about them.

onox:
> I've been waiting for your response for a week.

Sorry, I'm very busy with various things last couple of weeks, so haven't found much time to pay attention to it.

> If I understand well, this it is the only proposed change that has an effect on the behavior. Break a single wing above limits-vne-medium or limits-lift-medium, due to a strong roll moment. The question is: conditioned or not by the state of the other wing ?

As far as I remember, the idea was that if you exceed `limits-vne-high` a wing always breaks, but if you go just faster than `limits-vne-medium` than you need to have a strong roll moment plus an already damaged/broken other wing. Otherwise (if the other wing is still OK) a wing could become broken too easily.

> Or is it a sign that you completely reject this PR?

I don't outright reject it, it's just that the description of the PR suggests there are no functional changes (read: does the other wing has an influence on a wing), but this discussion shows there are.

onox:
Overall, I can live with the changes you have applied to the "Heavy damage" and "Light damage" sections, but not with the changes to the "Broken" section. My concern is that it shouldn't be too easy to break a wing.

Systems/c172p-damage.xml
dany93: Yes, that's on purpose. Rather, that _was_. Initially.
`This was in order to avoid breaking the right wing as soon as the left one was broken (e.g. by a roll moment)`

`But, from recent tests, I realize that it seems useless, the right wing holds as long as the `limits-vne-high` is not reached. I think that this condition can be deleted too.`
Thus I've removed it.

Systems/c172p-damage.xml
dany93: Corrected (If I understand well what "trailing spaces" are).
Removed, they were not on purpose. I hadn't seen them.

Systems/c172p-damage.xml
dany93: `wing-damage/right-wing LT 1.0` condition has been removed, as written above

Systems/c172p-damage.xml
dany93: Restored (it necessitates to be careful when changing the values, to preserve consistency)

dany93:
To make breaking a single wing by a strong roll moment more difficult:
Rather than a condition on the other wing, I've increased the roll-moment threshold (12000 for over-speed, 8000 for over-g, instead of 4000).

Also, as I found that _damaging_ a single wing was too easy, particularly at over-speed. Thus, I also increased the roll-moment threshold for these.

And please take your time if you have something else to do, at present I see why.
I have currently some problems too, which means I'm never sure when and if I will have time and availability

wlbragg:
@onox, I'd like to see this change implemented as I too have thought the wing damage and break was to easy. Any chance we're good to go here?

dany93:
Hi @onox, thank you.

onox:
Sorry it took such a long time, I only do aerial refueling in FG nowadays.

dany93:
No problem now that's OK. You're probably right to do this before the fuel price increases.

onox:
I'm actually flying the tanker: https://www.youtube.com/watch?v=q06wD6PgniE&t=886

PR #932, opened 2017-08-28
Bug 789 - Create a save mission state that includes position, orientation and heading 

### LINKED ISSUE #789: Create startup states and save mission state
> Thanks,
> Jay

wlbragg:
This is so simple to do. I can't believe it took this long for us to attempt to implement. It should fit nicely with the "save state of aircraft between sessions".

I am also implementing this on the AirCrane and the J3Cub.

Fixes #789

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -376,6 +376,7 @@ var thunder = func (name) {
     }, delay_seconds);
 };
 
+var persistent = getprop("/sim/model/c172p/persistent");
 var reset_system = func {
     if (getprop("/fdm/jsbsim/running")) {
         c172p.autostart(0);
@@ -393,6 +394,13 @@ var reset_system = func {
     props.globals.getNode("/fdm/jsbsim/pontoon-damage/right-pontoon", 0).setIntValue(0);
 
     setprop("/engines/active-engine/kill-engine", 0);
+
+    if (persistent) {
+        setprop("position/latitude-deg", getprop("/sim/model/c172p/currentlat"));
+        setprop("position/longitude-deg", getprop("/sim/model/c172p/currentlon"));
+        setprop("position/altitude-ft", getprop("/sim/model/c172p/currentalt"));
+        setprop("orientation/heading-deg", getprop("/sim/model/c172p/currenthead"));
+    }
 }
 
 ############################################
@@ -470,6 +478,14 @@ ad_timer.start();
 ############################################
 var global_system_loop = func {
     c172p.physics_loop();
+
+    persistent = getprop("/sim/model/c172p/persistent");
+    if (persistent) {
+        setprop("/sim/model/c172p/currentlat", getprop("/position/latitude-deg"));
+        setprop("/sim/model/c172p/currentlon", getprop("/position/longitude-deg"));
+        setprop("/sim/model/c172p/currentalt", getprop("/position/altitude-ft"));
+        setprop("/sim/model/c172p/currenthead", getprop("/orientation/heading-deg"));
+    }
 }
 
 ##########################################
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -399,6 +399,10 @@ var reset_system = func {
         setprop("position/longitude-deg", getprop("/sim/model/c172p/currentlon"));
         setprop("position/altitude-ft", getprop("/sim/model/c172p/currentalt"));
         setprop("orientation/heading-deg", getprop("/sim/model/c172p/currenthead"));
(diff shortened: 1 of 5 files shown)


wlbragg:
This is probably not ready to merge. I think it needs to be thoroughly tested first. I added it to the Cub which should help expose any unforeseen bugs over time of which I can respond to quickly if need be.
I'm currently trying to test improper game exits, game crashes, etc. to find any potential unforeseen issues.

gilbertohasnofb:
> This is probably not ready to merge. I think it needs to be thoroughly tested first.

How exactly should this be tested?

wlbragg:
I guess just use it and see if you can break it. I added some conditions to the saving of the location so it only saves your position if you are on the ground and close to stopped (and of course have the "Persistent position" check-box checked".
I did this so it will not save if your crash in flight or landing somewhere. Also it shouldn't compete with the saved or startup states work done prior. Like if you are using a "startup state" like approach and also have the "Persistent position" checked, it should ignore the persistent position and still start up on approach. I have actually tested that yet, but it should work that way.
I considered having to click a "Save" button each time you wanted to save it at a location but opted for this simpler approach.
I think there are a few ways you could do this...

1) the way it is now, if "Persistent position" is checked it will save and load that position if all conditions are meant.

2) you could have a save and load dialog requiring an action to both save or load a position.

3) you could have both methods with the "Persistent position" checked overriding the save/load GUI.

4) you could have a save and load dialog requiring an action to both save or load a position. It could have multiple save/load positions that you could choose from.

5) you could have a check for a saved position on startup and ask if you want to activate it.

I opted for the way it is because it was simple and effective with a low overhead.

wlbragg:
>Like if you are using a "startup state" like approach and also have the "Persistent position" checked, it should ignore the persistent position and still start up on approach. 

Now that I think about it, it may not ignore "Persistent position" when starting a save state that starts the aircraft stopped and on the ground. I'll test that now.

wlbragg:
This isn't playing nicely with GUI-Location/Select Airport. It needs a check for this condition/action if possible.

gilbertohasnofb:
@wlbragg sorry for my late reply. I will try to test this and report back here as soon as I can, but I will be very busy during these next weeks and I might not have a chance to do that soon.

wlbragg:
I can't figure out a way to catch the GUI-Location/Select Airport action so as to stop the repositioning on load if "Persistent Position" is checked. (GUI-Location/Select Airport uses an fgcommand("reposition") to move the aircraft to a new airport location. I may not be able to "automatically" have the aircraft start at the saved or "persistent" location. Instead we may have to have a Save/Load feature.

wlbragg:
I'm going to talk about this on the dev list and see if we don't want to make this a system wide feature in the Location Dialog.

onox:
I like the feature, but I managed to break it when I tried to spawn at a different airport (stored at KLSV, loaded at FHAW).

Nasal/c172p.nas
```diff
@@ -393,6 +393,17 @@ var reset_system = func {
     props.globals.getNode("/fdm/jsbsim/pontoon-damage/right-pontoon", 0).setIntValue(0);
 
     setprop("/engines/active-engine/kill-engine", 0);
+
+    if (getprop("/fdm/jsbsim/position/persistent")) {
+        setprop("position/latitude-deg", getprop("/sim/model/c172p/currentlat"));
```
onox: You have a lot of properties in the same directory that are related to each other (`/sim/model/c172p/current*`). Maybe you could group them together and move them to a separate directory like `/sim/model/c172p/current-position/*`)?

onox:
Maybe store the closest airport and check if the current airport (when spawning) is the same as the stored one?

onox:
(or measure the distance between the spawned position and the stored position)

wlbragg:
> I managed to break it when I tried to spawn at a different airport (stored at KLSV, loaded at FHAW

I have found it can break with two other actions.
1) Go to airport using GUI
2) Saving the position over water. I can't figure out why when saved over water it always drops the aircraft from about 10-20 feet and crashes, even though I save the altitude or make it 0.
I'll try breaking it the way you did also and see what that is all about.

>Maybe you could group them together and move them to a separate directory 
Will do.

I talked to James on the mailing list and he would like this done using the reposition cmd/code. Potentially that might fix all the bugs it has now. I'll keep working that angle and see where it goes.

wlbragg:
Thanks for looking at it and the comments.

wlbragg:
This needs to be put on hold for now and put to an issue for work on at a later date.

gilbertohasnofb:
@wlbragg sorry I have not tested this yet, I have been just so busy lately... once you solve this issue and make a new PR I will be happy to test it!

Issue #931, opened 2017-08-28
C172P JSBSim model fcs/left-aileron-pos-rad


okroth:
I went through the JSBSim parameters of the C172P model in FGFS to insert some force-feedback formulas (hinge moment * gear factor) and found a peculiarity:
The roll moment, side force, and yaw moment from aileron deflection use the left aileron deflection (in rad) only.
This will introduce an error, as the aileron is deflected downwards to -20 degrees, but only 15 degrees upwards. This is true in the real C172 the same way, to reduce adverse yaw.

Now, as only the left ailerons deflection is used in the formulas, turning the yoke clockwise (for a right turn) is 33% more effective then turning it counter-clockwise (for a left turn).
Possibly, one should not use the actual deflection of a single aileron, but the fcs/roll-trim-sum (computed as sum of fcs/aileron-cmd-norm and fcs/roll-trim-cmd-norm), of course with adjusted factors?

dany93:
You are right. I think that it is a historical flaw from Aeromatic (initially assuming that aileron deflection are equal for up and down ).
I'm going to see what can be done. Rather than taking some  roll-trim-cmd-norm, I'd prefer attempting to implement an average angle value for right and left, in order to keep the linear aerodynamic coefficients consistency (conventionally per radian if I understand well). Possibly with the same factor.
Looking at it, I see that there are other mistakes or weird code in the aileron deflection limits.

The ailerons are rarely pushed at full deflection, which may explain why this is not sensed at piloting.

okroth:
Hi,


Am 28.08.2017 um 10:39 schrieb dany93:
>
> You are right. I think that it is a historical flaw from Aeromatic 
> (initially assuming that aileron deflection are equal for up and down ).
> I'm going to see what can be done. Rather than taking some 
> roll-trim-cmd-norm, I'd prefer attempting to implement an average 
> angle value for right and left, in order to keep the linear 
> aerodynamic coefficients consistency (conventionally per radian if I 
> understand well). Possibly with the same factor.
> Looking at it, I see that there are other mistakes or weird code in 
> the aileron deflection limits.
>
> The ailerons are rarely pushed at full deflection, which may explain 
> why this is not sensed at piloting.
>



I found some more oddities with the rudder:
the side force from rudder (CYdr) is only 38.1% of the 'classic' 172P, but
the yaw moment from rudder (Cndr) is 162.8%, while
the roll moment from rudder (Cldr) is same on both.

[...]

PR #933, opened 2017-08-29
FDM: Fix L and R aileron deflection angles, create averaged property

dany93:
Fix for [JSBSim model fcs/left-aileron-pos-rad #931](https://github.com/c172p-team/c172p-detailed/issues/931).

Two issues:

1 - Aileron full deflection angles, should be: up = -15 deg, down = +20 deg.
Initially, the deflection angles were, with stick full left or right:
stick full left: left / right aileron -20 deg, +20 deg,
stick full right: left / right aileron +15 deg, -15 deg.

2 - The FDM (roll moment due to ailerons) was driven by the property `fcs/left-aileron-pos-rad`
Which this commit changes for `fcs/aileron-pos-rad-avgd` (averaged left and -right aileron deflections).
A function calculates `aileron-pos-rad-avgd = (left - right) / 2` (notice the minus sign).
With the proposed changes implemented, all that one has to do is to use the `fcs/aileron-pos-rad-avgd` property when needed, instead of `fcs/left-aileron-pos-rad`.
Of course, after having set the correct deflection angles for the aircraft (a bit tricky).

_(sorry for the noise in changes, I've also removed trailing spaces. I should have done it separately)_

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -750,8 +750,8 @@
                 <input>fcs/roll-trim-sum</input>
                 <gain>0.01745</gain>
                 <range>
-                    <min>-15</min>
-                    <max>20</max>
+                    <min>-20</min>
+                    <max>15</max>
                 </range>
                 <output>fcs/left-aileron-pos-rad</output>
             </aerosurface_scale>
@@ -759,8 +759,8 @@
             <aerosurface_scale name="Left Aileron Position Normalized">
                 <input>fcs/left-aileron-pos-deg</input>
                 <domain>
-                    <min>-15</min>
-                    <max>20</max>
+                    <min>-20</min>
+                    <max>15</max>
                 </domain>
                 <range>
                     <min>-1</min>
@@ -773,8 +773,8 @@
                 <input>-fcs/roll-trim-sum</input>
                 <gain>0.01745</gain>
                 <range>
-                    <min>-15</min>
-                    <max>20</max>
+                    <min>-20</min>
+                    <max>15</max>
                 </range>
                 <output>fcs/right-aileron-pos-rad</output>
             </aerosurface_scale>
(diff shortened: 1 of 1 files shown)


dany93:
I have a doubt. Possible mistake by exchanging the max values for up and down deflection angles.
I have to look at it.

dany93:
Fix full angle up and down: done, pushed.

@wlbragg 

I think that you have made the same mistake as me in Models/c172p.xml (lines 3587....).
Must be:
full up -20 deg
full down +15 deg
for each aileron.

or maybe use (?)
`fcs/left-aileron-pos-deg`
`fcs/right-aileron-pos-deg`

If you don't mind, I'd prefer that you possibly fix it yourself, I'm not sure to see everything in your damage model.

dany93:
@okroth,
@gilbertohasnofb or @onox will probably ask for your opinion before possibly merging this commit.

c172p.xml
```diff
@@ -1099,6 +1099,17 @@
             </product>
         </function>
 
+        <function name="fcs/aileron-pos-rad-avgd">
+            <description>Averaged aileron position</description>
+            <product>
```
onox: There's a `<avg>` element that you can use instead. See section 2.5.1 of the JSBSim Reference Manual.

onox:
@dany93 Please rename the branch to `bug-931` (for consistency) or `931-aileron-pos-averaged`. That makes it easier to see (on GH as well as in `git log`) which issue is being addressed by the branch.

Use: `git branch -m bug-931` in your `aileron-pos-averaged` branch.

c172p.xml
dany93: Done. (lines 1102 -1111)
Is there a more elegant way to do it? With this writing (because of the minus sign), is has the same number of lines as my initial code...

dany93:
Done, branch number included.
But if I do a Pull Request by myself (no bug issue number, not still), how should I do at creating my branch name? I have a Pull Request number just after...

dany93:
it seems that deleting the aileron-pos-averaged branch closed this Pull Request. How can I reopen it?

onox:
It seems that's not possible, PR is tied to the branch.

PR #934, opened 2017-09-02
No lighting on pontoon wake effect. Eliminate wingtip and tail sparks…

wlbragg:
I recently discovered this bug on the wingtip sparks. It makes no sense to me that it sparked when the condition for sparks was a product of contact[]/wow. When over water contact[0]/wow is always = to 0. A product with 0 in the equation should always equal 0. But none the less, we had sparks on the water. By adding fdm/jsbsim/position/solid to the product, it eliminated the sparks when over water but not when over land.

The lighting in all pray spray effects should be false.

### DIFF
--- a/Models/Effects/pontoon/left-wake.xml
+++ b/Models/Effects/pontoon/left-wake.xml
@@ -15,7 +15,7 @@
         <name>right-wake</name>
         <texture>wake.png</texture>
         <emissive>false</emissive>
-        <lighting>true</lighting>
+        <lighting>false</lighting>
 
         <condition>
             <and>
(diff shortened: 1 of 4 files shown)


onox:
These are two issues? Can you split them into two commits? Or do you want me to do that for you?

wlbragg:
>do you want me to do that for you?

If you can easily enough, yes please do. Otherwise maybe you can tell me how to back this out, what just cancel it and re-request the PR but with two commit this time?

onox:
I did:

```
git checkout Groundeffectbugs
git rebase -i HEAD~1 (change "pick" to "edit" and then save and exit the editor)
edited Systems/c172p-ground-effects.xml and removed the solid properties
git add Systems/c172p-ground-effects.xml
git commit --amend
edited Systems/c172p-ground-effects.xml and added the solid properties again
git add Systems/c172p-ground-effects.xml
git commit --author="wlbragg <your email address>"
git rebase --continue
git push --force
```

PR #935, opened 2017-09-10
931 aileron pos averaged

dany93:
Deleting the  'aileron pos averaged' branch (to modify the branch name) closed the #933 pull request.

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -1101,13 +1101,13 @@
 
         <function name="fcs/aileron-pos-rad-avgd">
             <description>Averaged aileron position</description>
-            <product>
-                <difference>
-                    <property>fcs/left-aileron-pos-rad</property>
+            <avg>
+                <property>fcs/left-aileron-pos-rad</property>
+                <product>
                     <property>fcs/right-aileron-pos-rad</property>
-                </difference>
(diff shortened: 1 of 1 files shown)


c172p.xml
```diff
@@ -1099,6 +1099,17 @@
             </product>
         </function>
 
+        <function name="fcs/aileron-pos-rad-avgd">
+            <description>Averaged aileron position</description>
+            <avg>
+                <property>fcs/left-aileron-pos-rad</property>
+                <product>
+                    <property>fcs/right-aileron-pos-rad</property>
```
onox: Did you try `<property>-fcs/right-aileron-pos-rad</property>` (without any spaces between - and the property)?

dany93: Yes.
It gives 

> Fatal exception, '-' found in propertyname after 'jsbsim' name must begin with alpha Il or '_'

onox: It seems the minus sign is only supported for `<input>` in the summer, pure gain, etc. components. I'll merge the PR.

dany93: I seemed to remember having seen this somewhere, but I couldn't find where.

dany93:
Thank you, @onox .

Issue #947, opened 2018-01-27
Spiraling propwash effect


dany93:
I wondered several times by observing the behavior, but now I'm almost sure that the spiraling propwash effect is not simulated by default in JSBSim (differently from the P-factor). Moreover, it is not included in Aeromatic. Thus, almost every aircraft is lacking it.
http://design-plane.blogspot.fr/2008/11/corkscrew-effect.html
This is the effect which makes the aircraft steer leftwards (yaw) at full thrust, low speed. Strongest when rolling on the runway, accelerating before takeoff.
This often disconcerts beginners in simulation. However, it is always present on single-propeller aircraft, more or less strong. From my IRL experience, strong enough to quickly take you out of of the runway if you don't correct it with the rudder pedals.

I think it is rather gentle in my proposal, compared with my observations (from memory) in RL ultralights. However, I do not know for the C172P. If C172p-like RL pilots could give their opinion, thank you. @tigert @Juanvvc @stuartbuchanan ?
Anyway, changing it later is easy.

My opinion is that merely feeling it before takeoff adds some realism.

tigert:
Yeah, right rudder is needed. Also it is present on climbing and turning to left crosswind leg (low speed, climbing turn to left) where you need no left rudder at all — or actually you just slightly ease the right rudder a to keep the ball centered.

tigert:
The rudder has a groumd-adjustable trim tab to set the cruise speed+power with pedals centered, ball centered. Low speed and high power setting does indeed require right rudder. It's not like it will fall out in a spin, the C172 is very forgiving, but you can see it from the slip ball and feel it in your butt.

dany93:
Thank @tigert for this accurate description of the behavior.
All of it looks like that I've seen on ultralights.

> The rudder has a groumd-adjustable trim tab to set the cruise speed+power with pedals centered, ball centered

On ultralights, too.
Thank you for this, too.
I think I may add a trim-preset for the rudder.

Now, the question is 'how much effect', to be realistic enough and not discouraging the beginners. And for not being too harsh for those (many) who have no rudder pedals...

tigert:
I think the answer is not very simple, because in addition to actually having proper flight controls in tye sim, also the actual travel movement can be different. Most joysticks have rather short movement so it is not too easy to fly accurately, and thus a small stick movement can be big in the simulated flight model. 

[...]

PR #948, opened 2018-01-27
add spiraling propwash effect (new)

dany93:
see issue #947

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -494,8 +494,8 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 
     <controls>
         <flight>
-            <aileron-trim type="double">0.017</aileron-trim>
-            <rudder-trim type="double">0.009</rudder-trim>
+            <aileron-trim type="double">0.02</aileron-trim>
+            <rudder-trim type="double">0.013</rudder-trim>
 
             <freeze-yoke type="bool">false</freeze-yoke>
             <aileron-cmd type="double">0.0</aileron-cmd>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -494,8 +494,8 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 
     <controls>
         <flight>
-            <aileron-trim type="double">0.027</aileron-trim>
-            <rudder-trim type="double">0.0</rudder-trim>
+            <aileron-trim type="double">0.017</aileron-trim>
+            <rudder-trim type="double">0.009</rudder-trim>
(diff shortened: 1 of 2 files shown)


onox:
I have limited knowledge of the FDM, but it seems to work and the code looks fine. Can be merged if no one has any objections.

dany93:
Thanks onox for watching. 
I have prepared a small variant, with:
- a bit more propwash effect,
- and a rudder preset-trim to have the slip-skid-ball centered in cruise.
If you can wait a bit....

dany93:
After reading @tigert comments,
- Increased a bit the spiral propwash effect,
- Added a rudder trim preset, to have the slip-skid-ball centered at cruise.

Your opinions on the propwash effect? The quantity is `Cnspw` <value> (line 1667), 0.01 (weak, previous value) or 0.02 (current proposal). Easy to change later.

HHS81:
@dany93 
Tried it.

I have no reallife experience, just know X-Plane's C172.
In X-Plane the effect is much stronger, especially at very low speeds. When I begin to acclerate, I need half of right rudder, the more speed I gain the less I need. 

Here in FGFS there is a little need of right rudder, but very minor, and not at low speeds.

dany93:
Thank you @HHS81.

This is also my feeling, but I hesitate to set it stronger, because of beginners.
However, at very low speed (almost stopped without brakes) and still at idle, it already starts steering leftwards.

From my memory in URL Ultralights, it is stronger too (I cannot give a value).

HHS81:
@dany93 
I think especially for beginners a realistic, and with that, strong effect should be set. So they can learn, the c172 is still very forgivable.
I'm thinking to implement the propwash to the C182S/T as well.

dany93:
I'm ready to set it stronger.  If others agree...

gilbertohasnofb:
> I'm ready to set it stronger. If others agree...

I do. I am all for trying to simulate the real aircraft as much as we possibly can.

dany93:
Spiral propwash effect made stronger (Cnspw value = -0.05).
Above this strength (e.g. -0.1), the floats and amphibious versions become uncontrollable on the water or on the ground at very low speed. If we wish to make the spiral propwash such stronger, the Cnspw value will have to be conditioned, weaker for these versions. However, that's only a circumventing.
Personally (but that's pure guess) I think that this value (-0.05) can be applied and submitted to practice and observation.

onox:
The changes are OK for me. I need to use more rudder, but I can takeoff and land.

dany93:
Thank you, onox.

PR #951, opened 2018-01-30
Fix visible yellow sparks while not moving if landed on a moving object

### LINKED ISSUE #950: Ground effects should depend on rollspeed-ms of gear, not the groundspeed
If you try to land on a moving carrier and damage your gear, the ground effects remain visible even after you come to a standstill on the flight deck (because the carrier keeps moving).

onox:
Fixes #950

### DIFF
--- a/Systems/c172p-ground-effects.xml
+++ b/Systems/c172p-ground-effects.xml
@@ -396,7 +396,8 @@
             <function>
                 <product>
                     <property>/fdm/jsbsim/contact/unit[6]/WOW</property>
-                    <property>/velocities/groundspeed-kt</property>
+                    <property>/gear/gear[0]/rollspeed-ms</property>
+                    <value>1.94384449</value><!-- m/s to knots -->
                 </product>
             </function>
         </fcs_function>
@@ -405,7 +406,8 @@
             <function>
                 <product>
                     <property>/fdm/jsbsim/contact/unit[7]/WOW</property>
-                    <property>/velocities/groundspeed-kt</property>
+                    <property>/gear/gear[1]/rollspeed-ms</property>
+                    <value>1.94384449</value><!-- m/s to knots -->
                 </product>
             </function>
         </fcs_function>
@@ -414,7 +416,8 @@
             <function>
                 <product>
                     <property>/fdm/jsbsim/contact/unit[8]/WOW</property>
-                    <property>/velocities/groundspeed-kt</property>
+                    <property>/gear/gear[2]/rollspeed-ms</property>
+                    <value>1.94384449</value><!-- m/s to knots -->
                 </product>
             </function>
         </fcs_function>
--- a/c172-sound.xml
+++ b/c172-sound.xml
@@ -157,6 +157,20 @@
                     <property>gear/gear[1]/wow</property>
                     <property>gear/gear[2]/wow</property>
                 </or>
+                <or>
+                    <greater-than>
+                      <property>/gear/gear[0]/rollspeed-ms</property>
+                      <value>5</value>
+                    </greater-than>
+                    <greater-than>
+                      <property>/gear/gear[1]/rollspeed-ms</property>
+                      <value>5</value>
+                    </greater-than>
+                    <greater-than>
+                      <property>/gear/gear[2]/rollspeed-ms</property>
+                      <value>5</value>
+                    </greater-than>
+                </or>
             </condition>
             <position>
                 <x>0.5</x>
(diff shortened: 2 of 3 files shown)


wlbragg:
I'll review this tomorrow (later today actually).

dany93:
I'm not sure that I've understood well the issue and the way to test it.
What are 'visible ground effects'?

On the Nimitz, if I break a gear (with "Damage" checked on, of course), the "Repair" button is still greyed, I cannot have it repaired (unless by unchecking "damage", which is not the right way).
`/gear/gear[0 to 2]/rollspeed-ms` = 0 ([0], broken) to 3E-3 or 4E-3,
Ground speed kt about 10

Was it the issue? Is it a correct test?

dany93:
I don't know what is wrong, what to look at.

What I have done:
- on Nimitz, break the right gear (just by taking hard turns).
- try "Repair": the button is grey. Ineffective.

Properties:
```
gear/rollspeed-ms = 6.95e-06
gear[1]/rollspeed-ms = 4.87e-06
gear[2]/rollspeed-ms = 0
unit[6]/WOW = false
unit[7]/WOW = false
unit[8]/WOW = true
groundspeed-kt = 9.99997
```

dany93:
Not a carrier or moving object issue!!!
I can get that in a runway (KSFO).

dany93:
OK for the "visible ground effects". I realized, only in the meantime, that it might be that. Sorry.
In this case, the "visible ground effects" issue on a moving object (carrier) is solved. In my test I only see smoke (not sparks) at the gear in the version before this fix, but now I understand that it was part of the issue.
Solved, no gear smoke when on the carrier with this fix.

OK too for the "Repair" conditions, it works if they are fulfilled (`/engines/active-engine/running = false`   is logical but not trivial).
.

dany93:
Is there any reason for not merging this pull request ? @wlbragg ?

wlbragg:
Sorry, looking at it now!

wlbragg:
OK, sparks are gone but the scrape sound is still active.

wlbragg:
c172-sound.xml is only using velocities/groundspeed-kt. I don't know if you want to do this in the sound.xml and use rollspeed or change the XML to use a function derivative  instead.

wlbragg:
Also any model part that can cause spark needs to be tied to the function as I just put it upside down on the carrier and the wings are sparking while not moving.

wlbragg:
Looking at the wings sound and damage code, it may not be as easy as there is no rollspeed to use. We're  going to have to come up with something else.

wlbragg:
Hum, while we're at it we need to check to see if we are "in" water and not allow any sound or spark in that crash condition.

I changed rumble and rumble-scratch to the following which fixes the sound on a crashed gear landing on the carrier.

~~~xml
<rumble>
            <name>rumble</name>
            <mode>looped</mode>
            <path>Sounds/rumble.wav</path>
            <condition>
                <or>
                    <property>gear/gear[0]/wow</property>
                    <property>gear/gear[1]/wow</property>
                    <property>gear/gear[2]/wow</property>
                </or>
                <or>
                    <greater-than>
                      <property>/gear/gear[0]/rollspeed-ms</property>
                      <value>5</value>
                    </greater-than>
                    <greater-than>
                      <property>/gear/gear[1]/rollspeed-ms</property>
                      <value>5</value>
                    </greater-than>
                    <greater-than>
                      <property>/gear/gear[2]/rollspeed-ms</property>
                      <value>5</value>
                    </greater-than>
                </or>
            </condition>
            <position>
                <x>0.5</x>
                <y>0.0</y>
                <z>-1.15</z>
            </position>
            <volume>
                <type>log</type>
                <factor>0.25</factor>
            </volume>
            <volume>
                <property>/sim/model/c172p/sound/volume-boost-doors</property>
            </volume>
            <reference-dist>2.5</reference-dist>
            <max-dist>200.0</max-dist>
        </rumble>

        <rumble-scratch>
            <name>rumble-scratch</name>
            <mode>looped</mode>
            <path>Sounds/scratch.wav</path>
            <condition>
                <and>
                    <or>
                        <!-- Broken gear touching the ground -->
                        <property>/fdm/jsbsim/contact/unit[6]/WOW</property>
                        <property>/fdm/jsbsim/contact/unit[7]/WOW</property>
                        <property>/fdm/jsbsim/contact/unit[8]/WOW</property>

                        <!-- Tail strike -->
                        <property>/fdm/jsbsim/contact/unit[3]/WOW</property>
                    </or>
                    <or>
                        <greater-than>
                            <property>/gear/gear[0]/rollspeed-ms</property>
                            <value>5</value>
                        </greater-than>
                        <greater-than>
                            <property>/gear/gear[1]/rollspeed-ms</property>
                            <value>5</value>
                        </greater-than>
                        <greater-than>
                            <property>/gear/gear[2]/rollspeed-ms</property>
                            <value>5</value>
                        </greater-than>
                    </or>
                </and>
            </condition>
            <position>
                <x>0.5</x>
                <y>0.0</y>
                <z>-1.15</z>
            </position>
            <volume>
                <type>log</type>
                <factor>0.25</factor>
            </volume>
            <volume>
                <property>/sim/model/c172p/sound/volume-boost-doors</property>
            </volume>
            <reference-dist>2.5</reference-dist>
            <max-dist>200.0</max-dist>
        </rumble-scratch>
~~~

dany93:
I'm wondering... The solid, moving object reference exists, it is already used in `/gear/gear[N]/rollspeed-ms`.  Would using it for creating some `/gear/gear[N]/scrapespeed-ms` property for the structure contacts be a big issue for the developers? Like Erik kindly did for the oil pressure vs engine RPM.

wlbragg:
>The solid, moving object reference exists, it is already used
>Would using it for creating some [snip] property for the structure contacts be a big issue for the developers?

We can ask, it would be a useful exposure to have. Because I really don't see another way other than polling the position of the airframe or structure over time to determine if it moved (ugly), and that still wouldn't solve for the landing surface moving.

onox:
Good idea to ask, but it doesn't seem like something that will be fixed at short notice. Can we merge this PR or are there still things that need and can be fixed?

wlbragg:
Yeah, I'm good with it being merged. Just don't flip your Cessna upside-down when trying to land on a Carrier.:)

PR #954, opened 2018-02-09
Enable walker views if and only if walker is switched on

### LINKED ISSUE #78: Walker views show the ocean if walker is not switched on in the help menu

onox:
Fixes #78


gilbertohasnofb:
@onox 👍

PR #957, opened 2018-02-10
New gyro sound

### LINKED ISSUE #846: Interior sounds
Some Cessna aircraft have sounds in the cockpit which we currently don't have, but I am not sure those apply to our models. For instance, I was looking for Youtube videos of the start up procedures of the 172 and 182 models and I have noticed that toggling the avionic switch causes a constant high pitched sound to be present in the cockpit (it never goes away). I have synthesized this sound for the c182s project, here is the result: https://github.com/HHS81/c182s/blob/Issue-151/Sounds/avionics.wav

Does anyone here knows if this sound apply to our plane? @Juanvvc @tigert ?

And does anyone know which other sounds are still missing so I can work on them? I am discussing the same thing concerning the 182 with Heiko in this thread here: https://github.com/HHS81/c182s/issues/151

gilbertohasnofb:
Closes #846 

To test: turn on the avionics switch with the battery on. See if the logic is reasonable, test if the sound volume is okay (in comparison with the engine, for instance), see if the sound position is reasonable (I positioned it somewhere in the panel).


gilbertohasnofb:
@onox I will be out during the whole day tomorrow, so if you spot any needs of code modification then feel free to simply change it yourself, else we might miss the release deadline.

onox:
I've increased the volume because I couldn't hear it at all. I'll merge this and the other PR's in the evening if no one has any objections.

gilbertohasnofb:
@onox thanks for taking a look at this

> I've increased the volume because I couldn't hear it at all.

I just tested it and I think it is much too loud right now. In the video posted by @algefaen the gyro sound is nearly imperceptible once the engine is on at idle, so I chose that value so that the gyro can be listened when the engine is off but nearly imperceptible after starting it. I am using headphones on my test and the original gyro volume was certainly clear to me, albeit soft. Do you think you could check this again?

gilbertohasnofb:
This is his video, by the way: https://www.youtube.com/watch?v=gg3q002qsyI

gilbertohasnofb:
@onox another idea would be to link the engine rpm to the gyro volume, so that the higher the RPM the lower the gyro would sound. This would give the desired effect and better simulate what we hear in RL.

Unfortunately I must leave now and I won't be back home until the evening 😞 So if you want to change something please go ahead.

onox:
In @algefaen's video the sound is indeed imperceptible when the engine is running. I think the engine sound of our c172p is maybe too soft at idle.

gilbertohasnofb:
> I think the engine sound of our c172p is maybe too soft at idle.

I had the same feeling while testing this. I will open an issue for that and investigate later.

Issue #860, opened 2016-11-14
small 3D issues with the external fuselage


gilbertohasnofb:
These are a couple of minor 3D issues that we currently have with our external fuselage:

- While testing the tutorials, the camera angle for removing the tie-down in the tail of the plane on the preflight tutorial shows a possible problem with the 3D model, see:

![fgfs-screen-001](https://cloud.githubusercontent.com/assets/5700795/20261406/e8d43086-aa5d-11e6-94ce-d644011e802b.png)

On the left side of the purple circle we can see through the fuselage! (click on the image to open it in higher resolution to see what I am talking about)

- problematic UV mapping near the aircraft nose in the left side, colors look really strange on all liveries:

![fg-0](https://cloud.githubusercontent.com/assets/5700795/20360200/12c1a6ae-ac33-11e6-8660-d770d5ef2cc0.jpg)

![fg-1](https://cloud.githubusercontent.com/assets/5700795/20360201/12c675d0-ac33-11e6-8b3b-13f933b42d08.jpg)

- the rudder shape is a tiny bit wrong, it is missing a half-circle hole, see:

![rudder](https://cloud.githubusercontent.com/assets/5700795/20360322/9259e5a2-ac33-11e6-86f9-0e9f0a2be341.jpg)

some other references:

![](https://upload.wikimedia.org/wikipedia/commons/e/ee/Cessna_172S_Skyhawk_SP_N5204A_tail_1.JPG)

![](http://truepcgaming.com/wp-content/uploads/2014/07/A2A-Accusim-Cessna-172-Review-1.jpg)

gilbertohasnofb:
The post above has been edited. 

@wlbragg would you maybe be interested in taking a look at this issue? These should be quite minor fixes I suppose.

wlbragg:
Yep, I'll look at it when I do the control wheel lock.

gilbertohasnofb:
👍 Thanks! Let me know if I can help somehow

gilbertohasnofb:
I am not sure if this has been reported before, but there are some issues with the bottom part of the fuselage:

![fgfs-screen-001](https://user-images.githubusercontent.com/5700795/36034101-e0dedc30-0daa-11e8-88a2-b7d90be131d7.png)

Is that a 3D model, UV mapping or effect problem? Perhaps it could be added to this issue or maybe we should create a new issue for it -- opinions?

wlbragg:
[...]

PR #960, opened 2018-02-12
Issue 860: Small 3D issues with the external fuselage

wlbragg:
This fixes a longstanding mesh issue at the tail of the fuselage. There were several vertices that were joined and moved to the same location that shouldn't have been creating unsightly seems on the underside of the fuselage. It is surprising this was the only noticeable artifact as this was pretty extensive damage to the mesh.
I would like to see this merged as soon as possible.


gilbertohasnofb:
@wlbragg Thank you so much for tackling this! I have tested it in the sim and I have found some issues with the new model, could you please take a look at these? The main issue is that some parts of the model are not smooth any longer but rather made out of "squares". Also, the wing texture is reverting back to the default model:


fg-0
fg-1
fg-2
fg-3
fg-4
fg-5
fg-6
fg-7

gilbertohasnofb:
> I think I can also model the semicircle at the base of the rudder before we merge.

That would be great! And do you think you can also take a look at these strange areas of the top of the nose which appear on both sides and with all liveries?:

gilbertohasnofb:
@wlbragg one more thing I just notices: the fuselage in the interior of the plane is also "squared", and the panel texture does not change according tot he livery any longer.

onox:
@wlbragg Which issue is this PR fixing? #680 has been closed a long time ago.

gilbertohasnofb:
> Which issue is this PR fixing? #680 has been closed a long time ago.

He meant https://github.com/c172p-team/c172p-detailed/issues/860 and not 680.

onox:
Okay. @wlbragg could you please rename the branch to `issue-860` then? That's needed so that Github puts the text "issue-860" in the commit message. I've used that multiple times to find issues/PR's on Github which modified/added certain files. Used it last week to find a comment explaining the license of a sound file I had copied to the p51d (for Stuart).

One more thing: if you put "Fixes #myissuenumber" in your PR's description, then Github will automatically close the linked issue when the PR gets merged. I've seen it a few times that issues were long closed but still open because the PR wasn't linked to the issue.

wlbragg:
@onox 
Sorry for transposing of the issue number!

@gilbertohasnofb 
Wow, I am glad you were paying attention to this. I can't believe I didn't notice the smoothing, texture and livery changes.
I have no idea what happened. Give me some time to look into this.
One thing I did already look at was those bright square anomalies
image
There is absolutely nothing out of the ordinary with the mapping or the texturing at those locations. From what I can tell it is caused by the angle of the face of those single squares. They simply catch the light differently then their adjacent neighbors. I will take another look at it while fixing the rest of it, just in case I missed something.

wlbragg:
I guess I did that wrong. I don't know how to rename the branch short of deleting it and pushing a new on. But that closed the PR which I didn't even realize I had actually done a PR on it. I thought it was just a branch with no PR yet. I don't want to loose all the issues @gilbertohasnofb has highlighted here.

wlbragg:
OK, I pulled a clean master and the Blend in it is just as screwed up as the Blend in this branch is. Again I don't know who is responsible or for how long it has been this way but it got pushed at sometime with some major flaws in it. So, it's going to take some time to sort it out and I am going to need @gilbertohasnofb help to critique it until we get it all straitened out. We certainly can't leave it like it is.

onox:
@wlbragg Just with `git branch Issue680 issue-860` and then `git push -u origin issue-860`

onox:
Added the missing `-m` in the previous comment.

onox:
Looks like it (non-smooth windows) happened somewhere between 85e271b (2015-11-18) and c520440 (Initial commit)

onox:
@NikolaiVChr Maybe this has something to do with the ac3d exporter?

onox:
721f6f7 bad (2015-07-19)
d0eca8a ok (2015-03-13)

onox:
750050d bad (2015-05-02)
0b953ec ok (2015-04-18)

onox:
65172de bad (2015-04-28)
57d8a71 ok (2015-04-26)

@wlbragg Seems like the problem was introduced in commit 65172de

gilbertohasnofb:
> Wow, I am glad you were paying attention to this. I can't believe I didn't notice the smoothing, texture and livery changes.

No worries, it's easy to not notice things when you work on the same stuff for many hours. That's what's good about having a fresh pair of eyes checking the PRs :smile:

> I have no idea what happened. Give me some time to look into this.

No rush. I will be happy to review this once you are done.

> There is absolutely nothing out of the ordinary with the mapping or the texturing at those locations. From what I can tell it is caused by the angle of the face of those single squares. They simply catch the light differently then their adjacent neighbors. I will take another look at it while fixing the rest of it, just in case I missed something.

That is really strange... In particular because if that was just a matter of angle, then sometimes those squares would have to be darker than the surrounding -- but they are always brighter! Could that be some bug or collision caused by the cockpit glass perhaps?

* * *

And many, many thanks for taking a look at this! I know how much effort these sort of issues require.

wlbragg:
OK, I guess the damage is done to this PR as it is

wlbragg:
There is no way to attach this PR thread to branch issue-860? All I accomplished by renaming was to effectively create a new branch here with no PR and this PR still attached to Issue680.

What should I do now, delete branch Issue680 as I did before which also closes this PR. Then make a new PR for issue-860 when ready to proceed?

wlbragg:
I am ready to work with @gilbertohasnofb on verifying all the corrections and I don't know where we want to work from. I don't think we should be working from this PR thread as it needs to either be attached to a new PR that will be created by branch issue-860.

wlbragg:
But I need to move forward so for now I'll stay here even though the commits only show up in the https://github.com/c172p-team/c172p-detailed/tree/issue-860 branch.

@gilbertohasnofb I have pushed a version smoothing the entire aircraft, every object. What I am concerned about is whether or not the correct mesh has all or only some of the mesh smoothed. I have no way to decipher that other than go back in history and check every last object. So, please look at this in the sim and give me your impression, both interior and exterior and tell me what you think. Is this back to where it is supposed to be?

wlbragg:
@NikolaiVChr I don't understand something I am seeing in Blender. If you check the Auto Smoothing checkbox in the Object Data tab it has the opposite effect of removing the smoothing. any idea what that is all about. That might be part of the issue as far as why it broke the effect. Most of the objects were checked to smooth but it had the opposite effect.

wlbragg:
@NikolaiVChr @onox  @gilbertohasnofb 

Model/c172p.xml references two groups of objects by their group name minus the .group extention...
GROUP_FrontPanels.group
rudder.group
The references do not include .group in the property tag and thus gives an error in the log as not being found and the references don't work. For example rudder is referenced in the xml in a pick animation as "rudder" and it doesn't work. The .ac exports the rudder group as rudder,group and if you use that name in the property (ie: rudder.group) it works. Anyone know if this is the way it has always worked or is there some other problem?

wlbragg:
I cannot figure out what is wrong with the wing livery not changing over, no log message, the code looks OK. I'm stumped. It is working in the c172p in fgdata.

Anyone have any ideas?

wlbragg:
@gilbertohasnofb 
the small square on the fuselage under the front window with the different lighting property on it, It is the UV mapping on the texture.
In other words it was apparently by design but we can get rid of it if you want.
We change the mapping from this...
http://i.imgur.com/5nf30mu.png
to  this...
http://i.imgur.com/KuyFPtW.png

onox:
> The references do not include .group in the property tag and thus gives an error in the log as not being found and the references don't work.

In `Models/c172p.ac` I see: `name "GROUP_FrontPanels". The same name is used in `Models/c172p.xml`. So messages about missing references should not occur?

onox:
> Then make a new PR for issue-860 when ready to proceed?

Yeah, just open a new PR.

onox:
In the old c172p.ac you have a material called `DefaultWhite.012` but I don't see it in the new one. Not sure if that's the problem.

onox:
@wlbragg I see `wing` instead of `wing_1` on line 6544 in `Models.c172p.xml`.

onox:
Okay, I think that's it. If I do `grep wing Models/c172p.ac` I see `wing`, `wing_0`, `wing_1` (twice) in the old file, but in the new file I see `wing.group`, `wing_0`, `wing_1.group`, `wing_1`.

onox:
Ugh, I do see the log messages about some missing references now. About some cones + GROUP_FrontPanels and rudder.

onox:
@wlbragg You have any idea why these `.group` objects are in the exported ac file?

wlbragg:
>In Models/c172p.ac I see: name "GROUP_FrontPanels". The same name is used inModels/c172p.xml`. So messages about missing references should not occur?

OK, in the fgdata .ac there is GROUP_FrontPanels but in our master .ac and in the issue-860 .ac the name was saved as GROUP_FrontPanels.group. That is the issue with all the missing objects and the wing livery problem. Why is my version of the exporter causing trouble? OK, thats my problem, I'll fix all that.

>In the old c172p.ac you have a material called DefaultWhite.012 but I don't see it in the new one. Not sure if that's the problem.

It's probably not a problem but I'll look into it.

PR #971, opened 2018-02-20
Refactor all mesh using Blenders smoothing feature

### LINKED ISSUE #118: Problems with the 3D model of the wings
The 3D model of the wings need some rework:

1) Glass reflection colliding with vent:

fgfs_screen_001

2) The wings are too thin, compare the photograph below with our plane:

3) Our wings are missing the drooped tips ("inverted winglets"), see:

### LINKED ISSUE #974: Better UV-map of normal map on tail
The UV-map of the normal map on tail is not looking too good:

fg-1

Is this perhaps a side effect of fixing #430? If so, do we need to change the textures themselves rather than the UV-map of the normal textures, @wlbragg ?

wlbragg:
Fixes #118 
Fixes #974

Not ready to merge.


gilbertohasnofb:
> Not ready to merge.

What is missing exactly?

As for the work, the aircraft definitely looks better, in particular the cockpit! The cockpit looks better in almost every single way. I would just perhaps unsmooth the toggles (e.g. lights, master switch, etc.) as those seem to look better before, though this is a matter of opinion I think (both ways look good).

The issues I have are with the exterior, in particular with multiple surfaces like flaps, leading edges, etc.: they do not match the wings and/or surroundings any longer:

fgfs-screen-001
fgfs-screen-002
fgfs-screen-003

the image you posted yourself shows very well the issue with flaps and leading edges (in the second row):



Also, the aircraft fuselage looks darker -- why is that?

wlbragg:
>What is missing exactly?

I wanted you to critique the smoothing first.

>Also, the aircraft fuselage looks darker -- why is that?

I'm not really sure unless we've changed lighting VS what is in fgdata (nothing I did recently) or it was simply my weather conditions between the two shots.

>issues I have are with the exterior, in particular with multiple surfaces like flaps, leading edges, etc.: they do not match the wings and/or surroundings any longer:

Yeah, that's because some was smoothed a little more than others. The original smoothing settings worked for things like the recessed fuel caps and removed some of the mesh artifacts where the VS meets the fuselage. But the higher smoothing settings made other areas look better except "artifacts" and now they don't match. So I put all the exterior back to what it was originally. It looks consistent now I think. I also changed the toggle switched to get more definition.

I repositioned the beacon light a tad bit. I think it looks just as bad if it is too high because it looks like it is above the aircraft, no good solution for this. See if you like the way it is any better. If so I think it is good to merge if you do

wlbragg:
I need this in master before I proceed to anything else. It gets too complicated if I have all these changes outstanding and try to start something else. I guess I have yet to master git work flow.

gilbertohasnofb:
No problems, I will review this still today when I get home

gilbertohasnofb:
@wlbragg everything is much better, but the leading edge still looks too different from the main wing. Do you think it is possible to match it at all with the wing? The horizontal stabilizer does not seem to suffer from that. Also, this is the case in all times of the day:

fg-2
fg-3
fg-4

By the way, I have found a couple of small 3D bugs and other places that can be improved but I will be opening new issues and/or commenting on older ones so that we keep things more or less organized, since this PR is rather about the smoothing effect, ok?

gilbertohasnofb:
@wlbragg one more thing, it looks to me that there is one face of this object below which is always dark and non reflective, regardless of the light direction (I tested it by flying it circles and comparing it to the adjacent face). Maybe I am wrong but it seems to me this face is different than the others, perhaps untextured or having no effects applied, not sure.

fgfs-screen-001

wlbragg:
Another good catch with the leading-edge. I somehow forgot to make it the same angle as the rest of the wing. Should be correct now.

gilbertohasnofb:
@wlbragg I will test right now and merge if all is fine, thanks for the quick fixes

gilbertohasnofb:
@wlbragg the antenna is fixed, but I still see the leading edge in a darker tone at any time of the day:

fg-0
fg-1

gilbertohasnofb:
@wlbragg Just tested the new bumpmap, looks really good. So just the leading edges are missing; are you able to reproduce what I see?

wlbragg:
>I still see the leading edge in a darker tone at any time of the day:

I'm thinking that may be normal and caused by the angle of the mesh face and not the smoothness angles?

Unless it is because we now have two separate edges there, back edge of leading and front edge of wing (where they meet), where it used to be no edge there for the smoothing effect to span over, if that makes sense.

I can test this by combining it back into one or making the leading-edge incorporate more area of the front part of the wing.

wlbragg:
I'm pretty sure it is because there is a distinct edge there now that the smoothing algorithm can't exploit. It can't go across two different objects to smooth it.

So we either live with it.
Put it back.
Or, first I'll try to add more of the front edge of the wing top and bottom to the leading edge so the smoothing algorithm has something to work with.

gilbertohasnofb:
> So we either live with it.

Hmm, I am not too happy about that. The dirt effect is cool and all but in the end it's just a gimmick that will be noticed in very few cases. I am not sure it is worth having an uglier wing which will be noticed in many more situations. It's your call, but my vote would be for a prettier wing than a prettier effect.

wlbragg:
>Or, first I'll try to add more of the front edge of the wing top and bottom to the leading edge so the smoothing algorithm has something to work with.

I think this worked well enough. When I said "put it back", what I meant was, we can simply not separate the leading edge and still use the dirt effect. We were only separating the leading edge so we could potentially have the leading edge get dirtier faster than the rest of the wing or use a different texture with it.

wlbragg:
See this is where I get in trouble with git and different branches. Now we made a change to this branch because it was a smoothing issue. But this change absolutely affects the "dirt" branch  as it is the leading edge (edge-dirt) and wing (wing-dirt) that we modified. Thus we test the dirt affect that branch doesn't have the changes.
How is one supposed to manage this?

gilbertohasnofb:
@wlbragg 

> I think this worked well enough. 

Yes it did, it looks really good again! :smile: 

> See this is where I get in trouble with git and different branches.

I think we need @onox for this one as I don't want to give out stupid or wrong advice.

Also, just one more thing I spotted: are the wing struts smoothed? They look a bit too 'polygonal' from the angle below:

fgfs-screen-001

wlbragg:
Yes, I had set the fuselage to 35 deg smoothing angle and that isn't enough to smooth out the edges very well. If I go to high I started getting ugly artifacts. I think I found a happy medium, 65 deg.
You know what is interesting is I never even know this stuff was available or how to use it. It makes a huge difference as we have found out, but requires lots of testing and trial and error.

onox:
@wlbragg With these binary files you might risk undoing previous changes. So do the blend and ac files in this branch also contain the changes you have made in the other branch?

wlbragg:
>do the blend and ac files in this branch also contain the changes you have made in the other branch

No, they have diverged.

wlbragg:
Please merge and close this PR. We can continue any Blender work under the Grain Effect PR.

gilbertohasnofb:
Just retested this, everything is looking good. Many thanks for all fixes.

legoboyvdlp:
Hi there, just a small issue here at the junction of the fuselage, horizontal stabalizer, and the part of the fuselage that rises towards the rudder. 
image

gilbertohasnofb:
@legoboyvdlp I don't think I understand the problem, could you try explaining it again? What do you think @wlbragg ?

wlbragg:
I see the texture line there and this area was reacting to the smoothing in less than desirable ways. So I am not sure there is much we can do other then change the UV map. See how the UV map is laid out. That is why there is that line. All you can do is try to blend those two parts of the texture together better. But then the left of that area where the pin-striping starts is going to mot match. You'd have to be very good at blending and fading.
tail

legoboyvdlp:
That's unfortunate, but thank you for looking into it.

gilbertohasnofb:
@wlbragg so if I understand this correctly the fix is to improve all textures by removing those dashed lines and extending the texture for the rear fuselage a bit? If so, I can work on it.

wlbragg:
I wasn't really looking at how I would fix it, but now that you mentioned it. The exact issue is what your saying.

(Bottom right image) the bottom edge of the face with the green line (looking down on the tail section) is in a darker shaded area of texture than the top edge of the face with the red line. Those two different texture areas would have to match exactly.
tail2

gilbertohasnofb:
I might try to fix these in the future, but it looks like an annoying one: the matching between the texture areas will be extremely hard if not impossible to match perfectly, and the shading is done by an overlay which is a pain to tweak. And this would have to be done to each and every single livery. I will give this a shot but I am not sure what I will be able to do.

gilbertohasnofb:
I've created a new issue for this livery problem, see https://github.com/c172p-team/c172p-detailed/issues/983

PR #972, opened 2018-02-20
Fix engine axis bindings for 2018.1 compatibility.

Zaretto:
2018.1 introduces new "all" properties for throttle, mixture and prop pitch this is the correct way[1] to interface with the axis based controls - use a listener on the -all properties as below. This will retain compatibility with previous versions of FG and 2018.1.

* /controls/engines/throttle-all
* /controls/engines/mixture-all

-------------
[1] The controls.throttleAxis, mixtureAxis, propellerAxis bindings should no longer be overridden directly - simply because the way they use cmdarg() makes it impossible to change the interface inside controls.nas

### DIFF
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -319,24 +319,25 @@ controls.throttleMouse = func {
 # 2018.2 introduces new "all" properties for throttle, mixture and prop pitch.
 # this is the correct way to interface with the axis based controls - use a listener
 # on the *-all property
-_setlistener("/controls/engines/throttle-all", func{
+setlistener("/controls/engines/throttle-all", func{
     var value = (1 - getprop("/controls/engines/throttle-all")) / 2;
     var new_value = std.max(0.0, std.min(value, 1.0));
     setprop("/controls/engines/current-engine/throttle", new_value);
-},0,0);
-_setlistener("/controls/engines/mixture-all", func{
+}, 0, 0);
 
+setlistener("/controls/engines/mixture-all", func{
     var value = (1 - getprop("/controls/engines/mixture-all")) / 2;
     var new_value = std.max(0.0, std.min(value, 1.0));
     setprop("/controls/engines/current-engine/mixture", new_value);
-},0,0);
+}, 0, 0);
 
 #_setlistener("/controls/engines/propeller-pitch-all", func{
 #    execWithLocalCmdArg(propellerAxis,getNormalisedEngineAxisValue("/controls/engines/propeller-pitch-all"));
-#},0,0);
+#}, 0, 0);
 
 # backwards compatibility only - the controls.throttleAxis should not be overridden like this. The joystick binding Throttle (all) has 
-# been replaced and controls.throttleAxis will not be called from the controls binding
+# been replaced and controls.throttleAxis will not be called from the controls binding - so this is to 
+# maintain compatibility with existing joystick xml files.
 controls.throttleAxis = func {
     var value = (1 - cmdarg().getNode("setting").getValue()) / 2;
     var new_value = std.max(0.0, std.min(value, 1.0));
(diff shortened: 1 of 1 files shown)


Nasal/engine.nas
```diff
@@ -316,6 +316,27 @@ controls.throttleMouse = func {
     setprop("/controls/engines/current-engine/throttle", new_value);
 };
 
+# 2018.2 introduces new "all" properties for throttle, mixture and prop pitch.
+# this is the correct way to interface with the axis based controls - use a listener
+# on the *-all property
+_setlistener("/controls/engines/throttle-all", func{
```
onox: Why are you using `_setlistener` instead of `setlistener`? The [wiki](http://wiki.flightgear.org/Using_listeners_and_signals_with_Nasal#setlistener.28.29_vs._setlistener.28.29) says `setlistener` should be used.

Nasal/engine.nas
```diff
@@ -316,6 +316,27 @@ controls.throttleMouse = func {
...
+    var value = (1 - getprop("/controls/engines/throttle-all")) / 2;
+    var new_value = std.max(0.0, std.min(value, 1.0));
+    setprop("/controls/engines/current-engine/throttle", new_value);
+},0,0);
+_setlistener("/controls/engines/mixture-all", func{
+
```
onox: Remove this empty line.

Nasal/engine.nas
```diff
@@ -316,6 +316,27 @@ controls.throttleMouse = func {
...
+    var value = (1 - getprop("/controls/engines/mixture-all")) / 2;
+    var new_value = std.max(0.0, std.min(value, 1.0));
+    setprop("/controls/engines/current-engine/mixture", new_value);
+},0,0);
```
onox: Please separate the arguments with a space here. Same for below.

Nasal/engine.nas
```diff
@@ -329,6 +350,8 @@ controls.adjMixture = func {
     setprop("/controls/engines/current-engine/mixture", new_value);
 };
 
+# backwards compatibility only - the controls.throttleAxis should not be overridden like this. The joystick binding Throttle (all) has
```
onox: The `master` branch is expected to be used with fgdata git and will be copied to fgdata when is going to be released 2018.2, so I guess the backwards compatibility stuff can be removed.

Nasal/engine.nas
Zaretto: I had run time errors because setlistener wasn't defined; I suspect it is the inclusion order - so I followed the example in other FGdata/Nasal modules

Nasal/engine.nas
Zaretto: This is runtime backwards compatibility for existing Input/joysticks/xml files that haven't been rebound to use the new methods. Once you change a binding with the joystick dialog it will of course fix that entry; but otherwise old or manually editted joystick XML files will not work.

onox: Could you add a comment that it's needed for old joystick XML files?

Nasal/engine.nas
onox: There are multiple `setlistener` calls already in the file so it would be strange to get runtime errors.

Zaretto: sorry my mistake; I copied the code from controls.nas and obviously it doesn't need to be a _setlistener.

Nasal/engine.nas
```diff
@@ -316,6 +316,28 @@ controls.throttleMouse = func {
...
+    setprop("/controls/engines/current-engine/throttle", new_value);
+}, 0, 0);
+
+setlistener("/controls/engines/mixture-all", func{
+    var value = (1 - getprop("/controls/engines/mixture-all")) / 2;
+    var new_value = std.max(0.0, std.min(value, 1.0));
+    setprop("/controls/engines/current-engine/mixture", new_value);
+}, 0, 0);
+
+#_setlistener("/controls/engines/propeller-pitch-all", func{
```
onox: The c172p doesn't have pitch control, so this commented code can be removed.

PR #981, opened 2018-02-22
Bug 976: Fix applying effect to renamed object 'extinguisher'

### LINKED ISSUE #976: Extinguisher object missing in master branch
Could not find at least one of the following objects for animation: 'extinguisher2'

onox:
Fixes #976 

Prevents FG printing `Could not find at least one of the following objects for animation: 'extinguisher2'` to the console. Interior effect should be applied again.

### DIFF
--- a/Models/Interior/FE-ELT/fe2.xml
+++ b/Models/Interior/FE-ELT/fe2.xml
@@ -3,13 +3,10 @@
 <PropertyList>
 
     <path>fe2.ac</path>
+
     <effect>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
@onox thanks for this fix. And what about the following console message: `Aircraft does not specify a minimum FG version: please add one at /sim/minimum-fg-version`. Do we want to do something about it or is it better not to define a minimum FG version?

onox:
The model is copied to fgdata, so we can safely assume a minimum FG version.

PR #988, opened 2018-02-24
Issue 124: Add engine damage and more advanced oil management

legoboyvdlp:
* Adds damage controller which adds damage as you exceed redline RPM
* Fail the engine based on low or high oil pressure
* Add more complex oil management from the C182, including the external heater

WIP-DO-NOT-MERGE ;)


Nasal/engine.nas
```diff
@@ -87,39 +87,73 @@ var primerTimer = maketimer(5, func {
...
         }
+        
+        # Consumption also raises with oil in service time (lower viscosity => more friction)
+        # (Oil should be changed at 50 hrs!)
+        # See: http://www.t-craft.org/Reference/Aircraft.Oil.Usage.pdf
+        # Hours:        0 |    10 |    25 |  50   |    75
+        # Add Qts/hr:   0 |  0.02 | 0.125 | 0.5   | 1.125
+		service_hours = getprop("/engines/active-engine/oil-service-hours");
+        service_hours_increase = 0.00020 * math.pow(service_hours, 2);
+        if (service_hours_increase > 1.5) service_hours_increase = 1.5;  # cap at that rate
```
onox: Can you use { } here? You can also write `service_hours_increase = std.min(1.5, service_hours_increase);`

Nasal/engine.nas
```diff
@@ -87,39 +87,73 @@ var primerTimer = maketimer(5, func {
...
+        if (oil_level > oil_full) {
+            consumption_qph = consumption_qph * 1.3;
         }
+        
+        # Consumption also raises with oil in service time (lower viscosity => more friction)
+        # (Oil should be changed at 50 hrs!)
+        # See: http://www.t-craft.org/Reference/Aircraft.Oil.Usage.pdf
+        # Hours:        0 |    10 |    25 |  50   |    75
+        # Add Qts/hr:   0 |  0.02 | 0.125 | 0.5   | 1.125
+		service_hours = getprop("/engines/active-engine/oil-service-hours");
```
onox: Indentation level of several statements here seems off?

Nasal/engine.nas
```diff
@@ -143,6 +177,71 @@ var oil_consumption = maketimer(1.0, func {
...
+var oil_refill = func(){
+    var service_hours = getprop("/engines/active-engine/oil-service-hours");
+    var oil_level     = getprop("/engines/active-engine/oil-level");
+    var refilled      = oil_level - previous_oil_level;
+    #print("OIL Refill init: svcHrs=", service_hours, "; oil_level=",oil_level, "; previous_oil_level=",previous_oil_level, "; refilled=",refilled);
+    
+    if (refilled >= 0) {
+        # when refill occured, the new oil "makes the old oil younger"
+        var pct = 0;
+        if (oil_level > 0) pct = previous_oil_level / oil_level;
```
onox: Here also add { }.

Nasal/engine.nas
```diff
@@ -143,6 +177,71 @@ var oil_consumption = maketimer(1.0, func {
...
+var rpm = getprop("/engines/active-engine/rpm");
+var damage = 0.0;
+
+var eng_damage_function = maketimer(1.0, func {
+	if (getprop("/engines/active-engine/damage_allowed")) {
+		rpm = getprop("/engines/active-engine/rpm"); # we need to update the RPM, but we do not set the variable here to avoid GC load
+	
+		if (rpm > 2700) {
+			damage = damage + ((rpm - 2700) / 100); # 1 point of damage per 100RPM over limit added per second, so the damage increases faster the more you go over the limit
+		} else
```
onox: The else block seems redundant.

Nasal/engine.nas
```diff
@@ -143,6 +177,71 @@ var oil_consumption = maketimer(1.0, func {
...
+# Initialize variables to reduce GC load
+var rpm = getprop("/engines/active-engine/rpm");
+var damage = 0.0;
+
+var eng_damage_function = maketimer(1.0, func {
+	if (getprop("/engines/active-engine/damage_allowed")) {
+		rpm = getprop("/engines/active-engine/rpm"); # we need to update the RPM, but we do not set the variable here to avoid GC load
+	
+		if (rpm > 2700) {
+			damage = damage + ((rpm - 2700) / 100); # 1 point of damage per 100RPM over limit added per second, so the damage increases faster the more you go over the limit
```
onox: Does `damage += ...` work?

Nasal/engine.nas
```diff
@@ -143,6 +177,71 @@ var oil_consumption = maketimer(1.0, func {
...
+        interpolate("/engines/engine/oil-temperature-env-diff", 0, 180); # hand over to jsbsim caluclation gradually over 2 minutes
+    }
+});
+
+# ========== engine damage due to over-RPM =========
+# Initialize variables to reduce GC load
+var rpm = getprop("/engines/active-engine/rpm");
+var damage = 0.0;
+
+var eng_damage_function = maketimer(1.0, func {
```
onox: There are tabs in this function.

Systems/engine.xml
```diff
@@ -536,6 +542,30 @@
                     <value>1.97</value>
                 </less-than>
                 
+				<!-- kill engine if the oil pressure is too low while the engine is running, according to Lycoming manaul 3-6 minimum oil pressure
```
onox: Please replace the tabs with 4 spaces.

Systems/external-heat.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0"?>  
+
+<!-- c182s external heater
```
onox: Could you add a copyright block (GPL 2+) here? You can find an example in some of the files in the Systems folder. Make sure to put Benedikt Hallinger as the author in the copyright line.

gui/dialogs/c172p-ground-equipment.xml
```diff
@@ -290,6 +290,36 @@
             
         </group>
     </group>
+	
+	<group>
+            <layout>vbox</layout>
```
onox: Is this 6 spaces indentation?

onox:
I'm cooking dinner, so I did just a quick code review. @gilbertohasnofb can you do some functional testing?

legoboyvdlp:
Hi @onox,

I just made some changes, but did not have time to push them before leaving... I will attempt to push them tonight.

Thank you for your codereview :) 

The only thing I will be adding after I finish bugfixing the oil management is the reduction of RPM as damage increases. I believe I will make the engine cut out at 300 damage levels, that is, 5 minutes of 100RPM over redline.  How does this sound? The RPM will begin to drop as you go past 100 damage levels, dropping 100rpm from 100 to 200 and then 500rpm from 200 to 250 and cutting out completely at 300... 
The damage level will also be saved... you have to hit repair if you go over redline :)

Nasal/engine.nas
legoboyvdlp: I have no idea, how exactly do you propose I do this?

gilbertohasnofb:
> can you do some functional testing?

I am not sure I will have time for this this weekend, these sort of tests take really long time and I am super busy with some ultra boring applications and forms... I will try to take a look at this asap, but it might be a while

legoboyvdlp:
Hi Onox, Gilberto, I just pushed my commit. I'll see if I can do the fixes suggested by Onox now.

Nasal/engine.nas
legoboyvdlp: I use tabs as standard... hard to forget old habits :)
Fixed, let me know if there are any more.

onox:
@legoboyvdlp There are still a bunch of tabs in the code. If you select a tab and then Ctrl + f in your favorite editor then you can find all of them :)

Nasal/engine.nas
```diff
@@ -143,6 +178,71 @@ var oil_consumption = maketimer(1.0, func {
...
+    if (getprop("/engines/active-engine/damage_allowed")) {
+        rpm = getprop("/engines/active-engine/rpm"); # we need to update the RPM, but we do not set the variable here to avoid GC load
+    
+        if (rpm > 2700) {
+            damage = damage + ((rpm - 2700) / 100); # 1 point of damage per 100RPM over limit added per second, so the damage increases faster the more you go over the limit
+        }
+    } else { # make sure damage is always 0 if the checkbox is disabled
+        damage = 0.0; 
+        if (getprop("/engines/active-engine/damage-level") != 0.0) {
```
onox: Isn't this if check redundant? If it's 0 you do nothing, but if it's not 0, you set it to 0. So you can just do the `setprop()` call without the if condition.

Is your concern performance? The function is only run once per second, so I don't think it will really matter. You can remove the if condition below as well.

Note the tie-downs run code every frame while they're visible and I haven't seen people complaining about performance drops.

legoboyvdlp:
You can now see the heater heating up the oil on the gauge, if you have external power, of course :)

@gilbertohasnofb apologies that I haven't gotten the RPM drop done yet. I'm rather busy, and never got round to it!

legoboyvdlp:
As for the heater, should it plug into the air intake / carb... or does it go in a special port...?

The property for winterkit is present: it just needs a 3D model.

gui/dialogs/c172p-ground-equipment.xml
```diff
@@ -314,6 +314,10 @@
                 <binding>
                     <command>property-toggle</command>
                     <property>/engines/active-engine/external-heat/enabled</property>
+                </binding>
+				 <binding>
```
onox: Last tabs remaining on this line (318) and line 293.

Systems/engine.xml
```diff
@@ -250,6 +250,15 @@
         <gain>
             <property>/engines/active-engine/low-oil-temperature-factor</property>
         </gain>
+		<input>
```
onox: Here some tabs as well.

Systems/fuel.xml
```diff
@@ -8,6 +8,7 @@ Fuel system for 2 different engines at choice, JSBSim FDM
 -->
 
 <system name="fuel">
+	<property type="bool" value="0">/engines/active-engine/already-started-in-session</property>
```
onox: Tab.

Systems/fuel.xml
```diff
@@ -132,5 +133,23 @@ Fuel system for 2 different engines at choice, JSBSim FDM
             <output>propulsion/tank[1]/external-flow-rate-pps</output>
         </summer>
     </channel>
+
```
onox: Here many tabs as well.

c172p-set.xml
```diff
@@ -678,6 +679,17 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <oil-level type="double">7.0</oil-level>
             <oil-lacking type="double">0.0</oil-lacking>
             <oil_consumption_allowed type="bool">false</oil_consumption_allowed>
+			<oil-powergain-factor type="double">1</oil-powergain-factor>
```
onox: Here also many tabs. You should see in Github that the indentation is odd compared to the surrounding code.

c172p.xml
```diff
@@ -1912,6 +1912,7 @@
     <system file="c172p-damage"/>
     <system file="c172p-sounds"/>
     <system file="c172p-heat"/>
+	<system file="external-heat"/> <!-- integrate external-heater, thanks to HHS for allowing use-->
```
onox: Tab.

gui/dialogs/c172p-oil-160.xml
```diff
@@ -64,6 +64,10 @@
                 <binding>
                     <command>dialog-apply</command>
                     <name>c172p-oil-slider-160</name>
+                </binding>
+				 <binding>
```
onox: Tabs on line 68, 91, 106.

gui/dialogs/c172p-oil-180.xml
```diff
@@ -65,6 +65,10 @@
                 <binding>
                     <command>dialog-apply</command>
                     <name>c172p-oil-slider-180</name>
+                </binding>
+				<binding>
```
onox: Tabs on line 69, 91, 92.

onox:
I went through all the changed files and hopefully catched all the tabs. Sorry for having to be such a coding style nazi, but if a mixture of tabs and spaces slip into the project, then code (diffs) can become more difficult to read.

Systems/external-heat.xml
```diff
@@ -0,0 +1,88 @@
...
+            <output>/engines/active-engine/external-heat/applied-degF</output>
+        </aerosurface_scale>
+        <fcs_function name="Oil temp plus external heat">
+            <function>
+                <sum>
+                    <property>/engines/active-engine/external-heat/applied-degF</property>
+                    <property>/engines/active-engine/oil-temperature-degf-indicated</property> <!-- compensated for env-temp and low oil, see engine.xml -->
+                </sum>
+            </function>
+            <output>/engines/active-engine/oil-final-temperature-degf</output>
```
onox: While you're removing the tabs, you might as well remove the trailing spaces on lines 1, 66, 68, 77, 78. If you do `git diff` in the terminal, they spaces should have a red background color, so they're easy to spot.

Systems/engine.xml
```diff
@@ -530,12 +536,36 @@
...
+					<equals>
+						<property>/engines/active-engine/ready-oil-press-checker</property>
+						<value>2</value>
+					</equals>
+					<less-than>
+						<property>/engines/active-engine/oil-pressure-psi</property>
+						<value>40</value>
+					</less-than>
+				</and>
+
```
onox: Trailing spaces on line 559, 568, and 624.

Nasal/engine.nas
```diff
@@ -291,12 +389,25 @@ setlistener("/controls/switches/starter", func {
             primerTimer.stop();
         }
     }
+    
+    # sorry - had to hack this to prevent coughing on startup due to the oil pressure simulation. Maybe this can be used elsewhere
+     if (getprop("/controls/engines/active-engine") == 0)
+       var rpm = getprop("/engines/engine[0]/rpm");
```
onox: Why don't you just do `var rpm = getprop("/engines/active-engine/rpm");` here?

Nasal/engine.nas
```diff
@@ -143,6 +178,65 @@ var oil_consumption = maketimer(1.0, func {
...
+# Initialize variables to reduce GC load
+var rpm = getprop("/engines/active-engine/rpm");
+var damage = 0.0;
+
+var eng_damage_function = maketimer(1.0, func {
+    if (getprop("/engines/active-engine/damage_allowed")) {
+        rpm = getprop("/engines/active-engine/rpm"); # we need to update the RPM, but we do not set the variable here to avoid GC load
+    
+        if (rpm > 2700) {
+            damage = damage + ((rpm - 2700) / 100); # 1 point of damage per 100RPM over limit added per second, so the damage increases faster the more you go over the limit
```
onox: So you get a linear increase of damage for the engine RPM above the threshold (2700 RPM)? I'm not completely sure about that. Why not have a slow increase for engine RPM near the threshold, but a much higher damage increase if you go far above the threshold? Exponential growth.

I think you could use a JSBSim/Autopilot table that defines the damage/second as a function of RPM-above-the-threshold. That is also easier to tune than having just a single formula. This way the data that controls how fast the damage increases (the table) is separated from the code that applies the damage (this timer function).

If you have a table, you could even use an integrator if that's what you prefer.

onox: I managed to get the 180 HP engine to 4600 RPM while diving, but that's ridiculously high. Perhaps create a table from 0 (2700 RPM) to 800 (3500 RPM).

legoboyvdlp: Thank you, I will attempt to implement your suggestions by Monday ;)

Nasal/engine.nas
legoboyvdlp: Because of "Nasal runtime error: nil used in numeric context
  at C:/Users/redpa/Documents/FlightGear/Custom Aircraft/c172p/Nasal/engine.nas, line 399
  called from: C:/Users/redpa/Documents/Programs/FlightGear 2018.1.1/data/Nasal/globals.nas, line 119"

legoboyvdlp:
May I propose that the formula be
(rpm-over-redline)² / 10000?

`100*100   =   10000 / 10000 =   1
500*500   =  250000 / 10000 =  25
1000*1000 = 1000000 / 10000 = 100
`

legoboyvdlp:
@dany93 
Hi Dany,
Onox redirected me to you ;)
What occurs in the real Cessna when you operate the engine above 2700RPM for too long?
Is there any limit for operating **at** 2700RPM -> e.g. CHT limitation?

How would you suggest I implement engine damage given a value between 0 (no damage) and 1 (full damage?

One suggestion is to do throttle = throttle * normalized_damage_value:
According to it0uchpods:

> Well, what I would do, is put some code in between the engine, and the throttle prop, that way, you can reduce it if the engine starts to fail, so the RPM and its power drop. Pretty simpl actually, isn't it? Versus just faking the RPM property,  you can reduce the power as well, no?

dany93:
@legoboyvdlp 
Hi lego,

For the engine, sorry I'm not the good one. I just have some RL experience on ultralights, with an instructor. Needless to say that I didn't experiment the over-regime just to know how long it lasts...
@tigert, @Juanvvc, @wezzer  @itOutchpods, are real C172 pilots.

About the way to implement this, I remember having suggested to @gilbertohasnofb the property
`volumetric-efficiency `under `/fdm/jsbsim/propulsion/engine[0] or [1]`. Decreasing this property made the engine slow down, or stop. He was looking for this about the carburetor icing.
He added a `volumetric-efficiency-factor` under` /engines/active-engine`.
The default value in JSBSim for volumetric-efficiency is 0.85 for normal functioning.

It might be possible to combine both, for the same effect on the the engine.

legoboyvdlp:
Many thanks... I will try that!

dany93:
See 
Nasal/engine.nas (lines 190 - 197)
Systems/engine.xml (lines 627 - 641)

Nasal/engine.nas
legoboyvdlp: I now have a rpm over redline function, and am now creating the table.

legoboyvdlp:
Thank you Dany! I have now only to create the function -> volumetric efficiency factor calculation and apply the factor to the engine ;)

dany93:
If you want to test its effect, you can do it on almost any JSBsim aircraft (most have a the `volumetric-efficiency` property (under `/fdm/jsbsim/propulsion/engine[0]) which is not driven by a nasal or other code. Which means that you can 'Set' any value in-Sim in the properties.

legoboyvdlp:
@gilbertohasnofb @onox @dany93 
I've just pushed my hopefully second-to-last commit. I now need a formula that increases rpm drop as the damage increases, exponentially, NOT linearly.

It's hard to notice the rpm drop in the dive, but as you pull the nose up, you definitely notice it.

AVOID going to volumetric efficiency 0.0: it causes FG to crash due to division by zero!

I next want to make the engine sieze if damage increases above 1000. It should ideally belch out black smoke from the exhaust and stop working completely :)

If someone could model a broken prop in jsbsim and in 3D **in the future**, it would be interesting to have the prop literally tear off if your RPM is too high.

dany93:
I have issues due to my lack of knowledge in git.

I'm used to test new codes from a branch on https://github.com/c172p-team/c172p-detailed, but I could not get a correct copy of your current code. It merges with my current contain from /c172p-team/c172p-detailed master.
At `git remote show https://github.com/legoboyvdlp/c172p-detailed.git`, I do not see your issue-124 branch.
Same with `git fetch https://github.com/legoboyvdlp/c172p-detailed.git`

I've tried a git pull into a (specially created) issue-124-lego branch.
On git pull `https://github.com/legoboyvdlp/c172p-detailed.git issue-124`, it merges it with my current contain from c172p team master.
At testing, I have /engines/active-engine/volumetric-efficiency-factor = 0.001.
Same for fdm/jsbsim/propulsion/engine[0] or [1]/volumetric-efficiency = 0.001
I think that it is due to my improper merging, I do not have the right files.

Can you (or @onox) tell me how to do?

It would have been easier (at least for me) with a branch created on c172p-team/c172p-detailed (which is our usual procedure).

Sorry

legoboyvdlp:
That's strange... it should not be 0.001, I do not notice that behaviour?

dany93:
Dany wrote:

> I think that it is due to my improper merging, I do not have the right files.

I have doubts about the git procedure that I followed. I think that I have a corrupted version.
My question was: was is the correct git procedure to get your branch content?

onox:
@dany93 Github provides instructions if you click the link to the right of the green "Merge pull request" button:

```
git checkout -b legoboyvdlp-issue-124 master
git pull https://github.com/legoboyvdlp/c172p-detailed.git issue-124
```

The checkout first creates a new branch (called `legoboyvdlp-issue-124`) branched off of `master`. The pull then pulls in the changes from his `issue-124` branch.

Systems/c172p-engine.xml
```diff
@@ -79,6 +79,60 @@
             </clipto>
             <output>/engines/active-engine/rpm-over-redline</output>
         </fcs_function>
+
+        <fcs_function name="damage level calculation">
+              <function>
+                <ifthen>
+                    <eq>
```
onox: Is the `<eq>` needed? Isn't just `<p>` enough (it should be if it's a boolean). Maybe rename it to `damage-allowed`.

c172p-set.xml
```diff
@@ -695,6 +695,10 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <carb_icing_allowed type="bool">false</carb_icing_allowed>
             <auto-start type="bool">false</auto-start>
             <complex-engine-procedures type="bool">false</complex-engine-procedures>
+			<carb-volumetric-efficiency-factor>0</carb-volumetric-efficiency-factor>
```
onox: More tabs here :sob:

Systems/c172p-engine.xml
```diff
@@ -79,6 +79,60 @@
...
+                            <property>simulation/channel-dt</property>
+                        </product>
+                    </sum>
+                    <v>0</v> <!-- this is the value if damage is disabled -->
+                </ifthen>
+            </function>
+            <output>/engines/active-engine/damage-level</output>
+        </fcs_function>
+		
+		<fcs_function>
```
onox: Tabs! :cry: In case you press the actual tab key for indentation: editors provide the ability to replace those automatically with a number of spaces. See the preferences/settings of your editor.

Nasal/engine.nas
```diff
@@ -293,6 +274,15 @@ var carb_icing_function = maketimer(1.0, func {
     };
 });
 
+# =========== engine damage =======================
+var damage_vol_eff_factor = 0.0;
+var damage = getprop("/engines/active-engine/damage-level");
+
+var engine_damage_function = maketimer(1.0, func {
```
onox: Why not convert this function to JSBSim as well? There's also a tab on the next line (282)

onox:
> If someone could model a broken prop in jsbsim and in 3D in the future, it would be interesting to have the prop literally tear off if your RPM is too high.

There's a 3D model of a bended propeller. (Visible if you crash into terrain for example). If the blades are separate objects then you could animate them getting torn off. However, in my experience with FG's submodels, they (high velocity submodels) sometimes become visible too late (several meters from the original spawn position).

dany93:
I followed the Github instructions (which finally is that I did previously and which gave my bad result).
I get a message in the nemo editor (in  French):

>  Please enter a validation message to explain how this merge is
 necessary, especially if it merges an upstream branch updated into a subject branch.

I don't know what to do, I exit from the message editor.
git branch: legoboyvdlp-issue-124
The merging is done (apparently on my master branch as seen from gitk).
It looks like I have a merging of issue-124 branch on my master branch instead of a separate legoboyvdlp-issue-124 branch
Same result,
/engines/active-engine/volumetric-efficiency-factor = 0.001.
fdm/jsbsim/propulsion/engine[0] or [1]/volumetric-efficiency = 0.001

Systems/c172p-engine.xml
legoboyvdlp: Richard helped me with this, he used this in his example. As for the name, alright, will do.

Systems/c172p-engine.xml
legoboyvdlp: Yes, I thought I had done that for each changed file, but obviously made changes afterwards! Old habits are hard to break ;)

onox:
When you pull in stuff from another repository, you do get a merge commit. But that's fine as long as you're in some random test branch.

onox:
You can delete `legoboyvdlp-issue-124` after testing.

dany93:
I've found the glitch!

If you launch the simulator with` Complex engine procedures and failures` **unchecked**, the `volumetric-efficiency` is 0.001. _This point is to be fixed_.

As soon as I check it on, the `volumetric-efficiency` goes to 0.85.
At over-regime, the values change as expected, the `volumetric-efficiency-factor` decreases, `damage-level` increases, `damage-volumetric-efficiency-factor` decreases, max power (RPM) decreases.
At least for what I've observed.

legoboyvdlp:
@onox would you mind checking my code? I've had a look, but I don't see anything glaringly obvious that could cause that behaviour if you start with the property unchecked.

dany93:
@legoboyvdlp 
In c172p-set.xml, you added (lines 711 - 712)
```
			<carb-volumetric-efficiency-factor>0</carb-volumetric-efficiency-factor>
			<damage-volumetric-efficiency-factor>0</damage-volumetric-efficiency-factor>
```
Which gives` carb-volumetric-efficiency-factor = 0 ` at sim launch if 'Complex engines procedures and failures' are unchecked. Hence `volumetric-efficiency-factor = 0.001`.

There was nothing like this in my initial version with @gilbertohasnofb  carb-icing only, and it worked. I guess the factor was declared and set elsewhere.

**Suggestions:** (to be tested, I don't know which is the best even if several can work)
1- 
```
			<carb-volumetric-efficiency-factor>0.85</carb-volumetric-efficiency-factor>
			<damage-volumetric-efficiency-factor>1</damage-volumetric-efficiency-factor>
```
or 2-
```
			<carb-volumetric-efficiency-factor>0.85</carb-volumetric-efficiency-factor>
			<damage-volumetric-efficiency-factor>0</damage-volumetric-efficiency-factor>
```
or 3-
**Nothing at all in c172p-set.xml**, and see how @gilbertohasnofb did with the carburetor icing, which worked. Probably the 0.85 factor was declared and set elsewhere.

_Detail: please do not use tabs for indentation.  Use 4 spaces. Mixing is confusing with some text software (and I think having read that it can cause errors)._

legoboyvdlp:
Setting it to <carb-volumetric-efficiency-factor>0.85</carb-volumetric-efficiency-factor>
	<damage-volumetric-efficiency-factor>1</damage-volumetric-efficiency-factor>

ought to do the trick, since carb-volumetric-efficiency is multiplied by damage-volumetric-efficiency

Systems/c172p-engine.xml
legoboyvdlp: @onox regarding the name, all the other allowed properties use an underscore. 
See engine.xml lines 696 to 717

Nasal/engine.nas
legoboyvdlp: Right.

c172p-set.xml
legoboyvdlp: fixed

Systems/c172p-engine.xml
legoboyvdlp: I checked without <eq>, did not work.

legoboyvdlp:
@gilbertohasnofb I'm ready to merge, if you could test please? Thanks!

legoboyvdlp:
Sorry - couple of bug fixes to do.

legoboyvdlp:
Alright, I'm done!

Systems/c172p-engine.xml
```diff
@@ -121,12 +121,36 @@
...
+                <min>0.0</min>
+                <max>1.0</max>
+            </clipto>
+            <output>/engines/active-engine/damage-volumetric-efficiency-factor</output>
+        </fcs_function>
+
+        <fcs_function name="summer for volumetric efficiency factor">
+            <function>
+                 <ifthen>
+                    <eq>
```
onox: Is this 3-space indentation. Can you really try to just use `<p>` instead of wrapping it in `<eq>`?

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,101 @@
...
+						<property>/engines/active-engine/damage-level</property>
+						<v>0.00166</v>
+					</product>
+				</difference>
+            </function>
+            <clipto>
+                <min>0.0</min>
+                <max>1.0</max>
+            </clipto>
+            <output>/engines/active-engine/damage-volumetric-efficiency-factor</output>
```
onox: Is this property only used in this file? Then you don't need this `<output>`. You can just use `fcs/volumetric-factor-calculation` in the fcs_function below.

These `<output>` properties are mostly only needed when it is used by Nasal listeners for example.

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,101 @@
...
+                    </sum>
+                    <v>0</v> <!-- this is the value if damage is disabled -->
+                </ifthen>
+            </function>
+            <output>/engines/active-engine/damage-level</output>
+        </fcs_function>
+
+        <fcs_function name="volumetric factor calculation">
+            <function>
+				<difference>
```
onox: Tabs here.

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,101 @@
...
+                <sum>
+                    <property>/engines/active-engine/rpm</property>
+                    <value>-2700</value>
+                </sum>
+            </function>
+            <clipto>
+                <min>0</min>
+                <max>10000</max>
+            </clipto>
+            <output>/engines/active-engine/rpm-over-redline</output>
```
onox: You can remove this `<output>` too. Rename the `name` attribute to something like `damage/rpm-over-redline`. Maybe there's a better prefix than `damage/`, you could look in the property tree in FG and see where the most appropriate place for the property is.

onox:
@legoboyvdlp I see a lot of `<output>` elements, but you don't really need those when you chain multiple JSBSim blocks. I recommend to also replace the spaces with a `-`in the `name` attributes and prefix it with `damage/` or `oil/` or something like that.

Take a look at `c172p-heat.xml` and `c172p-damage.xml` where everything has been prefixed with `heat/` and `damage/`.

onox:
My rationale behind prefixing the `name` attributes is to make the JSBSim system sort of an isolated module. Everything is still thrown into one big global property tree unfortunately (you cannot define a list of valid input and output properties for the module/system), so the prefixing is an attempt to keep it isolated as much as possible.

The properties in `/engines/active-engine/` are used by the 3D model, the sounds, flight recorder, web panel, etc.

legoboyvdlp:
I apologize, I never actually knew that about JSB! It makes sense, I will implement your suggestion. As for the tabs, I could have sworn I removed them all :flushed:

onox:
One of the few cases where you have to add an `<output>` is when you want to alias the property in the -set.xml file.

For example, in the P51D I have something like this:

```xml
        <switch name="contact/unit[15]/spark">
            <default value="0"/>
            <output>/gear/gear[15]/spark</output>
```

And then in the -set.xml file I use it like this:

```xml
<bool n="32" alias="/gear/gear[15]/spark"/>
```

dany93:
Sorry, I've just discovered which, I think, is the most realistic feature to simulate the engine seizing up.
`/fdm/jsbsim/propulsion/engine[0] or [1]/friction-hp`
`friction-hp` value is 1.5 by default.
At 60 - 65 (hp), the 180hp engine at full throttle stops.

No more interference with carburetor icing, the initial @gilbertohasnofb code can be restored.

Sorry one more time for being so late, I did not know this.

Another thing:
With or without complex engine procedure and failures  (+ 'Start with engine running', and 'Damage' ON, although probably not important)
```
Nasal runtime error: nil used in numeric context
  at /home/dany/FG/github/c172p/Nasal/engine.nas, line 381
  called from: /home/dany/fgfs2018-2-0-A/install/flightgear/fgdata/Nasal/globals.nas, line 119
```

Systems/engine.xml
```diff
@@ -528,14 +534,38 @@
...
                     <value>1.97</value>
                 </less-than>
-                
+
+                <!-- kill engine if the oil pressure is too low while the engine is running, according to Lycoming manaul 3-6 minimum oil pressure
+                in normal flight is 60psi, we use quite a bit below that to avoid excessive coughing during start. The engine will run for a while
+                probably at these low pressures, but if you have zero, you need to cut the engine fast! -->
+                <and>
+                    <property>/engines/active-engine/running</property>
+                    <equals>
```
onox: Why not replace this `<equals>` with `<not><property>/controls/switches/starter</property></not>` so that you don't need this `ready-oil-press-checker` hack?

legoboyvdlp: It sometimes kept coughing after the starter was disengaged, until the RPM, thus oil pressure had risen to a suitable value.

onox: What about `/engines/active-engine/cranking`?

Nasal/engine.nas
```diff
@@ -143,6 +178,46 @@ var oil_consumption = maketimer(1.0, func {
     }
 });
 
+# Oil Refilling
+var oil_refill = func(){
+    var service_hours = getprop("/engines/active-engine/oil-service-hours");
+    var oil_level     = getprop("/engines/active-engine/oil-level");
+    var refilled      = oil_level - previous_oil_level;
```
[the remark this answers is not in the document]
legoboyvdlp: Should have been obvious... thanks for pointing it out!

legoboyvdlp:
Onox,
I've implemented most of the things that you suggested. As for dany's suggestion, I would like to implement that in the future, in a more complex manner: the engine will not necessarily stop until the damage is extremely high. Wheras, the rpm will drop before the engine stops.

legoboyvdlp:
I will now see if I can move much of the oil simulation to jsbsim.

onox:
How did you come up with the 32.5 psi value? If I do a manual start (without your condition to kill the engine) the oil settles to about 26 PSI.

legoboyvdlp:
I used a value that was reasonably high (the green bar starts at 50psi, as far as I know) without causing the engine to die all the time. I would appreciate input on this value, should I reduce it to the mid 20s?

onox:
The POH section 2-5 (powerplant limitations) says oil pressure has a minimum of 25 psi (20 psi with a certain service kit) and a maximum of 115 psi.

onox:
Figure 2-3 in section 2-6 says a minimum of 25 PSI, normal between 60 - 90 PSI, maximum 115 PSI. With a certain service kit: min 20 PSI, normal 50 - 90 PSI.

dany93:
Dany wrote
> friction-hp value is 1.5 by default.
At 60 - 65 (hp), the 180hp engine at full throttle stops.

@legoboyvdlp wrote:
>  I would like to implement that in the future, in a more complex manner: the engine will not necessarily stop until the damage is extremely high. Wheras, the rpm will drop before the engine stops.

I guess you have understood well that by progressively increasing  `friction-hp`, the engine RPM progressively drops. Exactly as currently with `volumetric-efficiency`.
Or is it me who have not understood well what you are meaning?

legoboyvdlp:
I misunderstood what you are saying, I see what you mean now. I will switch the damage over to that property and restore the icing. Thank you!

legoboyvdlp:
Using a graphical calculator, I came up with this formula for the friction property:
image

legoboyvdlp:
@onox our 172 has a Rochester oil gauge with the 50psi arc minimum limit. Thus, the minimum is 20psi

dany93:
@legoboyvdlp wrote
> Using a graphical calculator, I came up with this formula for the friction property:

Why not? The first thing I checked is that it actually gives 1.5 for x=0.
(I almost did it with no calculator !!  :laughing: )

legoboyvdlp:
yes, because 0.625 * 2.4 = 1.5 :)

@dany93 my main problem is that while rpm reduces well, you normally get rpm overspeed in a dive. I had to slow below 90 knots before the engine would stop.

dany93:
It depends on the factor(s) that you set in the formula, that's a compromise.
I don't see this as a fundamental problem. IRL when the engine seizes up, if you add a prop windmilling moment it will seize up later. But take care to remain in realistic behavior. The pilot will not play like this and the true issue is that the engine is no longer able to pull the aircraft, and possibly ends by a stop. Even a lack of power can be dangerous.

dany93:
At first test it seems (at least) interesting. If the engine is too damaged, it goes on to run as long as you keep diving but you can no longer hold the aircraft in horizontal flight. Or the engine stops and cannot be restarted.

legoboyvdlp:
If you all are happy with it, I'm finished for now. It has taken a while, but at last it is done :)

dany93:
I will test it a bit more (however it is simple enough to observe).
Rather than diving on purpose with full throttle (which is a weird practice), I think the most likely is an over regime in horizontal or slightly climbing flight. In this case, after some time, the engine will no longer be able to maintain horizontal flight.

At least qualitatively it is a real improvement.
The remaining question is how much time the engine holds for a given over regime. But the code is here, easy enough to change later.

Useless to say that we are in a domain where people do not know well how things are going to be.

dany93:
@legoboyvdlp wrote:
> my main problem is that while rpm reduces well, you normally get rpm overspeed in a dive. I had to slow below 90 knots before the engine would stop.

After more test, I may understand better what you are meaning. In horizontal flight, you can be in over-regime at full throttle as long as the engine is not too damaged. When enough damage level  (partly seized up) is reached, the engine RPM slows down and you stop damaging more (if damaging is only due to over-regime). The result is that you never reach the total stop.

That makes me think that IRL, when the engine seizes up, not only the available power decreases but the temperature rises. And this **temperature rising (even at moderate RPM <  2700)** might well be the final cause of engine failure. Engine which will probably be impossible to restart because partly seized up.

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,100 @@
             </function>
             <output>/fdm/jsbsim/propulsion/engine[1]/air-intake-impedance-factor</output>
         </fcs_function>
-        
+
+    </channel>
+
+    <channel name="damage">
+        <fcs_function name="/damage/rpm-over-redline">
```
onox: Path should be relative. Since this function is in the `c172p-engine.xml` file, you might as well change the attribute to `name="engine/rpm-over-redline"`. (I just mentioned `damage/` and `heat/` as an example)

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,100 @@
...
+                <min>0</min>
+                <max>10000</max>
+            </clipto>
+        </fcs_function>
+
+        <fcs_function name="/damage/damage-level">
+              <function>
+                <ifthen>
+                    <eq>
+                        <p>/engines/active-engine/damage_allowed</p>
```
onox: Can you change this to `damage-allowed`?

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,100 @@
...
+                    <value>-2750</value>
+                </sum>
+            </function>
+            <clipto>
+                <min>0</min>
+                <max>10000</max>
+            </clipto>
+        </fcs_function>
+
+        <fcs_function name="/damage/damage-level">
```
onox: Change attribute to `name="engine/damage-level"`

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,100 @@
...
+                                  700   49
+                                  800   64
+                                  900   81
+                                 1000  100
+                                 1100  121
+                                 1200  144
+                                 1300  169
+                                 1400  196
+                                 1500  225
+                                 9999  300
```
onox: This row is a bit odd. Why 9999?

c172p-set.xml
```diff
@@ -601,6 +602,11 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <contamination_allowed type="bool">false</contamination_allowed>
         </fuel>
     </consumables>
+	
+	<damage>
+		<rpm-over-redline>0.0</rpm-over-redline>
```
onox: Tabs here.

c172p-set.xml
```diff
@@ -678,6 +684,17 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <oil-level type="double">7.0</oil-level>
             <oil-lacking type="double">0.0</oil-lacking>
             <oil_consumption_allowed type="bool">false</oil_consumption_allowed>
+            <oil-powergain-factor type="double">1</oil-powergain-factor>
+            <winter-kit-installed type="bool">false</winter-kit-installed>
+            <oil-temperature-env-diff type="double">0</oil-temperature-env-diff>
+            <oil-compensated-temperature-degf type="double">60</oil-compensated-temperature-degf>
+            <external-heat>
+               <enabled type="bool">false</enabled>
```
onox: Indentation needs two extra space.

gui/dialogs/c172p-oil-160.xml
```diff
@@ -65,6 +65,10 @@
                     <command>dialog-apply</command>
                     <name>c172p-oil-slider-160</name>
                 </binding>
+                 <binding>
```
onox: Unindent this 1 space.

gui/dialogs/c172p-oil-160.xml
```diff
@@ -84,6 +88,22 @@
         </group>
     </group>
     
+     <group>
+        <layout>hbox</layout>
```
onox: And since you're already fixing the last pieces of wrong indentation, you might add a space to this piece of code as well...

gui/dialogs/c172p-oil-180.xml
```diff
@@ -85,6 +89,22 @@
         </group>
     </group>
     
+     <group>
+        <layout>hbox</layout>
```
onox: Give this group some :heart: by adding some spaces.

Systems/external-heat.xml
```diff
@@ -0,0 +1,88 @@
...
+                <min>0</min>
+                <max>77</max> <!-- 68°F = 20°C --> <!-- TODO: I currently don't know how much heat the heater outputs! currently this gives a limit of -40°C after which priming gets really hard -->
+            </range>
+            <output>/engines/active-engine/external-heat/applied-degF</output>
+        </aerosurface_scale>
+        <fcs_function name="Oil temp plus external heat">
+            <function>
+                <sum>
+                    <property>/engines/active-engine/external-heat/applied-degF</property>
+                    <property>/engines/active-engine/oil-temperature-degf-indicated</property> <!-- compensated for env-temp and low oil, see engine.xml -->
```
onox: Isn't this faking the indicated oil temperature? That is, the actual engine temperature is without the external heat applied? Does JSBSim provide a way to actually add heat to the engine?

Systems/c172p-engine.xml
legoboyvdlp: I was not sure if it would just cut off at 1,500 and 225 damage levels per second, or if it would continue being interpolated past 1,500 rpm over redline.

Systems/external-heat.xml
legoboyvdlp: This is how it was done on the 182. I will look into it more.

legoboyvdlp:
@dany93 you are exactly right. I will be trying to simulate the temperature effect tomorrow. 
I will probably add some effects (black smoke, sounds) when the engine is severely damaged (siezes completely due to overtemperature or lack of oil). I thought I was finished, but I can always improve :smile:

Systems/c172p-engine.xml
onox: It should cut off at 1500.

onox:
> I will probably add some effects (black smoke, sounds) when the engine is severely damaged (siezes completely due to overtemperature or lack of oil)

That would be cool :smile:

legoboyvdlp:
@dany93 
I'll be making a seperate calculation for the effect of high temperature on engine wear (high viscocity -> less lubrication -> more wear, but long-term).
Should I just sum the two friction calculations and output to friction-hp?

Systems/c172p-engine.xml
legoboyvdlp: To be consistent with the other properties (line 696-707 engine.xml), no.

gui/dialogs/c172p-oil-160.xml
legoboyvdlp: or remove a space, in fact 😛

legoboyvdlp:
System transferred to relative properties ;)

dany93:
> ..... effect of high temperature on engine wear (high viscocity -> less lubrication -> more wear, but long-term).
Should I just sum the two friction calculations and output to friction-hp?

higher or lower viscosity?
As it adds an other kind of friction, I'd say that's correct. Of course, you will get a loss of available power.

legoboyvdlp:
image

Not quite right, but getting there
Can I increase the intensity of the smoke based on damage-level?

legoboyvdlp:
also, poor 172, getting a punishment tonight 😄

legoboyvdlp:
image

legoboyvdlp:
image

legoboyvdlp:
Mission accomplished: see the RPM gauge:
image

legoboyvdlp:
Fixed some trailing spaces, added smoke, and oil over-temperature now affects damage :)

legoboyvdlp:
Oops -- sorry for the merge conflict! Hopefully it won't be too much bother to resolve it...

I am now making the smoke more realistic in size.

legoboyvdlp:
@gilbertohasnofb @onox @wlbragg I'm happy to merge this, if you could check it please, and review?
I don't think there's anything left to do, but if there is let me know.

dany93:
I've just updated my fgfs git version to 01f8404 (from 22 march 2018 13:52).
I don't know if it comes due to this FG version, but I had never seen this before.

With this version of the C172 (issue-124), just after starting the engine, there is a brutal and strong enough nudge which makes it yaw leftwards and, very often, break one or several gears.
It does not seem to depend on the way the engine is started.

With this new fgfs version, I've tried several times with other C172 versions (my perso version, about 1 month old) and my master one (e98a983c6, 10 mar 2018), I haven't observed that.

legoboyvdlp:
That's wierd. I don't see how I could introduce that.

Models/Effects/damage-smoke/smoke.xml
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0"?>
+
+<PropertyList>
+
+    <params>
+        <engine>
+            <damage-smoke>/fdm/jsbsim/engine/damage-level</damage-smoke>
```
onox: This doesn't work over MP. Let's use the new 2017.2 protocol:

Can you change this to `<damage-smoke>sim/multiplay/generic/bool[0]</damage-smoke>`.

Then replace the whole `<and>` block in the `<condition>` below with just `<property alias="/params/engine/damage-smoke"/>`.

In the -set.xml file, add to the list of generic properties (at the end of the block):

```xml
<!-- Effects -->
<bool n="0" alias="/aircraft/effects/damage-smoke-visible"/>
```

Then create a JSBSim `<switch>` block somewhere and test for `/fdm/jsbsim/engine/damage-level GT 100` and `/engines/active-engine/running EQ 1`. Make sure to add `<output>/aircraft/effects/damage-smoke-visible</output>`.

onox: Also add in the -set.xml file:

```xml
<aircraft>
    <effects>
        <damage-smoke-visible type="bool">false</damage-smoke-visible>
    </effects>
</aircraft>
```

Models/Effects/damage-smoke/smoke.xml
```diff
@@ -0,0 +1,119 @@
...
+        </engine>
+    </params>
+
+    <particlesystem>
+        <name>startup</name>
+        <texture>smoke.png</texture>
+        <emissive>false</emissive>
+        <lighting>false</lighting>
+
+        <condition>
```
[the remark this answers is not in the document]
legoboyvdlp: for some reason Git didn't pick up I fixed this.

legoboyvdlp:
Fixed the merge conflict.

If there are no further problems, this can be merged.

Systems/c172p-engine.xml
```diff
@@ -62,7 +62,121 @@
...
+                </product>
+            </function>
+            <clipto>
+                <min>0.001</min>
+                <max>10000</max>
+            </clipto>
+            <output>engine/friction-hp</output>
+        </fcs_function>
+
+        <switch name="propulsion/engine[0]/friction-hp">
```
onox: This did not work as intended unfortunately. So you would have to change it back to `name=engine/engine-0-friction-hp` and add `<output>propulsion/engine[0]/friction-hp</output>`.

Same for the other engine.

onox:
@gilbertohasnofb What do you think of the smoke?

Why not reuse the exhaust smoke you see when you start the engine?

onox:
When I enable the externeal heater (in Ground Equipment), it shows a tube that goes through the fuselage. Is it meant to be connected to the exhaust or the inlets instead?

onox:
@legoboyvdlp What does "Oil in service (hrs)" mean? Is it the time until I need to refill the oil or the last time that the oil was replaced? I flew for something like 2 hours and it still said 0 after I landed. The property said 0.00027.

Maybe it should be done per engine? (so in `/engines/engine[x]/oil-service-hours`)

Does it respect "Complex engine procedures and failures"?

Another thing: if I click the "Repair" button, the damage doesn't get reset to 0.

@gilbertohasnofb Do we want to reset it to 0 when the user does an autostart?

legoboyvdlp:
> Does it respect "Complex engine procedures and failures" checkbutton?
It does.
> Another thing: if I click the "Repair" button, the damage doesn't get reset to 0.
It should now.

legoboyvdlp:
OK, the service hours (hours in service) is broken then.

dany93:
About the jolt (sometimes breakage) leftwards at engine start.
It is 100% reproducible in the same conditions.

On 22 mar 2018, @legoboyvdlp wrote
> That's wierd. I don't see how I could introduce that.

I've made tests to localise which file is responsible and how, when, it happens.

The jolt happens a few seconds after engine start (Auto or manual start, even start in the sim) when the aircraft begins to roll, still slowly.
With the 180 hp engine and 'Autostart', it happens 2 to 6 seconds after the engine start, left at idle.
In case of manual start with throttle at 0%, it does not happen. I have to add some throttle, make it roll slowly, to have it happen.
With the 160 hp engine and 'Autostart' (probably because the aircraft does not roll enough spontaneously), I also have to add some throttle to make it roll slowly.

Tests on the files:

1- Deactivate `engine.nas` ==> same jolt leftwards.
==> `engine.nas` is not the (direct) responsible

2- `c172p-engine.xml` (in Systems/ )
Delete all lines 68 - 160

```
    <channel name="damage">
        <fcs_function name="/damage/rpm-over-redline">
 (....)

        <fcs_function name="/damage/damage-level">
(....)
            <output>/fdm/jsbsim/propulsion/engine[1]/friction-hp</output>
(....)
```


==> **the jolt leftwards does NOT happen**.
**c172p-engine.xml is the direct responsible**.

Something bothers me by observing the properties.
The non-active engine has its `friction-hp = 0`. It comes from your code (only when lines 68 -160 in c172p-engine.xml are present).
I'm not sure whether it is important or not, but it is not justified and I do not like it. An engine with no friction at all can give a divergence (division by 0 or RPM divergence). And despite our C172 has one engine active at a time, the calculations sum the effects the 2 possible engines (giving no effect from the inactive one in _normal_ conditions).
Thus, the first thing to do IMO is at least to keep the basic `friction-hp = 1.5` for both engines whatever the active one is. And to test if it changes something.

legoboyvdlp:
Alright. I will fix this shortly. Thank you for reporting and suggesting how to fix it!

legoboyvdlp:
@dany93 
With active engine 160hp:
image

image


As seen, both friction properties are equal to 1.5. So I don't see how this could cause the problem?

legoboyvdlp:
I also just tested, and with the 180hp engine I rolled forward slowly, until I gently rolled off the right side of the runway due to a strong wind. I did not experience the jolt, even with autostart.

legoboyvdlp:
The 160hp also does not have any jolt.
Note, this was on asphalt. Perhaps grass is different.

legoboyvdlp:
I took off at Shoreham (grass) with 160hp, no problem.

dany93:
To check what I mean about the 'jolt' and friction-hp = 0 bug in your initial version,
do
`git checkout ff6fb309dbe56` ('switch to friction-hp', 2018-03-17)
and test it in the sim, see the props.

You (git) will be in 'detached HEAD',
Come back by
`git checkout issue-124` (or your own branch name)

legoboyvdlp:
OK, at least it is now fixed :)
And I think I saw it one time, but when I tested again a few days later, it seemed to have fixed itself...?
Anyway, ok, I will try and make it reset if save-sate is disabled. I will see how it is done with other things.

legoboyvdlp:
I just searched all the code for save-state, and I cannot see where is is used:
see: https://github.com/c172p-team/c172p-detailed/search?utf8=%E2%9C%93&q=save-state&type=

wkitty42:
On 03/29/2018 12:54 PM, Jonathan Redpath wrote:
> I just searched all the code for save-state, and I cannot see where is is used:


it is in the -set.xml file... search for aircraft-data ;)

gilbertohasnofb:
> What do you think of the smoke?
> Why not reuse the exhaust smoke you see when you start the engine?

Agreed, that would look better.

> Do we want to reset it to 0 when the user does an autostart?

Yes, I think we do. Autostart should solve all your problems, including any damage.

Also, I don't like that the toggle for enabling damage just says "Damage". It should say "Enable Damage (wings and engine)" or something like that.

onox:
Let's rename it to "Enable damage" then.

wlbragg:
Everyone is OK with one control over both types of damage. I only ask because I can see where someone might want the engine damage potential but not the structural damage.

gilbertohasnofb:
@wlbragg I think one toggle is fine, my issue was just with the label. IMO "Enable damage" does solve it.

legoboyvdlp:
Alright. 
Once all the bugs and problems and requests are done, could you merge this, then switch to Enable Damage, not do it now?
To avoid duplicate work.

gilbertohasnofb:
@legoboyvdlp or could you perhaps add it to this PR? It's super simple, just replace line 111 of `gui/dialogs/aircraft-dialog.xml` for:

```
                <label>Enable damage</label>
```

legoboyvdlp:
Right!
Do you prefer
<label>Enable damage</label>
or 
<label>Enable Damage</label>

?

gilbertohasnofb:
@legoboyvdlp I believe we are capitalizing just the first word in that dialog, but could you double check that when doing the change? Just keep things consistent.

gilbertohasnofb:
And thanks for the fix.

legoboyvdlp:
Confirmed, yes :)

legoboyvdlp:
Gilberto,
My smoke is based off the engine start smoke. It is impossible to make smoke like that while flying at a high speed: as speed increases, the trail gets longer. I think that my smoke is probably more appropriate than your smoke for this purpose, but if you prefer I can change it.

legoboyvdlp:
The only outstanding bug is the oil service hours.

legoboyvdlp:
I'm trying  service_hours_new = (service_hours / 3600) + 1; # add one second service time
now.
Hopefully it will fix it!

legoboyvdlp:
Unfortunately, not.
This time, it went to 1.00027 or something, but remained stuck there.
The oil consumption property is also static. I think something is not being updated.

onox:
@legoboyvdlp Maybe you can use the same technique that is used for the engines in `Nasal/engine.nas` (lines 15 to 51)?

legoboyvdlp:
Thank you for the suggestion, I'll try it!

onox:
@legoboyvdlp Did the technique from `engine.nas` help you to fix the problem? Otherwise the keeping track of the service hours should be removed (or is this vital? am I missing something?)

legoboyvdlp:
Not really. I think I'll remove it pending investigation, so this can be merged.

Nasal/engine.nas
```diff
@@ -128,24 +125,10 @@ var oil_consumption = maketimer(1.0, func {
             consumption_qph = consumption_qph * 1.3;
         }
 
-        # Consumption also raises with oil in service time (lower viscosity => more friction)
-        # (Oil should be changed at 50 hrs!)
-        # See: http://www.t-craft.org/Reference/Aircraft.Oil.Usage.pdf
-        # Hours:        0 |    10 |    25 |  50   |    75
-        # Add Qts/hr:   0 |  0.02 | 0.125 | 0.5   | 1.125
```
onox: You could probably leave these comments in the code in case it gets implemented in the future.

legoboyvdlp:
@onox alright, can do.
As for that, the previous oil level variable is changed below as far as I know, to equal to old oil level.

onox:
@legoboyvdlp I've chopped off the last 2 commits here locally, and then replaced:

```javascript
service_hours_new = (service_hours + 1)/3600;
```

with:


```javascript
service_hours_new = (service_hours * 3600 + 1) / 3600;
```

That seems to fix the service hours.

onox:
I've also replaced `%d` with `%.2f hours` in the GUI (and removed the `(hrs)` part)

onox:
Another problem I noticed is that `/engines/active-engine/oil-final-temperature-degf` (in `Systems/external-heat.xml`) is always 60 (because `oil-temperature-degf-indicated` used in the sum expression always has the value 60 defined in the -set.xml file)

Nasal/c172p.nas
```diff
@@ -12,6 +12,9 @@ var autostart = func (msg=1) {
     # Reset battery charge and circuit breakers
     electrical.reset_battery_and_circuit_breakers();
 
+	# Repair engine damage
+	setprop("/fdm/jsbsim/engine/damage-level", 0);
```
onox: This should be moved to `reset_all_damage` function in `Nasal/physics.nas`

onox:
Another thing: the damage-level keeps increasing even when RPM = 0. Should it instead only increase if RPM > 0?

legoboyvdlp:
I'm not sure how I missed that last one. 

Thanks for fix8ng the service hours!

Apologies for taking so much time over this... it is a learning esperience, but maybe sometime, in the next decade it will be finished ;)

legoboyvdlp:
@onox

gui/dialogs/c172p-oil-160.xml
```diff
@@ -93,23 +93,23 @@
...
         <text>
-            <label>Oil in service (hrs): </label>
+            <label>Oil in service: </label>
             <halign>left</halign>
         </text>
         <text>
             <label>123</label>
             <halign>left</halign>
-            <format>%d</format>
+            <format>%.2f</format>
```
onox: can you add ` hours` here?

onox:
Thanks for making the changes @legoboyvdlp. Could you strip commits 4 to 7, together they change nothing, so it's pointless to have them in your git history.

Do `git rebase -i HEAD~7`, you should see 7 lines containing:

```
Remove service-hours.
Re-add comment
Revert "Re-add comment" 
Revert "Remove service-hours."
Restore service hours, add warning in dialog
Move repair engine code to physics.nas
Fix the text in the oil dialog
```

Remove the top 4 lines then save and quit the editor. When you do `git log` you should see the 4 commits have disappeared. Then `git push --force` to force push your new history to Github.

legoboyvdlp:
@onox , Roger!

onox:
Thanks. In `Systems/c172p-engine.xml` you have tabs on lines 134, 165, 173. In `gui/dialogs/aircraft-dialog.xml` you have tabs on lines 168 to 172.

onox:
@dany93 @gilbertohasnofb When there is some damage (damage-level > 0 and friction-hp > 0), should the damage threshold (the property that affects the rpm-over-redline property) be reduced a bit? Otherwise you can just go full RPM while flying at a certain altitude, get some damage (which reduces the current engine RPM a bit) until the RPM has been reduced enough that the rpm-over-redline becomes 0 again.

This basically means you get some damage early in the flight, but you get to keep flying at 100% throttle (just with a little bit lower RPM).

onox:
I also notice that I'm flying horizontally with a negative pitch and AoA (-1.9 deg). (gross weight 1844 lbs and CoG 38.14 in). AoA should be positive?

onox:
@legoboyvdlp When I uncheck the "Enable damage" checkbox in the settings, the damage level should be reset to 0.

gilbertohasnofb:
> When there is some damage (damage-level > 0 and friction-hp > 0), should the damage threshold (the property that affects the rpm-over-redline property) be reduced a bit? Otherwise you can just go full RPM while flying at a certain altitude, get some damage (which reduces the current engine RPM a bit) until the RPM has been reduced enough that the rpm-over-redline becomes 0 again.

@onox I am really not familiar what of these two behaviours would be the most realistic, but it might make sense that a damaged engine will be more sensitive to RPMs that would otherwise be in the normal operating range. @dany93 do you have any idea about this?

dany93:
[On 17 April 2018, onox wrote](https://github.com/c172p-team/c172p-detailed/pull/988#issuecomment-382104069):
> When there is some damage (damage-level > 0 and friction-hp > 0), should the damage threshold (the property that affects the rpm-over-redline property) be reduced a bit? Otherwise you can just go full RPM while flying at a certain altitude, get some damage (which reduces the current engine RPM a bit) until the RPM has been reduced enough that the rpm-over-redline becomes 0 again.

[On 18 march 2018, dany wrote](https://github.com/c172p-team/c172p-detailed/pull/988#issuecomment-374018237):
_(I copy it because github hides many messages)_
> In horizontal flight, you can be in over-regime at full throttle as long as the engine is not too damaged. When enough damage level (partly seized up) is reached, the engine RPM slows down and you stop damaging more (if damaging is only due to over-regime). The result is that you never reach the total stop. Unless you dive on purpose, which is not a realistic behavior.

>That makes me think that IRL, when the engine seizes up, not only the available power decreases but the temperature rises. And this temperature rising (even at moderate RPM < 2700) might well be the final cause of engine failure.

wlbragg:
CAUTION: physics.nas has conflict, please resolve prior to merging so nothing gets overwritten.

dany93:
@onox wrote
> I also notice that I'm flying horizontally with a negative pitch and AoA (-1.9 deg). (gross weight 1844 lbs and CoG 38.14 in). AoA should be positive?

Cambered wings have a positive lift even at zero AoA. Here, the `Lift_due_to_alpha` becomes zero near -0.05 rd (-3 deg).

legoboyvdlp:
The damage threshold will reduce exponentially using an offset of up to 250RPM at 500 damage.

legoboyvdlp:
Using the online interface, I fixed the merge conflict. It had the side-effect of merging the **entire** master in, but hopefully that won't be an issue.

Nasal/c172p.nas
legoboyvdlp: Done. Could you remove this to remove the thing that says you are still requesting changes?

legoboyvdlp:
Looks like I fixed that bug where damage kept increasing with rpm 0 :)

legoboyvdlp:
@onox my attempt to add an offset to the damage threshold has a serious bug: if you throttle back, it does not respond to the throttle, the rpm offset continues to increase (the rpm decreases, but damage continues to increase, and the engine/rpm-over-redline-offset keeps increasing).

I will push it commented out, could you check it please?

onox:
I'll take a look at it in the weekend.

onox:
@legoboyvdlp The problem you described is causes by the last commit? Could you remove/revert it? (+ keep the <eq> that you added)

@gilbertohasnofb @dany93 Anything else that prevents this PR from getting merged?

gilbertohasnofb:
@onox not that I am aware of, but I haven't followed this PR too closely. But if @legoboyvdlp is done and you think it's all fine, then merge it by all means

dany93:
I tested it and it works, got it damaged via over-RPM up to engine failure (engine slowed, lost power, then stopped).

In the console I have (had it for a long time if I remember well)
```
Nasal runtime error: nil used in numeric context
  at ......c172p/Nasal/engine.nas, line 388
  called from:....fgfs2018-2-0-B/install/flightgear/fgdata/Nasal/globals.nas, line 119
```

dany93:
After more thinking,

@onox wrote
> I also notice that I'm flying horizontally with a negative pitch and AoA (-1.9 deg). (gross weight 1844 lbs and CoG 38.14 in). AoA should be positive?

@dany93 wrote

> Cambered wings have a positive lift even at zero AoA. Here, the `Lift_due_to_alpha` becomes zero near -0.05 rd (-3 deg).

What I think is that the manufacturers do their aircraft to fly close to 0 deg at the best cruise airspeed (less drag), which would give slightly negative at higher speeds.  (just a guess.)
In JSBSim, this resulting angle seems sensitive to things  which are not always very clear...

legoboyvdlp:
Yep, I'll be done. I'll just fix that bug, then this can be merged.

legoboyvdlp:
@onox it would be better to fix the bug instead of removing the offset since otherwise you get this behaviour:

_This basically means you get some damage early in the flight, but you get to keep flying at 100% throttle (just with a little bit lower RPM)._

Have you any suggestions?

onox:
I'll take a look @legoboyvdlp

onox:
Rename the following file from .txt to .patch:

[0001-Fix-engine-RPM-redline-offset.txt](https://github.com/c172p-team/c172p-detailed/files/1972852/0001-Fix-engine-RPM-redline-offset.txt)

Then apply it with `git am 0001-Fix-engine-RPM-redline-offset.patch`

legoboyvdlp:
Many thanks.

onox:
@legoboyvdlp Have you been able to apply the patch?

legoboyvdlp:
I'm afraid not - I was rather busy this week with a few exams. I'll get round to it tomorrow if possible.

legoboyvdlp:
@onox I applied your patch and fixed the merge conflicts. Should be ok now.

dany93:
If there were no functional changes since [my last test](https://github.com/c172p-team/c172p-detailed/pull/988#issuecomment-386056658) (02 May 2018, 7day ago) I have no objections.

dany93:
Checked with your last https://github.com/c172p-team/c172p-detailed/pull/988/commits/b6b6a718b034b613e539efb6b48b2ac827f3fc93 commit.
Correct. I no longer have the Nasal runtime error message.

Like you, I have the 
```
FGPropertyManager::GetNode() No node found for hydro/active-norm
Could not find at least one of the following objects for animation: 'PowerSwitch'
etc...
```
messages

wlbragg:
>I do have the following errors:

Investigating!

Nasal/engine.nas
```diff
@@ -357,6 +357,16 @@ var update = func {
             }
         }
     }
+	
+	if (getprop("/controls/engines/active-engine") == 0)
```
onox: Tabs.. again...

onox: I think you replace these 2 if's with just:

```javascript
var active_engine = getprop("/controls/engines/active-engine");
var rpm = getprop("/engines/engine", active_engine, "rpm");
```

wlbragg:
>FGPropertyManager::GetNode() No node found for hydro/active-norm

It's the "which came first, the chicken or the egg" syndrome. @onox, is it worth adding the fdm/jsbsim/hydro/active-norm property definition to c172p-set.xml to eliminate this error?

I'm still researching the other warnings.

onox:
If you really want to have that warning disappear, I don't think there's any other way.

wlbragg:
>I'm still researching the other warnings.

I have no explanation for any of the the following
Could not find at least one of the following objects for animation: 'NavUse100M'
Could not find at least one of the following objects for animation: 'NavUse10M'
Could not find at least one of the following objects for animation: 'NavUse1M'
Could not find at least one of the following objects for animation: 'NavUse100K'
Could not find at least one of the following objects for animation: 'NavUse10K'
Could not find at least one of the following objects for animation: 'NavStby10M'
Could not find at least one of the following objects for animation: 'NavStby1M'

However these two I think will be gone after PR #1094 and #1094 are merged 
Could not find at least one of the following objects for animation: 'PowerSwitch'
Could not find at least one of the following objects for animation: 'indicator.dots'

I will define fdm/jsbsim/hydro/active-norm in c172p-set.xml to eliminate that warning (already tested and verified).

onox:
These warnings are not really relevant for testing this PR though.

@wlbragg These warnings appear because you renamed several objects (".001" was added) in `kx165-1.ac` and `kx165-2.ac` in a commit (on April 29th) with the message "Absolute re-position"

wlbragg:
<These warnings appear because you renamed several objects

Oops, that was a  mistake. Thanks for finding it. I'll fix it in #1095

legoboyvdlp:
Looks like this is finished so if nobody has any objections feel free to go ahead and merge tomorrow once I fix that tab?

onox:
@legoboyvdlp Yeah, will merge it once that tab has been removed :smiley: You can `git commit --amend` and then force push if you want. (Make sure you do `git diff --cached` to check what you're going to amend to the last commit)

legoboyvdlp:
@onox I did as you said,
@gilbertohasnofb this is ready to merge, since you are online ;)

legoboyvdlp:
Standby, it didn't actually change anything, see https://github.com/legoboyvdlp/c172p-detailed/blob/issue-124/Nasal/engine.nas line 360.

legoboyvdlp:
All good now.

legoboyvdlp:
: grinning:

After 3 months, finally! Thanks for your patience and help ;)

onox:
It was a long journey indeed :smiley:

gilbertohasnofb:
@legoboyvdlp congrats on your PR, lego! :clap:

PR #989, opened 2018-02-24
Add force and moment induced by AI wakes.

bcoconni:
Hi there,

I added a feature to FlightGear v2017.3.1 that allows aerodynamic interaction between the aircraft and AI models. The feature is briefly described in a [dedicated article](http://wiki.flightgear.org/AI_wake_turbulence) in the FlightGear wiki.

All you need is to add external reactions that models the induced aerodynamics loads on the aircraft. The resultant force and moment will superpose to the rest of the forces and moments computed by the FDM.

The computations made by FlightGear assume that the force is located at the intersection between the main wing leading edge and the aircraft plane of symmetry. This is important for the resulting moment to be correctly computed. I tried to estimate that position for the c172p at x=-4.7in and z=50in but this might not be quite accurate. Let me know if that needs some correction.

The c172p being the flagship of FlightGear I would be honored that you include that feature as an example for other aircrafts modelers.

Thanks.

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -1901,12 +1901,12 @@
             </direction>
         </force>
         <force name="ai-wake" frame="BODY">
-          <location unit="IN">
-            <!-- Center of the main wing leading edge -->
-            <x>-4.7</x>
-            <y> 0.0</y>
-            <z>50.0</z>
-          </location>
+            <location unit="IN">
+                <!-- Center of the main wing leading edge -->
+                <x>32.0</x>
+                <y> 0.0</y>
+                <z>52.5</z>
+            </location>
         </force>
         <moment name="ai-wake" frame="BODY"/>
     </external_reactions>
--- a/c172p.xml
+++ b/c172p.xml
@@ -1900,6 +1900,15 @@
                 <z>1.0</z>
             </direction>
         </force>
+        <force name="ai-wake" frame="BODY">
+          <location unit="IN">
+            <!-- Center of the main wing leading edge -->
+            <x>-4.7</x>
+            <y> 0.0</y>
+            <z>50.0</z>
+          </location>
+        </force>
+        <moment name="ai-wake" frame="BODY"/>
     </external_reactions>
 
     <system file="bushkit"/>


dany93:
Hi @bcoconni,

_(Assuming the reference axes are those of the FDM)_
The 3D coordinates are hard to accurately locate, but I'd rather evaluate:

For the wing leading edge
x = 32 in (positive)
z = 50 to 55 in
If I've understood well...

In the FDM,
The propeller is at
x = - 37.7 in
z = 26.6 in

Empty CG
x = 39.06 in

Nose wheel contact (last, from @wlbragg )
x = -6.8 in      (cannot be close to the wing leading edge)
z = -20.3 in

Main wheel contacts
x = 58.2 in
z = -15.3 in or  -15.1 in

gilbertohasnofb:
@bcoconni thanks for the PR. Just one small thing: please use four spaces when indenting.

bcoconni:
Thanks for the feedback. So I updated the patch with a new commit:
- The force location has been updated according to @dany93 data
- The code indentation is using 4 spaces according to @gilbertohasnofb  request.

wlbragg:
@bcoconni I've been wanting this added to the c172p, thank you so much for taking care of this.

dany93:
Agreed.
Thank you @bcoconni .

onox:
I guess everyone is OK with merging this.

wlbragg:
@dany93 

The c182 team, @hbeni to be specific used our numbers for this when they added the AI wake effect, see:
https://github.com/HHS81/c182s/pull/468

Where did the numbers we sued come from? Why aren't they the mesuments as shown on the image I posted to the c182 issue above?

wlbragg:
Referencing  any gear contact points seems wrong to me as I set those based solely on trial and error. As far as using the FDM prop position as the reference, I now see where the numbers came from.
What is the propeller point referencing?

wlbragg:
Where did @bcoconni get the value x=-4.7in?

dany93:
@wlbragg 

I had a hard time trying to make the connection between the Blender and FDM coordinates.

Values can be approximate, I used a C172P sketch from a POH for this, that I tried to scale.

The C172S 552SP POH, Section 6 Weight and balance, PDF.148/384, gives a few values.

C172S-POH-DATUM
The reference DATUM is the FIREWALL FRONT FACE, LOWER PORTION.

The FG FDM takes this DATUM for its 0 reference.

Unfortunately, the Blender model is offset.

What I got:
Offset in X : about 34" (0.86m).
X(FDM) = X(Blender) + 34"

Z(FDM) = Z(Blender) + 28" (very approximate)

pitch offset about 3°. (FDM higher at the tail)

Perpendicular part location of the wing leading edge: 
From the POH (sketch above) : X(FDM) = 25.90" = 0.66m
X(Blender) = -0.157m = -6.18"
(-25.90" instead of -6.18+34 = 27.82, which gives 1.92" disagreement. I think due to scaling errors in the sketch.

Imperfect... I hope it can help.

Better to have a good agreement between the 3D model and the FDM, but the important are the FDM relative locations.
However, I agree, things would have been much easier if the 3D model had had the same references as the POH. And naturally, the FDM would have taken the same ones.

wlbragg:
@dany93 interesting that after many years of working on this aircraft I finally learned why gear bogie positions and contact points are nowhere near the model reference points. Not only this aircraft but virtually all aircraft I ever worked on. Not only that but that they were all slightly different. I asked many times and never got an answer. In fact it was probably one of the first serious questions I had when I started down the development road. Had I understood this years ago when doing the work, it would have been proportionately easier. I think the amphibious tuning would have made more sense as well.

I still can't get over why anyone would do it this way and not make everything match, It seems like a recipe for failure and something I would never do as a programmer unless it was required.

Anyway, thank you for finally providing me a glimpse of what I was missing. :)

dany93:
@wlbragg 
You are welcome. Glad that it could also help in more than this aircraft alone.

dany93:
@wlbragg 
It gives something like this.
(Qualitative, approximate, not to be scaled)
cessna-172p-3D-FDM-axes

dany93:
@wlbragg 
I never could find where these offsets are given.
The only general offset I find is in Models/c172-common.xml, lines 8-9,
```
    <offsets>
        <pitch-deg>-3.0</pitch-deg>
        <z-m>-0.065</z-m>
    </offsets>
```
Instead of about 28" = 0.71m in Z.
These offsets are really a mess...

wlbragg:
>I never could find where these offsets are given.

It's like when the modeler made the model they just picked a spot and said, OK this is where I'm starting at?

I have no idea why we are using a z offset here? That makes no sense other than someone got lazy and took a quick way to tune something. It could very well have been me.

Do you know if when using a "general offset" like that if all the animation points follow along, as in they don't need any farther adjusting. So if we set that general offset to something like...

x = -34
y = 0
z = -28

We get the model matching the FDM, kind of?

I'll have to test that theory.

dany93:
@wlbragg wrote
> I have no idea why we are using a z offset here? That makes no sense other than someone got lazy and took a quick way to tune something. It could very well have been me.

I don't think that is you. 
I have an old version (January 2013, FG 2.10). 
Observed in Blender, the axes location are the same or very close (slight difference on the propeller cone tip, which may have been redrawn). Wheel contacts with ground at Z=-1.34m. Wings at Z about 0.52m. X origin seems to be the the same. Very close if not exactly the same.

c172p.xml
```
        <contact type="BOGEY" name="LEFT_MAIN">
            <location unit="IN">
                <x> 58.2 </x>
                <y> -43 </y>
                <z> -15.5 </z>
```
Very close too.

And Models/c172p.xml
```
 <path>c172p.ac</path>
 <offsets>
   <pitch-deg>-3.0</pitch-deg>
   <z-m> -0.065 </z-m>
 </offsets>
```
Same offsets.

dany93:
> I have no idea why we are using a z offset here?

I think we need it, we cannot do otherwise to have the FDM and Model axes match.
My issue is not the offset, but the fact that I cannot find where it is written in the files.

To approximately localize the Model axes relatively to the FDM ones, I used a pragmatic technique:
I deleted every aircraft contacts in the FDM and added 4 contacts in the Z=0 plane. Seen in the simulator, the aircraft partly sinks in the ground, just as "floating" on these contacts. Which gives the FDM Z=0 plane seen in the 3D model. In principle...
A difficulty is that you have to set very hard spring coefficients for these contacts, otherwise you are misled by the spring compression. But too hard contacts prevents JSBSim from converging at start, or make the aircraft bounce. May be more or less circumvented by high damping coefficients.
You can also try to have the aircraft upside down, to do an average of compression effect.
Not easy.

The same method can be used for the X axis origin seen in the 3D model.

dany93:
@wlbragg wrote
> Do you know if when using a "general offset" like that if all the animation points follow along, as in they don't need any farther adjusting. So if we set that general offset to something like...
x = -34
y = 0
z = -28
We get the model matching the FDM, kind of?

As I wrote:
> My issue is not the offset, but the fact that I cannot find where it is written in the files.

The ideal would to apply a general offset to the 3D model in Blender to have it match the FDM axes. But I think it would lead to change many many coordinates for the dashboard, instruments and other objects.

I am much less used than you in animations, locations and such things like Blender practice. Thus, please take my opinion for what it's worth.

If everything where clear and perfect, you would be right. But, here, I think that these offsets will add to the already (hidden?) existing ones.
Otherwise, and from my understanding, I think this should work, that the internal, relative objects location would follow. Which makes me write this is that I've often seen that instrument coordinates are relative to the dashboard, dashboard coordinates relative to the c172p.ac main model, etc...
Also, my experience of changing such global offsets worked without issue for the internal objects.

wlbragg:
Well, that was a deep dive into this and greatly appreciated. I don't know if anything will come of it but it certainly helps me understand some long hidden peculiarities about tuning any external reactions.

Yes, I would like to check some of this in Blender now that I have a better understanding.

>The ideal would to apply a general offset to the 3D model in Blender to have it match the FDM axes. But I think it would lead to change many many coordinates for the dashboard, instruments and other objects.
>I am much less used than you in animations, locations and such things like Blender practice. Thus, please take my opinion for what it's worth.

Well, you know enough to know this would be way to late to try something like this. Although I think I have probably touched virtually every animation position at one point or another.

>My issue is not the offset, but the fact that I cannot find where it is written in the files.
>It is such a mess that I can find a Z offset between 28" and 34", depending on which representation I'm looking at

This alone might account for it. It's hard to quantify it with a range of offsets such as this. Or am I not understanding what your looking for? The z offset is about -2.56 inch somewhere inside that 28" to 34" inch range, no?

>I have no idea why we are using a z offset here?
>>I think we need it, we cannot do otherwise to have the FDM and Model axes match.

Is there technically any difference between using an offset of -2.56" or all the gear bogie's of  current z-pos - 2.56?

><pitch-deg>-3.0</pitch-deg>

The pitch is a different story, can you account for it? Where it comes from and why? I think were back to the origination of the model in Blender?

wlbragg:
I removed the offset and added it to the bogies and it all looks right. What is wrong with this approach to simply getting rid of that offset?

wlbragg:
If not to difficult to explain, what FDM elements in the FDM file represent that datum?

wlbragg:
I'm trying to visualize this data in blender and I am getting more confused as I go.

If all this was ideal, where would you locate the FDM datum in Blender, at 0,0,0? And then build everything around that?

dany93:
@wlbragg wrote:
> The z offset is about -2.56 inch somewhere inside that 28" to 34" inch range, no?
Yes, the z offset = -0.065m = 2.56" contributes to the total offset.

> Is there technically any difference between using an offset of -2.56" or all the gear bogie's of current z-pos - 2.56?

If we change the z-pos in the FDM
(if we change this)
```
        <contact type="BOGEY" name="LEFT_MAIN">
            <location unit="IN">
                <x> 58.2 </x>
                <y> -43 </y>
                <z> -15.3 </z>

```
we change the distance between the physical contacts (of course) and the CoG (and every part of the aircraft), that is the moments of lateral forces. Not important for a small amount, but we have to keep in mind that this distance is a physical quantity linked to the aircraft. If we use it as an adjustment variable for the ground height, it must remain consistent with the physical dimensions.

My opinion is that the FDM must be internally in accordance with the physical dimensions and part locations of the aircraft. After that, if (unfortunately) we have to fix the position from the ground, we must use the offsets.

> The pitch is a different story, can you account for it? Where it comes from and why? I think were back to the origination of the model in Blender?

I agree. I think that they built the aircraft in Blender from some drawing or picture, then realized that the pitch was not what they expected from other sources. Not easy. The wheel positions in drawings or pictures can include the spring compression. In FG, we need the uncompressed position, then the compressed position must give the correct pitch and height on the ground.

dany93:
@wlbragg wrote
> If all this was ideal, where would you locate the FDM datum in Blender, at 0,0,0? And then build everything around that?

_(once more, I've never built an aircraft , 3D model and FDM, from scratch. Thus, please take it for what it's worth... )_

The datum is defined by the manufacturer only in X coordinates (at my knowledge).
Yes, I would take this Datum as the FDM origin and Blender origin.
For the Z location, if the manufacturer gives one (for the seats and other part locations, to calculate the CoG location), I would take this one. Otherwise, the propeller axis seems a good reference.

dany93:
@wlbragg wrote:

> If not to difficult to explain, what FDM elements in the FDM file represent that datum?
> 
> ```
>         <location name="AERORP" unit="IN">
>         <location name="VRP" unit="IN">
> ```
> Or, how do these relate?

None of these represent the datum.
[JSBSim aerodynamics FG wiki](https://wiki.flightgear.org/JSBSim_Aerodynamics#Metrics)

> AeroRP - the location where the aerodynamic forces are applied to the model. If this location is not exactly equal to the center of gravity (CoG) then moments, or twisting forces, will be created. Conventional aircraft need AeroRP to be aft of the CoG for stability.
VRP - the world position reported by JSBSim to FlightGear is for this location in the structural frame. Hence, the VRP location should match the origin of the coordinate system(s) used to position views, sounds and the 3d model in the FlightGear XML files.

Firstly think of JSBSim as a stand-alone FDM where the relevant values are the relative coordinates (AERORP, CoG, surface locations,...). Relatives to any origin, datum or any other.
After that, to connect it to the world, the origin of the FDM or something connected with the aircraft (VRP or anything, possibly with offsets) must be transmitted in the scenery coordinates.

**The VRP is an important point**, I had forgotten this one. This idea came back to me last night.
```
        <location name="VRP" unit="IN">
            <x> 42.6 </x>
            <y> 0 </y>
            <z> 38.5 </z>
```
This was the "hidden" offset I was looking for. Added to the smaller ones 
```
        <pitch-deg>3.0</pitch-deg>
        <z-m>-0.065</z-m>
```
it gives, for the total offset, values which explain the ones that I gave above (approximately, at least).

Changing the Z VRP, you can have the aircraft above or sank in the ground,
Changing the X VRP: (42.6 initial) by rolling and watching the wheels in external, view,
- more than 42.6": the wheels "feel" a step (ground change) in advance, before them,
- less than 42.6": the wheels "feel" a step (ground change) late, after them.
- Moreover, due to the 3° offset angle, an X VRP offset makes the aircraft appear more or less above or below the ground surface.

> As well as how does the propulsion relate to the datum? is it offset from the datum?

The propulsion in itself does not relate to the datum. The propeller coordinates must be related to the aircraft coordinates, should it be via the datum or any other location taken as origin.
In our case, the propulsion (the propeller) is obviously offset from the C172P datum as define by the manufacturer. Its location matters because it can gives moments.

wlbragg:
Thanks @dany93 I been away for the last few days and just now read your replies. I want to really study what you posted. Again, thank you for taking the time to explain this. It is really nice to get some of the details explained.

PR #990, opened 2018-02-25
Issue 949: Switches in KMA 20 instrument move dials in radio instruments

### LINKED ISSUE #949: Switches in KMA 20 instrument move dials in radio instruments
this was originally reported on the forums here...
[https://forum.flightgear.org/viewtopic.php?f=14&t=33695](url)

i've reproduced it in both the current c172p in FGData as well as the current craft in master on this repo...

here's the steps i followed...

1. before loading, i went to my ~/.fgfs/aircraft-data and deleted my c172p.xml file so i would start clean...
2. i started at EGLL and used the autostart... make sure you set the parking brake!
3. hit F12 for the radio stack
4. looked up EGLL radio freqs
5. tune COM1 to 128.07 (Heathrow Information)...
6. i didn't hear anything
7. i used my mouse (left mouse button) and toggled the COM1 switch up (ON)...
8. i hear the ATIS information
9. i used my wheel button to click the switch back to the center position (OFF)...
10. i don't hear the ATIS information any more
11. zoom in close and toggle the COM1 switch
12. you can see the radio's volume switch jump from all the way down (actually 10%) to about 75%

there's more confusion in there between those two switches, too... this also happens when flipping the switch down to the headphone position and back up...

what should happen is that the COM1 switch should only affect the sound... it should be off, speaker or headphone and the radio's volume button should not move... it should stay at the same volume setting no matter which position the COM1 switch is in... yes, this means that you may flip the speaker on and be blasted by the radio at 100% volume... same with flipping the headphones on...

mathieugouin:
Fixes #949.

- Switching COM1 & COM2 on KMA-20 does not change COM volumes anymore.
- COM volume changes only when enabled for audio.

Note: please double-review my code as this is a first for me...


Models/Interior/Panel/Instruments/kx165/kx165-1.xml
```diff
@@ -12,7 +12,7 @@
         flightgear sessions
 
         aircraft.data.add(
-        "instrumentation/comm[0]/volume",
+        "instrumentation/comm[0]/volume-selected",
```
mathieugouin: volume-selected was chosen based on similar work done in c182s.

onox:
COM1 and COM2 are OK, but the NAV1 and NAV2 switches still control two dials in the radio instruments.

Models/Interior/Panel/Instruments/kx165/kx165-1.xml
onox: :+1:

mathieugouin:
Yes, I agree, but at least, the volume is not changed.  Only the "pull" for ident.  To me that was less of an issue compared to the volume knob being automatically rotated.

But I'm not sure what would be the desired behavior?

I see two potential solutions:
1. Behave like the ADF volume knob: no pulling when ident is activated by the KMA-20 panel.  However, this will not match the label "pull for ident"...
2. Implement the "real" knob pull action.  To hear the ident, KMA-20 switch would need to be in SPEAKER or PHONE mode (like right now) **plus** the NAV radio knob would need to be pulled in ident position.  This is the portion I need help...
3. (Keep as-is?)

mathieugouin:
I found this, good to have as a reference for the KMA-20: http://www.mft-bzn.com/pilotg~1.pdf

onox:
@gilbertohasnofb What do you think of this?

gilbertohasnofb:
@onox I will take a look at this tomorrow, it got quite late here already.

gilbertohasnofb:
Ok, so finally had a chance to take a look at this. The COMs are completely fixed on my side, thanks @mathieugouin. As for the NAVs, I now understand the automatic pulling and it's kind of a smart implementation to simplify things. But personally, I would rather go with solution 2) and force the user to pull the NAV radio knob themselves. The issue with this is that the knob is small and I wonder how the clickspots would work... I already find our light knobs a pain to deal with, so that tiny thing might prove really difficult to handle. If that would be the case, then solution 3) of keeping things as they are  would be fine with me.

Solution 1) is a no go for me.

What do you guys think? About the clickspots, does @wlbragg has experience with those or was someone else creating them, e.g. for the illumination knobs? If so, it would be nice to hear his opinion on this.

wlbragg:
The only way I could see to do solution #2 is to make a click spot over the text "pull/ident" that when clicked animates the knob to extend as if it had been pulled and then controls "ident". The click spot would be a toggle switch (pull/push). Otherwise I don't know how you would implement it.

gilbertohasnofb:
@wlbragg I think that solution could be quite confusing for anyone who doesn't know about it. The solution used by A2A on their FSX aircraft is to differentiate between left and right click, with the latter pulling a knob -- but I don't think that's feasible in FG. So with all that in mind, I am inclined to leave things as they are at the moment.

onox:
Is it possible to use some keyboard key like Ctrl to switch between pushing/pulling and turning the dial?

wkitty42:
On 02/28/2018 12:33 AM, onox wrote:
> Is it possible to use some keyboard key like Ctrl to switch between 
> pushing/pulling and turning the dial?


it should be possible... i think i've seen the [SHIFT] being used in one of the 
airliners... possibly josh's craft... i don't recall...

gilbertohasnofb:
I think using `Shift + click` would be the best solution. There are some properties for dealing with that, namely `<shift-action>`, `<shift-increase>` and `<shift-decrease>`. I think that the route to solve this would be to first disable the shift increase and decrease to that shift+click on the volume knob will not rotate it 10x more than normal, then use shift+click to toggle the knob position. What do you guys think?

gilbertohasnofb:
See http://wiki.flightgear.org/Knob_/_slider_animation#Shifted_Mode and https://forum.flightgear.org/viewtopic.php?f=4&t=21427#p195206

gilbertohasnofb:
I think perhaps @it0uchpods would know how this work as it seems some of his aircraft uses `Shift + left click` as a secondary function of a knob/toggle

it0uchpods:
@gilbertohasnofb @wkitty42  None of my planes uses Shift + left click -- not sure what plane you are using, but it isn't mine.
I do know how to use it however. What do you need?

Kind Regards,
Josh

gilbertohasnofb:
@it0uchpods I thought I read you writing somewhere in the forum that you were using/intending to use Shift + Click in your IDG-A32X, but perhaps that's my mistake.

> I do know how to use it however. What do you need?

Could you confirm that the correct route to implement this is to use `<shift-action>`? Also, is it necessary to disable the default behaviour of the Shift key with a knob which normally makes it turn 10x the default?

it0uchpods:
@gilbertohasnofb 
I did plan to use it, but did not implement this "laptop mode" yet. 

I'll look over it tonight when I am free.

Kind Regards,
Josh

gilbertohasnofb:
@it0uchpods Great, thanks.

@mathieugouin Would you like to implement this yourself?

mathieugouin:
@gilbertohasnofb Yes, I can try.  I quickly looked at the examples provided, but they were not obvious to me.  Is there a complete manual/reference/wiki of all the various bindings possible?

gilbertohasnofb:
> Yes, I can try.

@mathieugouin Great!

> I quickly looked at the examples provided, but they were not obvious to me. 

I also found those examples lacking. @it0uchpods said he knows how those work so it would be greatly appreciated if he could share some examples here!

it0uchpods:
Sorry @gilbertohasnofb I mistook what you said for shift + drag. I don't see any shift-click support in the knob animation. I did think up a workaround I was planning to use, but I don't remember what it was. After I sort the AP out, I'll work on that.

Sorry for the confusion.

Kind Regards,
Josh

mathieugouin:
From what I understand, there are two regions, one to the left and one to the right.  Could we map both left and right regions of the knob shift-click to toggle the knob pull in or out?

gilbertohasnofb:
In that case perhaps we should ask in the dev-list or in the forum. @mathieugouin, are your part of any of those? We could ask about the `Shift + click` problem or even ask more generically on how to tackle a small knob that can be both rotated and pulled as a toggle.

mathieugouin:
I'm not sure what the dev-list is, but I'm part of the FG forums.

it0uchpods:
I'm on the list, if you'd like me to ask.

The list is where the core-dev goes on.

J

gilbertohasnofb:
@mathieugouin We could try the forum, and info about the mailing list can be found here: http://wiki.flightgear.org/Mailing_lists

> I'm on the list, if you'd like me to ask.

@it0uchpods that would be really appreciated!

it0uchpods:
Sure, I'll put in a mail later tonight.

Kind Regards,
Josh

mathieugouin:
I have the a/c in FG in front of me right now, and it seems the volume knob can only be rotated to increase the volume by clicking.

gilbertohasnofb:
@mathieugouin I noticed that as well, those definitions are missing `<drag-direction>horizontal</drag-direction>`. Without that, left click or mouse wheel up increases the value, middle click or mouse wheel down decreases it. With that, you then have another option which is to click and drag.

mathieugouin:
That's strange, drag works on my side?

gilbertohasnofb:
> That's strange, drag works on my side?

Oh, I must test that. I just noticed that by looking at the code, perhaps that's used only to override the default behaviour then.

onox:
With a knob animation you can click, scroll, and/or drag.

mathieugouin:
Idea note to myself ;)

```
        <shift-action>
            <binding>
                <property>/instrumentation/nav[0]/TBD</property>
            </binding>
        </shift-action>
```

gilbertohasnofb:
@mathieugouin In the list, James has mentioned that it might be possible to use double click on the knob to toggle it and use simple click + drag (current behaviour) for rotation. As soon as I get a reply on how this is done I will post it here.

mathieugouin:
I realize that it will not be obvious to know that the knob is pulled or not.  Refer to attached screenshot.  See if you can tell which one of the NAV1 or NAV2 knob is pulled?

image

mathieugouin:
I got it !!! It works with shift-click!

mathieugouin:
I updated the hover tool tip to show the special instruction.  Do you think it can help?

image

mathieugouin:
I have updated my branch.  Let me know if this seems OK.

gilbertohasnofb:
> I realize that it will not be obvious to know that the knob is pulled or not. 

Possibly this is also an issue in RL. But I guess it's not much of a problem and frankly I don't see anything we could do to improve it.

> I got it !!! It works with shift-click!

Fantastic, great job! But I am afraid I won't be able to test it today, if nobody beats me to it I plan to take a look at it by tomorrow.

> I updated the hover tool tip to show the special instruction. Do you think it can help?

That helps a lot, great idea. But I would write `Shift + click` insted of `Shift-click`. What do you think?

mathieugouin:
> ...But I would write Shift + click insted of Shift-click. What do you think?

I agree, will update it!

onox:
@mathieugouin Thank you very much for implementing this :+1:

mathieugouin:
No problem, glad it can help!

If I find other issues in the c172p, what is the best route?  Through the FG forums?

gilbertohasnofb:
> If I find other issues in the c172p, what is the best route? Through the FG forums?

Since you are a GitHub user, the best way of reporting is by opening Issues here in our repo. Also, feel free to go through our standing issues and propose solutions as PRs, we really welcome those. Just keep in mind that there are a few issues that are very large and take long time to implement, so there is a chance someone is already working on those. But that should be obvious from the discussion under the issue itself, but feel free to ask if that's unclear. Many, many thanks once again.

Issue #790, opened 2016-07-08
absolute position for instruments in individual .ac files


gilbertohasnofb:
Thorsten wrote:

> That's caused by translating/rotating things into place rather than having them absolutely positioned in the ac file, not by including a separate ac file.

@wlbragg wrote:

> I guess I never really understood that and it is really helpful to know this. So technically in the c172p's case, all the eternal .ac instruments it uses (the ones that are local anyway), could be positioned absolutely in their ac file and then be included in the shadow effect to get a better visual. It would be some work but might be worth it. It for sure would be worth understanding this on any new build and start out using this method so you can include the entire instrument panel in your effects, good to know.

I personally think it would be really worth to do it, so that all instruments can respond properly to the effects. But maybe, if we are still considering fixing https://github.com/Juanvvc/c172p-detailed/issues/6 and https://github.com/Juanvvc/c172p-detailed/issues/216 (which are fundamental to have a properly looking cockpit IMO) then we maybe should wait for those before adjusting all instruments.

wlbragg:
Yes, for sure we would only want to do this once.

gilbertohasnofb:
:+1: So let's put this on hold for a bit

dany93:
_(I don't know if this is the best place...)_
I noticed that the 3D magnetic compass indication is wrong.
About 180° error near 90° and 270° mag heading (map),
Pretty well at 0° and 180° mag heading.

gilbertohasnofb:
@dany93 I don't think this is the place for that, this issue is just about the position of the instrument models in the .ac file. I will open a new one with your report.

dany93:
I had the impression that the code change was in [Move compass to absolute position....](https://github.com/c172p-team/c172p-detailed/commit/20fdd9612decd95e34f4bbe705928e26736035ff).
Hence this place for my message. I don't know if it can help, or if there is no relation (it can be a mistake on my part, then sorry).

gilbertohasnofb:
@dany93 I see. Nevertheless, @wlbragg is already working on a fix so this should be back to normal soon. Thanks!

wlbragg:
@gilbertohasnofb 
Yes

Issue #35, opened 2015-03-12
Instrument illumination is too strong during daytime


gilbertohasnofb:
XML14 - Instrument illumination is too strong during daytime. In the original c172p, one could hardly tell the difference between 0% and 100% instrument illumination during daytime. At night time, our project behave exactly as expected.

legoboyvdlp:
Close after merge of #1103?

gilbertohasnofb:
I think the instrument lights are still way too strong during the day. The dome lights are very faint which looks good. If we improve that then we could add this issue to the list of fixed stuff on that PR.

![fg](https://user-images.githubusercontent.com/5700795/41500987-bd7c5b26-7193-11e8-8282-e2d3f396dffc.jpg)

 @wlbragg would you consider that idea I floated of somehow lowering the intensity of the instrument lights during the day?

wlbragg:
Yeah, absolutely, I think i even know what property to use.

gilbertohasnofb:
Excellent, thank you!

wlbragg:
I think your going to like this, pushing now. Let me know if you think it needs adjusting.

PR #992, opened 2018-03-01
Add elevator trim tab

### LINKED ISSUE #259: Small improvements to the exterior 3D model
Create a hole for the wheels inside fairings:

unbenannt

Create a door handle:

unbenannt

### LINKED ISSUE #622: Create elevator trim tab
Create elevator trim tab in the 3D model.

### LINKED ISSUE #879: UV-map the non-Canvas version of Garmin GPS
In https://github.com/Juanvvc/c172p-detailed/issues/662, we spoke about porting the Garming GPS to Canvas (we currently even have a folder `Models/Interior/Panel/garmin196` for this porting). But while we don't get that done (and that would be a lot of work, I believe), we are stuck with the non-Canvas version which is functional but has some bug and problems (see https://github.com/Juanvvc/c172p-detailed/issues/392) as well as being in an unrealistic grey colour. 

Given that we have a better texture for the GPS, I think we should UV-map this older version to make it at least somewhat appealing (as far as I know, the Garmin 196 is always black). This version is not hosted locally though, but is part of the `Aircraft/Instruments-3d` folder in FGDATA, so whomever would like to improve it would have to push the improved texture and .ac file to that repository.

### LINKED ISSUE #986: Tips of new wings are not UV-mapped correctly
Tips of new wings are not UV-mapped correctly.

Before the new wing, as the textures were intended:

fg-5
fg-6
fg-7
fg-8
fg-9

After the new wing:

fg-0
fg-1
fg-2
fg-3
fg-4

### LINKED ISSUE #987: GPS should not be behind the amphibian cockpit bars
GPS should not be behind the amphibian cockpit bars:

fg-0

wlbragg:
Fixes #622 
Fixes #987
Fixes #879
Fixes #259 
Fixes #986


wlbragg:
Nothing fancy here. I didn't check the full range of the trim wheel adjustment. I used the existing "commented out" trim tab code that was in c172p.xml. All I did was change the center point of the rotation so it needs to be double checked.

onox:
The rotation axis of the animation needs to be adjusted.

wlbragg:
The animation rotation is fixed but I'm not convinced this is correct property to be using or it may have some issues. It appears that the property
controls/flight/elevator-trim
is rotating the entire elevator, is that the correct behavior?

gilbertohasnofb:
> It appears that the property controls/flight/elevator-trim is rotating the entire elevator, is that the correct behavior?

@wlbragg No, that is not correct. The animation should move only the trim tab.

dany93:
I had this thought once when doing such.

IRL, the trim tab effect is, **in flight**, to deflect the surface (here the elevator). Thus, it is equivalent to pull or push the stick or yoke by the pilot, the difference being that the pilot no longer has to hold the stick. The neutral point of the stick as felt by the pilot is shifted (offset).
Think of the (initially flat and symmetric) elevator (or rudder) as a cambered one with the trim. It becomes asymmetric.
If my reasoning is good (I think it is, but I'd prefer some critic) this rotation of the entire elevator is correct in flight.

Of course, this effect is not true on the ground, where only the weight acts on the elevator (no force due to airspeed). Not easy to do, I would tolerate this small issue.

gilbertohasnofb:
@dany93 I understand what you are pointing and I think you are 100% correct. But let's just remember that what we are implementing here is only the animation, the effects of using trim is already part of the aircraft behaviour, so a small deflection on the whole elevator surfaces at speed vs no deflection at lower speeds in the ground would just be "eye candy". In that sense, I am with you when you say that we can perhaps not worry too much about this and leave things as they are.

But what is currently incorrect with this PR is that the trim tab itself should move proportionally to the trim, regardless of speed or being airborne or not, and that should be corrected. This has nothing to do with the elevator being deflected or not.

dany93:
> ... the trim tab itself should move proportionally to the trim

I agree. Sorry, I did not realize that the issue was the **trim** animation because our c172 had no 3D trim to animate previously.
For the animation, mind a trap: when you deflect the trim tab backwards, the elevator moves upwards.

dany93:
> @wlbragg No, that is not correct. The animation should move only the trim tab.

That was the reason for my confusion. Not _only_ the trim tab (at least in flight).

I've observed the current status of the animation https://github.com/c172p-team/c172p-detailed/pull/992/commits/1f6c156fb27b40fb01cec28da4126ea3dc14348c. The trim tab is now animated. But I think (see my reasoning just above) that it is in the wrong way.

gilbertohasnofb:
> That was the reason for my confusion. Not only the trim tab (at least in flight).

Yes @dany93, you are completely right, when I wrote that I did not think about the deflection of the whole elevator surface that would be observed at high speeds. 

> I've observed the current status of the animation 1f6c156. The trim tab is animated. But I think (see my reasoning just above) that it is in the wrong way.

Yes, I do see now that the trim tab is moving. I had missed it as the trim tab moves very little in comparison with the whole elevator surface, and I just assumed the wrong property was being used. The trim tab does move, but I agree with @dany93 that its movement has to be inverted.

gilbertohasnofb:
Also, I noticed these errors on my log when using this branch:

```
In file /home/gilberto/WorkingFG/c172p/c172p/c172p.xml: line 1903
No direction element specified in force object. Default is (0,0,0).

In file /home/gilberto/WorkingFG/c172p/c172p/c172p.xml: line 1911
No direction element specified in moment object. Default is (0,0,0).
```

Is anyone else experiencing this?

dany93:
I also have these errors.
This is the part added by @bcoconni for the AI wake https://github.com/c172p-team/c172p-detailed/pull/989

dany93:
@wlbragg for the reasoning: (maybe unnecessary, but just in case)
Differently from a (possible) assumption, the trim tab effect is not to increase the elevator efficiency, it is to offset its position at rest, when the pilot does not hold the commands.

wlbragg:
>AI wake

That may be a default if no AI wake is present?

> it is to offset its position at rest, when th the commands are released.

Is the FG's implementation or real life?

wlbragg:
This is confusing me.  Let me see if I understand this correctly.

FG doesn't have FDM rmoment for elevator trim, it simply adjust the elevator moment to achieve the desired trim?

>it is to offset its position at rest, when th the commands are released.

FG's implementation, not real life?

>The trim tab does move, but I agree with @dany93 that its movement has to be inverted.

I disagree, unless were talking about something else. If turning the trim knob one way that raises the elevator, then the trim tab should also rotate in the same direction as the elevator, not opposite. Because if it was working as in real life it would be deflecting the as the elevator only in smaller proportions.


If we are simulating the

dany93:
>      it is to offset its position at rest, when th the commands are released.
> 
> Is the FG's implementation or real life?

I mean in real life (if my understanding is good).

My reasoning:
Imagine a flat, symmetrical airfoil (elevator here). Without pilot action, it stands parallel with the apparent wind.
After that, imagine a curved airfoil (like a wing). Or our elevator with its rear trim-stab deflected downwards. If held parallel with the apparent wind, the wind exerts a vertical force upwards. Hence, without pilot action, the elevator rotates around its hinges upwards until an equilibrium is reached. Which makes the aircraft lift up.

Am I clear ? Do you agree ?

FG translate this as an elevator deflection with control at neutral when some trim in applied (see FG Properties,/surface-positions : elevator-pos-norm = ), which is in accordance with reality as I describe it.

wlbragg:
It would be really nice if we could model this in the FDM because one of it's uses is apparently to offset the yaw moment due to spiraling propwash?

wlbragg:
I posted my last comment befor your reply, thus the lack of it tying into your last reply.

>Am I clear ? Do you agree ?

I don't disagree, but it is only on one elevator. I did a search on it and got a little bit different interpretation of what it is doing.

wlbragg:
I'm reading this now
http://www.marvgolden.com/downloads/dl/file/id/63/cessna_172_training_manual_book_excerpt.pdf

>The trim tab moves opposite to the control
surface, reducing the aerodynamic force on the control surface in order to hold
the selected position.

That's solved.

dany93:
> If turning the trim knob one way that raises the elevator, then the trim tab should also rotate in the same direction as the elevator, not opposite. Because if it was working as in real life it would be deflecting the as the elevator only in smaller proportions.

The trim knob does not raise the elevator, it rotates the trim tab. Consequently, without pilot action, the elevator takes another position, deflected even without pilot action. No need for a large elevator deflection, it is just to change the equilibrium position of the elevator, hence the AoA and airspeed for the aircraft with no pilot action.

You are reasoning as if the trim tab action were to increase the elevator efficiency. Which would make it harder to move. But it is not to increase the elevator efficiency, it is to offset its position at rest, when th the commands are released.

wlbragg:
OK, your last statement cleared it up for me ,and I can see why now it is opposite to the elevator action.
Let me find some verbiage I read before and maybe you can explain it to me. It is the only thing I am not understanding completely. If has to do with the yaw from prop.

Let me ask you this though. OK, I understand and agree with your explanation. But what about the difference between the left elevator surface with no trim tab and the right elevator with trim tab (of course only if the trim tab is actually being used).
That does have to affect yaw or roll?

wlbragg:
>It would be really nice if we could model this in the FDM because one of it's uses is apparently to offset the yaw moment due to spiraling propwash?

wlbragg:
Something else, I noticed a rudder trim tab on our aircraft, but no control mecognisum in the cockpit, why? Are we suppose to have it? The handbook says yes in the newer models of which our elevator trim knob is mimicking (new model).

wlbragg:
>The Cessna 172 aircraft has a manually operated rudder and elevator trim system. 

>A rudder trim system is installed in later models to provide a means of directional trim.

>The rudder trim, if installed, is operated by either a rudder trim control tab or control wheel mounted on the center control pedestal. 

I ask because according to our elevator trim wheel style and placement, we are using the "newer version" of it.

wlbragg:
Here is the article I was referring to...
https://www.quora.com/Why-is-the-elevator-trim-tab-on-a-Cessna-Skyhawk-on-the-right-side-only

dany93:
> If has to do with the yaw from prop.
.....But what about the difference between the left elevator surface with no trim tab and the right elevator with trim tab (of course only if the trim tab is actually being used).
That does have to affect yaw or roll?

_From my experience on IRL ultralights. I guess that it's also true for the C172 (confirmation ?)._
The elevator right and left surfaces are firmly attached together (by a rigid rod). On the ground, if you manually raise the right side, the left one follows. Like a single piece.
Nothing to do with  yaw or roll.

> It would be really nice if we could model this in the FDM because one of it's uses is apparently to offset the yaw moment due to spiraling propwash?

Spiraling propwash offset is modeled in the FDM. I have added a rudder-trim preset in the -set.xml file.
The rudder trim is not controllable in flight by the pilot. IRL, this trim preset is done on the ground, by twisting a metallic plate (a vertical stab attached to the rear of rudder).

wlbragg:
>On the ground, if you manually raise the right side, the left one follows. Like a single piece.
Nothing to do with yaw or roll.

But the different shape of the surface of each elevator should have something to do with yaw or roll as per the writeup I posted the link to above. I'm not talking about a difference between elevator positions, I get they are the same. Only that one elevator surface is effectively changing the airflow because of the trim tab.

>Spiraling propwash offset is modeled in the FDM. I have added a rudder-trim preset in the -set.xml file.

I didn't know that, good. If we stick to this setup then I think we include a GUI popup, using a hotspot on the rudder trim tab, that you can use to adjust that value (ie: bend the trim tab).

OR

>The rudder trim is not controllable in flight by the pilot. IRL, this trim preset is done on the ground, by twisting a metallic plate (a vertical stab attached to the rear of rudder).

As I tried to ask above, Definitively, "should" this model and version of the c172p have a cockpit control knob for the rudder trim. Why are we using a "newer version" elevator trim knob"? Is that indicative of a newer version or model c172p, one that might also have the rudder control knob in the cabin found on "newer versions" of the 172p?
I don't mean to beat a dead horse here, just want to make sure we have it right.

dany93:
> Here is the article I was referring to...
> https://www.quora.com/Why-is-the-elevator-trim-tab-on-a-Cessna-Skyhawk-on-the-right-side-only

Very interesting. That's the response to why the elevator trim tab is on the right side.

wlbragg:
@dany93 
Please forgive me for forcing you to take the time to give me basic lessons in control surface functions. Unfortunately it took both your explanations and my own investigation to get it straight.

dany93:
No problem.
If I (hope to) have a response about something, not exchanging it would be a pity.
Better: often, trying to explain something is a way to improve our own comprehension. Your link on the elevator trim tab position is a close example.

dany93:
> >     On the ground, if you manually raise the right side, the left one follows. Like a single piece.
> >     Nothing to do with yaw or roll.
> 
> But the different shape of the surface of each elevator should have something to do with yaw or roll as per the writeup I posted the link to above. I'm not talking about a difference between elevator positions, I get they are the same. Only that one elevator surface is effectively changing the airflow because of the trim tab.

I agree. But I also think that it is a very minor effect, compared with the multiple other effects for a single engine aircraft.

wlbragg:
>https://i.ytimg.com/vi/mIjcptMq45A/maxresdefault.jpg

Maybe, but it may be an important factor or effect to the ease of rolling down the runway.

I would be helpful to hear form a pilot with actual experience in this aircraft about this subject.

bcoconni:
> Also, I noticed these errors on my log when using this branch:
>```
>In file /home/gilberto/WorkingFG/c172p/c172p/c172p.xml: line 1903
>No direction element specified in force object. Default is (0,0,0).
>
>In file /home/gilberto/WorkingFG/c172p/c172p/c172p.xml: line 1911
>No direction element specified in moment object. Default is (0,0,0).
>```
>Is anyone else experiencing this?

Oops, sorry ! These warnings are harmless for the AI wake force and moment since their directions are computed by FlightGear and the content of the `<direction>` tag is ignored. This is the reason why I did not specify a direction for `<force>` and `<moment>`. However, when `<direction>` is not specified in an external reaction, JSBSim complains with the warning message above. Whether the AI wake is enabled or not, they can be safely ignored.

The other option to remove the warnings would be to specify dummy direction values for both the AI wake force and moment, something like below should silent JSBSim warnings.
```xml
<direction>
    <x>1.0</x>
    <y>0.0</y>
    <z>0.0</z>
</direction>
```

wlbragg:
@bcoconni 

Thank you for the quick reply ,I figured that is what that was. I guess the team can decide whether to use a dummy to silence the warning. Personally I like to define these things so we don't get unnecessary warnings.

gilbertohasnofb:
@bcoconni 

> The other option to remove the warnings would be to specify dummy direction values for both the AI wake force and moment, something like below should silent JSBSim warnings.

That would be ideal. Even though the warnings are harmless, in my opinion it's better not to pollute the log with warnings like that. Do you want to fix it yourself via a new PR?

algefaen:
> Maybe, but it may be an important factor or effect to the ease of rolling down the runway.
> I would be helpful to hear form a pilot with actual experience in this aircraft about this subject.

Hi, I've never heard of this being an issue or noticed this having any affect on yaw/roll. Adding / reducing power (propwash and torque) and changing AoA (P-effect) do affect it noticeably. I also think, as @dany93 
says, that any elevator trim effect on yaw/roll would drown in these other effects.  Re. rudder trim, the aircraft should fly relatively straight in cruise. If this is not the case, you can ask a mechanic to adjust the trim. It's not something you can do from the cockpit.

wlbragg:
>It's not something you can do from the cockpit.

In all versions? 
Please see page 31, I am trying to discover when this was installed and if our version is older than this or not?
http://www.marvgolden.com/downloads/dl/file/id/63/cessna_172_training_manual_book_excerpt.pdf

algefaen:
> In all versions?
> Please see page 31, I am trying to discover when this was installed and if our version is older than this or not?

I think that's slightly misleading, as it makes it sound like all newer c172 have rudder trim. I've flown 2 c172n, 1 c172p and 3 c172sp, and none of them has adjustable rudder trim. The pa28s and c206 I've flown do have it, and I've sometimes missed it in the skychicken... Long Vx climbs are hard for your right foot. (=

Note that unlike the elevator trim, the rudder trim (of at least the pa28) just apply a force on the pedals, they don't have have any adjustable trim tab that pushes on the pedals.

dany93:
[How The 4 Types Of Trim Tabs Work](http://www.boldmethod.com/learn-to-fly/systems/4-types-of-trim-tabs/)
See in particular the animation "1) Trim Tab Operation",
and "4) Ground Adjustable Tab"

wlbragg:
Thanks guys, I guess I'll just let this go then. 

@gilbertohasnofb  @onox 
I think it good to add a GUI for adjusting the rudder trim tab, if only to allow pilots to screw it up! :grin:

I'd like feedback on ...
1) Make it like the damage button, only available when on the ground?
2) What section to put it in in the Aircraft Options dialog, if at all, and what to call it?
3) How exactly to implement it?

Here are my thoughts...
1) I think a checkbox in the existing Aircraft Options, something like
Adjustable rudder trim
If checked, clicking the hot-spot on the rudder pops up the CGI to adjust it, only if on the ground (I'm not firm on that idea, it might be nice to be able to play with it while flying).
If unchecked, default rudder trim and no GUI available.

2) Or just have it always available by clicking on the rudder, no entry in the Aircraft GUI at all.
2a) Only a click-able hot-spot when on the ground.

3) Or only a button in the Aircraft Options dialog that pops up a GUI to adjust
3a) Only click-able when on the ground

wlbragg:
@dany93 

How is it that the 
controls/flight/rudder-trim
controls/flight/elevator-trim
controls/flight/aileron-trim
actually affect the aircraft when they are not referenced in the FDM section, only defined somewhere? Are they internal properties to FG?

Is the alieron-trim supposed to be mimicking a cable adjustment to the ailerons to give you some kind of a trim factor?

Personally I thing all three of these should be adjustable through the GUI system in one form or another so pilots can adjust these to suit their needs. I know they can set them in the set file (at least rudder and aileron) but that doesn't seem appropriate to force a code modification to achieve a basic aircraft flight tuning function.

Comments, anyone?

dany93:
**Adjusting the rudder trim tab:**
The condition 'only available when on the ground' seems logic.
Its restrictiveness can be circumvented by those who know, by manually setting a rudder-trim value in the Internal Properties.
**Important:** the setting must not prevent from manually adjusting it later even in flight, by any mean. Like currently with the one in the -set file. Even if that's cheating with realism, it is a very convenient way to test adjustments. Those who go in the Internal Properties know what they do.

(still cheating, I know...) I even have a rudder trim button on my joystick.

dany93:
> How is it that the
> controls/flight/rudder-trim
> controls/flight/elevator-trim
> controls/flight/aileron-trim
> actually affect the aircraft when they are not referenced in the FDM section, only defined somewhere? Are they internal properties to FG?

_(That's too easy, I'm afraid of not understanding the question)_
Yes, they are in the Internal Properties, they can be modified in the sim at any moment (Internal Properties, 'Set')
Pre-set in the -set.xml file (lines 497-498), but changeable.

dany93:
> Is the aileron-trim supposed to be mimicking a cable adjustment to the ailerons to give you some kind of a trim factor?

In FG, yes. Like an offset for the neutral command position.
I reality, that's exactly like the one for the elevator, a tab that you can bend when on the ground  for the aircraft which do no have this feature adjustable from the cockpit.

dany93:
> And by the way, on my last test the whole elevator surface was still moving together with the trim tab.

This is the behavior in flight. Now, they move in the correct direction.
Discussed [in this message](https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369549067). We suggested to tolerate it because the real behavior on the ground was difficult to simulate at the same time.

wlbragg:
>Why would you like to implement all these?

I don't mean in the cockpit.
We already have elevator-trim in cockpit = done.
Sounds like we have agreement for "simple" rudder-trim implementation, on ground, override-able at anytime in code for pros = will be done.
All that is left is the aileron-trim which we have access to already in code, just do the same for it as we are going to do for the rudder-trim, click-able hot spot where appropriate either on aileron, flaps or where linkage would be, simple slider gui. Nothing fancy or not appropriate to the model of aircraft.
Overkill? If we have things like oil changes, dirt, frost and fog, damage, shadows, then a functional, documented adjustment that probably every aircraft has had done to it at one time or another. I don't see it as overkill, but as basic functions.

wlbragg:
>real behavior on the ground was difficult to simulate at the same time.

So on the ground (no motion) just the elevator trim tab would move? The elevator surface wouldn't adjust until great enough airflow passed over it?

wlbragg:
>So on the ground (no motion) just the elevator trim tab would move?

I think we might be able to do this, with a filter or function.
@onox, job opportunity?

gilbertohasnofb:
@dany93 apologies, I have missed that message. But it seems to me that the deflection of the elevator surface is enormous comparing to the deflection of the tab. Shouldn't the animation be a bit more subtle?

> So on the ground (no motion) just the elevator trim tab would move? The elevator surface wouldn't adjust until great enough airflow passed over it?

Yep. The range could be multiplied by the airspeed times a factor and then capped within a certain range.

> I don't see it as overkill, but as basic functions.

@wlbragg Unless I missed something from the discussion (and if I did my apologies!) but I think that the difference is that there is no such thing as an aileron tab in the 172, while there is a rudder tab that can be adjusted on the ground and an elevator tab that can be adjusted in-flight. In other words: an aileron trim tab would be something not realistic as the aircraft does not have that. Or did I miss something and people do somehow manage to adjust the aileron trim on a real 172??

gilbertohasnofb:
And as for the GUI for the elevator trim, I think that's not really useful: we already have the trim control on the panel! I would suggest just a super simple dialog for the rudder tab, activated when clicking on it.

wlbragg:
>Personally I thing all three of these should be adjustable through the GUI system in one form or another 

First what I mean say after everyones input is...
Personally I think all three of these should be adjustable in one form or another.

>And as for the GUI for the elevator trim, I think that's not really useful: 

Correct, see above re-clarification.

>an aileron trim tab would be something not realistic as the aircraft does not have that. Or did I miss something and people do somehow manage to adjust the aileron trim on a real 172?

Yes, or I have it wrong. It sounds like they do adjust it in one of three ways. Trim tab somewhere on the wing (either manual bend or some models in cockpit?)
Or in adjusting the aileron linkage.
Or even adjusting the wing strut I believe.

dany93:
> And as for the GUI for the elevator trim, I think that's not really useful: we already have the trim control on the panel!
I agree.

wlbragg:
Sorry, I'm still not being very clear.
I think were done with the elevator trim adjustment availability, we have it in the cockpit.

We will be done with the rudder trim adjustment with the simple popup slider and click spot.

All we may want to do next is add the same simple solution for the aileron trim adjustment beings it is a thing that is set and we actually are setting it in the FDM as
/controls/flight/aileron-trim type="double" = 0.02

wlbragg:
I think you may have missed this comment or didn't understand what I meant.

>I don't mean in the cockpit.
We already have elevator-trim in cockpit = done.
Sounds like we have agreement for "simple" rudder-trim implementation, on ground, override-able at anytime in code for pros = will be done.

gilbertohasnofb:
> Yes, or I have it wrong. It sounds like they do adjust it in one of three ways. Trim tab somewhere on the wing (either manual bend or some models in cockpit?)
> Or in adjusting the aileron linkage.
> Or even adjusting the wing strut I believe.

If that's the case then I would be 100% with your idea; the thing is I absolutely never heard of this before. Both rudder and elevator have trim tabs (one adjustable from cockpit the other manually on ground), but shifting the linkage on the ailerons sounds like a hack rather than something intended! I would say that we need some opinions from people with flying experience.

My problem is simple: do pilots and/or mechanics adjust the ailerons in real life? If so, let's simulate that! If not, then let's not.

wlbragg:
>My problem is simple: do pilots and/or mechanics adjust the ailerons in real life? If so, let's simulate that! If not, then let's not.

Right, I totally agree. From what I read they do and there are three ways that it is done on the c172?.

>It sounds like they do adjust it in one of three ways.
Trim tab somewhere on the wing (either manual bend or some models in cockpit?)
Or in adjusting the aileron linkage.
Or even adjusting the wing strut I believe.

gilbertohasnofb:
@wlbragg My mistake then, didn't know this was a thing. I think I must have missed that message of yours.

dany93:
@wlbragg wrote:
> So on the ground (no motion) just the elevator trim tab would move? The elevator surface wouldn't adjust until great enough airflow passed over it?

Of course. On the ground, the elevator falls down due to its weight, and that's all.
I flight, the airflow gives the dynamic equilibrium.

gilbertohasnofb:
And so in that case, there won't be any click spots for the aileron trim GUI. So what about this: we could create a new "mechanical adjustments/options" menu or something like that and have things that pilots don't normally adjust or change. We could even move the engine selection over there. And so there we would find the rudder and aileron trims as a single slider each.

And we could still make the rudder tab clickable and that would open a dialog with only the rudder trim. This is similar to what I did with the luggage weight: there is a small dialog but one can also adjust those values in the fuel and payload main dialog.

gilbertohasnofb:
> On the ground, the elevator falls down due to its weight, and that's all.
I flight, the airflow gives the dynamic equilibrium.

Yeah, I did notice that in several photos. In case this will get implemented, please note that the deflection is relatively small, i.e. not the full range:

 



@dany93 does that also move the yoke then?

dany93:
@gilbertohasnofb wrote:

> But it seems to me that the deflection of the elevator surface is enormous comparing to the deflection of the tab. Shouldn't the animation be a bit more subtle?

I would tend to agree. Although I don't know for the reality. The question is: can an control surface be 100% deflected with only a trim setting? That seems a lot.
But we are in simulation and, in FG, changing the trim value comes to an offset for the control surface deflection. Hence the crude animation. Improving this is possible, but tricky. It depends on your motivation. And I could say nothing for the quantitativity.

wlbragg:
>changing the trim value comes to an offset for the control surface deflection. Hence the crude animation. Improving this is possible, but tricky.

@dany93 could you direct me where to look at this? Is this link to elevator control surface or offset done in the FDM or in FG source?

dany93:
@gilbertohasnofb wrote:
> @dany93 does that also move the yoke then?

Yes.
On the ultralights that I've seen, the elevator falls down full range. And the stick follows. One even have to be careful when checking the elevator amplitude on the ground (manually from the tail) because if you let it drop, you can damage the dashboard with the stick end going full forward and hitting.

wlbragg:
I see that elevator-trim changes elevator-pos-norm, do you know where that happens?

dany93:
@wlbragg wrote:
> I see that elevator-trim changes elevator-pos-norm, do you know where that happens?

Dany wrote:
> But we are in simulation and, in FG, changing the trim value comes to an offset for the control surface deflection. Hence the crude animation. Improving this is possible, but tricky. It depends on your motivation. And I could say nothing for the quantitativity.

wlbragg:
@dany93 
>changing the trim value comes to an offset for the control surface deflection

Where in the code does this happen, FG source or c172p.xml FDM.
I am more than motivated and interested in this. But I could speed up my own enlightenment with a bit of help!
It seems to me that this animation is simple wrong and I don't like it. I am sure there are at least a couple ways to fix it. A ack (which in my opinion it already is), or try to at least take a look at where this is happening and seeing if it truly is too difficult to fix.

gilbertohasnofb:
>> does that also move the yoke then?
> 
> Yes. The link is simple rigid cables, possibly with intermediate rods.

So in RL the pilot can have his hands on the yoke or off it. So if while taxiing your hands are off, the elevator falls and the yoke moves, on the other hand if you hold the yoke tight regardless of airspeed then the surface can be straight. So here is a catch: in FG we don't have a way of telling if the pilot's "hands" are holding the yoke or not.

To me, the solution is to drop the elevator surface only when the engine is off. If the engine is on, then it is assumed the pilot is inside and it is assumed that they might very well be holding the yoke. What do you guys think?

@wlbragg would you like to have a new issue opened for this?

dany93:
@gilbertohasnofb wrote:
> To me, the solution is to drop the elevator surface only when the engine is off. If the engine is on, then it is assumed the pilot is inside and it is assumed that they might very well be holding the yoke.

I'm not sure there is a satisfying solution. I'm wondering if you are not aiming to go too far.

Don't forget that, in every conditions even with the engine stopped, moving the commands makes the surfaces move.
IRL, often a first check is, possibly from outside the aircraft, move the stick (or the yoke) and look if the control surfaces move accordingly.

In FG, everyone does this with his joystick, looking at the aircraft from the exterior (or from inside). This possibility cannot be canceled.
Very different situations, natural in real life, difficult to conciliate in simulation...

gilbertohasnofb:
> I'm not sure there is a satisfying solution. I'm wondering if you are not aiming to go too far.
> [...]
> In FG, everyone does this with his joystick, looking at the aircraft from the exterior (or from inside). This possibility cannot be canceled.

@dany93 You are right, and I am happy leaving things as they are. The point you make about using joysticks is spot on. And even for mouse users, it would be super weird to have the yoke suddenly move every time you turn on the engine. Indeed that was a bad idea...

dany93:
>     > changing the trim value comes to an offset for the control surface deflection
> Where in the code does this happen, FG source or c172p.xml FDM.

It is in FG source code.
Perfectly satisfying from the simulation point of view.
Also close to reality in flight.
In flight, FG takes the control the surface deflections, and moves the aircraft accordingly. Same cause to effect in reality.

The only fail is for animations.
One possibility may be to animate the control surfaces via an intermediate 'animation' property. This 'animation' property would attenuate the control surface movement (in fact, it would decrease the offset) vs the trim value. Only for animation purposes
Just a first thinking....

wlbragg:
>The only fail is for animations.
...
Only for animation purposes.
It must decrease the trim effect on surface positions, but keep the amplitude due to controls (yoke).

Thanks @dany93 that really was my first question I needed answered. If I can deal only with the animation and leave the actual FDM alone then this maybe doesn't have to be complicated. I think I have enough information to make a decision as to what I am capable of doing, if anything.

>To me, the solution is to drop the elevator surface only when the engine is off.

@gilbertohasnofb I really don't think this is the worst idea there is. I think it would add to the authenticity and look really cool. It actually might be a very simple thing to do.
A simple condition that checks for engine off, WOW and 0 ground speed = sag the elevator. Any condition other than that and we leave it operate as default. The reason I think this is no issue to do is with that condition check it really doesn't matter if you adjust a control surface or not. The aircraft isn't doing anything. It's just for looks.

>@wlbragg would you like to have a new issue opened for this?

@gilbertohasnofb 
If you want to make a new issue for 
>So what about this: we could create a new "mechanical adjustments/options" menu or something like that and have things that pilots don't normally adjust or change. We could even move the engine selection over there. And so there we would find the rudder and aileron trims as a single slider each.

>And we could still make the rudder tab clickable and that would open a dialog with only the rudder trim. This is similar to what I did with the luggage weight: there is a small dialog but one can also adjust those values in the fuel and payload main dialog.

Go ahead, I still like this idea and it would be a good idea to re-factor the GUI and put it all together as you mentioned.

As far as drooping the elevator, I am going to try a simple condition and see if a) it works and b) if there is any complications that changes my mind about it. If you want to make a new issue for it that would be OK and we can cancel it if it doesn't pan out.

wkitty42:
On 03/03/2018 12:22 AM, wlbragg wrote:
>     To me, the solution is to drop the elevator surface only when the engine is off.
> 
> @gilbertohasnofb <https://github.com/gilbertohasnofb> I really don't think this 
> is the worst idea there is. I think it would add to the authenticity and look 
> really cool. It actually might be a very simple thing to do.
> A simple condition that checks for engine off, WOW and 0 ground speed = sag the 
> elevator. Any condition other than that and we leave it operate as default. The 
> reason I think this is no issue to do is with that condition check it really 
> doesn't matter if you adjust a control surface or not. The aircraft isn't doing 
> anything. It's just for looks.


don't forget about securing the yoke with the lock... doesn't the lock go 
through a hole in the shaft to hold it in place and keep it from turning? that 
would move the yoke to a point where the elevator may very well be level instead 
of drooping...

just thought i'd mention that to add to the brain game here :)

dany93:
(more thinkings...)
Dany wrote:
> The only fail is for animations.
> ...
> Only for animation purposes.
> It must decrease the trim effect on surface positions, but keep the amplitude due to controls (yoke).

Even that, I'm not sure that the animation (and FG) is so wrong.

Let's take an aircraft which is in equilibrium in cruise with zero rudder trim.
If you decrease the airspeed (reduce power), the aircraft must take a high AoA in order to keep horizontal flight. Possibly up to close to stall, where the elevator is almost 100% deflected upwards. Not recommended. However, you are close to this during the final leg.
To keep the aircraft at this high AoA (airspeed, attitude, glide slope) the pilot has two options:
- permanently pull the yoke (which is uncomfortable and unstable),
- or adjust the elevator trim in order to have the yoke almost neutral (easier, you just make small rectifications).
In the second case, the elevator is not far from 100% deflected upwards, **due to the trim**. The yoke position is almost 100% pulled, accordingly. Without pilot action.

gilbertohasnofb:
@ wlbragg it would really look cool to have the elevator fall down on the ground, but the issue of that surface being connected to the yoke is really problematic. As I said, it will look super weird if the yoke moves by itself at every engine start, and also users with joystick will nevertheless override any control surface position that we change.

We could of course just make the elevator surface animate down, regardless of the yoke position. Kind of a compromise that would look good, but then again would imply that the yoke and elevator are not directly connected.

algefaen:
I'd vote against the elevator dropping at low / 0 airspeeds. As @gilbertohasnofb  says, it's connected to the yoke. I would find it strange if the yoke in the aircraft did not match my joystick position. If playing without a joystick I would still be surprised if arrow keys didn't directly affect the elevator. One of the pre-takeoff checklist items are "flight controls free and correct", which IRL means glancing back at the elevator to see that it actually moves when you move your yoke. (IFF we had had proper force-feedback support then it could make sense that the force-feedback drops the elevator by moving the joystick down. But I guess the industry has given up on force feedback.)

dany93:
@wlbragg,
My current understanding on the trim effect on elevator deflection in FG.
(if it can help and _to be confirmed_)

Internal Properties, [interval]
`/elevator` [-1, 0, +1] whatever `elevator-trim`
`/elevator-trim` [-1, 0, +1]

`/fdm/jsbsim/fcs/elevator-cmd-norm` [-1, 0, +1] whatever` elevator-trim`

`/fdm/jsbsim/fcs/pitch-trim-cmd-norm` = `elevator-trim` [-1, 0, +1]


Calculated in the FDM (lines from 674, `<flight_control name="FCS: c172">`)

(`/fdm/jsbsim/fcs/`)
`elevator-cmd-norm` + `pitch-trim-cmd-norm`, clip to [-1, +1] ==> =` pitch-trim-sum`

`pitch-trim-sum` is limited inside [-1, +1], and affected by the `elevator-trim` value, which gives a trim offset.

If X = `elevator-trim value`, the `pitch-trim-sum` interval is [-1+X, X, 1], the central value X being taken for `elevator` = `elevator-cmd-norm` = 0
The trim effect is that one end of the interval is no longer reached, and we have an offset for elevator = 0
Example :
`elevator-trim` = +0.3 ==> `pitch-trim-sum` interval is [-0.7, 0.3, 1], with 0.3 for `elevator` = 0 = `elevator-cmd-norm`

After that, `pitch-trim-sum` is converted into degrees [-28, +23 for trim = 0], radians, `elevator-pos-norm`. Like `pitch-trim-sum`, these last ones are affected by the trim offset.


**Elevator Animation:**
**If you wish to modulate the trim effect seen on elevator vs airspeed** (trim effect between 0 at rest and full effect at moderate to high airspeed),
I think that the starting quantity is `pitch-trim-cmd-norm` = `elevator-trim`
Create a property `animation-pitch-trim-cmd-norm` = `pitch-trim-cmd-norm x K`, with `K` between 0 (at rest) and 1 (with airspeed).
To be added to `elevator-cmd-norm` for the elevator animation. Maybe clip the sum inside [-1, +1] or the corresponding degrees.
Thus, the elevator deflection **seen** would be `elevator-cmd-norm` at rest (usually expected value) and `elevator-cmd-norm` + `pitch-trim-cmd-norm` in flight.

Hard to explain... Please tell me if mistakes or unclear.

wlbragg:
@dany93 
Thanks, I'll have to look that over carefully before I can comment.

>I 'm wondering how many users notice this anomaly in the elevator animation vs trim...

Probably not that much as you really only would notice it at rest and if your moving the trim at the same time. It seems "difficult" to see it without going outside normal viewing positions and at the same time adjusting the trim wheel, not the easiest thing to do. These minor issues take up enormous amount's of time. But some of them just "stick in your craw". :smile:

dany93:
@wlbragg 
Something else.

1 - The elevator deflection is clipped inside [-1,  +1], trim effect included.
Consequence 1: if the trim is close to or equal to + or -1, the yoke has no effect at one end (the elevator is already full deflected).
Consequence 2 for the animation: if you cancel the trim effect on the ground for the elevator deflection, the elevator will be seen close to neutral even with full trim (OK) but the yoke will have no effect in one direction.

2 - @Dany wrote
Create a property `animation-pitch-trim-cmd-norm = pitch-trim-cmd-norm x K`, with K between 0 (at rest) and 1 (with airspeed),
To be added to `elevator-cmd-norm`.

Other way, pragmatic, not very clean but may be easier:
The elevator animation is currently done by `surface-positions/elevator-pos-norm` (although I don't know where this property is calculated).
For the elevator tweaked animation, use `surface-positions/elevator-pos-norm - pitch-trim-cmd-norm x K` (which come to subtract the trim effect for the animation on the ground)

wlbragg:
Once again I am getting to fragmented so here is what I would like to do.

This PR can be pushed once reviewed. It will close issue #622, issue #987 and issue #879.

I would like to potentially make a new issue for Fix elevator trim tab animation (maybe). That has dropped down on  my priority list but I want to preserve @dany93 explanation of how it works.

I would like to resume work under a new PR that will encompass #790, #6 and possibly #35.

wlbragg:
Adding the door handles

fgfs-screen-017

Why do we have hot spots for the handles as the entire door is a hot spot?

door

wlbragg:
Door handles and fairing cutouts are done so I ready to have this PR merged and closed.

wlbragg:
I decided to do the shadows on the pedals as they wont be affected by any panel work.
Also, door handles and fairing cutouts are done so this PR is ready to merged and close.

gilbertohasnofb:
> Why do we have hot spots for the handles as the entire door is a hot spot?

Maybe we should have hotspots only on the handles and not on the whole door.

wlbragg:
>door handles and fairing cutouts are done

I forgot to animate the door handles, I would like to do that before closing this PR.

>Maybe we should have hotspots only on the handles and not on the whole door.

We could, I'm not sure what would be best.

gilbertohasnofb:
> We could, I'm not sure what would be best.

I like the idea of using the door handles, but it's up to you if you think that's good or not. @onox some opinion?

onox:
I don't have an opinion about it. But what's that weird sticky thingy to the right of the radio's? Is it for the GPS instrument? I think it should be hidden when the GPS is not visible.

gilbertohasnofb:
> what's that weird sticky thingy to the right of the radio's? Is it for the GPS instrument? I think it should be hidden when the GPS is not visible.

Yeah, that's the new GPS mount. @wlbragg do you think you can hide it together with the GPS? 

Also, I noticed another issue with the UV-mapping, this time with the interior of the doors:




These should be mapped to the door, more or less like this:

wlbragg:
@gilbertohasnofb 
Wow, I didn't see any of you or @onox last posts and decided to remove the glass and door  from the "pick" open animation. Also I noticed the garmin mount and pushed a condition for it as well. I'm getting nervous, I think we're starting to send some strange ESP thoughts across the globe.

I'll look at the interior UV issue now.

wlbragg:
I think we need to do something to make it stand out like this.
 I can map it to the same texture as the dark part at the bottom of the door.
cessna172_i2

wlbragg:
What about this?

fgfs-screen-002

gilbertohasnofb:
> What about this?

That looks fine, but I think they are also fine as they are, and often they have the same colour as the door. This is OH-CTL in RL:

_dsc8055

wlbragg:
fgfs-screen-003
fgfs-screen-004
fgfs-screen-005

wlbragg:
OK, your way looks best I think.
fgfs-screen-008
fgfs-screen-009
fgfs-screen-010

wlbragg:
OK, done!

gilbertohasnofb:
@wlbragg Ok, so I will work on those wing tips. After that can we merge this or is there still anything else missing?

wlbragg:
I am doing the exterior door handle animations right now. That's all I got.

gilbertohasnofb:
@wlbragg Should I wait for you or can I start working on the texture? Will you need any textures for that?

wlbragg:
I am all done. Have at it.

gilbertohasnofb:
@wlbragg I have some good and some bad news for you. The good is that the new wing tips look great, thanks for fixing the UV-map. The bad news is this:

fgfs-screen-001

So the GPS and the amphibian gear are colliding now... Since the GPS is in such a fantastic position for all other variants, being able to be used by both pilot and copilot, it would be a pity to change its position. So what about changing its place _only_ for the amphibian variant? This way we have an awesome GPS location for all others. Would that be too difficult to implement?

Other than that I guess we are done.

gilbertohasnofb:
@wlbragg sorry, one more thing. For the sake of consistency, would you be willing to add a latch to the baggage door and use that to pick the animation? The latch looks identical to the one you already modelled, so we could just copy and paste it.

wlbragg:
OK, I think that is it. I'm finished as far as I know.

wlbragg:
@gilbertohasnofb 
Why did you decide to leave the blue on the end of the default wing livery? It is not there in the real life paint job.

screenshot_2018-03-05_18-49-12
screenshot_2018-03-05_18-47-51

gilbertohasnofb:
@wlbragg sorry, this was an oversight of mine, fixing it now

gilbertohasnofb:
@wlbragg Done. Thanks for fixing the GPS, looks perfect now. And the baggage door also looks great, I like how you managed the clickspots by the way, with the whole interior of the baggage door.

And here is a little surprise, I just managed to finish a new livery I have been working on during the last days, hope you guys will enjoy it. It has a very old livery feel to it, and the cockpit colours have those terrible old 172 colour combinations (based on photographs of the real thing): 

fg-1
fg-2
fg-0

gilbertohasnofb:
@wlbragg @onox From my side, this is ready to be merged. See if you guys need to check anything more, else you can just merge this.

wlbragg:
Looks great!

I'll be in touch with you before long about the cockpit modifications. I need to make sure I understand what it is we need to do. I looked over the thread about it and I think I understand the problem.

I guess we can let @onox take the final look and merge this PR.

gilbertohasnofb:
> I'll be in touch with you before long about the cockpit modifications. I need to make sure I understand what it is we need to do. I looked over the thread about it and I think I understand the problem.

Sure. This week I am a bit busy until Friday, I have a supervisory meeting coming up and let's just say I wish my research topic was in aircraft modelling and texturing :wink: So I will be online these next days and happy to clarify whatever is necessary about that thread, but probably I won't be having too much free time to test stuff and give thorough feedback. But from the weekend on I am "completely" free. 

> I guess we can let @onox take the final look and merge this PR.

Sounds good :+1:

onox:
I like the new livery. Looks vintage :)

About the trim tab: if I move the elevator down and then set the trim wheel to -1, the elevator moves to the horizontal position. Is this intended?

onox:
Shouldn't it be that the trim wheel just moves the trim tab (rotation animation) and then, in the FDM, the trim tab modifies the elevator deflection (depending on airspeed).

What do you think @dany93 ?

onox:
Or is it just the "depending on airspeed" that is not implemented in the FDM? I didn't look at the code :sleeping:

onox:
I'm probably forgetting that we're in a simulator without force feedback. If the FDM part is okay @dany93 , then the PR can be merged as far as I'm concerned.

gilbertohasnofb:
The effects of the trim are all fine, the issue is the animation. @wlbragg
spoke of using a filter to map the airspeed into max elevator deflection
due to trim, so that 0 kias = no deflection due to trim. I think this needs
to be taken care before merging this PR

wlbragg:
@onox 

> If the FDM part is okay @dany93

I think it has been established the the FDM is correct. The issue as @gilbertohasnofb stated is only in the animation. We should be able to filter the animation based on airspeed without touching the FDM.

wlbragg:
@gilbertohasnofb 
What do you think?
https://www.dropbox.com/s/0ygakjvgmsyigc7/2018-03-07%2000-21-22.mkv?dl=1

wlbragg:
I finished both window frames including tying them to liveries. Also, look closely at the door seems, where the door would seal with the fuselage. There were huge gaps on both doors (looking from the outside). I re-aligned the doors and remove those gaps.

wlbragg:
This aircraft is looking sharp!
fgfs-screen-002

wlbragg:
This aircraft is looking sharp!
fgfs-screen-002

dany93:
@onox wrote:
> Shouldn't it be that the trim wheel just moves the trim tab (rotation animation) and then, in the FDM, the trim tab modifies the elevator deflection (depending on airspeed).

Yes, but the relation between the trim value and the elevator deflection is not depending on airspeed in the FDM (differently from IRL). Which depends on airspeed is the trim setting made by the pilot to compensate the force to exert.
IRL, the trim wheel just moves the trim tab (by a pair of cables), whatever the elevator position is.
On the ground at rest, changing the trim wheel has no effect on the elevator position.
When flying IRL, the trim tab exerts a moment on the elevator, which offsets its neutral position as seen from the pilot. Enabling the pilot to fly with almost no action.
In FG FDM, the trim tab gives an offset on the elevator **position** for a given yoke JS position. At rest as in flight. That's an important difference from IRL, where the trim gives an offset only in flight. Moreover, IRL, the change in "neutral" (that is: with no force from the pilot) elevator position is associated to a change in yoke position. Nothing like this in FG, where we have, not only no force feedback, but also no position feedback.
From the FDM point of view, the trim compensation has a correct effect (although not perfect because it "cuts" the elevator deflection on one side, differently from IRL).
From the animation point of view, the trim effect is qualitatively correct _in flight_, but not at all at rest on the ground.
@wlbragg was considering fixing this animation but the textures and 3D modeling were eating all his time.

Still farther, at rest on the ground, the elevator drops down and the yoke position follows (if the yoke is not held). But this point will not be fixed in FG animation. Too tricky, too much drawbacks, incompatible with our JS which have no force feedback, no position feedback.

https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369549067
https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369551181
https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369557362
https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369596454
https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369640834
https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369645751

https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-369933270
https://github.com/c172p-team/c172p-detailed/pull/992#issuecomment-370158266  --> detailed FDM + animation point of view.

> Or is it just the "depending on airspeed" that is not implemented in the FDM?
I'm probably forgetting that we're in a simulator without force feedback.

I guess the above responses will help.

gilbertohasnofb:
> What do you think?

Looks amazing, thanks a lot! I did notice those gaps in the door and I am happy they are fixed. But I have a couple of comments:

- perhaps the user should only be able to open the windows when clicking from the interior side of the door, since in RL they don't seem to be operable from the outside. Is this possible to implement?
- I noticed that the UV-mapping of the exterior part of the baggage door is not aligned very well with the livery. Do you think you could try to improve it a tiny bit? If so, use the 4X-CHV livery as reference, and make very small variations as the UV-map is nearly perfect already.

So there is only that plus the elevator animation with trim.

wlbragg:
>perhaps the user should only be able to open the windows when clicking from the interior side of the door,

I would have never thought of that detail :smile:
I Suppose it isn't too much trouble to go to interior view before your allowed to open your window. As you said, you can't do it in RL, unless your a locksmith or a repo-person. :rofl:

gilbertohasnofb:
> you can't do it in RL, unless your a locksmith or a repo-person. 

Exactly! :rofl: 

Oh, and one more thing, possibly my very last comment about the external model. Almost all liveries suffer from a bit of an ugly UV-mapping on the top back of the fuselage. I tried fixing it by adjusting textures, but that's not really possible. All liveries that have stripes on the sides should have the same angles, so fixing one should fix all of them, or at least improve it a bit. Do you think you could take a look at this? See:

fg

gilbertohasnofb:
@wlbragg two more small details:

1) the triangular face on the wing tip side of both ailerons is missing: 

image

2) in FG we can click through surfaces, so by making the back side of the baggage door a huge clickspot then we are effectively making the whole front side a clickspot as well. Perhaps we could leave only the handle then, which can be used from both sides.

gilbertohasnofb:
@wlbragg I will try to correct this once again from the texture side. Your first two images above are really helpful, would you mind sending me another image like the first one but from the other side? If I can see the faces of the polygons I might be able to tweak it.

wlbragg:
yuck3

wlbragg:
The images showing the "ugly" effect (not the "straight" effect) is my remapping that I just now did. I just now re-mapped it because when I originally corrected the fuselage tail issues I never really remapped it. When I originally corrected the tail I did have to move the tail end vertexes which would have potentially thrown off the mapping. That wasn't the issue in this case however, but I opted to re-map it just o be safe.

By the way, the baggage area UV mapping is corrected. That was a bit of a challenge to get it correct on all liveries. I thing whenever we cut the mesh (for the door handles) it can change the surrounding mesh like that. Once again, good catch.

I'm not sure how i missed the aileron ends, but those are corrected also, however not pushed yet.

wlbragg:
See edit above.

wlbragg:
@gilbertohasnofb 
Hold on for a second. We have a problem.

gilbertohasnofb:
@wlbragg Ok, I think I might be able to improve it by tweaking the textures, so don't change the UV-map of the fuselage at those points. I will wait for you before pushing any modifications.

wlbragg:
I have no choice, I have to have the mapping line up with the rest of the fuselage or the shaded graduation won't match, and it is really a tricky and a precise mapping. When I get the graduation to match, then it will be up to you to make the rest of it work. 
grad

Also I saw another UV issue I need to clean up. This stuff is really tedious to get perfect.
vs

gilbertohasnofb:
Since you brought this up, here are two more issues with the UV-mapping:

1) bottom part of the right side of the rudder with OH-CTL (careful if you decide to fix this as it will affect all other liveries) 

fgfs-screen-004

2) rudder trim


fgfs-screen-003

In order to see the last one, you will need the updated `tail.png` of the new livery, which I still did not push it. Simply copy the file below as `tail.png` and substitute `Models/Liveries/N73429/tail.png`

tail

gilbertohasnofb:
I will wait for you before doing any more work on the textures apart from this new tail, I noticed that I had forgotten to add the light green stripe to the rudder. After you are done, I will try to fix things on my end.

gilbertohasnofb:
One more thing, about this:



Yeah, it would be great if this gets cleaned up but priority should go for the side view, as this issue above is only noticeable from a very specific angle and zoom. What I mean is: don't compromise the side view to fix the matching above!

wlbragg:
I'm going to have to spend some time on this. I'm not sure how much I can do to correct this "ugly" part. I actually made everything else worse in trying to. It is a very difficult part to get right. So I have to start over and pull what is in the current 622 branch and start over. Starting over may mean doing nothing as you suggested originally. I actually got the "ugly" part to look really good, but it messed up the gradient and the bumpmap in the default livery.

I looked at filtering the elevator trim animation. @dany93 is right, I don't think it is worth it to even attempt to correct the animation when there is no airspeed. The immediate problem I ran into was even if I filter a correct animation, then you effectively remove the yoke effect on the animation. It is very complicated as @dany93  implied.

@onox, if you want to try to create the solution, go ahead, otherwise I am done with it and it will have to be left as is.

gilbertohasnofb:
@wlbragg Ok, let me know if you need some help testing the UV-map. And as much as it would be nice to have these issues fixed, please know that I am aware of how tricky this is. The best approach when 3D modelling is to first do a proper UV-map which will _never_ be changed, and then texturize it. But since we inherited a problematic model, there are already several passes in which I attempted to fix things via textures. I highly doubt we will ever be able to get rid of all these minor issues with the tweaks we are doing, but we can try at least to minimize them to the best of our abilities. But please don't lose any sleep over small imperfections, we have lived for many years with a _much_ worse looking aircraft and this PR is already making it look so much better... :wink:

wlbragg:
>I think I can fix that matching issue on the vertical stabilizer of 4X-CHV.

I figured you could and it was creating too many issues trying to fix the mapping.

I'm all done!

wlbragg:
Oh, I forgot to push my last bit of work, hope I am not too late.

By the way, there is no way to limit clicking on the window latch from the outside. Same thing as the interior baggage door. But I did limit the baggage door pick to the handle.

Yes, comments at #992 were completed. The baggage pick will be in this last push. Aileron end also, seems I lost it during one of my reworks.

gilbertohasnofb:
> Oh, I forgot to push my last bit of work, hope I am not too late.

Not at all, I need to go to sleep already. Tomorrow my day starts quite early, so I will take a look at the textures tomorrow. 

> By the way, there is no way to limit clicking on the window latch from the outside. Same thing as the interior baggage door. But I did limit the baggage door pick to the handle.

No problems, I thought that would be the case. Thanks anyway!

wlbragg:
Now I am done, have a good night!

gilbertohasnofb:
@wlbragg I just pushed the textures fixes. I tried my best on the top back of the fuselage and it does look better, albeit not perfect. But I guess we can live with that. I also fixes the tail matching in the 4X-CHV as well as with ZS-MDF and N73429. So from my side I am done.

But I did notice two small UV-mapping problems in the back of the aircraft which perhaps should not be too much trouble to get fixed. They can be seen using the new livery N73429:

fg-0

Other than that, I guess we are all set and done. I will open a new issue for the elevator movement with trim so that perhaps one day we can think of working on that.

wlbragg:
Only you could find the UV mismatch on the hidden side of the rudder. :smile:

gilbertohasnofb:
> Only you could find the UV mismatch on the hidden side of the rudder. smile

In RL I am pedantic and annoying, but here I get to use my powers for a good cause :wink:

gilbertohasnofb:
@wlbragg I still found two small problems. 

1) the back side of the fuselage which is hidden by the rudder is still not matching the livery on the top part, and has a white spot on the bottom (see red squares)

2) the inside part of the rudder cable boxes should be black and use that texture I made for them, or did I understood something incorrectly and you want them in the fuselage colour? (see purple arrow)

fg-0
fg-1

This is the texture I am talking about:

fg-2

wlbragg:
>the inside part of the rudder cable boxes should be black 

Oops, messed that up moving it to match on the outside.

wlbragg:
OK, should be good now.

gilbertohasnofb:
@wlbragg That's perfect, thanks for the fixes. I just did some last texture fixes to improve a bit more the part of the fuselage hidden by the rudder and now all liveries are looking good! 

> Beings @gilbertohasnofb and I have touched it so much it would be best if you have the last review and merge it when your ready.

@onox Yes, this is ready to be merged. It would be great if you make a last review in case we missed something, but I think it's looking all good now.

wlbragg:
@onox, sorry you waited too long and I got bored. This is something I've wanted to fix for a long time. Maybe I can add the disk brakes in the future, but for now I think this will do.
@gilbertohasnofb if you get a chance please review and check for any anomalies in the rest of the mesh. It seems like every time I add something, something else gets reassigned a texture or a light setting. I'm starting to get gun shy.
fgfs-screen-001
fgfs-screen-002

wlbragg:
fgfs-screen-003

onox:
I didn't see @gilbertohasnofb's comment until half an hour ago. I'm not the one to look for when the mesh needs to be reviewed. @gilbertohasnofb can do that just fine, so he can merge the PR when he's OK with everything.

onox:
Those black thingies go through the main gear a bit.

wlbragg:
If your talking about the steps, that is a bolt that goes through the strut with a nut on the end (I didn't actually add a nut) to hold it on the strut.

onox:
OK.

gilbertohasnofb:
@onox Not a problem, I will do a last review and merge this myself then.

> if you get a chance please review and check for any anomalies in the rest of the mesh. It seems like every time I add something, something else gets reassigned a texture or a light setting. I'm starting to get gun shy.

@wlbragg what exactly should I be looking for? What was the last issue you found? You posted three screenshots but I don't understand what you were trying to show me, sorry. Once this is clear, I will do a last review and hopefully merge this :smile:

gilbertohasnofb:
@wlbragg I see now by the name of your commit that you are talking about the landing gear strut. So:

1) the new struts look great! But now that they are looking so realistic, perhaps we could add the brake box to the wheel as well. It's just a simple cube and you could use the same metallic texture as the exhaust pipe.




2) the fairings could also do with a little bit of improvements, as the new strut is joining it in a weird way. See how the connection between the strut and the fairing have a kind of 'shell'. Would it be possible to modify the fairing model as to include that?



This is how it looks right now:

fgfs-screen-001

3) the fuselage should have a small hole in the shape of a half circle where the main gear connects to it, see:

image
image

wlbragg:
OK, ready for review.

gilbertohasnofb:
@wlbragg wow, these last modifications made such a difference to the gear, they are looking stunning!!! Thanks for your hard work once again

onox:
Yeah, nice work @wlbragg  :smiley:

PR #995, opened 2018-03-03
Fix wrong production year in xml files

### LINKED ISSUE #985: C172P, landing light in the wing or the nose?
This is probably a stupid question, but I understood our 172P to be a 1981 model? Supposedly, "in 1982, the "P" saw the landing lights moved from the nose to the wing to increase bulb life", according to Wikipedia - could our landing light actually be that of the 1982, not the 1981? Or are we actually a 1982 172?

gilbertohasnofb:
Fixes #985 

* Fix wrong production year in `c172p-set.xml`
* Removed the word `detailed` from the description in `c172p-set.xml` as this has become the de-facto default aircraft and not a separate "detailed" model which would need to be differentiated
* Improves description in `c172p.xml`, as `Cessna 172P Skyhawk` is better than `Cessna C-172`

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -19,7 +19,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 <PropertyList include="Aircraft/Generic/Human/Include/walker-include.xml">
 
     <sim include="c172p-views.xml">
-        <description>Cessna 172P Skyhawk (1981 model, detailed)</description>
+        <description>Cessna 172P Skyhawk (1982)</description>
 
         <long-description>The Cessna 172 Skyhawk is a four-seat, single-engine, high-wing fixed-wing aircraft. First flown in 1955 and still in production, more Cessna 172s have been built than any other aircraft.</long-description>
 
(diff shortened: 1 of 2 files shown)


c172p-set.xml
```diff
@@ -19,7 +19,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 <PropertyList include="Aircraft/Generic/Human/Include/walker-include.xml">
 
     <sim include="c172p-views.xml">
-        <description>Cessna 172P Skyhawk (1981 model, detailed)</description>
+        <description>Cessna 172P Skyhawk (1982)</description>
 
         <long-description>The Cessna 172 Skyhawk is a four-seat, single-engine, high-wing fixed-wing aircraft. First flown in 1955 and still in production, more Cessna 172s have been built than any other aircraft.</long-description>
```
onox: You could write '172P' instead of '172' in the long descriptions?

onox: Just '172' is fine too.

onox: Maybe just leave it at '172' since it refers to the series.

gilbertohasnofb: Yeah, adding a P would make the second phrase puzzling since the model P was not first flown in 1955.

PR #996, opened 2018-03-04
Add keyboard shortcut for the flashlight

gftg85:
I don't know if this change is acceptable upstream or useful to other people, but it has been very useful to me in night flights.

### DIFF
--- a/c172p-keyboard.xml
+++ b/c172p-keyboard.xml
@@ -47,6 +47,16 @@
         </binding>
     </key>
 
+    <key n="102">
+        <name>f</name>
+        <desc>Toggle flashlight</desc>
+        <repeatable type="bool">true</repeatable>
+        <binding>
+            <command>nasal</command>
+            <script>c172p.toggle_flashlight()</script>
+        </binding>
+    </key>
+
     <key n="108">
         <name>l</name>
         <desc>Increase Panel lighting</desc>


wlbragg:
Looks and works well. I think this is a good addition.

gilbertohasnofb:
@gftg85 Looks good to me. Just quick question though, is the shortcut `f` not used for anything else? I just find it surprising that a non-combinatoric shorcut (i.e. without `Ctrl` or `Shift`) was available. But if that's the case, I am all for using it!

gftg85:
@wlbragg, @gilbertohasnofb,

I did not see `f` listed as one of the `common aircraft keys` nor as one of the `basic simulator keys`, but that is not necessarily enough to answer if it's actually available.

In `c172p-keyboard.xml`, `f` is not used.

I don't know where to check for other shortcuts that might be in use (I'm quite a newbie in fgfs :))

gilbertohasnofb:
> I did not see f listed as one of the common aircraft keys nor as one of the basic simulator keys

I also checked and `f` is not listed anywhere. A good place to check is `keyboard.xml` which lists all shortcuts in use, and indeed `f` is free to be used. Thanks a lot for your contribution!

@wlbragg @onox how can we inform the user about this shortcut? Is there a way of having our own copy of `common aircraft keys` help dialog so that we could add it there?

onox:
The key should be documented in `c172p-help.xml`

gilbertohasnofb:
Ok, see https://github.com/c172p-team/c172p-detailed/pull/997

PR #1002, opened 2018-03-09
Correct the latitude

edward-p:
The exit condition by latitude is wrong in 2017.3.1

### DIFF
--- a/Tutorials/landing.xml
+++ b/Tutorials/landing.xml
@@ -236,7 +236,7 @@ You start the tutorial cruising at around 1000 ft and 90 kts. To begin your desc
             <condition>
                 <less-than>
                     <property>/position/latitude-deg</property>
-                    <value>41.539</value>
+                    <value>41.519</value>
                 </less-than>
             </condition>
         </exit>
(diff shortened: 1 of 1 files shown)


wlbragg:
@edward-p  thank you for the correction. After testing I think we could adjust even more. Would you please try these number and if you agree please resubmit.
495
487
484

edward-p:
@wlbragg
I will, when I have time.

edward-p:
@wlbragg 
I flyed again and I just find that the latitude is decreasing when approaching the runway.
I think the first and the second need increasing， the last needs decreasing, because of the difference between `<less-than>` and `<greater-than>`
I've resubmited.

wlbragg:
I haven't really touched any of these tutorials so I don't know that I should be the one to have the final say in merging this. However, this is a very simple coordinate location change and from what i saw we could potentially keep adjusting this forever and not get it perfect. So as far as I am concerned, the last fix by @edward-p is good enough. If I here no objections I am going to merge this soon.

wkitty42:
On 03/18/2018 10:32 PM, wlbragg wrote:
> If I here no objections I am going to merge this soon.


i don't hear* any objections and i've been here for a while :)

wlbragg:
Thank you @edward-p

PR #1007, opened 2018-03-15
Improve ground objects

### LINKED ISSUE #1000: Improve ground objects
Improve ground objects

1) the hose is not properly connected to the aircraft
2) the textures of the cones and the fuel tank are quite bad, the latter being particularly low res
3) the ladder could also use some texturing
4) we could attempt to add the reflection effect to them, as to match the look of the aircraft

fg-1
fg-2
fg-3

gilbertohasnofb:
Fixes #1000 

:warning: :warning: :warning: Not ready to be merged yet! :warning: :warning: :warning:


gilbertohasnofb:
Ok, I started working on this and here is what I did so far:

* new textures for everything but the power unit
* fixed the power unit cable not properly touching the aircraft fuselage

fg-0
fg-1
fg-2
fg-3


missing:
* better UV-map of the fuel tank
* any modelling improvements that @wlbragg might think of, though I think these models are decent enough for their purpose
* can ground equipment objects be saved between sessions?

@wlbragg about the UV-map, could you give me a hand? I needed a new texture for those ugly black boxes which are above the wheel, so I had to steal a bit of the metallic texture (it used to be double the current side). So this is the new texture:

ref

And here are the problematic areas that need to be re-UV-mapped:

fg-4

wlbragg:
>can ground equipment objects be saved between sessions?

Yes, we can do that.

wlbragg:
>can ground equipment objects be saved between sessions

Admittedly I had never used this equipment before, so now that I have I think I get what your asking for here which is way simpler than what I was imagining. You want to make your GUI selection choices as to what you have showing persistent, right?

Yes we can easily do that. 

I think one check box called "Persistent" that when checked will make any items checked persist to the next session. Sound right?

wlbragg:
Hum, thinking more about this, this equipment is always located next to the aircraft no matter where it is. If your going to make it persistent, don't you want it positioned off the runway in an appropriate spot like at a parking position? If you do that then it would be difficult to position the aircraft to the objects (especially the ladder) when taxing back to the airport parking. So in that event ideally they would be movable.
I guess where I am going with this is we can make it as simple or complex as we want. The code I use for cargo lets you position this type of object anywhere you want with the shift-left-click of the mouse and then it remains persistent at that spot.  Do we want to do something like this instead?

gilbertohasnofb:
@wlbragg these ground equipment behave in a complete different way than other objects, as (if I am not mistaken) they are permanent in their position within a FG session. The whole code was imported from the c182s and I am quite clueless on how it works, to be frank. Concerning the persistence, that's not a big deal to be honest.

gilbertohasnofb:
You missed this spot here, this should be metallic as well:

fg

And would it be too much trouble to make the wheels a little bit more high-poly? They are the only thing that look quite terrible in this model, everything else looks fine.

gilbertohasnofb:
By the way, please do a `git pull` as I modified the texture for the fuel tank a tiny bit.

gilbertohasnofb:
@wlbragg much better, thank you so much

wlbragg:
I hope you OK with it, I meant to explain why I only did so little but forgot to. It was the pain to gain ratio. I still think maybe in the future we could make these objects relocatable and persistent like I do my cargo in the Aircrane. It would be cool to be able to bring the services to any location and leave them there for as long as you like.
I studied the existing objects and code and it is using all the same techniques I am familiar with. So it really wouldn't be difficult to achieve the desired outcome.

gilbertohasnofb:
I am totally ok with what we got so far. If you ever feel like giving a shot on making these objects permanent between sessions then by all means go for it.

PR #1016, opened 2018-03-18
Sound improvements

### LINKED ISSUE #959: engine sound too soft at idle
The engine sound of our c172p is maybe too soft at idle.

### LINKED ISSUE #1015: Gyroscope sound can be heard outside the aircraft
Playing the sound only while inside (or perhaps, near the aircraft?) would help?

Can still here the gyro from over here:
image

gilbertohasnofb:
Fixes #1015 
Fixes #959 

* gyro sound now decreases with distance, being barely audible from the outside views
* engine sound is louder at idle
* engine sound is a bit higher pitched

How to test it:
* test the gyro sound from several distances outside
* test that the gyro sound decreases in volume if you go to cockpit view and move the view outside
* test engine volume and pitch at idle
* test engine volume and pitch on the green arc
* test engine volume and pitch on a suicidal dive

### DIFF
--- a/c172-sound.xml
+++ b/c172-sound.xml
@@ -905,6 +905,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-radio-lighting</name>
             <mode>in-transit</mode>
@@ -921,6 +922,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-horizon-offset</name>
             <mode>in-transit</mode>
@@ -937,6 +939,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-heading-offset</name>
             <mode>in-transit</mode>
@@ -953,6 +956,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-heading-bug</name>
             <mode>in-transit</mode>
@@ -969,6 +973,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-altimeter</name>
             <mode>in-transit</mode>
@@ -985,6 +990,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-nav0-radial</name>
             <mode>in-transit</mode>
@@ -1001,6 +1007,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-nav1-radial</name>
             <mode>in-transit</mode>
@@ -1017,6 +1024,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-adf-rotation</name>
             <mode>in-transit</mode>
@@ -1033,6 +1041,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-egt-rotation</name>
             <mode>in-transit</mode>
@@ -1049,6 +1058,7 @@
                 <factor>0.2</factor>
             </volume>
         </dial>
+        
         <dial>
             <name>dial-trim-wheel</name>
             <mode>in-transit</mode>
@@ -1305,6 +1315,7 @@
                 <property>/sim/model/c172p/sound/volume-boost-doors-thunder1</property>
             </volume>
         </thunder1>
+        
         <thunder2>
             <name>thunder2</name>
             <mode>once</mode>
@@ -1314,6 +1325,7 @@
                 <property>/sim/model/c172p/sound/volume-boost-doors-thunder2</property>
             </volume>
         </thunder2>
+        
         <thunder3>
             <name>thunder3</name>
             <mode>once</mode>
@@ -1562,7 +1574,6 @@
             <reference-dist>2.5</reference-dist>
             <max-dist>200.0</max-dist>
         </damage-gear-right>
-
     </fx>
 
 </PropertyList>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
@onox so can you just go over the code to check I didn't do anything explicitly stupid?

@wlbragg @legoboyvdlp @wkitty42 can one or more of you guys take a look at this? See the first post for a recommendation on how to test this

gilbertohasnofb:
@dany93 would you also be able to take a look at this?

wkitty42:
@gilbertohasnofb  i think i can test this... not sure, though... i know i need to pull a new update... then what? checkout bug-1015 and see what happens while running through the list above?

gilbertohasnofb:
@wkitty42 If you have a clone of this repo, just update your master with a `git pull` then do a `git checkout bug-1015`. That will change your local repo to the branch `bug-1015`. Then just fire up the sim. :smile:

c172-sound.xml
```diff
@@ -149,7 +152,6 @@
         
         <gyro>
             <name>gyro</name>
-            <type>avionics</type>
```
onox: Why did you remove `<type>` here? AFAIK it's used so it's volume can be controlled in FG's sound settings dialog.

onox: Okay, you've already answered this in the issue.

gilbertohasnofb: I did that because if a sound is of type <type>avionics</type> then it is not affected by the max-distance for some unknown reason to me. It was the only way I could get this to work... see https://github.com/c172p-team/c172p-detailed/issues/1015#issuecomment-373998045

gilbertohasnofb: I think it's a small price to pay in order to have the sound volume working in all views, which is more important to me. What do you think?

onox: :+1:

wkitty42:
@gilbertohasnofb yes, that's what i've done... i got into the craft, turned on the battery switch ad listened but didn't hear anything... so i turned on the avionics switch and still didn't hear anything... so i went ahead and started the craft... ALT on, mixture to ~80, throttle to ~15, pump 3 times, turn the key to Both and hold 's' until it started...

with the engine running, it is a little quieter than when outside... zooming in and out makes the engine sound louder and softer...

i pulled the mixture to 0 and the throttle to 0... engine shut down... still not hearing anything... check my system's sound mixer and raise my fgfs volume slider to max from ~75... NOW i hear a really high pitched whine... if i turn everything off except the battery, i still hear it... turn off the battery, it goes away... i tried to hear a difference between inside and outside but when i scroll the wheel, it says i'm 6 meters and won't let me get any closer... i can't hear the whining at this 6 meter distance...

i started the engine again and let it warm up... then ran the throttle to max until the temp reached 245 on the guage... then i throttled back down to ~12 and i realized that i do hear that high pitched whine now... using 'v' to step out of the craft, i don't hear the whine... 'V' back into the craft and i do hear it... it is like a mosquito but a bit quieter...

i haven't tried actually flying, though... just what i can do on the ground for the moment...

kinda sucks that i have to have my system's mixing board set with FGs output maxed as well as having to max the sound control in the sim, too...

how's that?

gilbertohasnofb:
> NOW i hear a really high pitched whine...

Yes, that's the gyro and it's very soft. When the engine is on it should 99% cover it.

> it says i'm 6 meters and won't let me get any closer...

When in the cockpit view you can move your view position to any point and distance, which is how I tested it.

> how's that?

That's helpful, many thanks for taking a look at this.

@wlbragg @dany93 Would any of you give me your opinion as well, in particular when it concerns the new slightly higher pitch sound for the engine? Then we can merge this and move on.

wkitty42:
On 03/18/2018 06:09 PM, Gilberto Agostinho wrote:
>     how's that?
> 
> That's helpful, many thanks for taking a look at this.


it has been a while... i've been trying to keep up with you guys but you're fast 
and seem to have a lot of stuff done before i even get up :laughing:

ok, so i'm going to "git checkout master" and i should return to the ""normal"" 
craft, right? then an update will bring in all the stuff just recently merged? i 
think that's the right path to follow... i am learning a bit as time moves on 
hahaha...

gilbertohasnofb:
> ok, so i'm going to "git checkout master" and i should return to the ""normal"" craft, right? then an update will bring in all the stuff just recently merged? i think that's the right path to follow... i am learning a bit as time moves on hahaha...

Yes, that's correct. If you don't have any local modifications, commits or anything like this then you can just switch back and forth between branches using `git checkout <branchname>` without any fear of having issues.

legoboyvdlp:
I'll be away tomorrow morning, but can test in the afternoon ;)

legoboyvdlp:
Shouldn't the engine noise change as you rotate around the aircraft?

legoboyvdlp:
Otherwise, I'd say 10/10.

gilbertohasnofb:
> Shouldn't the engine noise change as you rotate around the aircraft?

That's beyond what FG sound system is capable of. We only take into consideration the distance and your relative angle to the sound source (i.e. the volume changes according to distance and the panning changes if the source is on your front, left or right).

gilbertohasnofb:
If there are no other comments I will just merge this.

legoboyvdlp:
Alright! 

I did think we had doppler, though. I know it works in flyby 🤔 

Also, freesound.org is quite useful:
https://freesound.org/people/JLWSound/sounds/322321/

Plenty of CC0 sounds that could be reused in future. I believe CC0 = public domain?

gilbertohasnofb:
> I did think we had doppler, though. I know it works in flyby 

Sure, but Doppler only affects the sound if the source is moving in relation to the observer, which is not the case with these types of views. And to implement Doppler for the helicopter view while panning around would be just silly IMO. 

> Also, freesound.org is quite useful:

Half of our sounds sources come from there, though I normally have quite a bit of work improving, cleaning and looping them :wink: But anyway, this is off-topic, this PR is about fixing our current engine sound which was too soft and having low pitch, not about replacing it.

PR #1018, opened 2018-03-19
Bug #771

### LINKED ISSUE #771: Propeller disk looks red under certain circumstances
Propeller disk looks red under certain circumstances: try loading the plane in the weather preset "low pressure system" and advance the time from dusk to evening by holding the key `t`, advance the time towards evening. At a certain point, the back of the propeller disk look red, while the front looks normal. See:

fgfs-screen-003

fgfs-screen-005

wlbragg:
Fixes #771 

Changed material settings on Propeller.Fast to remove red tint during low light.


onox:
It's no longer red: more yellowish/greyish instea, but I'll let @gilbertohasnofb decide whether the PR can be merged.

gilbertohasnofb:
Hmm, I still see a red propeller. See:

fgfs-screen-002

I also don't think this here is looking too good, the nearly fully black spinner with a bright yellow propeller:

fgfs-screen-001

gilbertohasnofb:
I am also thinking of improving the propeller disk texture in this PR. I had written:

> And by the way, our propeller disk could do with some improvement. Would it be possible to have the front of the disk mapped to a different texture than the back? And could the front disk depend on livery? This way I can try to reproduce how the propeller in each livery looks like.

@wlbragg what do you think of that? See the front and back of the propeller disk on the c182s for an example of what I am considering to do.

wlbragg:
>Would it be possible to have the front of the disk mapped to a different texture than the back?

This is done correct?

>And could the front disk depend on livery? This way I can try to reproduce how the propeller in each livery looks like.

Yes, as far as I know, you create texture for it and we can organize them in the livery code.

So we have the two items above front/back of prop, livery dependent blur, and now also front and back of blur prop with separate mapping and livery.

Yes to all of it as far as I am aware.

I don't see either yellow or red anymore?

I'll be away till late this afternoon or evening. If you make any test textures for the blur I can hook them up and create a two sided blur disk. One thing I noticed while looking at this compared to the 182 is our blur is more transparent, which I like better. So maybe we can keep an eye on that as we proceed.

gilbertohasnofb:
>> Would it be possible to have the front of the disk mapped to a different texture than the back?
> This is done correct?

I think you are mistaking the propeller while still with the propeller disk. There is only one single propeller disk blur used for all liveries, front and back, as far as I can tell. And I don't know of any option I could use in the xml livery files to select between them.

> So we have the two items above front/back of prop, livery dependent blur, and now also front and back of blur prop with separate mapping and livery.
>
> Yes to all of it as far as I am aware.

Great, so I will start working on them as soon as I can. 

> I don't see either yellow or red anymore?

Let me know if you need help reproducing this. For those screenshots, I loaded the c172p on SBTA at dusk, set advanced weather to Low Pressure and slowly moved the time forward.

> I'll be away till late this afternoon or evening. If you make any test textures for the blur I can hook them up and create a two sided blur disk. One thing I noticed while looking at this compared to the 182 is our blur is more transparent, which I like better. So maybe we can keep an eye on that as we proceed.

I will work on this, but I would suggest we first try to solve the red/yellow tint issue before changing the textures, else we might lose the reproducibility of the issue. So I will work on this but not push it yet, all right?

onox:
Do we not merge this PR and just close it then? @wlbragg is already working on the mesh in another branch.

gilbertohasnofb:
@onox for me whichever route is best for him is also best for me. He just need to make sure that he brings his modifications over to the other branch.

PR #1019, opened 2018-03-19
Refactor damage model

wlbragg:
Fixes #319 
Fixes #592 
Fixes #895 
Fixes #1027
Fixes #1029 
Fixes #1031
:warning: Not ready to merge yet :warning:

PR #1034, opened 2018-03-29
Rgb to zero

### LINKED ISSUE #771: Propeller disk looks red under certain circumstances

wlbragg:
This fixes the prop permanently, as in the Blend so we don't have to remember to change the .ac file.

Fixes #771
 
Ready to merge after testing.


wlbragg:
I thought I would add this in while I was at it.
This drag-tweak-factor allows the aircraft to slow down more realistically on the water. As it was, the drag while satisfactory for high speed as in landing and touchdown but was nowhere near enough for after the aircraft looses most of its energy from its reduced speed. The aircraft would unrealistically continue to glide on the water past the point where it should really start to slow down. This simulates the effect of considerable drag as it settles in the water. It's somewhat of a cheat because it uses the engine rpm to decide when to apply drag VS any kind of speed measurement. The reason using speed wouldn't work is because you couldn't accelerate past the applied drag to achieve takeoff. Using speed worked against itself. This was an effective way to achieve the effect and was fully adjustable.

onox:
@wlbragg You're addressing different issues in the same PR. That's a no no :stuck_out_tongue: 

`git format-patch -1`
`git reset --hard HEAD~1`
`git push --force`

`git checkout -b issue-1037 master`
`git am 0001-Add-drag-tweak-factor-logic-to-hydrodynamics.patch`
`git push -u origin issue-1037`

wlbragg:
:fearful: :sob:

Thank you! You make things so easy, what you think I'm lazy? :rofl:

wlbragg:
FYI:
It required
git push --force --set-upstream origin rgb-two-zero
I'm not sure why this is as I was on rgb-two-zero branch at the time?

onox:
@wlbragg For some reason the "RGB to 0 for prop blur" commit is also in the `issue-1037` branch. I'll remove it there and force push to that branch.

gilbertohasnofb:
@wlbragg @onox What's the situation of this PR, is this ready for review and possibly merging or do you guys still need to do some cleaning on the commits?

wlbragg:
Nope, it's ready to merge.
 All it does is set the FastProp material setting in Blender to rgb = 0, 0, 0 so whenever we export the .ac it will retain that setting, which effectively removes the portion of the atmospheric lighting that is causing the red/yellow tints at low light conditions.

gilbertohasnofb:
@wlbragg thanks!

gftg85:
Is it possible that this merge reverted the fixes created by commit ID bc5aca41dce051ffca8c429a2c5d4537c61755cd?

I noticed that the sharp edge got back, after I updated the repository.

gftg85:
Hrm, maybe not this merge, but the merge of the previous pull request (#1019)

wlbragg:
Yes, unfortunately we had to revert the glass change because the rain effect was basically broken by setting up the glass that way. It's either a sharp edge in the glass corner or a really ugly looking rain streak effect.

PR #1036, opened 2018-03-29
Make some objects not clickable in external views

### LINKED ISSUE #1026: Some objects are clickable from the exterior
image

This is just a "nice-to-have" thing, but it would be nice if the air vent and window handles, at least, didn't work from the outside, unless the door is open. If you agree, could I add that to the animation?

legoboyvdlp:
Fixes #1026

- Window Handles, Door Handles, and Air Vents do not work without a window or door open in external views

Sorry - got it sorted now :)

### DIFF
--- a/Models/Interior/Panel/Instruments/AI/AI.xml
+++ b/Models/Interior/Panel/Instruments/AI/AI.xml
@@ -99,6 +99,9 @@
     <animation>
         <type>knob</type>
         <object-name>AdjustKnob</object-name>
+        <condition>
+            <property>sim/current-view/internal</property>
+        </condition>
         <action>
             <repeatable>true</repeatable>
             <binding>
--- a/Systems/c172p-doors.xml
+++ b/Systems/c172p-doors.xml
@@ -39,12 +39,28 @@
                 /sim/model/door-positions/leftWindow/position-norm gt 0
                 /sim/current-view/internal eq 0
            </test>
+           <test logic="AND" value="1">
+               /sim/model/door-positions/rightDoor/position-norm gt 0
+               /sim/current-view/internal eq 0
+           </test>
+           <test logic="AND" value="1">
+                /sim/model/door-positions/rightWindow/position-norm gt 0
+                /sim/current-view/internal eq 0
+           </test>
         </switch>
 
         <switch name="doors/windowR-enabled">
            <default value="0" />
            <test logic="AND" value="1">
                 /sim/current-view/internal eq 1
+           </test>
+           <test logic="AND" value="1">
+               /sim/model/door-positions/leftDoor/position-norm gt 0
+               /sim/current-view/internal eq 0
+           </test>
+           <test logic="AND" value="1">
+                /sim/model/door-positions/leftWindow/position-norm gt 0
+                /sim/current-view/internal eq 0
            </test>
            <test logic="AND" value="1">
                /sim/model/door-positions/rightDoor/position-norm gt 0
@@ -73,12 +89,28 @@
                 /sim/model/door-positions/leftWindow/position-norm gt 0
                 /sim/current-view/internal eq 0
            </test>
+           <test logic="AND" value="1">
+               /sim/model/door-positions/rightDoor/position-norm gt 0
+               /sim/current-view/internal eq 0
+           </test>
+           <test logic="AND" value="1">
+                /sim/model/door-positions/rightWindow/position-norm gt 0
+                /sim/current-view/internal eq 0
+           </test>
         </switch>
 
         <switch name="doors/doorR-enabled">
            <default value="0" />
            <test logic="AND" value="1">
                 /sim/current-view/internal eq 1
+           </test>
+           <test logic="AND" value="1">
+               /sim/model/door-positions/leftDoor/position-norm gt 0
+               /sim/current-view/internal eq 0
+           </test>
+           <test logic="AND" value="1">
+                /sim/model/door-positions/leftWindow/position-norm gt 0
+                /sim/current-view/internal eq 0
            </test>
            <test logic="AND" value="1">
                /sim/model/door-positions/rightDoor/position-norm gt 0


onox:
I think it's not intuitive that a door/window being open controls whether the vents can be moved. It would be better if the vents can only be moved for an internal view and just not depend on the state of the windows/doors.

The windows are ok.

The door handles on the interior side should be clickable when the door is open.

wlbragg:
Why all the formatting modifications in areas that don't apply to this issue? I see corrections to many formating mistakes that are not in master to start with, only your local copy, I presume. Can you start with a clean copy so as not to be so confusing?

I am not understanding the entire premise of the issue. When adding some of the physical objects like door handles, etc. @gilbertohasnofb suggested and I agreed that interior "pick" objects should be deactivated from the exterior.

I agree with @onox that if the door is open then the interior door and window handles are OK to be hot as well as any potentially any other logical objects. But that starts to get complicated really fast and touches a lot of code if it encompasses everything.
So I would limit it to doors and maybe windows. I think vents would not have much impact with the doors open anyway.

onox:
> Why all the formatting modifications in areas that don't apply to this issue? I see corrections to many formating mistakes that are not in master to start with, only your local copy, I presume. Can you start with a clean copy so as not to be so confusing?

@legoboyvdlp made a single commit on top of `master`. So the formatting mistakes that he corrected are in `master`. Having said that, it's usually better not to touch code that isn't related to the issue.

> that interior "pick" objects should be deactivated from the exterior.

Is it possible nowadays to disable the showing of the yellow wireframe by disabling the pick animation? That would be better.

legoboyvdlp:
Hi there,
I just used Notepad's blank tabs / space converter tool. So I did not intentionally fix them.

Also, the vents are very prominent and noticeable things. So I thought that they should also be included.

gilbertohasnofb:
> It would be better if the vents can only be moved for an internal view and just not depend on the state of the windows/doors.

> I am not understanding the entire premise of the issue. When adding some of the physical objects like door handles, etc. @gilbertohasnofb suggested and I agreed that interior "pick" objects should be deactivated from the exterior.

Yeah, I also think that's a better and less cryptic idea. Apologies for green lighting your proposal, @legoboyvdlp, should have thought of that.

> I agree with @onox that if the door is open then the interior door and window handles are OK to be hot as well as potentially any other logical objects. But that starts to get complicated really fast and touches a lot of code if it encompasses everything.
> So I would limit it to doors and maybe windows. I think vents would not have much impact with the doors open anyway.

:+1:

legoboyvdlp:
So I should remove the vents?

And what about the formatting fixes? It would make no sense to remove them or undo them, and would be quite impossible to do as well. ;)

gilbertohasnofb:
My opinion is this:

* interior door handle clickspot should work only if door/wind is opened.
* windows are fine
* all else should be disabled when in exterior view IMO. 

@onox @wlbragg what do you think?

legoboyvdlp:
I'd make the windows only work if the door or window is opened as well, otherwise, that sounds fine.

legoboyvdlp:
What is the status of this? You want me to remove the vent, and disable everything else whenever the view is not inside the aircraft?

gilbertohasnofb:
@legoboyvdlp Sorry, I seem to have overseen your last messages. 

> What is the status of this?

My opinion still is this:

* interior door handle clickspot should work only if door is opened.
* I like your idea about the windows: they can only be opened or closed from inside or when the door is opened.
* all else (vents, panel, etc.) should be disabled when in any exterior view IMO. 

What do you think?

legoboyvdlp:
Alright, reverting the vents, and disabling everything else once I get a chance ;)

gilbertohasnofb:
@legoboyvdlp I don't know if I am doing something wrong when testing this branch from your repo, but I don't see any of those features implemented:

fg-0
fg-1

legoboyvdlp:
Try to click them ;)

I had no idea how to get rid of the yellow highlight with the property, though.

gilbertohasnofb:
> Try to click them ;)
> 
> I had no idea how to get rid of the yellow highlight with the property, though.

Hmm, I see. So it means that the yellow clickspots as well as the cursor changes are still happening, but they are only disabled. That's already something but I was so much hoping for something else. @onox do you happen to know if we can do anything further or is this as much as we can have?

onox:
Adding a condition to the knob/pick animations doesn't help?

Zaretto:
Whether or not to disable the yellow highlighting based on the condition is a subject for the dev. list as it has wide implications. I'm also not exactly sure how to do this.

legoboyvdlp:
If it is a matter of FlightGear, SimGear, etc changes, then this can be just left alone till the next version of FlightGear, if that ever happens. ;)

gilbertohasnofb:
> Try to click them ;)

I just tried and I have access to everything from the outside, i.e. I can move the vents, change all dials in the cockpit, open the windows while the doors are closed. I really don't know if I am doing something wrong to review this, but what I did was to create a brand new brach and pull from your repo, using:

```
git checkout -b legoboyvdlp-issue-1026 master
git pull https://github.com/legoboyvdlp/c172p-detailed.git issue-1026
```

@onox are you able to test this PR?

legoboyvdlp:
I'm not sure what is happening. It worked here.
And that method of pulling should have worked.

onox:
@gilbertohasnofb If you do `git log` you should see the last commit is "Pick animations of..." just like here on Github. Then you know you have the last commit of his branch.

@legoboyvdlp You have introduced tabs in many files again. You should really make sure you switch your editor to spaces before you start editing files.

If you use Linux or MacOS (if they provide GNU's grep) then you can use `find . -type f -name "*.xml" -exec grep -HP '\t' {} \;` to find all the files.

onox:
@legoboyvdlp Have you checked the commit before you pushed? Github shows a large diff for `Models/Interior/Panel/Instruments/vor/vor.xml` and now it's full of tabs.

legoboyvdlp:
@onox are you sure?
I tested about 10-15 lines, and they all have tabs. Actually, perhaps it was using tabs before, and I switched them to spaces...?

I use the NotePad ++ tabs to spaces function, and it does nothing on taht file you mentioned, which further backs up my belief there are no tabs there 😕

gilbertohasnofb:
> are you sure?

On the top of this PR, here in the GitHub interface, look for the 'files changed' tab, search for the `vor.xml` onox mentioned then try highlighting the indents before and after your PR. Tabs have indeed been introduced.

> tabs to spaces function

Lego, the best approach is to set up your editor to use spaces instead of tabs and not to use these automatic functions, else you end up potentially modifying files not related to your PR.

legoboyvdlp:
Sorry... I was testing the wrong branch, and the old vor.xml did not have tabs, obviously.
As for Notepad ++, I don't think there is any way to automatically use spaces, but if there is I'll add it...
Thanks!

gilbertohasnofb:
> As for Notepad ++, I don't think there is any way to automatically use spaces, but if there is I'll add it...

See: https://stackoverflow.com/questions/8197812/how-does-one-configure-notepad-to-use-spaces-instead-of-tabs

gilbertohasnofb:
> However, I do not really wish to do this, since the IDG-A32X uses tabs, and not spaces.

You can just flip that toggle whenever you are editing in one project or other. Or you could install a second editor and use it only for the c172p. I personally love and swear by Atom.

wkitty42:
FWIW: i've seen and been in this tabs vs spaces mess for 30+ years... the discussions are always the same... it boils down to tabs will always lead to problems in formatting... you just don't have that problem with spaces... just count them out... i learned how to type by writing basic, asm and pascal code... pascal taught me two space indention and i write all my code with that... even perl, bash, php and others... the problem with tabs is they're never the same width everywhere... in one editor, they may be two spaces, in another they may be four... it is painful and i try to stay out of these types of discussions... the tab key is too far away for my touch typing...i have to stop and look for it... on the other hand, my thumbs have callouses from tapping the space bar twice for each indention... i love them and never have to look for the space bar :laughing: 

that's it, i'm done with that :smile_cat:

onox:
You're not supposed to press the spacebar two or four times, but just configure the editor you're using to insert the right amount of spaces when you press the Tab key (which is perfectly reachable on the keyword IMO).

Anyway, the important thing is to be consistent and either use only spaces or only tabs in your project.

onox:
The good:

- Air vents can no longer be moved in an external view
- Windows and interior door handles cannot be openen in an external view if the door is closed

The bad:
- In an external view, windows and interior door handles should work if the door is currently *open*
- Horizonal offset knob AI and the volume dials of the radio's still work in external view

The ugly:
- Tooltips are missing for:
  - primer
  - master alt
  - master bat
  - avionics switch
  - circuit breakers
  - lighting switches
  - parking brake handle
  - carb heat slider
  - flaps handle
  - little knob to the right of the horizontal offset knob in the AI
  - windows
  - door handles
  - Show/hide yoke
  - place/remove control lock
  - show checklists

legoboyvdlp:
I'm not aware of a way to add the tooltips, unless I make a seperate animation with the tooltip for each object. Would you like me to do this?

I must have accidentally disabled the door and window handles again... :)

onox:
@legoboyvdlp This looks like a bug in FG

@Zaretto It seems that if a `<condition>` is added, the tooltip is never shown, even when the condition is true.

legoboyvdlp:
@gilbertohasnofb I couldn't figure out how to push, but otherwise, Atom is very good and useful!

@onox should be fixed.

gilbertohasnofb:
@legoboyvdlp Check this out: https://stackoverflow.com/questions/33325204/how-can-i-run-git-commit-and-git-push-in-atom But that said, I tend to use a terminal for git commands and Atom 'just' for programming and visualizing what is modified, committed or untouched in the directory tree.

legoboyvdlp:
This _should_ be ready to merge.

onox:
We should wait with merging this PR until FG is fixed so that tooltips are shown when the condition is true.

onox:
@gilbertohasnofb Are you able to reproduce the problem I had with the tooltips? See https://github.com/c172p-team/c172p/pull/1036#issuecomment-380166649

gilbertohasnofb:
@onox not really. I just tried with the baggage door. To open it, you can only use the door handle, whose `hovered` definition on line 3650 of `Models/c172p.xml` (on `master`) has no `<condition>`. The tool tip shows fine to me. To close the baggage door, you can click anywhere on it since we now have a hidden hotspot which matches the whole door. It's `hovered` definition on line 3699 has a `condition` and yet the tool tip also shows fine to me. Could you test using this same example I gave? Because if you can also see this tool tip the same as I do then there must be other things in order to reproduce this bug.

onox:
I cannot download the patch from @legoboyvdlp's repository.

legoboyvdlp:
Hmm - I'll change my remote and push this branch if thats ok? I'm afraid I can't do it until Monday though.

gilbertohasnofb:
@onox my test example above was using our `master`, so you don't need his patch to confirm what I said.

gilbertohasnofb:
@legoboyvdlp what's the situation with this branch? Are we still planning in working on this and merging this? If so, please be _ultra_ careful not to regress any work done in those xml files, this branch is old and reflects a much older state of our `master`.

legoboyvdlp:
Not too sure. I understood FG changes might be needed to make this work?

gilbertohasnofb:
@legoboyvdlp I agree, I think there could be a label for the pick animations that could be set to `internal`, `external` or `both` or something like that. I also find it too hacky to add hundreds of conditionals... 

So in that case should we close this PR or is there some work here that is still worth being merged? @onox any opinions on this?

onox:
@legoboyvdlp Can you create a new branch `issue-1026` in this repository and add your changes?

legoboyvdlp:
I don't seem to have this branch anywhere anymore unfortunately!
Sorry...

onox:
@legoboyvdlp You could create a new branch and look at the changed files in this PR.

PR #1048, opened 2018-04-05
Fix-GPS-display

### LINKED ISSUE #1045: Cannot see text in GPS screen
I believe that is caused by the black background color and black front color.

fgfs-screen-430

wlbragg:
Fixes #1045 

Don't really have an answer as to why this was broke. I simply re-imported the original .ac saved it and exported it again and it is fixed. I don't know it the .ac got corrupt somehow or what.

This should be ready to merge into master, please verify and then merge.


gilbertohasnofb:
@wlbragg Thanks for taking a look into this. This certainly fixes the GPS for me as well. The only issue I see is that the screen goes from a normal looking grey into a full white when under sunlight, and the latter doesn't look very natural. Would you have any idea how could we improve this?

fg-0
fg-1
fg-2

gilbertohasnofb:
@wlbragg Just checked and that's a texture issue, the backgrounds of individual screens are white but in reality this GPS has those greenish screens:



Keep this PR opened and I will fix that myself.

gilbertohasnofb:
@wlbragg I just pushed new green screens, but for some reason they are not working in the sim. Compare:

screens

with:

fg-0
fg-1

Could it be that `screens.png` and `symbols.png` are being loaded from the gps at FGData and not from our local copy for some reason? I remember that being a problem with the GPS exterior texture as well.

wlbragg:
Yes, I think your right, let me see if this is an easy fix or not.

gilbertohasnofb:
Or are the screens above 'greener' but too bright? Could it be something to do with the material of the screen?

gilbertohasnofb:
Great. I am leaving now but I will check this again in the evening, if there is any need to tweak the textures further or to test and merge this.

wlbragg:
@gilbertohasnofb you pushed them to the wrong path. We are using the GPS located at 
Models/Interior/Panel/Instruments/garmin196/
not
Models/Interior/Panel/garmin196/

But the support .xml's are pointing to the fgdata location and I think I need to change them to our local version.

wlbragg:
I'll wait for you to change the textures. I copied the your new texture over to my local version to test, you may want to go darker, I can barely notice the change in the sim.

wkitty42:
On 04/05/2018 02:25 PM, wlbragg wrote:
> @gilbertohasnofb <https://github.com/gilbertohasnofb> you pushed them to the 
> wrong path. We are using the GPS located at
> Models/Interior/Panel/Instruments/garmin196/
> not
> Models/Interior/Panel/garmin196/
> 
> But the support .xml's are pointing to the fgdata location and I think I need to 
> change them to our local version.


do we have a local version because it is easier to fix locally and then adjust 
for pushing to FGData when completed?? why aren't we just using the one in 
FGData and fixing that one?

gilbertohasnofb:
@wlbragg why the hell do we have two local copies of the GPS then? That makes no sense. I moved my textures to the correct folder and I am also deleting that other folder, hope that is all right. The textures are now working fine, so this PR is ready to merge if everything is fine on your side.

> do we have a local version because it is easier to fix locally and then adjust for pushing to FGData when completed?? why aren't we just using the one in FGData and fixing that one?

Because fixing things directly in FGData is a pain in the neck and affects a million aircrafts at the same time. Locally we can do whatever we judge to be good for our aircraft. But keep in mind that whomever is interested in the work we do here can simply merge it back into FGData.

wlbragg:
>why the hell do we have two local copies of the GPS

I think the second one, the one were not using, was meant to be for the canvas version. We can get rid of it until someone is ready to convert it.

gilbertohasnofb:
> I think the second one, the one were not using, was meant to be for the canvas version. We can get rid of it until someone is ready to convert it.

Well, in that case I think it's right of us to have gotten rid of it, that's really messy. First, as far as I know nobody started working on the Canvas port so there is no need to have a folder for that. Secondly, the paths become confusing, the model location as well, etc. Thirdly, that should _never_ be on the `master` as the Canvas port is not a ready feature yet. So I think removing it is the right decision and if one day anyone feels like tackling the Canvas port, there is an opened issue for that and the person can just create a new branch for that.

@onox do you agree with this? If so, I guess this is ready to be merged, right @wlbragg ?

wlbragg:
>I guess this is ready to be merged, right @wlbragg

Yes

PR #1053, opened 2018-04-07
issue-1050

### LINKED ISSUE #308: Clicking "Repair" button doesn't set craft to upright position
Now that we have the "Repair" button in the "Aircraft Options" window, I find myself using it more and more to try to correct things from accidents but some things it just doesn't do and that means that one has to exit or reset the sim.
1. Craft is upside down
2. Fix all instruments that may be broken (had the horizon get broken the other day after a nasty)

Those two are all I can think of at the moment... I'm sure there may be others, too.

### LINKED ISSUE #913: Shadow remains visible but unoriented upon water upside down
I'm looking for a good property, maybe an fdm/attitude, to use in the shadow condition code. What I want is one that will work for upside down. The contact points don't register WOW over water so we're getting a shadow after crashing upside down, which is a common attitude when trying to land upon water with gear down in the amphibious configuration.

### LINKED ISSUE #1050: Repair option is available even with damage turned off
in the aircraft options, the repair button is still available and working even when damage is turned off...

20180406-c172p-repair-available-without-damage-option-01

wlbragg:
Fixes #1050 
Fixes #308
Fixes #913 

This PR also correct the issue of a broken shadow when aircraft is at extreme orientations.

### DIFF
--- a/Models/Effects/shadow/3damphibiousshadow.xml
+++ b/Models/Effects/shadow/3damphibiousshadow.xml
@@ -39,6 +39,12 @@
                     <property>/fdm/jsbsim/unit[12]/WOW</property>
                 </not>
                 <not>
+                    <property>/sim/rendering/shadow</property>
+                </not>
+                <not>
+                    <property>/fdm/jsbsim/orientation/upside-down</property>
+                </not>
+                <not>
                     <property>/fdm/jsbsim/crash</property>
                 </not>
                 <not>
--- a/Models/Effects/shadow/3dbush26shadow.xml
+++ b/Models/Effects/shadow/3dbush26shadow.xml
@@ -39,6 +39,9 @@
                     <property>/sim/rendering/shadow</property>
                 </not>
                 <not>
+                    <property>/fdm/jsbsim/orientation/upside-down</property>
+                </not>
+                <not>
                     <property>/fdm/jsbsim/unit[12]/WOW</property>
                 </not>
                 <not>
--- a/Models/Effects/shadow/3dbush36shadow.xml
+++ b/Models/Effects/shadow/3dbush36shadow.xml
@@ -39,6 +39,9 @@
                     <property>/sim/rendering/shadow</property>
                 </not>
                 <not>
+                    <property>/fdm/jsbsim/orientation/upside-down</property>
+                </not>
+                <not>
                     <property>/fdm/jsbsim/unit[12]/WOW</property>
                 </not>
                 <not>
--- a/Models/Effects/shadow/3dbushdefshadow.xml
+++ b/Models/Effects/shadow/3dbushdefshadow.xml
@@ -39,6 +39,9 @@
                     <property>/sim/rendering/shadow</property>
                 </not>
                 <not>
+                    <property>/fdm/jsbsim/orientation/upside-down</property>
+                </not>
+                <not>
                     <property>/fdm/jsbsim/unit[12]/WOW</property>
                 </not>
                 <not>
--- a/Models/Effects/shadow/3dpontoonshadow.xml
+++ b/Models/Effects/shadow/3dpontoonshadow.xml
@@ -39,6 +39,12 @@
                     <property>/fdm/jsbsim/unit[12]/WOW</property>
                 </not>
                 <not>
+                    <property>/sim/rendering/shadow</property>
+                </not>
+                <not>
+                    <property>/fdm/jsbsim/orientation/upside-down</property>
+                </not>
+                <not>
                     <property>/fdm/jsbsim/crash</property>
                 </not>
                 <not>
--- a/Models/Effects/shadow/3dskishadow.xml
+++ b/Models/Effects/shadow/3dskishadow.xml
@@ -39,6 +39,12 @@
                     <property>/fdm/jsbsim/unit[12]/WOW</property>
                 </not>
                 <not>
+                    <property>/sim/rendering/shadow</property>
+                </not>
+                <not>
+                    <property>/fdm/jsbsim/orientation/upside-down</property>
+                </not>
+                <not>
                     <property>/fdm/jsbsim/crash</property>
                 </not>
                 <not>
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -19,8 +19,10 @@ var reset_all_damage = func
     setprop("/fdm/jsbsim/pontoon-damage/left-pontoon", 0);
     setprop("/fdm/jsbsim/pontoon-damage/right-pontoon", 0);
 
-    setprop("/orientation/pitch-deg", 0);
-    setprop("/orientation/roll-deg", 0);
+    if (getprop("fdm/jsbsim/orientation/upside-down")) {
+        setprop("/orientation/pitch-deg", 0);
+        setprop("/orientation/roll-deg", 0);
+    }
 }
 
 var repair_damage = func {
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -18,6 +18,9 @@ var reset_all_damage = func
     # Pontoons
     setprop("/fdm/jsbsim/pontoon-damage/left-pontoon", 0);
     setprop("/fdm/jsbsim/pontoon-damage/right-pontoon", 0);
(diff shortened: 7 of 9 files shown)


wlbragg:
We can make this as simple or complicated as we want.

The damage system is there to allow the user to simulate damage that is possible in RL.
The repair option is there to allow repair any damage that my have occurred.

Is the repair option supposed to simulate RL or simply allow users to repair damage whenever they want. I don't think this was ever discussed.

I noticed some time back that if the engine is running you cannot repair the aircraft.
I just now added that limit to it the Damage check box is not selected.
You might consider not to allow it if the aircraft is not on the ground or water.

The reason I bring this up is the most simplistic way to implement righting an upside down aircraft is to change pitch and roll to 0. Two simple lines of code. From my limited tests, it works in most cases, even when flying. It is unrealistic looking though, but then again so is repairing the aircraft anywhere than in a hanger or similar location.

So what are we expecting here, how realistic, how much effort to expend on this?

wkitty42:
On 04/07/2018 03:50 PM, wlbragg wrote:
> Is the repair option supposed to simulate RL or simply allow users to repair
> damage whenever they want. I don't think this was ever discussed.

there was some talk about it way back... AIR, the repair option was to allow the 
craft to be repair so you don't have to exit the sim and restart it to go flying 
again... this being along the same lines as having the ability to change the 
engine and landing gear without having to exit and reload using a different 
-set/xml file...


> I noticed some time back that if the engine is running you cannot repair the 
> aircraft.


yes, you can't repair it in real life when it is running, either...


> I just now added that limit to it the Damage check box is not selected.
> You might consider not to allow it if the aircraft is not on the ground or water.


awww... there goes the trick of breaking the wings off in a dive, shutting down 
the engine, repairing the craft and restarting the engine to hopefully recover 
before you hit the ground :lol:  i'm joking!! hahaha...


> The reason I bring this up is the most simplistic way to implement righting an 
> upside down aircraft is to change pitch and roll to 0. Two simple lines of code. 
> From my limited tests, it works in most cases, even when flying. It is 
> unrealistic looking though, but then again so is repairing the aircraft anywhere 
> than in a hanger or similar location.
> 
> So what are we expecting here, how realistic, how much effort to expend on this?


the way i see it, if the craft is crashed and on the ground, broken into itty 
bitty pieces, repairing the craft should fix whatever is broken and right the 
craft... if there was a special repair hanger at each airport, i'd say to 
reposition the craft to that hanger, too... but there isn't any special repair 
hanger at all airports so we just repair the craft where it is... then maybe we 
can get it into a clear area and take off again... i dunno...

i was going to write "get it fixed so you could lock it down for when you come 
back to the sim at a later time" but then i remembered that starting the sim 
gives you a good craft anyway... however, if you crash and repair everything, 
you can then teleport to an airport and take off from there... i don't think 
that teleporting after a crash will right the craft but i admittedly haven't 
tested that, either...

i think consistency was the main thing i was after, here... if we crash, we can 
repair but we don't get flipped upright... what's the use of repairing in that 
case? so anyway, i thought getting repaired and getting set back on your gear 
would be a GoodThing<tm>... then you can decide what to do after you finish 
laughing at your crash :)

gilbertohasnofb:
@wlbragg Thanks for implementing this, I just tested it and it works very well.

> The reason I bring this up is the most simplistic way to implement righting an upside down aircraft is to change pitch and roll to 0. Two simple lines of code. From my limited tests, it works in most cases, even when flying. It is unrealistic looking though, but then again so is repairing the aircraft anywhere than in a hanger or similar location.

Let's just not overcomplicate anything. Just having the aircraft upright is already fine :+1: 

But one thing could perhaps be improved: try clicking on the damage toggle to switch it on and off when the aircraft is _not_ broken and it will bounce in a super buggy way. Is there a way of detecting whether the aircraft is currently broken or not so that switching that toggle only activates the repair + flipping upright when the aircraft is broken? If not, then my opinion is that disabling the Repair button and connecting the Damage toggle to an automatic repair is not worth the introduction of this bug, and so I would revert the toggle and button behaviours.

wkitty42:
On 04/08/2018 08:35 AM, Gilberto Agostinho wrote:
> Let's just not overcomplicate anything. Just having the aircraft upright is 
> already fine 👍
exactly!

> But one thing could perhaps be improved: try clicking on the damage toggle to 
> switch it on and off when the aircraft is /not/ broken and it will bounce in a 
> super buggy way. Is there a way of detecting whether the aircraft is currently 
> broken or not so that switching that toggle only activates the repair + flipping 
> upright when the aircraft is broken?

is there not a ""global"" variable named (eg:) "craft_is_broken" that gets set 
any time there's any damage? then one only need to check

   if craft_is_broken then do_damage_repair;

or something like that... every block of damage code would need to set 
craft_is_broken to true no matter what it was set to when arriving in that code 
block... then do_damage_repair would set craft_is_broken to false at the end 
before exiting out of its routine...

   routine do_damage_repair;
   begin
     fix_wings;
     fix_gear;
     recharge_battery;
     craft_is_broken := FALSE;
   end;

yeah, pseudo code but the idea should be clear enough ;)

wlbragg:
@gilbertohasnofb yesterday I introduced a new property called /fdm/jsbsim/orientation/upside-down in the PR for removing the shadow when the aircraft is upside-down. If we use that property as a condition on whether to allow the flipping then I think it solves the problem. Only if the aircraft is <-90 or >90 pitch, same for roll, will the aircraft pitch and roll be reset. The only problem is like a dummy I pushed it through the hydro tuning PR VS the the one for the flipping the aircraft on repair. Maybe @onox could coach me on backing it out of that one PR, or I might be able to figure it out on my own. Then I will push it instead to the damage PR and make the correction and add the check for the upside-down condition to it. Will kill two birds with one stone, the upside-down damage repair and the upside down shadow off.

Let me see if I can figure out how to back it out of the hydro PR. Before @onox has to hold my hand. :smile:

wlbragg:
Oh, I wanted to mention. If your flying inverted, diving/climbing or rolling past 90 to -90 then the the aircraft will still reposition roll and pitch when either the "Damage" box is checked or "Reset" button is pressed. But that is a feat in and of itself and I don't think is an issue.

wkitty42:
On 04/08/2018 02:26 PM, wlbragg wrote:
> Oh, I wanted to mention. If your flying inverted, diving/climbing or rolling 
> past 90 to -90 then the the aircraft will still reposition roll and pitch when 
> either the "Damage" box is checked or "Reset" button is pressed. But that is a 
> feat in and of itself and I don't think is an issue.


am i misreading this? these repairs and flipping back over should only happen 
when one clicks the [REPAIR] button and when one turns OFF the damage option...

wlbragg:
Read it again. :small_airplane: 
>when either the "Damage" box is checked or "Reset" button is pressed.
So I'd say, yes, your misreading it.

wlbragg:
To expand slightly...
For normal flight envelope the roll, pitch reset shouldn't happen when either repair choice is actioned.

gilbertohasnofb:
Testing it now. 

> If your flying inverted, diving/climbing or rolling past 90 to -90 then the the aircraft will still reposition roll and pitch when either the "Damage" box is checked or "Reset" button is pressed. But that is a feat in and of itself and I don't think is an issue.

Yeah, this is just a ludicrous case. Let's just call that an Easter egg and not bother.

gilbertohasnofb:
Perfect, many thanks @wlbragg

wkitty42:
On 04/08/2018 02:59 PM, wlbragg wrote:
> Read it again. 🛩
> 
>     when either the "Damage" box is checked or "Reset" button is pressed.
>     So I'd say, yes, your misreading it.


damage checked tells me it is enabled... makes it sound like if you're flying 
and doing tricks, it will automatically flip you over... i think you even said 
that...

dany93:
What will happen if we break a wing? The aircraft spins and any roll angle can happen... 
Maybe also (I've not checked the roll angles, everything can happen) if we drop a wing in stall and spin, even with no damage.

wkitty42:
@gilbertohasnofb thanks... that explanation is a bit clearer than what i originally read...

PR #1058, opened 2018-04-15
Add factor to mag-compass

### LINKED ISSUE #1057: Magnetic compass indication bug
@dany93 wrote:

> I noticed that the 3D magnetic compass indication is wrong.
> About 180° error near 90° and 270° mag heading (map),
> Pretty (nearly) well at 0° and 180° mag heading.

wlbragg:
Fixes #1057

### DIFF
--- a/Models/Interior/Panel/Instruments/mag-compass/mag-compass.xml
+++ b/Models/Interior/Panel/Instruments/mag-compass/mag-compass.xml
@@ -19,10 +19,8 @@
         <type>rotate</type>
         <object-name>Ring</object-name>
         <property>/instrumentation/magnetic-compass/indicated-heading-deg</property>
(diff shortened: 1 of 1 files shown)


wlbragg:
@dany93  please test and verify this corrects the issue in #1057.

dany93:
Tested, correct.
Thank you @wlbragg.
At seeing the axis coordinates, I believed it would be more complicated than a simple -1 factor...

wlbragg:
I don't know if I accidentally removed it or if moving the object to an absolute position required me adding it. When I noticed that it started at the correct location for the default of 0 and at 90 it was 270 instead, I put 2 and 2 together. I didn't really think it was going to be that simple either.

PR #1063, opened 2018-04-18
FDM: make the aircraft less sensitive to controls

### LINKED ISSUE #1047: Is the pitch too sensitive?
Very little yoke movement (I believe only 5~10 millimeter) at pitch axis would cause head up and down very rapidly and abruptly, and it also cause VSI change huge amount, from ±500 to ±1000, does it real? 

I also test Cessna 337, CAP10B and DR400, these aircraft use jsbsim as well, but doesn't experience this problem. 

And c182s has the same problem too, but not so obvious and making me dizziness.

To re-produce, you have to fly longer and without AP all the time, such us fly from PHOG to PHNL, and using a accuracy joystick such T16000M.

dany93:
Fixes #1047

### DIFF
--- a/Models/Interior/Panel/Instruments/tc/tc.xml
+++ b/Models/Interior/Panel/Instruments/tc/tc.xml
@@ -43,8 +43,7 @@
         <type>rotate</type>
         <object-name>Ball</object-name>
         <property>/instrumentation/slip-skid-ball/indicated-slip-skid</property>
-        <factor>20</factor>
-        <!-- ini 5.729 -->
+        <factor>5.729</factor>
         <!--
            From the source for slip_skid_ball.cxx, the number returned is (-z_accel/y_accel)*10.
            For small theta, theta~=tan(theta) when theta is in radians.  So the angle of a "ball" 
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -501,8 +501,8 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 
     <controls>
         <flight>
-            <aileron-trim type="double">0.02</aileron-trim>
-            <rudder-trim type="double">0.013</rudder-trim>
+            <aileron-trim type="double">0.044</aileron-trim>
+            <rudder-trim type="double">0.02</rudder-trim>
 
             <freeze-yoke type="bool">false</freeze-yoke>
             <aileron-cmd type="double">0.0</aileron-cmd>
--- a/c172p.xml
+++ b/c172p.xml
@@ -1371,14 +1371,8 @@
                     <property>aero/qbar-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
-                    <table>
-                        <independentVar>aero/beta-rad</independentVar>
-                        <tableData>
-                              -0.3490	0.0322
-                              0.0000	0.0000
-                              0.3490	-0.0322
-                        </tableData>
-                    </table>
+                    <property>aero/beta-rad</property>
+                    <value>-0.092</value>
                     <!--
                           stall and spin (1):
                           via diedra effect.
@@ -1386,13 +1380,13 @@
                           and worsens the spin in a skidding turn.
                           But if too great, prevents the spin departure in a slipping turn
                     -->
-                    <table>
+                    <!-- <table>
                         <independentVar lookup="row">aero/alpha-wing-rad</independentVar>
                         <tableData>
                             0.279     1
                             0.297     3.5
                         </tableData>
-                    </table>
+                    </table> -->
                 </product>
             </function>
 
@@ -1530,7 +1524,7 @@
                     <property>aero/qbar-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/cbarw-ft</property>
-                    <value>0.1000</value>
+                    <value>0.1</value>
                 </product>
             </function>
 
@@ -1543,7 +1537,7 @@
                     <sin>
                         <property>aero/alpha-rad</property>
                     </sin>
-                    <value>-1.8000</value>
+                    <value>-1.8</value>
                     <!--
                         stall and spin (5):
                         horizontal tail stall (for flat spin)
@@ -1728,8 +1722,8 @@
                     <table>
                         <independentVar>aero/function/total-wing-damage</independentVar>
                         <tableData>
-                            0.0000	1
-                            1.0000	0
+                            0.0000  1
+                            1.0000  0
                         </tableData>
                     </table>
                 </product>
--- a/c172p.xml
+++ b/c172p.xml
@@ -26,7 +26,7 @@
         <vtailarea unit="FT2"> 16.5 </vtailarea>
         <vtailarm unit="FT"> 0 </vtailarm>
         <location name="AERORP" unit="IN">
-            <x> 44 </x>
+            <x> 40.6 </x>
             <y> 0 </y>
             <z> 59.4 </z>
         </location>
@@ -1018,7 +1018,7 @@
         </alphalimits>
 
         <hysteresis_limits unit="RAD">
-            <min>0.25</min>
+            <min>0.15</min>
             <max>0.297</max>
         </hysteresis_limits>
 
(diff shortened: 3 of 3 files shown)


dany93:
Every feedback has its interest, even if those of C172P pilots are more valuable. The criteria on which the critics are based are important too.
We are all more or less with partly subjective, based on memory, etc... feelings. Even on my part, only having a limited experience on Ultralights. Part of the performances can be quantified and measured, not all.

tonghuix:
I flew it from PHNL to PHMK, cruise at 3500ft AGL, IAS 110kt. The whole flight I feel is less sensitive and no any abrupt movement has been encountered, even at low fps. 

During cruise I tested *Accelerate Stall* (45 deg bank, 65kt stall speed), *Power-on Stall* with clean configuration (45kt stall speed), *Power-off Stall* with landing configuration (full flap, 40kt stall speed). And I also test slow flight. All works normal and perfect!

However, I found the plane is very hard to enter *spin*, I only flew into small radius circled down several times,  even at full stall and secondary stall with full rudder. I remember it was very easy to enter spin before.

dany93:
>However, I found the plane is very hard to enter spin

Are meaning _flat_ spin?
Yes, I didn't try to keep that working. In fact, it was an extra (kind of for fun) and I don't know how "easy" it is to have it flat spinning IRL.

When the main part of the FDM is fixed and stabilized, I will see if I can make it work again.

tonghuix:
Oh not only flat spin, any spin i cannot entering yet...

I tried some method which are introduced in some videos: 

- https://www.youtube.com/watch?v=DeTwuSo4xDE
- https://www.youtube.com/watch?v=52tPNkBcfmg
- https://www.youtube.com/watch?v=4dSrjVR0MvE&t=197s

dany93:
Very interesting videos. I'll keep them in my bookmarks.

With the version you have now, I could have (with default load):
- (soft and not certain) spins in horizontal flight with engine at idle,
- frank spins, both sides, in turns with moderate throttle (1500 RPM),
- but not in turns at idle.
- And no flat spins.
Mind: do these tests with elevator trim at zero (or close to). In FG, the trims prevent reaching full foil deviation in one direction (differently from IRL, where you only need more force).

But that should improve with the version I'm working on now.

tonghuix:
I also discover a awesome feature, adverse yaw which is not experienced before! When I use a uncoordinated turning, which is only  bank the plane no rudder or no enough rudder, the head will 'swing' to opposite direction of bank.

Even same as the video shows: https://www.youtube.com/watch?v=UV8xcm5xsuo (he fly a C152)

Is this a feature that you add this branch?

dany93:
I didn't _add_ adverse yaw, only it was too weak in the previous version. Constant, almost enough for cruise, too low for climb and approach (higher AoA).
It is constant in your version, I'm working on it just now, to have it depending on the AoA.
It reminds you to correctly use the rudder. Interesting.

wlbragg:
The video's are great, I'm working on the first one and our model performs almost identical.

dany93:
Changed AERORP (@okroth [indications](https://github.com/c172p-team/c172p-detailed/issues/1047#issuecomment-382449833)) and adverse yaw (table vs AoA).

Adverse yaw: use the rudder, watch the slip-skid ball. Particularly at low speeds.
Stall and spin behavior: recovered at 90%. Flat spin included, recoverable

tonghuix:
Good job!! I just want to test the slide slipping now.

tonghuix:
Wait some times please, more testing is needed. Especially someone who pilot the real one.

dany93:
At the condition that a real pilot accepts to spend time at doing tests...
In FG, anyone does that he _likes_ to do or that he deems as important, not what other people ask him to do.

Anyway, even as it is, I'm convinced that it is a true improvement from the initial version. Unless there is an important defect, merging it does not prevent from further refinements (if they can be done and there is someone wishing to do them).

wlbragg:
@onox, please merge after review if you wouldn't mind.

onox:
I find it difficult to do a quick descent on final approach (left aileron + right rudder (that's called side slipping?)). Is that how it's supposed to be?

Did @okroth test the FDM changes?

dany93:
Yes, that's side-slipping.

Difficult to give a quantitative response compared with reality. At 90 kt...80 kt it is easy to stabilize and control by almost only acting on the ailerons. From the videos that I've watched, this is the speed at which it is done.
Below, that's a subtle balance between rudder, ailerons and pitch angle (to keep airspeed). From the FDM point of view, the roll due to side-slip becomes more and more important at high angles of attack. IMO, the justification is that, at lower speed, the higher wing is shielded by the fuselage, down to a kind of one-wing stall.
Personally I like this difficulty, but I can't respond more precisely.

No, @okroth did not test it, despite [my request](https://github.com/c172p-team/c172p-detailed/issues/1047#issuecomment-382398158).

tonghuix:
I found a document about Spin Characteristics of Cessnas: 
https://www.kevincfi.com/files/pdf/manuals/Cessna/Spin%20Characteristics%20of%20Cessnas.pdf

onox:
@okroth Are you able to test the FDM changes? If not then we'll merge it, since I've no objections anymore.

dany93:
Anyway, @onox , your question on [difficulty at forward slip](https://github.com/c172p-team/c172p-detailed/pull/1063#issuecomment-383403234) is not lost. Later if needed, I hope that I can make it easier, I have to look at it.. but with more time. The problem is that one change has an influence on other handling (cross-coupling of parameters),  which means doing many other tests each time. By never being certain to find a compromise...

dany93:
Changes to make forward slip more stable (even with no flaps, which is generally recommended for forward slip).
Probably closer to real behavior IMO. Correct range: 80 kt... 70 kt (I think, less than 90 kt).

tonghuix:
Yes good job, the aileron turns more efficiency and the adverse yaw is more obviously too.

dany93:
Thank you for the feedback.
For the adverse yaw and coordinated turn that's weird... You need a bit less aileron deflection for the same turn, and I increased the adverse yaw just proportionally to compensate for the less aileron deflection needed. That should just give slightly more aileron efficiency, or less joystick for the same effect.
But if your feeling is positive, that's the point.

tonghuix:
I retest again. The aileron just in perfect. But adverse yaw is a problem, even at 10 deg bank (around 90kt), should apply rudder, even a little, to compensate adverse yaw too. And it is not use any rudder at all before.

dany93:
> And it is not use any rudder at all before.

I don't understand.
I've reloaded and tested the previous version (dcb8591), I see no difference.
At 90 kt, horizontal flight with engine (about 2000 RPM), starting from 0° roll,
then rolling 10° or even 20°with no rudder action:
- rolling leftward gives about 1 slip-ball displacement (temporary, during the aileron action),
- rolling rightward gives about 3/4 to 1 slip-ball displacement.

On both versions.

tonghuix:
Yep, and sorry that I did not test with previous version, so now it is the same.

But I noticed that the Dutch Roll is more obvious indeed.

tonghuix:
@dany93 could you please reference the c182s's FDM parameters?  I just fly c182s it is much more stable. 

It is balanced with adverse yaw and aileron efficiency.

dany93:
@onox, @wlbragg, do you object merging this PR?
As it is, I think that it brings an important improvement, and I see no regression compared with the current master version.
As far as this issue subject, it is fixed.

dany93:
@tonghuix,
The C182s is slightly (not much) more stable at cruise by letting the controls loose (after good trimming).
I don't know why, the main parameters are the same, based on Megginson C182S. A few more detailed tables for the C182, but IMO not enough to be sensible. Possible minor improvements for the future.
C182 propeller: variable pitch. Can it do some airspeed regulation?

tonghuix:
I only test at some speed, such 110kt, 100kt, 80kt, using prop lever to control speed. 

However, I don't know more about FDM difference between 172 and 182, but the 'feeling' is obviously.

okroth:
Could it be the propeller induced speed that is used in elevator and rudder effect computation?

dany93:
It probably has an effect, but as a shift of the values if everything is stabilized.
If I accurately trim the C172 (using the autopilot to set the trim, then disconnecting it), the aircraft is _perfectly_ stable. Which means that it is stable. Only, the trim may be slightly more tricky to manually set (even this, I'm not sure). When I manually trim it, its deviations are weak and slow. That's quibbling.

tonghuix:
Well OT, I also think the trim wheel is too slow..

gftg85:
Hi,

I noticed a difference with these changes applied and I would like to ask if this is indeed the expected behaviour...  I noticed this difference during take-off.

Before this change, I used to set throttle to 90%, then take-off. Right after take off, the aircraft would start to climb at a vertical speed that varies between 500 and 1000 feet/second for some time.  The variation becomes smoother with time, until it kind-of-automatically settles,  During this period, airspeed would be around 80 kias.

After this change, the climbing becomes a lot harder to manage.  Right after take off, the vertical speed varies a lot more (i.e.: outside the 500 to 1000 feet/s range that it used to be contained with), and it doesn't seem to settle.

Did anybody else noticed the same behavior?
If so, is it intended? (Maybe Cessna 172P is harder to fly than I imagined (based on my simulation-only experince (I'm not a real pilot))).

tonghuix:
Hi @gftg85 ,

I think takeoff speed is indeed correct after the change, due to the reality is. 

https://www.youtube.com/watch?v=0Wp0-l2_3t4 (jump to 9:50)

https://www.youtube.com/watch?v=BTpWlPL9zUE (jump to 8:10)

dany93:
It is not intended, it is the final result of the stability.
I have never piloted a C172, but I know from RL ultralights that you have to actively control the climbing speed after takeoff. You never let it go on its own.

gftg85:
OK, it looks more like these videos.  :)
So, the aircraft is indeed harder to pilot than I initially thought, hehehe.

Thanks for your kind replies.

Issue #1023, opened 2018-03-20
Improve 3D model and texture of all levers


gilbertohasnofb:
Improve 3D model and texture of all levers (including throttle, mixture, primer and carb heat):

![aad](https://user-images.githubusercontent.com/5700795/37676517-05ba8ff2-2c70-11e8-8243-0320a506e516.jpg)

![aaf](https://user-images.githubusercontent.com/5700795/37676518-05e4c1fa-2c70-11e8-9b65-f646c65e2e60.jpg)

![aag](https://user-images.githubusercontent.com/5700795/37676519-05ff85a8-2c70-11e8-80f3-59c6aae49668.jpg)

![abh](https://user-images.githubusercontent.com/5700795/37676520-062db824-2c70-11e8-8341-f4a5ebff19ce.jpg)

![_full](https://user-images.githubusercontent.com/5700795/37676521-064e29e2-2c70-11e8-98ab-f7a7de066abe.jpg)

![aaj](https://user-images.githubusercontent.com/5700795/37676587-3ac6ca30-2c70-11e8-9ca3-ebf5681f1f32.jpg)

![](https://user-images.githubusercontent.com/423703/37675452-3db04c04-2cb0-11e8-8836-b2ee6e94988f.png)

* throttle and mixture need a better 3D model, the handle needs a bit more detail and be a bit longer perhaps. Their bases (where the lever 'enters' the cockpit) need some rework.
* primer needs a better 3D model (handle is too shallow and not looking very good) + also a better texture
* carb heat needs a little bit more thickness, the handle is too big and the length when extended is way too short. The base needs to be resized as well, it's too big. Also the handle does not seem fixed which is why the 'square' handle does not necessarily point at 0 degrees nor 45 degrees with the horizon. Ours is ar 45 degrees, perhaps we could twist it a bit more?  
* consider improving the illuminations knobs as well

Related to: https://github.com/c172p-team/c172p-detailed/issues/802

gilbertohasnofb:
@wlbragg just notice two more things: our carb heat lever looks too short and the bit where you hold looks too large. See this excellent video at around 6:00 (this is a valuable source for a lot of things concerning the interior)

https://www.youtube.com/watch?v=sLFX33Ogmus

gilbertohasnofb:
I've updated list of things to do in the issue description

gilbertohasnofb:
Also, this issue is related to https://github.com/c172p-team/c172p-detailed/issues/802

[...]

PR #1066, opened 2018-04-22
Issue 6: Interior panel too low

### LINKED ISSUE #6: Panel is too low
3D02 - fuselage around the panel is way too large. Compare the pictures below:

http://i.ytimg.com/vi/LHPTqAe09Fo/maxresdefault.jpg
http://cdn1.airplane-pictures.net/images/uploaded-images/2010/1/25/76689.jpg
http://s23.postimg.org/u4ky33ygb/fgfs_screen_006.png
http://s2.postimg.org/pi120hoqh/fgfs_screen_004.png

some more images from outside shows that the panel raises above the height of the nose, so either our nose is too high or our panel is too low. See:

http://img.planespotters.net/photo/266000/original/PH-LFA-Private-Cessna-172_PlanespottersNet_266397.jpg
http://www.m0a.com/wordpress/wp-content/uploads/2009/01/cessna-172m-m0a.jpg
https://www.flickr.com/photos/14575229@N06/13820582295

### LINKED ISSUE #790: absolute position for instruments in individual .ac files

### LINKED ISSUE #1075: Control lock floating when yoke is hidden
Clicking on the control lock from the pocket on the left side of the pilot's seat when the yoke is hidden causes the lock to be floating in the air. The control lock should be hidden from the pocket when the yoke is hidden and vice-versa.

### LINKED ISSUE #1079: autopilot buttons are not animated when you click on them
The autopilot buttons are not animated when you click on them. They should quickly be toggled, just like the rest of the radio stack buttons.

### LINKED ISSUE #1080: Improve connection between wing and pillar
See picture here: 
_c203651 resized rotated

1. The connection between wing and pillar should more smooth. I think it is now missing something.

2. Add a L-shaped pipe, I assuming it is used for airflow intake to fuel tank.

3. Ring for tie-down should be a circle shape

_c203653 resized

Note: I took photos to a *C172R* at 2018 Zhengzhou Airshow, but it is very similar exterior.

### LINKED ISSUE #1081: Cowling should protrude around the edges
I apologize for bringing up an exterior issue, but I think it would be nice to make the cowling look like a piece of metal kind of above the rest of the fuselage. The cowling is probably about 3-4mm out from the rest of the fuselage as seen in the images in #1080. I am not explaining this well, but I'll make a diagram some time showing what I mean.

wlbragg:
Fixes #6 
Fixes #1075 
Fixes #790 
Fixes #1080
Fixes #1081
Fixes #1079 

:warning: :warning: :warning: work in progress :warning: :warning: :warning:


gilbertohasnofb:
@wlbragg I added a warning to your first post above

gilbertohasnofb:
@wlbragg let's move the communication to the PR over here as this makes things easier. So: the default view looks amazing now, just covering the thick fuselage by raising the panel a bit really makes it look absolutely stunning! :clap: 

If it would be possible, I would still tweak a bit the angle of the glareshield. IMO it should enter the fuselage at a lower angle, and this way it helps cover how thick the fuselage is on the bottom of the glass. See:

image

This might result in the glareshield needing some tweaking on the sides, but that shouldn't be too much trouble I suppose, right? What do you think?

The pedestal looks excellent and the current height is perfect. We are so nearly there :heart:

wlbragg:
OK, I had to edit the glare shield quit a bit to make if fit better. It's now a nice tight fit and effectively shaved the fuselage thickness down quite a bit.

gilbertohasnofb:
> OK, I had to edit the glare shield quit a bit to make if fit better. It's now a nice tight fit and effectively shaved the fuselage thickness down quite a bit.

I will be able to test this only in the evening, but I already want to say many, many thanks! The fuselage thickness is a problem everywhere (windows for instance), but the main problematic area was above the cockpit. If the panel now covers that then this solves the main problem we had. I have a great feeling about this PR! :smiley: 

I will report back in the evening.

wlbragg:
Here's a preview.
glare1

wlbragg:
>are we going to use this as a chance to change them to absolute positions?

Now is our chance to do this for sure. After working with it some, I have discovered some limitations. Anything that is rotated is best not to try to place the shadow effect on. But I think it should still look better to add shadow to what we can. So yes, lets reposition absolute ans as we go add the shadow effect to the .xml for the object.

wlbragg:
I'm thinking we add shadow effect to all non-movable objects and maybe material effects to those objects that move to soften their impact in the scene. We may have to experiment on the first few objects until we get it the way we want.

wlbragg:
Look at the mag-compass for an example. Note, non-movable objects get c172p-interior.eff
and movable objects get c172p-interior-radiance.eff.

tonghuix:
It is a chance to re-model some control lever as #1023 . And I think the mag-compass needs re-model too.

gilbertohasnofb:
> Now is our chance to do this for sure. After working with it some, I have discovered some limitations. Anything that is rotated is best not to try to place the shadow effect on. But I think it should still look better to add shadow to what we can. 

@wlbragg I agree. Also, I think that for the sake of consistency we should move all instruments to absolute position, even the ones in which will pose problems with the shadow effect. Also, if in the future the ALS framework manages to handle those, it will save us a lot of pain of having to move half of the instruments again.

> It is a chance to re-model some control lever as #1023 . And I think the mag-compass needs re-model too.

@tonghuix We are aware of a bunch of issues related to the cockpit and we plan to tackle those, but first things first. We need to first have everything in the correct place before we start changing the 3D model or fixing other things, ok? But no worries, those will eventually get done

dany93:
@wlbragg @gilbertohasnofb 
Just a few words which bring nothing (sorry for the noise) but encouragement.
I understand nothing on why you reposition every instrument and all in absolute position (shadows in ALS?), but just know that I am admiring when I read the amount of work that you are doing.
 :sweat_smile:

gilbertohasnofb:
@dany93 the merit of 99% of the work here goes to @wlbragg, I've been pretty much useless in this PR so far (busy as hell in RL and useless in Blender)

wlbragg:
Thanks @dany93! Yes it is solely for the ALS shadows to apply to everything possible. The pain to gain is almost a wash because we would have to reposition them in the xml anyway because of the panel movement. The additional work is in re-acquiring all the rotation points which is relatively easy as everything is already in place.
Still not a perfect visual, but I hope it an improvement.

@gilbertohasnofb don't sell yourself short. I think we both know that what is done so far is likely only the beginning. After all, you do have a gift for seeing what needs to be done and affecting that change. :smile:

legoboyvdlp:
Many thanks @wlbragg and @gilbertohasnofb for all this work :D

gilbertohasnofb:
@wlbragg :clap: :clap: :clap:

PR #1067, opened 2018-04-22
mooring-system-update

wlbragg:
@onox this fixes the inconsistent spawning behavior. You may have thought that was fixed but it wasn't. I had a speed limit condition on the "mooring-allowed" flag that would allow auto spawning on startup but it was killing the spawning from the "Go to airport" GUI reposition system. Anyway that is fixed. I also figured out a way to limit the "Go now" button's availability.

Please test and merge when ready.

### DIFF
--- a/Nasal/mooring.nas
+++ b/Nasal/mooring.nas
@@ -20,9 +20,25 @@
 #init if allowed and called for
 setlistener("/sim/signals/fdm-initialized",
     func {
-        if (getprop("/controls/mooring/automatic") and getprop("/controls/mooring/allowed")) {
-            seaplane = Mooring.new();
+
+        var presets = props.globals.getNode("/sim/presets");
+        var seaplanes = props.globals.getNode("/systems/mooring/route").getChildren("seaplane");
+        var harbour = "";
+        var airport = presets.getChild("airport-id").getValue();
+        setprop("/controls/mooring/port-available", 0);
+        if(airport != nil and airport != "") {
+            for(var i=0; i<size(seaplanes); i=i+1) {
+                harbour = seaplanes[ i ].getChild("airport-id").getValue();
+                if(harbour == airport) {
+                    setprop("/controls/mooring/port-available", 1);
+                }
+            }
         }
+        settimer(func{
+            if (getprop("/controls/mooring/automatic") and getprop("/controls/mooring/allowed")) {
+                seaplane = Mooring.new();
+            }
+        },1.0);
     }
 );
 setlistener("/controls/mooring/go-to-mooring",
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -113,15 +113,6 @@
                 <binding>
                     <command>dialog-apply</command>
                 </binding>
-                <binding>
-                    <command>nasal</command>
-                    <script>
-                        c172p.repair_damage();
-                        electrical.reset_battery_and_circuit_breakers();
-                        c172p.click("engine-repair", 6.0);
-                        c172p.dialog_battery_reload();
-                    </script>
-                </binding>
             </checkbox>
 
             <button>
@@ -131,7 +122,6 @@
                 <pref-height>28</pref-height>
                 <enable>
                     <and>
-                        <property>/fdm/jsbsim/settings/damage</property>
                         <not>
                             <property>/sim/freeze/replay-state</property>
                         </not>
@@ -831,19 +821,10 @@
                 <pref-height>28</pref-height>
                 <enable>
                     <and>
+                        <property>/controls/mooring/allowed</property>
                         <not>
                             <property>/sim/freeze/replay-state</property>
                         </not>
-                        <or>
-                            <equals>
-                                <property>/fdm/jsbsim/bushkit</property>
-                                <value>3</value>
-                            </equals>
-                            <equals>
-                                <property>/fdm/jsbsim/bushkit</property>
-                                <value>4</value>
-                            </equals>
-                        </or>
                     </and>
                 </enable>
                 <binding>
--- a/Systems/bushkit.xml
+++ b/Systems/bushkit.xml
@@ -355,7 +355,8 @@ Extra weight and drag due to bush wheels, floats and aircraft with 180 hp engine
             <test logic="AND" value="1">
                 bushkit GE 3
                 bushkit LE 4
-                /velocities/groundspeed-kt LE 1
+                /controls/mooring/port-available EQ 1
+                hydro/active-norm EQ 0
             </test>
             <output>/controls/mooring/allowed</output>
         </switch>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -570,6 +570,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <automatic type="bool">false</automatic>
             <anchor type="bool">false</anchor>
             <go-to-mooring type="bool">false</go-to-mooring>
+            <port-available type="bool">false</port-available>
         </mooring>
     </controls>
 


gui/dialogs/aircraft-dialog.xml
```diff
@@ -113,15 +113,6 @@
                 <binding>
                     <command>dialog-apply</command>
                 </binding>
-                <binding>
```
onox: Why did you remove this?

Nasal/mooring.nas
```diff
@@ -20,9 +20,25 @@
...
+        setprop("/controls/mooring/port-available", 0);
+        if(airport != nil and airport != "") {
+            for(var i=0; i<size(seaplanes); i=i+1) {
+                harbour = seaplanes[ i ].getChild("airport-id").getValue();
+                if(harbour == airport) {
+                    setprop("/controls/mooring/port-available", 1);
+                }
+            }
         }
+        settimer(func{
```
onox: What's the purpose of waiting a second here?

wlbragg: The new logic is used to set the "/controls/mooring/port-available" property. That property is used as a condition for the availability of the "Go now" button and also used in Systems/bushkit.xml mooring-allowed switch.
I was unable to leverage the second block of logic, which basically does the same thing, as it was only called when either the user ("Go now" button) or the system (on-reload) was actually invoked. I needed it available at all times for the two previously mentioned uses.

gilbertohasnofb:
Just saw the GUI for a first time in another branch (the panel raise branch). The question is: do we really need these info when using a non-float variant? I also find a bit weird that these are displayed in the aircraft options menu, as these are not options but rather variables. Finally, do we need to inform the user about those at all? Does a pilot needs these in RL and if so how they are measured? 

fg

wlbragg:
>The question is: do we really need these info when using a non-float variant?

I agree. What do you suggest? Hide it unless a float variant is selected?

>do we need to inform the user about those at all? Does a pilot needs these in RL and if so how they are measured?

Yes and no. I real life some of this will be known by pre-planning. In real life you have all your senses and visual clues as to your attitude, speed, direction, water flow, etc: not so much in a sim. So yes, in the sim especially, this is information that is required to successfully land on the water and set an anchor. However how and when we display it is another matter. If your using a float variant it needs to be able to be displayed on demand.
There are also physical conditions that have been implemented before your allowed to anchor such as aircraft heading in relation to wind direction, forward speed (a reading there is no other way to estimate without a display).

gui/dialogs/aircraft-dialog.xml
onox: That's for the repair button, this is for "Damage" checkbox. If you take a look at the changes here on Github, you'll see red lines; code that has been removed. Github doesn't make things up :wink:

gilbertohasnofb:
> I agree. What do you suggest? Hide it unless a float variant is selected?

That's an option. Another option would be to add a new dialog to the c172p menu instead of using the aircraft options dialog. And this dialog would only be available when a float variant is selected. Either way is fine with me.

> Yes and no. I real life some of this will be known by pre-planning.

Yes, I thought that would be the case. Makes sense that the user would need this info, all good by me.

wlbragg:
@gilbertohasnofb @onox I'm really not sure what to do here with the mooring menu and display data. How to break it up and where and how to visualize it.

Should we keep the "Spawn at seaport if available" and "Go now" button where they are?
Move the "Set mooring anchor" and display data to it's own pop-up dialog?
Where should the choice to pop-up the new dialog be located, as a line item in the "Cessna C172P" drop down menu list, grayed out if not in a float variant?
If so where in the list order?
If we want to get fancier and hide the choice of the float pop-up and data unless a float variant is selected, anyone know of some example dynamic menu work?

gilbertohasnofb:
What if we keep the "Spawn at seaport if available" and "Go now" button where they are and just create a new dialog called Mooring Parameters with only those four variable values. This dialog will then be greyed out if not on a float or if on air. What do you say?

> If we want to get fancier and hide the choice of the float pop-up and data unless a float variant is selected, anyone know of some example dynamic menu work?

That would be cool but I am only familiar with the traditional GUI. @legoboyvdlp would you happen to know any aircraft with that feature?

wlbragg:
>create a new dialog called Mooring Parameters with only those four variable values. This dialog will then be greyed out if not on a float or if on air. What do you say?

Yes, sounds good.
Where exactly do I add the "Mooring Parameters" menu choice (which existing menu and position in that menu?). Also you left out the "Set mooring anchor" check box which currently is grayed out unless in a float variant. Where is it to be, leave it where it is now? It needs to be fairly prominent so users know of its availability. I suppose it could either go with "mooring Parameters" menu or stay where it is at.
Maybe the "Mooring Parameters" menu choice should go where the "Set mooring anchor" is now (grayed out unless in float variant) and move/add the "Set mooring anchor" check box to the new "Mooring Parameters" pop-up?

wlbragg:
>if we want to get fancier and hide the choice of the float pop-up

I actually know how to do it when creating an entire menu dynamically. I'm just not sure you can do it without recreating the entire menu dynamically, not just the line items you want to add. We can pass on this for now. Sounds like it may be more trouble than it's worth.

gilbertohasnofb:
> Where exactly do I add the "Mooring Parameters" menu choice (which existing menu and position in that menu?)

Possibly just after Panel in a browser but before the About this aircraft section. Also, what about moving the Autostart to the first position, above the Aircraft Options? 

> I suppose it could either go with "mooring Parameters" menu or stay where it is at.

Your call. Both make sense to me, pick one that you prefer.

> Sounds like it may be more trouble than it's worth.

Yes, let's just stick to the traditional GUI then.

gilbertohasnofb:
@wlbragg just checked your last commit, I like very much how things are looking now.

gilbertohasnofb:
From my side it's all fine, let's just see if @onox has any comments

onox:
There's still some Nasal code for the "Damage" checkbox that has been removed...

gui/dialogs/aircraft-dialog.xml
wlbragg: @onox, I really can't remember why I may have removed that code in this PR. However, after testing, this is the correct behavior. I know we were not happy with the way the "Damage" check box worked. Because, when you were flying along and you unchecked the check box, the engine and everything would be reset. We didn't like that. If the aircraft is not damaged and you decide to un-check the box and not allow damage to occur anymore, nothing should happen in the sim. You should keep on flying and simply not allow damage anymore. But we also wanted the un-checking of the "Damage" check box to repair the aircraft it it was damaged at the time you un-check it so as not to leave the aircraft in a damaged state with no option to repair because you turned off "Damage". I probably figured out this needed to be removed to get everything working correctly (which it does) and decided to do it in this PR instead of creating another PR. I must have forgotten to note it in the commit message and for that I apologize. The damage code and logic in the menu and this PR is correct. SO GET OFF MY BACK! :rofl: Just kidding!
Again I am really sorry for how disjointed this looks and that I don't have a precise reason for doing it here. I'm sure there was a reason I did it here and at this time. It's just been too long since I worked on it to remember exactly why here and not in a more appropriate PR. More than likely it was because I had already closed out the appropriate PR and we figured out this still needed to be done. @gilbertohasnofb  you were involved in the discussions that lead to the final outcome of the damage system menu logic, maybe you can test this to confirm that it is exactly the behavior we wanted. I just tested it and works as I described above which is what I remember we wanted. This change also went hand in hand with https://github.com/c172p-team/c172p-detailed/pull/1053

wlbragg:
@onox please see my note above. Bottom line is that code needed to be removed and the damage logic is correct.

onox:
@wlbragg I read your explanation above. I hope you understand that changing something unrelated to the mooring system is confusing. Also I read "My copy shows that binding starting at 137" like as if you thought that the removed binding and the binding at line 137 were somehow the same, even though they're 2 different bindings (although effectively doing the same thing).

Anyway, I still found a bug:

1. Enable the "Damage" checkbox
2. Start the engine
3. Ctrl + u
4. Crash the plane
5. Disable the "Damage" checkbox (then the wings and the gear gets repaired, but not the propeller)
6. Try to do autostart

Autostart will then fail because the propeller is still damaged.

wlbragg:
>I hope you understand that changing something unrelated to the mooring system is confusing

Absolutely, thus my apologies above.

>Also I read "My copy shows that binding starting at 137" like as if you thought that the removed binding and the binding at line 137 were somehow the same, even though they're 2 different bindings (although effectively doing the same thing).

Yes, after looking through the code more thoroughly I realized they were to different bindings and the first was indeed missing. Which refreshed my memory as to why it was removed.

>Anyway, I still found a bug:

Interesting, I did exactly as you did and it works for me.
I did however do your exact test prior to typing my explanation, took it one step farther and made sure the aircraft landed upside down and broken. In that condition it doesn't right the aircraft. But we allow the repair button to always operate regardless of the "Damage" flag, which does now still correct everything and rights the aircraft.
I'm not sure why your prop wasn't repaired in your test?

Further testing revealed the bug, it is inconsistent. Investigating...

wlbragg:
@onox OK, done. I think this is what we want. It now also rights the aircraft if upside down but does nothing if there is no damage.

I again apologize for doing this here without explanation. I am pretty sure I must have stumbled on to what needed to be done while working on the mooring options, specifically when working on the caching of the damage condition. No excuse. :sob:

wlbragg:
@onox some of the issues here are the fragmentation or divergence of two damage systems, the physical and engine damage. They are intertwined and also separate. The original "Damage" and "Repair" options were intended and setup for the "physical" damage to the aircraft. Since it's inception they were leveraged to be used for the engine damage and repair. After looking at the flow of all the systems i see there is room for improvement in the engine damage logic. We managed to move almost all the physical damage logic out of physics.nas and into JSBSim channel logic, see c172p-damage.xml. Yet a substantial amount of the engine damage logic is still done in physics.nas using loops and listeners. There is also a control parameter driven in a logic gate in engine.xml. That was the actual property that wasn't getting reset on the un-checking of the "Damage" check box.

PR #1071, opened 2018-04-29
FDM: decrease one-wing-damage factor

### LINKED ISSUE #1070: FDM: decrease one-wing-damage factor
The "one-wing-damage" factor had been set (at 0.12) such as the aircraft could still be controlled with only the ailerons (without rudder correction).
Since the recent adverse yaw enhancement, this control is very difficult, almost above the full aileron efficiency.
The aircraft can be controlled without problem by using the ailerons + rudder, but it is more tricky and (IMO) the light damage on a single wing is surprising enough when it occurs without adding this difficulty. Unless you think differently?

This PR sets the one-wing-damage at 0.085.

If you wish to easily test the effect, you can manually set the damage factor in 
`/fdm/jsbsim/wing-damage/left-wing`
`/fdm/jsbsim/wing-damage/right-wing`

Please @onox; if you can check that I've forgotten nothing, thank you.  I've not done the corresponding files myself.

dany93:
Fixes #1070

### DIFF
--- a/Systems/c172p-damage.xml
+++ b/Systems/c172p-damage.xml
@@ -284,17 +284,17 @@
             <!-- ================================================================== -->
 
             <!-- Overspeed -->
-            <test logic="AND" value="0.12">
+            <test logic="AND" value="0.085">
                 /velocities/airspeed-kt GT damage/limits-vne-low
                 damage/roll-moment GT 8000.0
-                wing-damage/left-wing LT 0.12
+                wing-damage/left-wing LT 0.085
             </test>
 
             <!-- Over-g -->
-            <test logic="AND" value="0.12">
+            <test logic="AND" value="0.085">
                 damage/force-aero-lbs GT damage/limits-lift-low
                 damage/roll-moment GT 8000.0
-                wing-damage/left-wing LT 0.12
+                wing-damage/left-wing LT 0.085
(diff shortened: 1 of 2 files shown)


onox:
@dany93 Are these floating point numbers used in the FDM somehow? I've always wondered why we didn't just use integers (0 = ok, 1 = broken, 2 = damage light, 3 = damage heavy, etc.). Seems less error prone.

onox:
Okay, I can answer my own question: yes.

onox:
Have you checked the code base for "0.12"? I don't have access to a computer with FG on it right now.

dany93:
> Have you checked the code base for "0.12"?

Sorry, I don't understand...

dany93:
In this branch, I find no occurrence of "0.12" in the c172p directory.
However, I don't understand the connection with:
>  I checked Systems/damage.xml and that file should still work.

dany93:
I think I understand now for the Systems/damage.xml file. I had some difficulty to understand the logic.
It is for the animation, correct? The wings appear damaged (any value, same position) or broken, correctly.

onox:
> It is for the animation, correct?

Yes.

dany93:
Thank you, @onox .

PR #1078, opened 2018-05-01
FDM: pitch moment due to flap deflection

### LINKED ISSUE #1073: FDM: Enhance pitch behavior when flap extended
Quoted from 'Airplane Flying Handbook' by FAA:

> Flap extension has a definite effect on the airplane’s pitch behavior. The increased camber from flap deflection produces lift primarily on the rear portion of the wing. This produces a nose-down pitching moment; however, the change in tail loads from the downwash deflected by the flaps over the horizontal tail has a significant influence on the pitching moment. Consequently, pitch behavior depends on the design features of the particular airplane.

Because of this, in C172 when flap just extended, the plane will nose up immediately, but will nose down soon. Pilot should push nose down and up to counteract flap effect synchronously maintaining in normal approach path.

dany93:
Fixes #1073

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -1624,6 +1624,7 @@
                             30.0000    -0.1140
                         </tableData>
                     </table>
+                    <value>0.7</value> <!-- tests, adjustments -->
                 </product>
             </function>
         </axis>


dany93:
You seem to ignore the principles of mechanics. The velocity with which the nose moves is the result of moments, timing for the flaps, inertia and damping. The cause is a moment. The rest is the aircraft, whose behavior is correct by elsewhere. I didn't do the timing for the flaps, but I think it's correct. The amplitude is there. A faster pitching up could be the result of a higher moment, thus more amplitude. I set it for the amplitude. Unless they push the yoke to limit the pitching up in the video, I don't know.

Maybe I did the test at 80 kt, I don't remember.

If you want, it's simplistic to make tests. I've written the `<value>0.7</value>` factor on purpose for this.
```
            <function name="aero/coefficient/Cmdf">
                <description>Delta_pitching_moment_due_to_flap_deflection</description>
                <product>
                    <property>metrics/cbarw-ft</property>
                    <table>
                        <independentVar>fcs/flap-pos-deg</independentVar>
                        <tableData>
                            0.0000	    0.0000
                            10.0000	   -0.0654
                            20.0000	   -0.0981
                            30.0000    -0.1140
                        </tableData>
                    </table>
                    <value>0.7</value> <!-- tests, adjustments -->
                </product>
            </function>
```

dany93:
With what settings in the FDM? (`<value>0.7` ? `<time> ` for 10 deg?)

tonghuix:
No any additional edit after the commit. I feel it already perfect.

dany93:
Which means that we can ask for this PR to be merged.
If no objections, with the extra factor
 `<value>0.7</value> <!-- tests, adjustments -->`
I prefer letting it because I have the impression that this setting is sensitive to others in pitch. It makes changes easier without redoing the table each time, and by keeping the values from Megginson.

dany93:
Thank you, @onox .

Issue #1089, opened 2018-05-08
main wheels struts should be moved downwards


tonghuix:
I try to lower them, but it cause the wheel sink into earth and wheels rolling axis not changed.

So I think it should be more work to do to fix this major issue.

Ref picture:

![4681705](https://user-images.githubusercontent.com/423703/39759339-063ff848-5305-11e8-9b1e-1d78e5acae61.jpg)

![4763925](https://user-images.githubusercontent.com/423703/39759340-064232e8-5305-11e8-940a-e0b4c997e848.jpg)

wlbragg:
@tonghuix what do you mean by two high? Are the wheels themselves to tall? Is the airframe to far off the ground? You need to be more specific as to what you mean. The two images are not enough data to illustrate the issue.
Moving gear touches many areas of code and in no easy task, It requires a though understanding of the systems it involves and how it is all tied together.

tonghuix:
When I trying remodeling the connection between wheels struts and fuselage, I found these connection located too high. It should down at the corner of fuselage, due to this also cause the whole wheel too high.

I tried, so I do understand it relates many code, and needs much more work to do, so I mark this issue is Major problem...

wlbragg:
@tonghuixa a comparison picture is worth a thousand words. I understand now.
![gear](https://user-images.githubusercontent.com/11069680/39770396-87506770-52b4-11e8-9149-53f086254293.jpg)

We need to discuss this.
What happens to the struts if I simply reattach the top of the struts to the correct location, That is a trick that would require now wheel adjustments whatsoever, only a mesh change. But then do they become to short? Does the fuselage actually need to be raised?

wlbragg:
Comparing the photo with our aircraft it looks like our fuselage needs to be raised. I also noticed the tail is to tall as in thick.

gilbertohasnofb:
Ok, so since we are busy with the interior model at the moment let's leave this opened for the time being, ok? This one will need a lot of testing as it will affect a lot of things.

@tonghuix please do not report things as 'major issues', a neutral report with a picture is a thousand times more useful to us. Thanks!

wlbragg:
I wanted to show you a quick test. this compares the struts as they are now compared to a slight adjustment that really only required some rotation of the top of the gear stuts. What do you all think?
![fgfs-screen-001](https://user-images.githubusercontent.com/11069680/39784013-99e737e2-52dc-11e8-8cb6-6fb894e6835e.png)
[...]

PR #1092, opened 2018-05-08
Lower gear strut to fuselage connection point 

wlbragg:
Fixes issue #1089 

Still needs texture work,


gilbertohasnofb:
@wlbragg I will work on this now, should be done still tonight.

gilbertohasnofb:
Did you check if this change affects anything else since the WOW is moved downwards?

wlbragg:
>since the WOW is moved downwards?

That's the beauty behind this, nothing changed except the strut connection point. No other adjustments are needed. The wheels, brakes everything stayed the same. All I did was rotate the strut assemblies from the "axle" point to effectively lower the connecting point.

wlbragg:
Although, now that you mentioned it. Technically the 3d shadow for all the variants should be remade as the model has been changed (not Just this last PR but the others as well). I doubt anyone will notice any discrepancies however.

gilbertohasnofb:
I see! That's brilliant :smile: 

> Although, now that you mentioned it. Technically the 3d shadow for all the variants should be remade as the model has been changed (not Just this last PR but the others as well). I doubt anyone will notice any discrepancies however.

We maybe need to adjust a lot of things with the 3D shadow since we did so many modifications, but to be honest I don't think that's noticeable at all. We might as well just leave it as it is.

wlbragg:
I just noticed a couple other issues created by all the recent changes. Here are the proposals...
skis
float

gilbertohasnofb:
@wlbragg I just pushed the fixed liveries. Concerning the issues above, I am not sure I understand the diagrams very well. But I am all for using the new connectors in the pontoon and amphibian variants. And what about the ski one, would that be applicable as well?

I am probably signing off for tonight but tomorrow we can continue with this work. If you get these things done and give me a green light I will then review and merge this tomorrow during one of my breaks.

wlbragg:
@gilbertohasnofb  I figured this would only make sense to me. Basically I am going to move the float struts to the new positions and add the new connecting caps on them as well. The ski axle and bracing need reworked and simplified now that we actually have an axle on the left and right gears. I plan on doing this tonight so it will be ready for you to merge tomorrow.

tonghuix:
The main wheels are still a little bit high, I suppose move down 0.0456 alongside Z axis is better.



Please compare the distance between wheel strut's mount point on fuselage and edge of door. Should be downward only the height of a row of rivets,

fgfs-screen-502

wlbragg:
>The main wheels are still a little bit high,

@tonghuix that image you posted is one before the struts were moved. This is how they look now. It is going to have to be good enough.

stu1
stu2

wlbragg:
@gilbertohasnofb  I am finished here. We can move onto the the interior again.

tonghuix:
I test it now perfect!

gilbertohasnofb:
@wlbragg I noticed that lately we are not deleting these branchs that we used to work on the 3D model. Can I delete all of them? From my side it's a yes, as I think we don't plan to make more stuff on these branches and future work will be done in new branches.

wlbragg:
Yes, I noticed that same thing yesterday. I see you deleted them and that is fine with with me. I just now deleted the last one.

wlbragg:
Looks just about right. Thanks @tonghuix for revisiting this issue. It is so much better than before. I'm happy with it now.
For comparison..,
final

gilbertohasnofb:
Looks so much better indeed!

legoboyvdlp:
I apologize for bringing it up here, but I do not want to start another issue for something that is perhaps not an issue at all - if it is I will create a new issue.

image

Is it just me or the camera angle, or is our propeller a bit too thin in blade width near the spinner?

gilbertohasnofb:
@legoboyvdlp we do have some small issues with the nose, even though it looks much better than before. Please do open a new issue for this propeller thing and we can consider it in the future. Right now I really want to focus on the interior things, as @wlbragg and I agreed. Make a new issue so this doesn't get lost.

PR #1094, opened 2018-05-09
Issue 1076: Improve DME clickspots

### LINKED ISSUE #1076: Improve DME clickspots
Improve DME clickspots. See image below:

image

The DME ON/OFF should be mapped to the toggle itself (green arrow) and not to the ON/OFF labels (green rectangles). Also, the DME mode is already mapped to the button in the display (red arrow) so the clickspots at the KTS and MIN labels (red square) should be removed. The idea is that the radio stack should behave as a real life one so clickspots should always be mapped to toggles and switches whenever possible, never to labels.

ltoenning:
Fixes #1076

- Removed the clickspots on the on/off labels of the DME-power-switch and replaced it by the real switch.
- Removed clickspots on KTS/MIN labels. The button for that is already correctly implemented.

### DIFF
--- a/Models/Interior/Panel/Instruments/dme/dme.xml
+++ b/Models/Interior/Panel/Instruments/dme/dme.xml
@@ -54,45 +54,7 @@
 
     <animation>
         <type>pick</type>
-        <object-name>on</object-name>
-        <visible>true</visible>
-        <action>
-            <button>0</button>
-            <repeatable>false</repeatable>
-            <binding>
-                <command>property-assign</command>
-                <property alias="../../../../params/power-btn"/>
-                <value>1</value>
-            </binding>
-            <binding>
-                <command>nasal</command>
-                <script>c172p.click("dme-power")</script>
-            </binding>
-        </action>
-    </animation>
-
-    <animation>
-        <type>pick</type>
-        <object-name>off</object-name>
-        <visible>true</visible>
-        <action>
-            <button>0</button>
-            <repeatable>false</repeatable>
-            <binding>
-                <command>property-assign</command>
-                <property alias="../../../../params/power-btn"/>
-                <value>0</value>
-            </binding>
-            <binding>
-                <command>nasal</command>
-                <script>c172p.click("dme-power")</script>
-            </binding>
-        </action>
-    </animation>
-
-    <animation>
-        <type>pick</type>
-        <object-name>PowerSwitch</object-name>
+        <object-name>PowerSwitch.Lever</object-name>
         <visible>true</visible>
         <action>
             <button>0</button>
--- a/Models/Interior/Panel/Instruments/dme/dme.xml
+++ b/Models/Interior/Panel/Instruments/dme/dme.xml
@@ -20,8 +20,6 @@
 
     <effect>
         <inherits-from>../../../../Effects/interior/c172p-interior</inherits-from>
-        <object-name>on</object-name>
-        <object-name>off</object-name>
         <object-name>Label3</object-name>
         <object-name>n1</object-name>
         <object-name>hld</object-name>
@@ -54,7 +52,7 @@
 
     <animation>
         <type>pick</type>
-        <object-name>PowerSwitch.Lever</object-name>
+        <object-name>PowerSwitch</object-name>
         <visible>true</visible>
         <action>
             <button>0</button>
--- a/Models/Interior/Panel/Instruments/dme/ki266.xml
+++ b/Models/Interior/Panel/Instruments/dme/ki266.xml
@@ -144,8 +144,6 @@
     <animation>
         <type>pick</type>
         <object-name>ModeSwitch</object-name>
-        <object-name>Min</object-name>
-        <object-name>Kts</object-name>
         <visible>true</visible>
         <action>
             <button>0</button>


wlbragg:
@ltoenning 
On the other half of this instrument "dme.xml" there is no knob animation for the toggle switch only clickspotsl. There is a knob animation for the SourceSelector and also two unneeded clickspots for it called SourceSelectorPick.L and SourceSelectorPick.R. which can be removed. 
If you want to fix the other half of this instrument and add an animation for the On, DME, OFF toggle switch we can also remove those pick as well as the SourceSelectorPick.L and SourceSelectorPick.R which are already accounted for.

ltoenning:
Do you mean this toggle switch:
unbenannt

If I'm not interpreting it wrong there is a "rotate" animation in the dme.xml from line 39 to 53. When I'm clicking on the switch (and I should have removed the OFF/ON-label clickspots; at least the aren't working anymore after my change... ) it is actually moving correctly up or down.

I will take a look into the SourceSelector later today.

wlbragg:
>If I'm not interpreting it wrong there is a "rotate" animation in the dme.xml from line 39 to 53. 

Yes, but that is only the rotation animation that animates the toggle when the clickspot is pressed, controlled by the
~~~
<property alias="../../params/power-btn"/>
~~~
property.

We can remove the on and off clickspots in the code and their on/off picks in the blend/ac and only use the PowerSwitch.Lever as the pick. It already has a pick code block in the dme.xml but it is named incorrectly. It is named PowerSwitch but should be PowerSwitch.Lever;

wlbragg:
I misspoke, the On/Off labels in the Blend/ac need to remain as they are the labels. They were also being used as a pick in the code wihixh is not necessary if the toggle switch was working.

wlbragg:
I think some of this malfunction is due to the exporter being used is not supporting the .group function. Both the toggle switch and its base  were in a group called PowerSwitch. So I think both PowerSwitch.Lever and PowerSwitch.Base were acting as PowerSwitch in that particular pick code block. I don't like the idea of having to use an "older" .AC importer to export our .ac files. Over time new functionality is added to the current AC exporter and it would be of no benefit if we have to use an antiquated .ac exporter. I am curious though as to why this functionality was removed from the import/export script in the first place?

wlbragg:
No problem. However after some consultation what needs to happen is in the Blend/ac the group name should be changed to PowerSwitch then code block that was using PowerSwitch for the object name will work for both PowerSwitch.Lever and PowerSwitch.Base which is the way it should be. That way if you click on either the toggle or its base it will work. It just gives you a slightly larger area to click on.
I'll take care of changing the Blend file on another branch/PR that we are about to set up if you want to change code in dme.xml to repfect this.
The bottom line would be for you to...
1) remove on/off picks
2) continue to use PowerSwitch for the object name ion the PowerSwitch pick animation

I'll do the rest in the Blend/ac

ltoenning:
I hope it is now resolved.

wlbragg:
@ltoenning nope, it seams I faild to communicate this effectively.
Let me ask you, can you work in Blender, something simple like deleting an object?
Also the on/off are labels and still need to be in the effect definition only the on/off and "pick" animation are to be removed. The renaming of the "pick" animation PowerSwitch.Lever to PowerLever is correct.

gilbertohasnofb:
> Let me ask you, can you work in Blender, something simple like deleting an object?

@wlbragg I think you are complicating things. You could as well just send him the correct Blender file and he commits, but then again that's a bit of an inelegant solution. I think what @onox wrote in the other thread sounds like the best approach. Or we can just simplify this for one single time and do the following: I will pull your branch then manually add these modifications from this PR there, test it, then if all is fine I will merge both of them. Would that be fine with you?

gilbertohasnofb:
If you give me green light to go with what I proposed last I can test and merge this still tonight.

wlbragg:
@gilbertohasnofb 
>I will pull your branch then manually add these modifications from this PR there, 

That is basically what I proposed in the other thread, that I just manually add his changes to the other PR, but as you stated, that removes the credit from going to him. Doesn't your solution do the same thing? If we're going to do that then I should add his work in the other PR and we can continue with that PR for the rest of the interior work, which would be my first choice. I can add credit to him in the commits I do for him.
Then, next time something like this comes up I will make sure I don't tack on mesh work to some one else's PR. I didn't think this would be as much trouble as it is turning out to be.

So, as far as I am concerned, the ideal solution is to add his work with credits to the other PR and continue with thar PR. Especially as this PR is along the same line, "interior work".

wlbragg:
@ltoenning is this OK with you?

ltoenning:
> Let me ask you, can you work in Blender, something simple like deleting an object?

I've some basic experience with Blender but I think the best would be, as written, that you will manual copy my edits into your PR and pull them.

gilbertohasnofb:
> Doesn't your solution do the same thing? 

Nope, I would add manually just to test then I would revert that. Then I would merge both PRs, crediting each of you for the commits each one did. But it's up to you, I am fine with whatever you decide!

onox:
@wlbragg Just make sure the e-mail address is the one he used in his commits so that Github will recognize it.

wlbragg:
@onox the email @ltoenning is using in the commit that I can see is a "number+username"@users.noreply.github.com. Is that going to be the correct email?

@gilbertohasnofb I'll make sure I do each commit separately.

wlbragg:
@ltoenning I added your work directly to #1095 Thank you for your efforts. Sorry for my making this harder than was needed. As you can see I am still learning.
You can simply close this PR as we have your changes now.
Thanks again.

PR #1102, opened 2018-05-12
kx165-2-regression

### LINKED ISSUE #1101: Regression issue: NAV2 standby & active frequency readout not displayed
Currently only displays 000.000 in the digital readouts.

wlbragg:
Fixes #1101 

Forgot to export kx165-2.ac in "Correct automatic Blender versioning error in kx165-2"
Thank you @mathieugouin for catching this.
This should be ready to merge after testing and verification.


gilbertohasnofb:
@wlbragg There is still a problem. When you scroll the large knob to change the stand by frequency, it flicks between 100 and 110. I suspect someone changed the value of the delta caused by a rotation of that knob to be 10 instead of 1.

onox:
The problem is that `kx165-2.ac` contains "NavStby1M." (notice the dot at the end)

gilbertohasnofb:
The radios are not fine but this branch breaks the flap texture for some reason.

fgfs-screen-003

wlbragg:
Geez! There is a mistake in the commit history. Either i need instruction on how to remove that 2nd commit called kill, or I need to cancel this PR and start over.

wlbragg:
@onox can I do a git reset hard at the prior commit, do my correct commit, then force push?

wlbragg:
@gilbertohasnofb  see, we don't need @onox anymore, he can retire now!
Pleas test and verify all is well!

PR #1107, opened 2018-05-13
Correct VOR needle center points

### LINKED ISSUE #1106: VOR Needle behavior change
Hi,

Pull request #1066 changed the behavior of the needles of the VORs.

Before the pull request was applied, the needles moved entirely to the left and to the right (with an angle), just like the picture below:

fgfs-screen-009-crop

After the pull request, the needles still work, but they only spin around the center of the localizer and glide slope axis, as below:

fgfs-screen-011-crop

I checkout some commits out to test which one introduced the new behavior, and it seems that commit ID 7c607459f8d17e29cfd2f88c609d00779a724ca6 did it.

Checking out the previous commit, i.e.: commit ID 79848618edc1faa4d6f5dfa306bf9b695c3b1e70, restores the behavior of the needle, but it moves the entire VOR display down, as shown by the picture below (I understand that this change is within the pull request and you were fixing the entire panel. :)):

fgfs-screen-010-crop

Is this a regression?
Thank you.

wlbragg:
Fixes #1106


gilbertohasnofb:
@wlbragg it's all good now, but why do you always have these `Merge branch` commits? I think because we are kind of in a rush to fix this I will just go ahead and merge this but I think if this is being caused by work left over in other branches then maybe you need to use `git status` more often. I religiously use it before any `git pull`

wlbragg:
>but why do you always have these Merge branch commits? 

In this case it was because I was in a hurry and I think your last commit on the panel_parts.png snuck in between my commits. I was desperate to get it in before you were done for the night so I didn't take the time to correct the history.

PR #1115, opened 2018-05-25
FDM: fix aileron trim preset, introduce spiral propwash tuning

### LINKED ISSUE #1112: P-factor may be too excessive
Hi there,
I noticed that XDraconian made a complaint about the new c172p FDM with the RC version,

>  C-172 has excessive P-Factor and/or Left-Turn Tendency.  My gear is packed up, so I'm flying with auto-coordinated mouse control.
> Mitigation: None.  I was able to takeoff, but for novice pilots this would be unusable.  Not sure if anyone else has experienced this problem - my configuration is non-standard so I don't want to rule out that this may be isolated to my build.

Well, personally, I think the propwash is a bit too strong, but that's just the feeling I get. I imagine this is because he is using auto-coordinated mouse control.

it0uchpods as well, told me that the p-factor was far too much, and the engine seems "weak". Well, I have a hard time sometimes to maintain speed / rate of climb at a suitable level if I am heavy and took off from a short runway, but I can't say I have noticed a weak engine. He does say the problems are "new", so it might have been introduced by my issue-124 - I trust not!

### LINKED ISSUE #1113: FDM: fix right turning tendency
Now you will feel the plane has a little bit RIGHT turning tendency, yes RIGHT direction not left! Very little left rudder could compensate it easily.

As #1112 mentioned.

### LINKED ISSUE #1114: Create a Welcome easy-FDM for beginners.
Follows https://github.com/c172p-team/c172p-detailed/issues/1112#issuecomment-392001045
and https://github.com/c172p-team/c172p-detailed/issues/1113#issue-325643076
particularly
https://github.com/c172p-team/c172p-detailed/issues/1112#issuecomment-391721130
https://github.com/c172p-team/c172p-detailed/issues/1112#issuecomment-391528826
https://github.com/c172p-team/c172p-detailed/issues/1112#issuecomment-391797365

dany93:
Fixes #1114
Fixes #1113 

See #1112

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -514,7 +514,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 
     <controls>
         <flight>
-            <aileron-trim type="double">0.044</aileron-trim>
+            <aileron-trim type="double">0.022</aileron-trim>
             <rudder-trim type="double">0.02</rudder-trim>
 
             <freeze-yoke type="bool">false</freeze-yoke>
@@ -1049,6 +1049,11 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
                 <damage type="bool">true</damage>
                 <damage-flag type="bool">false</damage-flag>
             </settings>
+            <aero>
+                <coefficient>
+                    <spiral-propwash-coeff type="double">1</spiral-propwash-coeff>
+                </coefficient>
+            </aero>
         </jsbsim>
     </fdm>
     <!-- End Damage Mod -->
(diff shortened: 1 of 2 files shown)


onox:
I've removed the #1112 from being automatically closed by Github, since the propwash hasn't actually been reduced by this PR (just added the ability to tune it).

dany93:
OK, thank you @onox

tonghuix:
After this commit was merged, I could say both joystick and keyboard+mouse flying are more easier. And it is more better than joystick that I use mouse flying.

PR #1118, opened 2018-05-28
setting propwash coefficient to 0.5

gilbertohasnofb:
Fixes https://github.com/c172p-team/c172p/issues/1112


gilbertohasnofb:
This PR contains @dany93's fix for the propwash by setting this coefficient to a different default value which feels better overall. Once this is merged and that issue is closed, I would suggest that we open a new issue asking for RL feedback on our FDM in particular with this effect, so that we can still tweak it to a more appropriate value in the future.

@dany93 Thank you so much for all the effort you put into this.

PR #1120, opened 2018-06-01
Fix issue 1119

### LINKED ISSUE #1119: Missing property /engines/active-engine/damage_allowed
image

I could not reproduce, but I added some code to engine.nas, which the reporting user tested and confirmed that it works.

I will open a branch and PR shortly.

legoboyvdlp:
Fixes #1119


legoboyvdlp:
Sorry... Just bumping this in case you didn't notice.

onox:
These merge commits do not belong in a branch for such a simple fix. Can you try to just set a default in the `c172p-set.xml` file instead of using Nasal?

legoboyvdlp:
I'm afraid I don't really know how to remove them...

This code is what worked for the reporter - but if you really think it is best I can certainly try. ;)

onox:
I would just delete the whole branch, update your master, create a new branch off of master and then modify `c172p-set.xml`.

Issue #1124, opened 2018-06-14
Add Digital Clock


legoboyvdlp:
![image](https://user-images.githubusercontent.com/10434911/41433445-eb537de2-7010-11e8-9bdc-2f3746d3ece7.png)

![image](https://user-images.githubusercontent.com/10434911/41433353-8eef1d18-7010-11e8-862c-71c9119476b4.png)

![image](https://user-images.githubusercontent.com/10434911/41433343-8677e7e6-7010-11e8-9df2-a9f2d0e77bed.png)

The POH of the C172P that I have has a description of a digital clock: how would you feel about replacing our analog clock with this one? It seems to be designed to fit precisely in the same "hole" as the analog clock.

![image](https://user-images.githubusercontent.com/10434911/41433498-0f95b2f6-7011-11e8-87aa-761fb9b71eaf.png)
![image](https://user-images.githubusercontent.com/10434911/41433506-165e0c14-7011-11e8-9084-c8000642b7f3.png)

![image](https://user-images.githubusercontent.com/10434911/41433517-1bd7e44e-7011-11e8-9b57-74da063aa815.png)

![image](https://user-images.githubusercontent.com/10434911/41433549-37f92354-7011-11e8-8dd0-f7523004265b.png)
![image](https://user-images.githubusercontent.com/10434911/41433559-3d5fe224-7011-11e8-8baf-622bbd2602c7.png)

wlbragg:
@legoboyvdlp thanks for the detail, we have this in the works somewhere, we're going to use canvas for it. I started on it but got distracted on other stuff. You might search the issues for a duplicate of this?

gilbertohasnofb:
> we have this in the works somewhere

Do we? I don't recall that. What we recently spoke about using Canvas for was regarding the thermometer, not digital clock.

wlbragg:
Oh, ya, your right. My brain is turning to mush. I confused the two.

>The POH of the C172P that I have has a description of a digital clock: how would you feel about replacing our analog clock with this one? It seems to be designed to fit precisely in the same "hole" as the analog clock.

As far as replacing it, I would be against that idea. But as an option, I would be totally for it!

gilbertohasnofb:
[...]

PR #1129, opened 2018-06-20
Issue 1124

legoboyvdlp:
Add digital clock: #1124


gilbertohasnofb:
@legoboyvdlp this commit history is a mess, lego! :scream:

legoboyvdlp:
I know - that is the problem with forks 😂

wlbragg:
@legoboyvdlp 
I have employed the use of "git stash" and "git apply" and there is also the one @onox showed us but I can never remember it so I mostly use "stash/apply". When you need to update your branch you "stash" your current work and then after you pull and merge you "apply". I also mostly use "Git GUI", there is a feature that lets you add those commands under "Tools".

wlbragg:
After reading @gilbertohasnofb comments in another thread, I am not positive this will help you.

algefaen:
You can also maintain a cleaner history by using `git rebase`, as long as you're not re-writing history that you've already shared with others. Hint: `git pull -r` does the same as pull, just rebasing instead of creating merge commits.

wlbragg:
git rebase that's the one, what so hard about that to remember? I always wondered if I had to use any other commands with it. So for example using "GIT GUI" if I were to make a "tool" that effectively did "git pull -r", even if I had committed to my local branch, I could pull changed and place my commits in order?
Sounds simple enough, thanks.

legoboyvdlp:
@algefaen would you have any idea how to rebase it now that I have pushed these merge commits?

Or - you could do this:
create a new branch in c172p-team/c172p: issue-1124
pull the commits from my branch

run 
`git checkout interior-updates
git cherry-pick 0bc70d8
git cherry-pick b0edb52`

This **should** bring only the two important commits from my branch
Please check with someone more experienced (onox? algefaen?) before doing this ;)

gilbertohasnofb:
@legoboyvdlp alternatively, given how simple these commits are, you can simply copy your files to a safe place (make sure you copy everything) and then delete your branch and close this PR. Then create a new branch based on whatever you want and start again.

legoboyvdlp:
Problem is, my interior-updates branch is the one where all these merge commits are. So, I think I want to clean up that branch first then redo this pull request?

Or if it is easier for someone with push access here to just do the two cherry picks from branch issue 1124 to interior-updates?

If so I can write the commands needed tonight.

gilbertohasnofb:
@legoboyvdlp one more thing, I noticed that you use the folder `clock` for both clocks. Don't do that, keep one model per folder please as all other models have individual folders. Perhaps `digital-clock` would be better.

wlbragg:
@legoboyvdlp 
"IN THE FUTURE", to prevent this (bold is to make sure you understand this for future work and doesn't apply to cleaning this up).

If you have local changes you haven't committed locally yet and no local commits, use...

git stash
git pull
git apply

If you have no local uncommitted changes but you have local commits that haven't been pushed, use...

git pull -r 

If you have some local uncommitted changes and some local committed changes both that have not been pushed, use...

git stash
git pull -r 
git apply

I just tested this and it appeared to work as I expected.

Barring no other feedback or solution, I think I would go with @gilbertohasnofb suggestion and...
save your work, 
delete your branch,
make a new one from interior-updates,
add your changes,
request a merge to interior-updates.

And it anything changes in interior-updates before you request the merge, use my suggested steps above to get those changes depending on the state of your branch at the time.

After you make the request I can merge it into interior updates and apply all the effects to it.

If you want to learn how to add those effects yourself, at least the xml portion, let me know and I can walk you through it. Adding the instrument to the lm.blend is not so simple, there are a few steps and then the baking is complicated because of all the different issues we have.

legoboyvdlp:
Ok - can do :)

However - the problem is, when I pull from here to another fork, it creates these merge commits. That's the source of all this mess in the commit history. I'm not sure how to avoid that. 

Anyway, I'll do this tomorrow. As for adding the effects, I can do the XML part, sure, but it would make merging more difficult as it would require you baking to avoid the various issues. That's why I suggested originally to push the files to here in a pull request, then let you bake the lightmaps. But regardless let me know which is easiest for you. I'm happy doing it if it's easier.

wlbragg:
>However - the problem is, when I pull from here to my fork, it creates these merge commits. That's the source of all this mess in the commit history. I'm not sure how to avoid that.

If I stash my uncommitted work before a pull then apply, I have no merge request.

If I git pull -r with local commits I don't get a merge request.

 I think that is the key unless your doing something I am not understanding. I am not asking for PR into the interior-upgrades branch, maybe that is the difference?
or if

onox:
@legoboyvdlp What you can do (in the future) is `git pull --rebase upstream some-branch`. Make sure you have a remote called `upstream` that has our repository as URL. After you have done the pull, you can push to `origin` (your fork on Github).

PR #1131, opened 2018-06-20
FDM: P-factor (propeller effect) change from 5 to 59 or 60

### LINKED ISSUE #1130: P-factor (propeller effect)
I'm proposing a P-factor value which (hopefully) will be closer to reality. The previous one was much too weak, almost unnoticeable.

The P-factor is a propeller effect which makes the aircraft steer or yaw leftwards with AoA at high RPM (or steer with any apparent-wind misorientation). It adds to others such as the spiraling propwash effect, although on tricycles the spiral propwash effect is predominant when rolling before takeoff, whereas the P-factor becomes more efficient at climbing (due to AoA).
[Asymmetric Thrust Explained](http://www.qmfc.org/school/asym.htm)

An other improvement (I think) is that this implementation enables me to decrease the spiraling propwash effect while having this well known leftward yawing at low speeds, high AoA (rolling for takeoff, climbing).
Then, the (previous) strong tendency to steer left at very start of rolling, full thrust, is considerably decreased. In particular, even the amphibious no longer needs differential braking to hold the aircraft in line.

Please @algefaen, I would be very pleased if you could test it with these proposed values.

If you wish to test with other settings or analyze by selectively canceling the effects, the key-values are (for 180 hp engine)
- The P factor in c172p.xml, `<engine> / <thruster>`
line 620
`                <p_factor> 60 </p_factor>`
(must be set before launching, not very convenient but still easy),

- The spiral propwash, which can be set in the simulator:
fdm/jsbsim/aero/coefficient/spiral-propwash-coeff
or in c172p-set.xml,
line 1055
`                    <spiral-propwash-coeff type="double">0.25</spiral-propwash-coeff>`

I've set these values in order to try being in (qualitative) accordance with [your report](https://github.com/c172p-team/c172p/issues/1114#issuecomment-397380057), to get some effect when climbing.
As they are,
- at takeoff, 20% right rudder is needed at very start of rolling, quickly decreasing to 10%,
- at climbing at 60kt, about 20 to 25% right rudder is needed to keep the slip-ball centered.

In fact, those are estimations and I'd like to know if they fit your subjective opinion, if you can.

If you are not sure at this day, don't worry, you can see that the values are very easy to change in the code if needed later. The harder is in-flight testing.
For now the first question is, is it an improvement? (even with not definitive values)

dany93:
Fixes #1130

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -1052,7 +1052,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             </settings>
             <aero>
                 <coefficient>
-                    <spiral-propwash-coeff type="double">0.5</spiral-propwash-coeff>
+                    <spiral-propwash-coeff type="double">0.25</spiral-propwash-coeff>
                 </coefficient>
             </aero>
         </jsbsim>
--- a/c172p.xml
+++ b/c172p.xml
@@ -13,7 +13,7 @@
         <!-- Source: http://forum.flightgear.org/viewtopic.php?f=25&t=21664&start=45 -->
         <!-- this file with comments for stall and spin to help -->
         <!-- further modifications for c172p-detailed 2015 - 2018 -->
-        <!-- with help and extra informations from okroth -->
+        <!-- with help and extra informations from okroth and algefaen -->
         <description> Cessna 172P Skyhawk </description>
     </fileheader>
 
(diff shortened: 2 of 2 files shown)


onox:
@dany93 Does this still need any work?

dany93:
@onox 
I was just waiting for @algefaen confirmation ([this message](https://github.com/c172p-team/c172p/issues/1130#issuecomment-404294369)).
As he wrote 
> I think this is pretty good now

On my part, it is OK for being merged.

Thank you for following it.

onox:
Just flew it without any problems with mouse and keyboard.

dany93:
Thank you, @onox and @algefaen .

PR #1135, opened 2018-06-30
Wrong mesh in vor2.blend

wlbragg:
@gilbertohasnofb I'm not sure how it happened, but I noticed one of the vor's (vor2) contained a KX165 in the blend file instead of the vor2 object. So I imported the vor2.ac and saved it as the new vor2.blend. I'd really like to see this merged ASAP.

Thanks


gilbertohasnofb:
@wlbragg checking it now

PR #1136, opened 2018-06-30
Panel glass effect

### LINKED ISSUE #13: glass effect over the gauges
3D11 - can put "glass effect" over the gauges. What do you think ?
- Gilberto: that would be great!!

### LINKED ISSUE #1139: Attitude indicator updates
i'm copying this from email since i can't find the discussion... i think it might have been part of a PR but i don't know and can't find it...

On 07/01/2018 12:32 PM, wkitty42@gmail.com wrote:
> On 07/01/2018 11:20 AM, Joshua Davidson wrote:
>> Yes, that was an example picture. I've never seen an attitude indicator matching the one we
>> have in FG anywhere on the internet (I've searched long and hard).
>
> i don't know where it came from but the attitude indicator we're using looks
> to be a 3D model... at least it has an .ac file... the ""problem"" seems to
> come from the texture file for it...

so, i've done this...

20180701-c172p-attitude-indicator-reworked-colors-01

20180701-c172p-attitude-indicator-reworked-colors-02

i can make a PR for this texture if folks want me to... or i can get it to @gilbertohasnofb  and let him clean it up a little bit? i used gimp and got pretty lucky doing the circle selection for the blue outer ring so i could flood fill it... there's still some anti-aliasing in it but this looks like a good start for an hour's work...

i have no idea about that "post light" or why it is located where it is... that's something for @wlbragg to look at... but first we should get this texture in place ;)

### LINKED ISSUE #1142: Tooltip for the dome light is incorrect
Issue slipped through:
image

The _WHITE DOME LIGHT_ and _WHITE DOME INTENSITY_ positions should be reversed 😉

wlbragg:
Fixes #13
Fixes #1139
Fixes #1142


wlbragg:
Somethings not right here. I am not getting the same effect I did before.

wlbragg:
OK, I needed to make some changes to the material settings for this type of glass. The reflection strength is really the only thing we can change in the effect itself. All the other adjustments are going to come from material settings.

wlbragg:
@onox, I was reading up on rebase VS a revert commit and I really don't know which would be best in this case, if either. I kind of blew off the fact that I had the "Wrong mesh in vor2.blend" commit in this branch and really don't think it should be here as it is going to be merged in #1135 PR.
I still don't have a grasp on the subtleness of this. If that same commit is left here and also merged in #1135 PR, what happens with it? Does it duplicate and become two identical commits or is GIT smart enough to know it is the same commit and thus no need to rebase this branch?
What should I do in this case?

gilbertohasnofb:
@wlbragg I just tested this in the sim and the effect is fantastic. I do have a couple of comments though:

* the glass on the AI is having some Z-fighting with the bottom bit. The glass should not cover anything below the yellow line:

image

* same for the EGT:

image

* would it be possible to try to tweak the effect or the reflection map so that the dark shadow places are a bit darker? See the image below, the top is my screenshot, the bottom is tweaked a bit in GIMP:

image

* about the AI, I noticed that the top part should be a white triangle and not a single line like we have. How can we improve this? Currently that part is a 3D model and not a simple texture.

image

wlbragg:
>would it be possible to try to tweak the effect or the reflection map so that the dark shadow places are a bit darker?

@gilbertohasnofb I think that is the function of the reflection strength. I turned it down to .5. Try turning it back up to 1 or .9 and see if that makes the change you want.
It is in Models/Effects/interior/interior-glass-reflection-panel-front.eff
It is also in Models/Effects/interior/interior-glass-reflection-panel-side.eff but that effect is used only by the OAT and it has issues because it is movable. It looks like the effect turns on and off depending on where the OAT is moved to. So I may have to remove the effect from that instrument.
Not to confuse the reflection strength with Fresnel or diffuse or specular lighting effects. Those are what is causing the cloudiness or "glass" effect. I have Fresnel and/or diffuse turned way down so the glass stays more transparent in all light. I have specular up around .8 but I am not sure its effect is pronounced in this case at the angle we usually are looking at the glass from.
Where this effect will shine the most is with that view feature that moves the pilot view point perspective when you turn or apply G force. then you will see the reflection moving in the glass and it should look incredible.

I'll correct the glass on the AI and EGT.
On the EGT, are we sure the glass is supposed to stop at that half panel? I'm only asking because that half panel in also recessed past where the glass would be attached (at least as it is designed now). So there shouldn't actually be any z-fighting right now, but yes, the glass would be of no real uses there. I'm wondering if that panel shouldn't be recessed but should be flush with the case as the AI horizontal part is?
The AI, absolutely, I agree it is supposed to stop where you indicated.

wlbragg:
@gilbertohasnofb I increase the reflection strength and pushed it. I think it did what you wanted without increasing having any effect on transparency.

wlbragg:
>Where this effect will shine the most is with that view feature that moves the pilot view point perspective when you turn or apply G force. then you will see the reflection moving in the glass and it should look incredible.

Aw, disappointing, that effect is changing the rotation and not the pilot position.That is really disappointing. I can see why it is done that way, to simulate the pilot turning their head and I know good technique in to not lean your head when turning, But there needs to be a slight bit of head position adjustment in that effect as far as I am concerned.

gilbertohasnofb:
> I think that is the function of the reflection strength. I turned it down to .5. Try turning it back up to 1 or .9 and see if that makes the change you want.
> [...]
> I increase the reflection strength and pushed it. I think it did what you wanted without increasing having any effect on transparency.

Cool, I might play around with those values soon, but the results right now are already so good, it's really subtle and very realistic!

> I'll correct the glass on the AI and EGT.

Thanks, they look better already!

> On the EGT, are we sure the glass is supposed to stop at that half panel?

Yes, see:



There is still the triangle thing in the AI which needs some remodelling and I also noticed in the image above that the yellow needle should be quite longer than the white one in the EGT.

> Aw, disappointing, that effect is changing the rotation and not the pilot position.That is really disappointing. I can see why it is done that way, to simulate the pilot turning their head and I know good technique in to not lean your head when turning, But there needs to be a slight bit of head position adjustment in that effect as far as I am concerned.

That is a pity indeed...

wlbragg:
@gilbertohasnofb the AI and EGT are both finished now.

gilbertohasnofb:
@wlbragg Looks amazing. What's is left to be done with this? I love the effect and it is subtle enough, IMO this is quite ready for a merge

gilbertohasnofb:
But I see you have a wrong commit here: 'Wrong mesh in vor2.blend'

gilbertohasnofb:
@wlbragg does the glass effect consider the angle of viewing in order to intensify the reflections? E.g. if you look directly perpendicularly to a gauge you should see little reflections, but if you look at a shallow angle you should see quite a lot. That could be cool to implement if doable, perhaps Thorsten would be interested in doing this for the shader behind the glass effect.

wlbragg:
@gilbertohasnofb sorry about the out of place commit. I got rid of it with a rebase and forced push, so beware your branch is now diverged.
This should be ready to merge.

gilbertohasnofb:
@wlbragg Great. Concerning my last comment about the angle, that's something that could only be done in the shader side, right?

wlbragg:
>Concerning my last comment about the angle, that's something that could only be done in the shader side, right?

Yes, although that should already be included in the effect.

I wonder if it is more pronounced because I have Fresnel, diffuse material setting and transparency set so low to imitate more of a anti-glare type of material? If those values were higher then the extreme angle would probably whiten more and maybe hide that reflection some?

wlbragg:
@gilbertohasnofb I am treading on dangerous ground here. I need this PR merged and closed before we merge and close the NEW AI PR.
Either that or I need to delete that branch and do it all from here. I forgot I had worked on the AI mesh in this PR.
Let me know your preferences. It doesn't matter to me. I'll make sure everything is in sync if you merge this one first.

wlbragg:
@gilbertohasnofb heads-up, now we need to wrap up the AI changes before you can merge this PR.

gilbertohasnofb:
Just confirming: besides the AI fix is there anything else you plan to work on this PR? I will push the updated textures here shortly.

wlbragg:
No, I think that is all.

gilbertohasnofb:
@wlbragg Ok the new AI texture is pushed, all that is missing is the new bake of the panel lightmap. By the way I love the new AI model, that glass bit looks amazing!

gilbertohasnofb:
@wlbragg I also found a major bug: the glass in the gauges is also having the rain effect applied to them!

wlbragg:
@gilbertohasnofb I went ahead and edited that light map. Believe me, it is easier to edit it to redo it at this point. Especially that particular light map. It was a special process I did to create it. I had to merge a few separate bakes to create it. Clone tool is your friend in this case.

>I also found a major bug: the glass in the gauges is also having the rain effect applied to them!

Ah, I'll see what I can do for that.

>I also feel that the background of the AI is way darker than the new piece even though they use the same colors in the texture. Is this something you can try to improve a bit in the model itself? 

Yes, I made them different with a material setting. I'll either make it closer to the same or match or use the same material settings if I can't get them really close. That is why I was hoping you would make the colors off the textures ever so slightly different shades so I could remove the different material settings as they are inconsistent depending on the light.

Let me pull your latest push, then I'll add my changes to it and see where we are at.

wlbragg:
@gilbertohasnofb did you forget the panel texture sheet, to remove the cage knob AO from it?
I fixed the light map and just now pushed it as well as removing the cage system from the AI. I also think I removed the rain effect from the panel glass, would you please confirm that for me?

gilbertohasnofb:
> Clone tool is your friend in this case.

I just asked for the rebake because clone tool is never perfect, but I understand that was easier and you did a great job :clap: 

> did you forget the panel texture sheet, to remove the cage knob AO from it?

Ops, I did. Just pushed a new commit fixing all panel textures.

> Oh and I made the material settings between the roll plate and the pitch plate different by only .010 so please verify that made that close enough for your liking, thanks.

Looks very good now.

Are you ready to merge this?

wlbragg:
Yes, so long as you verified the rain on the panel glass is indeed gone.

wlbragg:
Just saw the bad normals on a couple panel bolts on the interior, middle bolts. I have no idea when that could have happened?

Anyway it's fixed and pushed.

gilbertohasnofb:
> Yes, so long as you verified the rain on the panel glass is indeed gone.

This is me taking off from water with the amphibian:

image

wlbragg:
Arrg, so it is still there!

gilbertohasnofb:
@wlbragg but not for rain, only for water take offs and landings (just tested it again)

wlbragg:
OK, I already fixed it. I'll push now.

gilbertohasnofb:
Bravo!

PR #1138, opened 2018-07-01
Particle lighting control system

### LINKED ISSUE #782: Create a particle lighting control system
Per Thorsten's advice I want to change the emissive value to true on all smoke and water particle effects, same as @onox did for the prop spray. I didn't want to potentially waist time with this without prior consent.

I tested the difference on all the "same" j3cub effects, sunset, clear skies. It appears you lose any atmospheric lighting changes in the effect (or they are very subtle) but we won't get any of the ugly weird color effects you might see at other specific lighting conditions.

wlbragg:
Fixes #782


gilbertohasnofb:
@wlbragg How do you suggest I test this? What sorts of particles should I be looking for? You fixed the particle system for what kind of situations and at what time of the day?

legoboyvdlp:
Is damage smoke included?

wlbragg:
>Is damage smoke included?

Yes

@gilbertohasnofb all this should be doing currently is causing any particle effect to change in visibility with the lighting conditions and slightly in color with terrain type.
Currently land snow and water are the only color changes.
Right now that is all that is being done.

If on land then the surface type and moisture conditions is also available and we could adjust  color and the density of particle effects. I'm not sure I want to bother with "density" as that would require modifying every particle effect file. But color would only require a small amount of code in the .nas file to convert the reported terrain type to color. The terrain lookup array in the .nas file is currently converting the terrain type to a density value VS a color value. 

Right now it is hard coded only to adjust for the 3 main land cover types, snow, land or water and it is not adjusting for ground cover such as grass VS dirt VS tarmac.

wlbragg:
>basically I need to test the particle effects when landing/taking off from land, snow and water in both day and night and see if all looks good, is that correct?

Yes, but don't merge yet. I need to get rid of some superfluous code. Don't let that stop you from testing though as it is just dead code I was holding onto in case I wanted to go farther. I think I have talked myself out of any more detail, for now. So you won't be waisting your time testing it.

wlbragg:
@gilbertohasnofb this is now ready to merge when your done testing it.

gilbertohasnofb:
@wlbragg All looks good to me

PR #1141, opened 2018-07-02
Add Aero Tech LC-2 digital clock 

legoboyvdlp:
Fixes issue #1124

### DIFF
--- a/Models/Interior/Panel/Instruments/digital-clock/digital-clock.xml
+++ b/Models/Interior/Panel/Instruments/digital-clock/digital-clock.xml
@@ -130,7 +130,7 @@
         </axis>
     </animation>
     
-    <!-- <animation> disabled until setting time and date is coded
+    <animation>
         <type>pick</type>
         <object-name>KnobLeft</object-name>
          <action>
@@ -138,10 +138,10 @@
             <repeatable>false</repeatable>
             <binding>
                 <command>nasal</command>
-                <script>clock.astrotech.set_time();</script>
+                <script>clock.astrotech.left_knob();</script>
             </binding>
         </action>
-    </animation>-->
+    </animation>
     
     <animation>
         <type>pick</type>
@@ -151,7 +151,7 @@
             <repeatable>false</repeatable>
             <binding>
                 <command>nasal</command>
-                <script>clock.astrotech.select_display();</script>
+                <script>clock.astrotech.select_mode();</script>
             </binding>
         </action>
     </animation>
@@ -164,7 +164,7 @@
             <repeatable>false</repeatable>
             <binding>
                 <command>nasal</command>
-                <script>clock.astrotech.control_action();</script>
+                <script>clock.astrotech.right_knob();</script>
             </binding>
         </action>
     </animation>
--- a/Nasal/clock.nas
+++ b/Nasal/clock.nas
@@ -13,7 +13,7 @@
         m.et_running=0;
         m.et_frozen=0;
         m.et_elapsed=0;
-        m.ft_start_time=0;
+		m.et_pause_time=0;
     m.ft_running=0;
         m.modetext =["LT","ET"];
         m.lc2 = props.globals.initNode(prop1);
@@ -35,7 +35,7 @@
     },
     
 #### displayed mode  ####
-    select_display : func(){
+    select_mode : func(){
         if(me.set_mode==0){
             me.MODE +=1;
             if(me.MODE>1)me.MODE -=2;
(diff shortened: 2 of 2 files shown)


gilbertohasnofb:
@legoboyvdlp ok, so let's catch up here: how is the implementation of the other buttons and functions doing? You mentioned having a problem resetting the timer, is this still the case?

legoboyvdlp:
Yep, I still have a problem with pausing and unpausing the timer. However, the buttons for the timer should be implemented correctly.

gilbertohasnofb:
@legoboyvdlp once you are done with the tutorials do you think you can take a look at this? This would be a cool feature to add :smile:

legoboyvdlp:
Mmm... I'm a little busy right now but I'll see what I can do this afternoon!

legoboyvdlp:
Yeah, that's right - (thanks for the texture ;) )

I would need lightmaps for the dome and red flood light - I don't think it would need a post light. I would expect the backlight and LCD display would be controlled by the radio panel knobs (just like the other avionics). No rush on that either.

gilbertohasnofb:
Indeed you don't want a post lights, so @wlbragg will need to add a conditional since the default clock has one. As for the lightmap I can create one manually or @wlbragg could create one in Blender (the latter option would look more accurate).

wlbragg:
>I would need lightmaps for the dome and red flood light 

We'll do them when your finished with it. I need it complete to do the light map correctly

gilbertohasnofb:
@wlbragg No rush but my understanding is that the model and textures are completely done, the only thing missing is some improvements to the functionality of the clock (i.e. Nasal coding).

wlbragg:
OK, then I can get it prepared.

gilbertohasnofb:
@legoboyvdlp @wlbragg do you guys think this feature can be done before the release? I think it would be really cool to have the option of a digital clock!

gilbertohasnofb:
Also, remember to make the choice of clock persistent, i.e. saved state between flights

wlbragg:
@legoboyvdlp you still have plenty of time, I think we decided the 2nd was our cutoff.
I was going to look at this tonight for the light map portion. But if you can get any outstanding nasal coding done, that would be great.

wlbragg:
@legoboyvdlp I got the light map and the glass panel effects done and will push shortly.

@gilbertohasnofb I noticed something that is really bugging me, it is what i consider to be a mistake in the light maps. The case I put around all the instruments I don't feel is shaded correctly for the post or overhead dome lighting, the red flood I think is OK as the whole black panel surface and anything dark doesn't show well i reed light anyway.
But as far as the other two lighting modes, I think it is wrong. 

I wanted your opinion. 
I don't know that i got the correct amount of light cast to them or if it is close enough?
Do you think I am correct about the current mapping being wrong?

Here is what is involved and how it would be changed, It's not a perfect solution but I think it is on the right path.
It's the small rings, they were a dark purple. Basically we change them to the same pink shad as the blocks around them, I'm not convinced that pink is the best color, It may be something close to that though.
vsi
alt

This shows the area I think is wrong.
lm3

What is wrong is the face of the case that is facing the front is dark, It should be catching more light from the overhead dome for sure and maybe the post lights depending on where the post lights are places. If the post lights are inside the case then no, they would be darker, but if they are built into the panel recesses (I suspect they are) then yes, they would catch a lot of light from the post lights. Iin fact we would really want to make that ring graduate from brighter on top to darker on the bottom ( I think we could do that maybe except the size of the ring might make graduating it difficult.
Dome lighting, the VSI and ALT have the corrections, the rest of the instruments are still wrong.
dome

Post lighting, again the VSI and ALT have the corrections and the rest of the instruments are wrong.
post

Maybe this isn't worth the effort, but I noticed it and the perfectionist in me is rearing it's ugly head.

I posted this here as this is probably were I would introduce the changes.

wlbragg:
Here is a side by side comparison. Right side is the corrected version. Don't look at the overall glare on the glass when comparing as that was just a slight difference in my light settings between the two shots. What I am getting at is only compare the face ring of the case taking any overall lighting adjustments into consideration.

Dome
badome

Post
bapost

gilbertohasnofb:
@wlbragg I see your point and I also had noticed this, I was just not sure it was worth the effort. As a fix, are you suggesting redoing the light maps or fixing them by hand? If the first, then please be careful as some of the light maps have been tweaked by hand. Here is the list I have of the files I manually altered:

```
lm_bat.png
lm_egt.png
lm_fuel-selector.png
lm_hi.png
lm_interior.png
lm_ki266.png
lm_magsw.png
lm_panel-parts.png
lm_panel.png
lm_pedals-copilot.png
lm_pedals-pilot.png
lm_yoke-l.png
lm_yoke-r.png
```

wlbragg:
>As a fix, are you suggesting redoing the light maps or fixing them by hand?

It would be by hand. It sch a small area on each light map that needs changed and those areas were all changed at least once before by hand. I debated whether this is worth it or not. 

Beings it is a relatively small change I think maybe I will go ahead and do it. It's only on the gauges and it is a relatively easy fix.

legoboyvdlp:
I've got it working - and if you press the RH button in clock mode you now get the month and day displayed :)

wlbragg:
@legoboyvdlp I'm really sorry for hijacking this thread with light map issues, it's only because I noticed it while in this branch making the light map for the digital clock.

Are you able to push your changes to legoboyvdlp-issue-1124 PR?

legoboyvdlp:
I thought I did!
Let me check where I _did_ push it...

legoboyvdlp:
@wlbragg see #1183 .

legoboyvdlp:
Just rename that PR and add 'Fixes issue #1124'

wlbragg:
Using New Aero Tech LC-2 digital clock #1183 PR instead of this remote.

PR #1145, opened 2018-07-03
adding HDG knob for HI

### LINKED ISSUE #1144: OBS knob of HI should have different label
Our model of HI only seldom has an OBS knob, but here is a photo of a model P sporting one:

But the label of the OBS knob is different than we have. It should be orange, labelled HDG and with a different symbol:

gilbertohasnofb:
Fixes #1144


gilbertohasnofb:
@wlbragg I also adjusted the lightmap, all you need to do is apply the correct label to the rightmost knob of the HI. See issue description if you need reference photos or more info

gilbertohasnofb:
@wlbragg the sides of the new knob has a problem in the lightmaps, notice how that's the only dark knob: 

fg-0
fg-1
fg-2

wlbragg:
@gilbertohasnofb Oh, yeah I moved the mapping to the new knob, the light map needed adjusted. I took care of it.

gilbertohasnofb:
@wlbragg Thanks for the help with this one!

PR #1147, opened 2018-07-12
Improve 3d objects

### LINKED ISSUE #544: Add a maximal speed in which the doors and baggage compartment door can be opened
Add a maximal speed in which the doors and baggage compartment door can be opened. Currently they can be opened at any speed and in-flight.

### LINKED ISSUE #802: Check dimensions of the levers
Heiko has found a schematic of the throttle and mixture levers of a 172. Values are in inches:

Could anyone who knows Blender well check if our values are correct?

### LINKED ISSUE #1023: Improve 3D model and texture of all levers

### LINKED ISSUE #1030: Improvements to the top of the vertical stabilizer
Hi there,
A few more problems:
image

Here we see that there are some strange lines on the top of the vertical stabalizer, with the N35799 livery.

OH-CTL was the other livery which I tested, which also presents the error:
image

Also, the model of the nav light is very unsmooth and not a great model:

This shows the top of the tail and light of a real Cessna:
image

I circled a few things that in particular would be nice to have modelled:
image

### LINKED ISSUE #1046: Wire antenna from tail fin to cabin roof
are you sure that the C172P has that type of antenna?
As far as I know, the C172P uses the King KR-87 ADF receiver, and the antenna for this is a complex antenna (KA44 or KA44B) assembled from the two orthogonal ferrite beads for the magnetic part and a small amplified antenna for the electric reference. The KR-87 does not use a wire antenna for the phase reference.

So there shouldn't be a wire antenna along the airframe as in the preview pictures.
(There are pictures in the internet that show a wire, these may use another ADF)

### LINKED ISSUE #1093: Propeller 3D Model - potentially too thin at the root
image

Is it just me or the camera angle, or is our propeller a bit too thin in blade width near the spinner?

### LINKED ISSUE #1096: Add Low Voltage LED to the Panel
To the right of the battery voltage indicator there should be a small red button or LED (not certain which) with the legend LOW VOLTAGE.

I would appreciate anyone who knows about the 172 or has access to one giving some information about this - is it a light or a button? 

Apologies for scanty information, but this does not have to be added until enough is known about it and what it does...

### LINKED ISSUE #1098: Yoke is missing PPT cord
There is a noticeable is the telephone cable type thing under the yoke:

image

Probably connects to the PTT button?

### LINKED ISSUE #1110: Add sunvisors above the pilot and copilot seats
We could add sunvisors above the pilot and copilot seats:

They could be adjusted using click and drag, so that their position can be adjusted to any angle like in the c182s. Actually we could even use that model as a base but the 172p should have non-transparent sunvisors.

### LINKED ISSUE #1122: Improve Avionics Switch and Autopilot Circuit Breaker
image

Our current switch and breakers look like this:

image

The part which could stand improvement is

image

- Make a hole in the avionics switch face - currrently it is solid
- Add the autopilot circuit breaker's unique head

image

You can see more in 1080p at https://www.youtube.com/watch?v=IP2PIjqf-Wc

### LINKED ISSUE #1137: Strobe lights dim or too weak
Would someone like to verify if our strobe lights have gotten too dim? Is it just me or my system or are they dimmer than they used to be?

If they appear correct to everyone else we can close this. They really seem dim to me.

### LINKED ISSUE #1148: Magnetic compass missing reference line
Our magnetic compass is missing reference line which is used to read the value. This is the model in real life:

and here is a screenshot of ours:

### LINKED ISSUE #1150: Landing Light Color
image

Some prankster seems to have painted our landing light bulbs? 🤣

### LINKED ISSUE #1151: Ammeter is the wrong model
Our ammeter is the wrong model. The 172 model P did not have a gauge showing the voltage, it was just an ammeter showing only the electric current. This is how it looks:

 

I checked at least 20 photos and 100% of all model P's had this same model and none had a gauge for the voltage. So should we remake this model?

Some reference shots:

### LINKED ISSUE #1154: Improve KMA20's marker sensitivity toggle animation
Improve KMA20's marker sensitivity toggle animation. All other toggles can be flicked either by clicking above or below them (there are two hidden squares as click spots) but also by hovering the mouse above the toggle and moving the mouse wheel up and down. The marker sensitivity toggle does not behave this way as it has a more complex behaviour. Ideally they should all be consistent.

### LINKED ISSUE #1155: Add black panel behind broken wing
I think we should add a simple black panel behind the broken wing. Currently you can see the wing interior through the holes but that doesn't look very good, that should be pitch dark. Here is what we have:

image

And this is what I propose:

image

a simple black object behind the broken wing structure on each wing should do the trick.

### LINKED ISSUE #1156: Remove shadow effect from radio digits
I don't think the radio digits should be affected by the shadow effect since they emit light themselves. Under the same RADIO LT settings, compare these two situations:

1) no shadow, all digits have the same light intensity:

image

2) under COMs and NAVs under shadows are dimmer, ADF is still bright:

image

### LINKED ISSUE #1157: Interior door handle animation is wrong
Interior door handle animation is wrong, see this video at 0:24: https://www.youtube.com/watch?v=kfGN_QEg1UU

We need to fix the following: 
* the pivot point for the interior handle is in the middle of the object, not in the back. We may need to 'carve' a bigger hole in the doors in order to accommodate the  movement
* the interpolation table should include more items as the movement of the door is very snappy and includes two stages. 
* when closing the door, the door handle animation should not take place.

### LINKED ISSUE #1159: Window handle base seems to be double mapped
Window handle base seems to be double mapped. Not the handle itself but the base (circle behind it). I suspect this because when the red flood light is on, it has a very strong red glow while nothing around it is even remotely red.

fgfs-screen-001

@wlbragg I have already pushed a new texture for this object and updated the lightmap to fix this issue, just map each of the bases to these, please:

image

### LINKED ISSUE #1162: Autopilot Panel Lighting
With the radio knob set to the minimum position shouldn't the autopilot's display also be fully bright (not the knobs)?

image

wlbragg:
Fixes #1023 
Fixes #802
Fixes #1096
Fixes #1110
Fixes #1148
Fixes #1122
Fixes #1137
Fixes #1098
Fixes #1151
Fixes #1154
Fixes #1093
Fixes #1150 
Fixes #544
Fixes #1030
Fixes #1046 
Fixes #1156 
Fixes #1159 
Fixes #1157 
Fixes #1155 
Fixes #1162

:warning: :warning: :warning: 
      DO NOT MERGE YET!
:warning: :warning: :warning:


gilbertohasnofb:
@wlbragg I hope you don't mind but I merged the branch `Issue-544` into this one since I started working on the door animations as part of this. I need to tweak some of the code affecting both branches, so I decided to merge them and close that other PR.

gilbertohasnofb:
Please do a `git pull` when confirm you have all the current development + the doors closing due to airspeed commits as well.

wkitty42:
why are we using directory traversal schtuff in our code? i just noticed that we have some in the latest diff for this issue... we should never ever ever be using directory traversal to get to something... not for something on the storage device or in the property tree... ../../../../foo **will break** at some point... if we can traverse back to something, we can certainly find its base name and start from there without a leading slash...

eg:
`<inherits-from>../../../../Effects/interior/interior-glass-reflection-panel-side</inherits-from>`
should be
`<inherits-from>Aircraft/c172p/Models/Effects/interior/interior-glass-reflection-panel-side</inherits-from>`
or where ever it is actually based from...

gilbertohasnofb:
@wkitty42 I think we always used this kind of relative paths. But I don't understand why you think we should not use it nor how it will break at some point. As long as the relative path is correct it should function as expected, right? Can you give an example where it would fail?

wlbragg:
@wkitty42 yes, please do enlighten me. I think of this as more of a guaranteed "relative" path structure VS an absolute path. If there is some inherent issue using this in FlightGear (MP for example or c++ internal path structure) then I need to know. Because in general I have always preferred relative path statements. I have found the opposite to be true, where using absolute paths can fail unless the application code accounts for the root directory. All things being equal traversing back through the directory structure from the calling directory should never fail as it is relative and will never change unless the directory structure changes in which case both styles will fail.

wkitty42:
On 08/04/2018 12:56 PM, Gilberto Agostinho wrote:
> @wkitty42 <https://github.com/wkitty42> I think we always used this kind of 
> relative paths. But I don't understand why you think we should not use it nor 
> how it will break at some point. As long as the relative path is correct it 
> should function as expected, right? Can you give an example where it would fail?


i'll just start a new issue because the discussion really doesn't belong in this 
PR and it could get fairly long... i do know that on the dev list, discussion 
about this has preferred to not use directory traversal "aka ../../../" and that 
the preference is to anchor and then traverse up to the desired directory... 
never traversing backwards with "../../../"...


i know that it will break if, as in the example i gave, the Effects directory is 
moved up a level or back a level... looking at the example code taken from the 
PR, one cannot even tell if the Effects directory is in the c172p's tree or if 
it is traversing outside of the c172p into one of FG's other directories that 
also contain an Effects directory... then there's the situation that other 
references to this very same Effects directory may use only one "../" to get to 
it while others may need "../../" or more to get to it...


give me some time to start a new issue... or maybe you can start a new issue and 
move or copy these related posts into it from this PR where they'll get lost 
really quickly...

gilbertohasnofb:
@wlbragg What's still missing before we can merge this?

wlbragg:
@gilbertohasnofb I have nothing except #1149 and I don't think that should stop us from merging.

I moved on to working on the Amphibious Tutorial which I will post in a new PR when ready.

I also would like to move on to the startup states and possibly the variants.

gilbertohasnofb:
@wlbragg One last request before we merge this. Would it be possible to add the glass effect to the radio stack as well? See this photo:



Basically all the readouts have a glass in front of them. Would look so pretty!

gilbertohasnofb:
@wlbragg I also pushed a new commit improving the radio stack readouts, they all now have a consistent colour (the DME and AP were a bit off):

fg

wlbragg:
>Would it be possible to add the glass effect to the radio stack as well? 

Sure, lets try it. I actually thought about going this as well.

gilbertohasnofb:
@wlbragg Thanks!! maybe this would be as easy as adding a glass pane just above the readouts (as little z difference as possible but non-zero as to avoid z-fighting).

wlbragg:
@gilbertohasnofb I think this is done.
I realized while doing it our hobbs meter had a few issues. No lightmap effect and the lighting was incorrect (it shouldn't have any). Also mine didn't work at all. It had a bunch of faced that were hidden from view I removed and I resized the dials a bit to make them fit better. Please make sure it is working correctly. Also the transponder had some lighting affect that I don't think was appropriate, please double check that as well.

gilbertohasnofb:
@wlbragg wow, what a difference that makes to the radio stack, thank you so much for this. The glass effect on the gauges is by far my favourite effect in our aircraft. Also the Hobbs is looking good under the dome lights now 👍

PR #1152, opened 2018-07-26
Issue 544

### LINKED ISSUE #544: Add a maximal speed in which the doors and baggage compartment door can be opened

gilbertohasnofb:
Fixes #544

### DIFF
--- a/Systems/ground-effects.xml
+++ b/Systems/ground-effects.xml
@@ -708,7 +708,15 @@ Under the GPL. Used by shadows under ALS -->
                             <difference>
                                 <value>1.0</value>
                                 <product>
-                                    <property>/velocities/airspeed-kt</property>
+                                    <sum>
+                                        <property>/velocities/airspeed-kt</property>
+                                        <sin>
+                                            <product>
+                                                <value>1.07</value>
+                                                <property>/velocities/airspeed-kt</property>
+                                            </product>
+                                        </sin>
+                                    </sum>
                                     <value>0.01</value>
                                 </product>
                             </difference>
@@ -736,7 +744,15 @@ Under the GPL. Used by shadows under ALS -->
                             <difference>
                                 <value>1.0</value>
                                 <product>
-                                    <property>/velocities/airspeed-kt</property>
+                                    <sum>
+                                        <property>/velocities/airspeed-kt</property>
+                                        <sin>
+                                            <product>
+                                                <value>1.38</value>
+                                                <property>/velocities/airspeed-kt</property>
+                                            </product>
+                                        </sin>
+                                    </sum>
                                     <value>0.01</value>
                                 </product>
                             </difference>
@@ -764,7 +780,15 @@ Under the GPL. Used by shadows under ALS -->
                             <difference>
                                 <value>1.0</value>
                                 <product>
-                                    <property>/velocities/airspeed-kt</property>
+                                    <sum>
+                                        <property>/velocities/airspeed-kt</property>
+                                        <sin>
+                                            <product>
+                                                <value>0.73</value>
+                                                <property>/velocities/airspeed-kt</property>
+                                            </product>
+                                        </sin>
+                                    </sum>
                                     <value>0.02</value>
                                 </product>
                             </difference>
--- a/Systems/ground-effects.xml
+++ b/Systems/ground-effects.xml
@@ -692,4 +692,92 @@ Under the GPL. Used by shadows under ALS -->
         </output>
     </logic>
 
+    <!-- ================================================================== -->
+    <!-- Max range for doors and baggage doors  -->
+    <!-- ================================================================== -->
+
+    <filter>
+        <name>Effective range for right door</name>
+        <type>gain</type>
+        <input>
+            <expression>
+                <product>
+                    <property>/sim/model/door-positions/rightDoor/position-norm</property>
+                    <max>
+                        <min>
+                            <difference>
+                                <value>1.0</value>
+                                <product>
+                                    <property>/velocities/airspeed-kt</property>
+                                    <value>0.01</value>
+                                </product>
+                            </difference>
+                            <value>1.0</value>
+                        </min>
+                        <value>0.5</value>
+                    </max>
+                </product>
+            </expression>
+        </input>
+        <output>
+            <property>/sim/model/door-positions/rightDoor/position-norm-effective</property>
+        </output>
+    </filter>
+
+    <filter>
+        <name>Effective range for left door</name>
+        <type>gain</type>
+        <input>
+            <expression>
+                <product>
+                    <property>/sim/model/door-positions/leftDoor/position-norm</property>
+                    <max>
+                        <min>
+                            <difference>
+                                <value>1.0</value>
+                                <product>
+                                    <property>/velocities/airspeed-kt</property>
+                                    <value>0.01</value>
+                                </product>
+                            </difference>
+                            <value>1.0</value>
+                        </min>
+                        <value>0.5</value>
+                    </max>
+                </product>
+            </expression>
+        </input>
+        <output>
+            <property>/sim/model/door-positions/leftDoor/position-norm-effective</property>
+        </output>
+    </filter>
+
+    <filter>
+        <name>Effective range for baggage door</name>
+        <type>gain</type>
+        <input>
+            <expression>
+                <product>
+                    <property>/sim/model/door-positions/baggageDoor/position-norm</property>
+                    <max>
+                        <min>
+                            <difference>
+                                <value>1.0</value>
+                                <product>
+                                    <property>/velocities/airspeed-kt</property>
+                                    <value>0.02</value>
+                                </product>
+                            </difference>
+                            <value>1.0</value>
+                        </min>
+                        <value>0.0</value>
+                    </max>
+                </product>
+            </expression>
+        </input>
+        <output>
+            <property>/sim/model/door-positions/baggageDoor/position-norm-effective</property>
+        </output>
+    </filter>
+
 </PropertyList>
--- a/Systems/ground-effects.xml
+++ b/Systems/ground-effects.xml
@@ -714,7 +714,7 @@ Under the GPL. Used by shadows under ALS -->
                             </difference>
                             <value>1.0</value>
                         </min>
-                        <value>0.5</value>
+                        <value>0.51</value>
                     </max>
                 </product>
             </expression>
@@ -742,7 +742,7 @@ Under the GPL. Used by shadows under ALS -->
                             </difference>
                             <value>1.0</value>
                         </min>
-                        <value>0.5</value>
+                        <value>0.51</value>
                     </max>
                 </product>
             </expression>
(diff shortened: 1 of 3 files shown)


gilbertohasnofb:
@wlbragg this is ready to be tested and possibly merged. To test, open both doors and the baggage door and accelerate the aircraft, the doors will start to close as airspeed raises. At ~50 kts the doors will have reached nearly closed position, but on purpose I left them a bit opened, see:

fgfs-screen-001

I added drag proportionally to the door positions so that accelerating with the opened doors makes a substantial difference. Windows cause a minuscule amount of drag, absolutely imperceptible. But please do test these scenarios.

wlbragg:
@gilbertohasnofb how cool, I will test this between now and tomorrow.

wlbragg:
I couldn't wait to test. That is really cool.
There is something I would change if it is practical but I am not sure how you would do it.
It doesn't feel totally realistic to have the doors open and close symmetrically synced with the air speed.
Let me see if I can explain.
The most critical loss of realistic immersion for me is basically at the speeds where the doors are moving because of the change in speed. Specifically, if your increasing in speed and the doors are slowly closing, then you reduce the speed and the door slowly open, That is not realistic to me. I don't think that would happen, I don't think the doors would open unless you drastically reduced speed with the brakes maybe.
I think we need two mechanisms. Closing with airspeed but not opening with lack of airspeed. Maybe once they start closing the only way they can be opened is either with the handle or breaking.
Does that male sense? Do you think that is practical to implement?
I would even think it better to only close with airspeed and open with handle VS opening with decreasing airspeed. That part, opening with decreasing airspeed needs to be removed.

One other comment, would it be difficult to scale the closing weighted to close faster once the door starts to move closed. In other words, not so linear but weighted  more on the closed side?

The Idea is really cool though, I also like leaving the door ajar a bit.
Trying to open the doors when in flight and not being able to except slightly also felt right.

With all that said, I realize we really shouldn't be leaving the doors open in the first place so it isn't like this is a feature that is going to be used consistently.

gilbertohasnofb:
@wlbragg Done and pushed, I think this looks much better. There might still be some room for improvement but for such a small effect I think this might already suffice. What do you think?

gilbertohasnofb:
In case you are interested, I basically added a small factor to the airspeed using a sinusoidal function and each door will have a function with different rates. Here is the superimposed graph of all three of them: http://fooplot.com/#W3sidHlwZSI6MCwiZXEiOiJ4K3NpbigxLjA3KngpIiwiY29sb3IiOiIjMDAwMDAwIn0seyJ0eXBlIjowLCJlcSI6Ingrc2luKDEuMzgqeCkiLCJjb2xvciI6IiMwMDA0RkYifSx7InR5cGUiOjAsImVxIjoieCtzaW4oMC43Myp4KSIsImNvbG9yIjoiI0ZGMDAwMCJ9LHsidHlwZSI6MTAwMCwid2luZG93IjpbIi0zMi4xNzg4MTkxNzk1MzQ3MyIsIjY0LjY3ODcyODU4MDQ3NDYxIiwiLTIuNTg4Mjg1OTIzMDA0MTIxIiwiNTcuMDE2MzU4ODUyMzg2MjYiXX1d

wlbragg:
That did improve the overall effect by un-syncing the left and right doors. Nice technique to have in the arsenal.
Unfortunately that wasn't exactly what I was talking about.
Here is how I think it should work. I also think it may require less code and less advanced code then you have already invested in. Or maybe just a parameter of breaking as a condition on the opening side of the animation.

First, in a car for example, once the wind is at a sufficient speed the doors shut fairly rapidly. That is if you put you foot into it and accelerate like your getting on a highway or up to speed in excess of 25mph+. In an aircraft taxiing, I don't know. But for sure once on the runway and throttle to full they're going to shut in a hurry as you build speed.
Then if you start to slow down, gradually, I don't think they are going to open. Only if you apply brake or any drag that slows you down faster than pure wheel friction on a hard surface do I think they would open once closed.
By tying the doors open and close to the speed, it isn't as realistic as it could be. Opening should not be tied to speed, only to deceleration.
I think the closing portion is OK, I would still like to see it weighted faster as it starts to close, especially if still accelerating.
The opening I think should either be removed or only work if brake is being applied above a certain level. Make sense?

wlbragg:
@gilbertohasnofb after more testing it is real close, especially if you can untie opening from speed and/or adding braking condition before allowing opening. That would also introduce another randomness type of effect to the animation if you know what I mean.
I think that is all it would take to make this look right to my perception.

gilbertohasnofb:
> First, in a car for example, once the wind is at a sufficient speed the doors shut fairly rapidly. 

Perhaps I can considerably diminish the threshold value from 50 kts to let's say 10 or 15 kts.

> Opening should not be tied to speed, only to deceleration.

Oh man, I just really don't even know where to start with that. There is no such opening or closing as far as the code is concerned, it's just changing the max range. Would you have some concrete suggestion on how to implement what you have in mind?

gilbertohasnofb:
oh wait, I have an idea. I think I need two filters: one enabled only under braking and one enabled only when NOT braking. But the door will flick between the positions given by the two filters -- how to make it smoothly change from one to another? I will give this a thought and try it later today.

gilbertohasnofb:
@wlbragg That didn't work either. There are too many problems. For instance, we want the doors to shut if the player applied brakes or parking brake, but not if airspeed is low (else the doors won't open on the ground with a little breeze). This then becomes a a huge mass of conditions, and worse (as I expected) one cannot use a gain filter like this and simply smoothly changes the values when a filter is enabled. As soon as a filter is enabled, it will start to apply its formula to whatever output paremeter it has, regardless of its current value.

I frankly think this is way beyond any reasonable gain/pain ratio. So unless there is some easy technique to achieve this out there, I would say that we can live with the current animation given that this is just a tiny special effect. Else I will end up having to fake a physical system using only xml and animations just for the sake of dealing with doors, which 99.999% of players will have shut all the time while the engine is on.

gilbertohasnofb:
@wlbragg I will try one more thing, please don't merge this yet.

wlbragg:
@gilbertohasnofb looking at the filter I agree there is no easy fix except, don't use 
~~~
<property>/velocities/airspeed-kt</property>
~~~
use something like 
~~~
<property>/accelerations/pilot/x-accel-fps_sec</property>
~~~
and condition to use it only if it exceeds X amount of acceleration or deceleration

gilbertohasnofb:
@wlbragg the issue is that a filter is simply outputting a property in relation to others. The conditional is trivial, the difficulty is to create something like a smooth animation. I have managed to create a second filter which is enabled only under breaking, but there are two caveats:

```
    <filter>
        <name>Brakes close right door</name>
        <type>noise-spike</type>
        <max-rate-of-change>0.01</max-rate-of-change>
        <enable>
            <condition>
                <and>
                    <less-than>
                        <property>/accelerations/pilot/x-accel-fps_sec</property>
                        <value>0.0</value>
                    </less-than>
                    <greater-than>
                       <property>/sim/model/door-positions/rightDoor/position-norm-effective</property>
                        <value>0.51</value>
                    </greater-than>
                </and>
            </condition>
        </enable>
        <input>
            <value>0.51</value>
        </input>
        <output>
            <property>/sim/model/door-positions/rightDoor/position-norm</property>
        </output>
    </filter>
```

caveat 1) it will jump to `0.0` since the input is just a `<value>` of `0.0`. AFAIK there is no way around and this looks terrible
caveat 2)  _any_ time the velocity decreases the acceleration figure will read a negative value, so right now the door is banging even if you just lower the throttle a bit. That is not the same as hitting the brakes. Of course we can try to find a sweet value but that will be tricky to say the least.

I am really not sure how to work around this without a complex Nasal function, and frankly we don't want to hog our system just for some door animation easter egg.

gilbertohasnofb:
Note that in my filter above was hoping that `<max-rate-of-change>0.01</max-rate-of-change>` would make the `<output>` smoothly move to the value of `0.51`, but that's not the case. `max-rate-of-change` works only when you have a property changing over time, not for a fixed value, hence the door teleporting to the close position.

wlbragg:
One more thought, why can't pilot/x-accel-fps_sec be used as a trigger to close the door or open the door depending on a predetermined accel or decel value? It just triggers a one time animation sequence that either opens or closes the door. By closes I mean slightly ajar like you already do.

pseudo code being something like...

If pilot-accel > X and door-open then
     adjust /sim/model/door-positions/rightDoor/position-norm to simulate closing
else
If pilot-accel < X and door-open then
     adjust /sim/model/door-positions/rightDoor/position-norm to simulate opening

and...

if airspeed > X then door open not allowed

To limit opening the door past a predetermined airspeed.

Are we making this more difficult than it needs to be? Or is this idea even more complex that I am imagining?

Because it seems like the problem lies in trying to use the same incremental property to incrementally control the action both negative and positive. Instead it seems like it needs to be a property trigger to set in motion a complete animation sequence.

I feel like I am completely trying to take you off the course you set out on and don't want to cause you any frustration. So if this doesn't make sense to you, or seems to be too much of a change in direction, simply go with the best implementation you already came up with. It'll work and it is a vast improvement over doing nothing.

gilbertohasnofb:
> But, how hard would it be to at least speed up the closing once it starts? Would that also have the opposite effect of speeding up the opening then as well?

There is no way of speeding up the animation only for closing. I am actually tricking the door system: as far as the sim is concerned, the doors are fully opened, I am just hacking what the sim understands as fully opened position. So at any point in time the position of a door is just proportional to the current airspeed, there is no such thing as opening or closing at all.

> Instead it seems like it needs to be a property trigger to set in motion a complete animation sequence.

That's exactly the problem, that's not how filter works. They don't trigger stuff, they just map according to a formula, constantly. Enabling a filter simply means enabling a specific mapping, but you can't smoothly interpolate from the previous value, which is why the door teleports to the new position. 

> Are we making this more difficult than it needs to be? Or is this idea even more complex that I am imagining?

You are suggesting a completely different implementation which can't be done with filters as far as I know, and what you suggest is also fairly complex. I am afraid this can only be created with a fairly complex Nasal function, and I am not too happy about the prospect of adding more nasal code for something so small.

> I feel like I am completely trying to take you off the course you set out on and don't want to cause you any frustration. So if this doesn't make sense to you, or seems to be too much of a change in direction, simply go with the best implementation you already came up with. It'll work and it is a vast improvement over doing nothing.

I am totally fine discussing this further or taking some more time before merging this, no worries about that. It's just that I don't know I frankly have no clue how to implement what you are saying without a fairly complex Nasal system. And I feel that's a really overkill for a tiny 'easter egg'. But we can discuss this further or ask around if you wish to, I am really fine with that :smile:

wlbragg:
@gilbertohasnofb  I'm still looking at this. I haven't as of yet come up with a better solution.

gilbertohasnofb:
Closing this due to https://github.com/c172p-team/c172p/pull/1147#issuecomment-409261774

PR #1163, opened 2018-08-08
Create tutorials for float variants

### LINKED ISSUE #1054: Create tutorials for float variants
It would be nice to have a take off and a landing tutorials for the float variants.

wlbragg:
Fixes #1054


wlbragg:
@gilbertohasnofb I am extremely happy with this now. As far as I am concerned (barring any obvious bugs) this is finished.
I took the approach that by the time a pilot would be learning how to fly and land an amphibious aircraft they really should have the basics down. These two tutorials teach you all you have to know to operate the amphibious portions of the aircraft. But it doesn't hold your hand in the basic operation of an aircraft.

gilbertohasnofb:
@wlbragg Ok, I tried the take off one and I had a couple of issues and I also have some recommendations:

* first, when I selected the tutorial, the aircraft was crashed then repaired, but the camera stayed floating above it and I had to press Ctrl+C to go to the cockpit:

fg-0

* 'turn pontoon master power switch on' sounds confusing to me. Why not call it landing gear?
* I had auto-started the engine when I was told to go through my preflight checklist (yeah, I shouldn't) and then the tutorial got stuck. Once I killed the engine by pulling the mixture the tutorial moved to the next item which was Start the engine. Maybe you need to add some conditionals there to make sure that if the aircraft already is in a certain state then the tutorial can move along.
* I got completely stuck on the Continue taxiing towards water runway 4W/22W... how am I suppose to know where the runway is exactly? I was pointing to 040 but apparently in the wrong place:
fg-2

* You need to remove the capitalization from Taxiing in the previously mentioned phrase.

Since I got stuck at locating the runway I had to stop there :cry: I will test now the landing one.

wlbragg:
For both taxing to water runway and to mooring location, there is an issue with using the available targets - taxi1 tools that come with the tutorial code.

I know that is a week spot, you have to know the general area where those targets are.
In both cases the tutorial lines you up right off the bat using...

Taxi towards
~~~
<error>
            <message>You're off course. To reach the runway turn to the left.</message>
            <condition>
                <less-than>
                    <property>/sim/tutorials/targets/taxi1/direction-deg</property>
                    <value>-20</value>
                </less-than>
            </condition>
        </error>
        <error>
            <message>You're off course. To reach the runway turn to the right.</message>
            <condition>
                <greater-than>
                    <property>/sim/tutorials/targets/taxi1/direction-deg</property>
                    <value>20</value>
                </greater-than>
            </condition>
        </error>
~~~
You are not allowed to leave that part of the tutorial until you are lined up. As soon as you trigger that then you get (in both landing and takeoff)...

Continue Taxing towards where the target changes to 
~~~
<exit>
            <condition>
                <less-than>
                    <property>/sim/tutorials/targets/taxi1/distance-m</property>
                    <value>100.0</value>
                </less-than>
            </condition>
        </exit>
~~~

So once you get the "Continue Taxing towards" if you simply stay on the course at that point, you will trigger the next step. Can you try it again and tell me what the last step was that you got stuck on, was it
Taxi towards
 or
Continue Taxing towards

Because that will make a difference on what I need to do.


If you combine those two into the same step, then they all have to be correct at the same time and it just refused to work. Which is why I eventually had to separate them

I don't want to go down a rabbit hole so it would help if you could try again and tell me which of those two steps were the main problem.

Is there a way to show a visual clue as to direction in the tutorial code?
Or, is there a way to capture a property and then message that property in the tutorial. I know you can use nasal to capture any property but I don't know if there is a print statement to display a value of a property in the tutorial, that is what I need.

The rest of your comments duly noted and I'll correct them.

Also I prefaced the tutorial with basically you have to look at the chart and know where the runway is, problem is we don't provide the chart. I also prefaced the tutorial with you have to know approximately where the mooring location is but that isn't even on the chart, that is a best guess based on some photos of a seaport charter service there in the lagoon. (hint) it is right in front of where the buildings are in the water on the edge of the lagoon. Looks like a boat ramp and then a flooded building. 
I could calculate direction to both the mooring and runway entry point, I have the tools available in the tutorial code, I just don't have a way to display them in the tutorial message. Unless that is allowed using in-line concatenating techniques, I never actually tried it.
Or being allowed to message a property value, I never tried it either.
Either one of those would solve this.

gilbertohasnofb:
@wlbragg I think the main issue is that it took me a while to get to the take off part and by then I had drifted away too much from the runway area, and I had no way of knowing where that is. The options you propose might help but, frankly, do we really need to control the area where the player is taking off from? What about just letting them take off with a certain heading, regardless of the actual position of the aircraft? I understand that using a water runway in RL should be precise (except if you use your private lake), but then again in RL you have charts and other stuff to rely on. As for the landing, the tutorial could ask the player as the single last item to both "move closer to the mooring area and activate the mooring option" but your condition would NOT include the position, just whether mooring is activated; this way the tutorial will terminate regardless of where the player activates it, but they will at least make the effort of getting close to the shore. I think that tutorials are there to help players learn the systems more than anything else, so it would be very handy to simply introduce them this way. What do you think? 

> The rest of your comments duly noted and I'll correct them.

Great, once you push a fix I will test again!

wlbragg:
>I had auto-started the engine when I was told to go through my preflight checklist (yeah, I shouldn't) and then the tutorial got stuck. Once I killed the engine by pulling the mixture the tutorial moved to the next item which was Start the engine. Maybe you need to add some conditionals there to make sure that if the aircraft already is in a certain state then the tutorial can move along.

This one is a must. if you start your engine before you release the mooring line bad things will happen, think about it in a real life situation.
The tutorial does tell you not to start your engine if you do it in that situation. So I guess I am not sure what your asking for, if you start your engine in the wrong step to go to the next step? I can't because the rope will still be connected and you will crash when you try to taxi.
So for this one I added a message to "Turn it off or you will damage your aircraft" if you try to start it before your supposed to.

I am pretty sure I addressed everything else. I am really happy with it now, Using the heading bug to direct everything works like a charm. I set it constantly to direct your every move. It felt pretty foolproof to me.

Please try them again and let me know if everything is OK.

gilbertohasnofb:
I agree with you about the engine.

> I am pretty sure I addressed everything else. I am really happy with it now, Using the heading bug to direct everything works like a charm. I set it constantly to direct your every move. It felt pretty foolproof to me.

Great, testing now!

gilbertohasnofb:
Ok, tested the landing one. A couple of thoughts:

* I think it's not 'cross leg' but 'crosswind leg'
* when I was told to change course to 050 for landing, I did so immediately. Look where the tutorial pointed me to:
fg-2
* I don't like the 'continue taxiing ...' followed by 'hint: look at the heading bug'. A single 'the heading bug will guide you to the mooring location' would be better
* I followed the heading bug and look what happened:


fg-5
fg-6
fg-7
fg-8
fg-9

Apologies if I am asking a lot of changes, I know you want to get done with this but these tutorials are such wonderful tools for beginners to learn that I think they are worth the time and effort to tweak them to the best we can.

wlbragg:
>why is the damage off by default? 

The tolerances for damage are "sometimes" exceeded when spawning on the water so I have to turn it of as soon as "fdm_init", and then reset damage in case it broke while spawning.
I don't see damage occur very often after you already have spawned, but it does happen. Most often it is if the aircraft hits the buoy too hard or the wave oscillation is just the right amount.
We might revisit the gear damage tolerances again, I think they are too low as I can still break a gear relatively easy if I brake hard. I really don't know how hard you have to brake to actually break a gear. I think it would be almost impossible. For one the shock should help mitigate the possible damage. I don't know that the front gear has any brake applied to it. Logic says it doesn't and so it should continue to roll during heavy braking and the shocks should do their job. If it did have brakes then it should just skid and again the shocks should help mitigate damage.
I'm not certain that is the cause of the damage on the water but I thin it is. This might require some decent testing to determine what is breaking in these cases.

>are you sure that switch on the landing gear is a ON/OFF switch? Because the texture says DIM. Isn't it perhaps a two position switch for the light intensity, given the word 'dim'? We need to research this. Even if that is a on/off switch, is ON position down (unlike say the avionics switch)?

I just recently read something that indicated a power switch to the amphibious gear circuitry which is what made me decide to hook that up. I don't remember exactly what it said but it convinced me that there is a switch somewhere. I'll see if I can find it again.

...Time passes....

I couldn't find it again :cry: 

Not understanding the language (see the rest of the object names) I wouldn't jump to the conclusion the "dim" doesn't also fill the place of on/off.
Yes we need more definitive proof, you have me second guessing my decision again. Until proven otherwise I would like to leave it operational as this piece of circuitry would most certainly have a breaker switch in line, it would have to or you would potentially drain the battery. Plus I doubt it would meet specs with no circuit breaker of any kind. Nut if this is a "dim" switch then it needs to be converted and we need to install a power switch somewhere.
As far as the switch position, I can flip it.

>First message I get is to taxi to runway 4W. Only then I am told to use the heading bug to orient myself as a hint. I think much simpler would be a single item 'Taxi to runway 4W, use the heading bug to guide you'

I hear you, but I am reluctant to use "hint: heading bug" directions in anything but the errors if the user is not lined up. Technically you can get to the runway and mooring positions without ever getting that hint at all. I think it comes up quick enough if your off course or don't know where to go. And if you do know where to go then it won't come up at all.

>Ok, tested the landing one.
...
    when I was told to change course to 050 for landing, I did so immediately. Look where the tutorial pointed me to:
 
Yes, the landing requires an aggressive bank to be close to lined up. How should I indicate that, should the tutorial offer a bank angle? What is the most an aggressive bank angle allowed?

>why the take off course is 070 if we are on runway 4W? Shouldn't it be something closer to 040?

I've been caught, I guess I can't have my cake and eat it too with realism, it either is real or not!

Yes, the takeoff heading does not match the runway, I need to change it. I had already got all the positions before I realized it was wrong and didn't want to do it again. But I have done it now enough that it's not that big of a deal to do it one more time.
With that said, when I change the takeoff heading it will be "exactly" the same as the landing. That "crosswind: leg will be so short that you won't have time to do any flap and gear changes (like landing is so short a crosswind because the land and water runways are so close together)

So...
>I think flaps and gear should be set on the downwind leg, not on the base leg. There is not enough time by then.

won't apply anymore and it will have to be done on the approach.
Also, you didn't mention the time crunch for deploying flap and gear for the landing and it is technically the same distance on approach? But I agree on the takeoff there is a bunch of time in the crosswind leg. However that will change when I change the takeoff course. there will be no time in the crosswind leg.

>I followed the heading bug and look what happened:

Where did you land?
land

Your pictures that run aground don't look anything like the area your supposed to be pointing at. You should be seeing the airport buildings?
lan2

wlbragg:
@gilbertohasnofb FYI: I am making most of the changes you requested/suggested.
When I am done you can pull, critique again and comment.

wlbragg:
@gilbertohasnofb please pull and try again. I think this covers most of it. It works well for me.

gilbertohasnofb:
> I don't see damage occur very often after you already have spawned, but it does happen

Yeah, it did happen to me once. Maybe this would be a case to change the damage variables if that property that checks if the plane is on the water is set to `true`? This way the plane would still break if one crashes during a water take off or landing but would not break just from mooring. Do you think this would be possible? If so we can open a new issue for that.

> We might revisit the gear damage tolerances again, I think they are too low as I can still break a gear relatively easy if I brake hard. I really don't know how hard you have to brake to actually break a gear.

I would say that a gear collapse should happen solely during a _very_ bad landing with very high vertical speed.

> I just recently read something that indicated a power switch to the amphibious gear circuitry which is what made me decide to hook that up. I don't remember exactly what it said but it convinced me that there is a switch somewhere. I'll see if I can find it again.

p. 102 of the manual (https://www.wipaire.com/parts_manual/model-2350-parts-manual-2/) talks of a switch dimmer. I don't think that's a on/off switch then, I suspect the gear is just attached to the avionics switch.

> Yes, the landing requires an aggressive bank to be close to lined up. How should I indicate that, should the tutorial offer a bank angle? What is the most an aggressive bank angle allowed?

The difficulty is that I don't think anyone would take off from that concrete runway and then land on that water runway, there is simply not enough time in RL. I see now from you screenshot what was happening, I was landing much to the right of that area due to taking longer time to reach 1000 ft. Two solutions come to mind, but you might not like either: a) start the tutorial in the air and b) use my previous suggestion of not requiring a landing on a specific area, just _any_ water landing. Option b) isn't bad if you think that the role of the tutorial is to teach beginners how to handle the aircraft and its procedures and not to teach them about how things are done with this specific airport. As I said, people taking off from lakes or rivers in RL (or even the sea in some cases, as is the case in Brazil) would not use a specific water runway but just land as close as possible to where they want to be.

> I've been caught, I guess I can't have my cake and eat it too with realism, it either is real or not!

Hehehe

> I had already got all the positions before I realized it was wrong and didn't want to do it again. But I have done it now enough that it's not that big of a deal to do it one more time.

Great! But this means you will hate my option b) above even more...

> Your pictures that run aground don't look anything like the area your supposed to be pointing at. You should be seeing the airport buildings?

That's because I was landing waaaay away from the correct place. Sorry I just find it quite confusing to 'find' the water runway.

wlbragg:
>Maybe this would be a case to change the damage variables if that property that checks if the plane is on the water is set to true? This way the plane would still break if one crashes during a water take off or landing but would not break just from mooring. Do you think this would be possible? 

Yes, I think so. 

>If so we can open a new issue for that.

Let do make a new issue for it and even if we don't get it in this release we will in the next. In this new issue lets also include making the gear breakage harder on land as well.

>the manual ... talks of a switch dimmer. I don't think that's a on/off switch then, I suspect the gear is just attached to the avionics switch.

Then I will make it a dimmer switch
I am also going to tie the entire gear panel into the avionics switch then. It needs to be on a circuit with a breaker, one we can shut off without killing the engine.

>The difficulty is that I don't think anyone would take off from that concrete runway and then land on that water runway, there is simply not enough time in RL.

Your probably right, but I am going to leave it like this, I really like the flow now.
Here is my reasoning why I want to use it the way it is and I hope you can agree this is sound enough logic to leave it as is (after you test it again for any other bugs)

First...
>Yes, the landing requires an aggressive bank to be close to lined up. How should I indicate that, should the tutorial offer a bank angle? What is the most an aggressive bank angle allowed?

I was wrong about this. It does not require an aggressive bank angle or turn to achieve the runway lineup, as a matter of fact it is just about perfect with a casual 180.

Second...
Instead of only using 1000 ft altitude for the trigger to make the 180 I am now using a longitude as well. I can make the point as far out as we want to allow time for the approach procedures.
The way it is now, after my last push, I actually like the limited reaction times in this tutorial and here is why I like it so much.
By the time a pilot takes on a float plane, I would imagine they have a bunch of experience under their belt. Think about it, like you said "people taking off from lakes or rivers in RL", fishing lakes, hunting camps, etc. All these can be situations requiring a very good skill set, fast reactions, steep and extremely short approaches. So with that in mind the simple lazy 180 and a mildly aggressive approach sounds about right for a tutorial to get your feet wet in using the amphibious.
These two tutorials cover "every" float action we offer. So in my opinion, these two tutorials, as they are written now, are more than adequate to serve their purpose.
The changes you have suggested and that I have incorporate into this latest version, have helped make it pretty darn effective and just know that I have a lot time flying and tuning these float variants and I am extremely happy with the information presented by these tutorials.
So I urge you to pull the latest, and try to follow the tutorial again with all the changes we have made and all I have said in mind. See if you can complete it. With my experience level of flying and testing this system, I am pretty satisfied where were at.

Now changes you need to know about also contained in the PR...

I re-factored the dim switch to be a dim switch.
I tied the gear panel into the avionics switch
I re-factored the ALS procedural glare effects a bit for better readability and to allow for some "use" statements in some of the effect files pertaining to the dimmer switch.
I had to add the avionics switch logic condition to the gear up/down logic and the mooring logic as well as the tutorials.

Whew, I think that is it.

wlbragg:
@gilbertohasnofb one more little note, this is just to maybe save you some time and to head off a requested change. Depending on the angle that you are taxiing to the mooring location, if you are in a hurry (are going to fast) you may not have time to shut the engine off and turn into the wind. There is some common sense required to execute a mooring/anchoring operation. The distance was really tight, even for me who has some experience and was expecting it if I cam in too fast. I think maybe a little to tight, so I gave it a little more room. Still just know if your being "foolish" (too much speed) in your operations, you may get burned.

gilbertohasnofb:
@wlbragg I created an issue for the damage problem: https://github.com/c172p-team/c172p/issues/1175

Everything you mention looks good to me and the tutorials are nearly perfect now. I have just a last handful of tiny comments for you to consider:
* why do you use heading `043.4`? You should pick either `043` or `044`, this notation you use is not standard and the minimal degree used in headings is 1 degree AFAIK.
* you say press `R` to lower the water rudder. Perhaps better should be `Shift+R`. I know that FG likes to use capital letters for shift but I find that not very intuitive. Just an idea though..
* In the take off tutorial, it would be better if you would write `180 degree` instead of just `180`,  like you used in `Tutorials/amphibious-landing.xml` (by the way these 180 degree turns are wonderful, they really make everything simpler!)
* Maybe it's worth mentioning in the description of both tutorials that the water runway 4W is parallel to the 4L and 4R and is located in the small water section in the airport layout. It would be so cool if we could add an image to the description such as the one below, but 

* in the water landing tutorial, I approached the mooring place and I had a speed of less than 7 kts, but I was being told this:
fgfs-screen-006
How am I suppose to face the wind if I was told to turn off the engine before hand? Should I just use the water rudder and have a lot of patience for the aircraft to slowly turn?

So these are all very minor things for you to consider, the only one I would really insist is changing that `043.4`, all else is up to you.

wlbragg:
@gilbertohasnofb great critiques. I incorporated all of them.

>In the water landing tutorial, I approached the mooring place and I had a speed of less than 7 kts, but I was being told this:
"Face into the wind"
How am I suppose to face the wind if I was told to turn off the engine before hand? Should I just use the water rudder and have a lot of patience for the aircraft to slowly turn?

It did go on to say "using momentum", so the answer is yes to your question. But, after hearing you say it, it make much better sense to make the turn facing into the wind under power and then shut off the engine. That is how one would do it in RL. So I changed it to shut off the engine after turning into the wind.


One other note about the [R] [r].
I have always seen it done like above, but I have no problem changing it if you really think it is easier to read the other way.
I did have to find a happy medium though because the choices are...


So I think what you must be looking for is
Shift + [r] and [r] , yes?

blah blah blah Shift + r (for upper case) would be OK but 
blah blah blah r (for lower case) doesn't show well.
Shift + R (for upper case) and only r (you can't use R without the "Shift" in front to indicate lower case) doesn't work.
I don't think
Shift + [R] and [r] 
is right either

wkitty42:
On 08/11/2018 06:29 PM, Gilberto Agostinho wrote:
>   * why do you use heading |043.4|? You should pick either |043| or |044|, this
>     notation you use is not standard and the minimal degree used in headings is
>     1 degree AFAIK.


if you use the F11 dialog to enter the heading value for the heading bug, you 
can easily put in tenths and hundredths of degrees... the same goes for most of 
the fields in that dialog but the heading bug field is about the only one that 
works with the c172p... at least the last times i fly long flights with it, the 
heading bug was the only one that i was able to use because the others related 
to speed and pitch are/were not linked together...

gilbertohasnofb:
@wlbragg Excellent, thank you so much for your patience and for all changes. I think this was really worth it, these tutorials are wonderful now!

> if you use the F11 dialog to enter the heading value for the heading bug, you can easily put in tenths and hundredths of degrees... the same goes for most of the fields in that dialog but the heading bug field is about the only one that works with the c172p...

@wkitty42 That's just silly from FG, then. Not even airliners use this sort of precision, this is simply unheard of in aviation. Headings are always given to a precision of 1 degree.

wkitty42:
On 08/12/2018 08:13 AM, Gilberto Agostinho wrote:
>     if you use the F11 dialog to enter the heading value for the heading bug,
>     you can easily put in tenths and hundredths of degrees... the same goes for
>     most of the fields in that dialog but the heading bug field is about the
>     only one that works with the c172p...
> 
> @wkitty42 <https://github.com/wkitty42> That's just silly from FG, then. Not 
> even airliners use this sort of precision, this is simply unheard of in 
> aviation. Headings are always given to a precision of 1 degree.


whatever, man... i call it precision mathematics and there's more than just 
airliners modeled in FG... heck, you try following an AI on a ballistic flight 
across the country, EG: KATL to KLAX... you hardly can without being able to 
specify decimal values in the controls... they'll drift right away from you 
without such precision capabilities... i've spent hours and hours and hours 
following AI craft and regularly use tenths and hundredths of degrees to try 
match their course as closely as i can while i do other things like learning how 
to sidle up close to them and loop from wing to wing over their top, down and 
under their belly...

would you also make the same statement knowing that FG fully supports eight (8) 
decimal places for decimal degrees? we're talking DDD.dddddddd... that's getting 
down to millimeter and sub-millimeter placement of objects depending on your 
location on this sphere we call Earth... if we go to degrees and decimal 
minutes, DDD MM.dddddddd, that's even finer placement control...

https://en.wikipedia.org/wiki/Decimal_degrees

anyway, whatever, man...

wkitty42:
On 08/12/2018 09:22 AM, Gilberto Agostinho wrote:
>     i call it precision mathematics and there's more than just airliners modeled
>     in FG...
> 
> My point is that this is not just airliners, it's GA, it's military aviation, 
> it's everyone. Aviation deals with 1 degree of precision when it comes to headings.


i know this... you seem to be being too realistic in this case... no one has 
said that /you/ need to use decimals... personally, i don't see any problem with 
them if they lead you directly to your destination... we're flying artificial 
craft in an artificial world all in a computer...


>     heck, you try following an AI on a ballistic flight across the country, EG:
>     KATL to KLAX...
> 
> That doesn't sound like a realistic real-world scenario to me. I suppose that if 
> you are intercepting a ballistic missile cross country in RL you would be first 
> guided roughly to the area where your paths intersect and then you'd use your 
> jet fighter radar to get a lock on the target.


nothing was said about a ballistic missiles... i said the AI craft fly a 
ballistic course... that terminology comes from the AI traffic management source 
code or at least the author's ow written words... "Ballistic course", in this 
case, means they leave their departure airport and fly directly to the arrival 
airport... they do not use waypoints or SID or STAR... they do appear to fly out 
on standard departure tracks but only until they reach a certain altitude and 
then they turn to their direct track to the destination... i think, but have not 
looked specifically, they may also use great circle calculations when the code 
generates the ""flight plan"" for the craft to follow... it isn't a real flight 
plan like we humans would create... it is more a set of coordinates and if you 
ever go and follow an AI craft, you can easily see them rocking left and right 
as they stay on their flight track...


>     i've spent hours and hours and hours following AI craft and regularly use
>     tenths and hundredths of degrees to try match their course as closely as i
>     can while i do other things like learning how to sidle up close to them and
>     loop from wing to wing over their top, down and under their belly...
> 
> But again, that's because FG is a game.


Blasphemy!


> Flying in formation in RL is done 
> visually. Nobody is expected to follow aircraft using autopilot or ATC 
> instructions with many decimal places!


you play your "game" your way but don't force your methodology on everyone else, 
ok? ;)


>     would you also make the same statement knowing that FG fully supports eight
>     (8) decimal places for decimal degrees?
> 
> I don't know what those are used for,


uhhh... look at the HUD some time... oh wait, a real c172 doesn't have a HUD... 
yeah, ok but still there's like 10 different ways to indicate longitude and 
latitude numbers... they're at least used for placing objects in scenery... some 
of us use them for exact positional data... yes, you can position your craft or 
building or trees with millimeter precision if you want...


> my point is only concerning navigation. Nagivation uses the format xxx
> ranging from 000 to 359. If that property is used with that precision for
> navigation then it is being implemented differently than real world
> navigation.

i know what your point is/was... i was trying to show you that there is more 
available than what you are seeing or using... i've shown you the pond and the 
fresh water in it... it is up to you if you want to drink it ;)

gilbertohasnofb:
> Blasphemy!

It's you who is pointing that you want to use FG in your own way which is different than how things are done in RL. But let's maybe just leave it here :smile:

wlbragg:
@wkitty42 for what it's worth, the point @gilbertohasnofb was making is that "displaying" that kind of precision in that instance was wrong and I whole heartedly agree with him.
Now if I needed to use that precision in the code to achieve a desired outcome. then that would be one thing. But in the context of a heading description in a tutorial, it was wrong and not to spec. It was the correct call to ask for it to be changed.

>Blasphemy!

I laugh every time I see the word "game" used anywhere near the FlightGear crowd. It's like triggering a Bass to bite.

wkitty42:
On 08/12/2018 11:58 AM, Gilberto Agostinho wrote:
>     Blasphemy!
> 
> It's you who is pointing that you want to use FG in your own way which is 
> different than how things are done in RL.

no sir, that is incorrect... all i did was to point out that while decimal 
precision may not be used in real life for headings and other positional data, 
the precision IS available in this simulator... then i gave a couple of examples 
of how i use it at times... nothing more...

the blasphemy comment above was targeting specific statement that FG is a game, 
which it is not...


> But let's maybe just leave it here 😄


yes, let us do that ;)

gilbertohasnofb:
It's a simulator if you use it to simulate RL procedures but it's a game if you fly a dragon in a heading with seventeen decimal points of precision 😜

wkitty42:
On 08/12/2018 12:47 PM, wlbragg wrote:
> @wkitty42 <https://github.com/wkitty42> for what it's worth, the point 
> @gilbertohasnofb <https://github.com/gilbertohasnofb> was making is that 
> "displaying" that kind of precision in that instance was wrong and I whole 
> heartedly agree with him.


agreed... i didn't know that it was being displayed to the user...


> Now if I needed to use that precision in the code to achieve a desired outcome. 
> then that would be one thing. But in the context of a heading description in a 
> tutorial, it was wrong and not to spec. It was the correct call to ask for it to 
> be changed.


agreed there, too...


>     Blasphemy!
> 
> I laugh every time I see the word "game" used anywhere near the FlightGear 
> crowd. It's like triggering a Bass to bite.


i know! it is fun ain't it?! hahaha...

PR #1166, opened 2018-08-09
Add cowl plugs

### LINKED ISSUE #881: Add cowl plugs and/or engine covers
~~What about adding cabin and engine covers to the aircraft? See:~~

We could add cowl plugs to our aircraft as part of securing it:

Alternatively we could also have the option of a full engine cover, but that might be completely unecessary:

Previously I had suggested cabin covers but I don't think those are typically used in the 172, the photos below are from Cessnas 150/152. Let's just scratch this idea.

### LINKED ISSUE #1168: Sun Visor Mapping
image

Sorry - still not quite right! 😀

wlbragg:
Fixes #881 
Fixes #1168


gilbertohasnofb:
@wlbragg Done. Please test it. I've done the following:

* new texture
* implement logic. Test both by clicking on the cowl as well as by using the aircraft dialog. Test that it behaves as expected when securing aircraft is ON or OFF
* cowl plugs state should be saved between flights
* autostart removes the cowl plugs
* forgetting cowl plugs will cause engine to cough

This is ready to be merged. Once this is done I will add the cowl plugs to the tutorials and checklists.

gilbertohasnofb:
@wlbragg I don't want to push you or anything but as soon as this is merged I can work on the splash screens.

wlbragg:
One thing I already noticed is when you leave the plugs in and manually start the engine, it coughs (great effect :smile:) but if I then open the menu to try to remove them with the checkbox, the checkbox is cycling between grayed out and active.

Also there is a tiny bit of UV mapping bug work I see and I want to correct that.

I'll do that right now locally and have it standing by to merge into the PR when you are done with any other changes such as the menu issue I mentioned above.

wlbragg:
Oh, just dawned on me, You could never do that with the engine running!

wlbragg:
@gilberto, it's the coughing that makes the menu choice cycle between active and not. I guess that is just a side effect and something we live with?

gilbertohasnofb:
> Also there is a tiny bit of UV mapping bug work I see and I want to correct that.

Cool, let me know if the textures need any change. I hope you will find them satisfactory, I thin they are looking good enough.

> Oh, just dawned on me, You could never do that with the engine running!

Exactly! That's the only securing item that I used the engine running property as a condition for toggling 😄 

> it's the coughing that makes the menu choice cycle between active and not. I guess that is just a side effect and something we live with?

Yeah, the issue is that the coughing function basically disables the the engine for some short time then turns it on again. It's a side effect of that simple implementation and I don't think we can easily solve that other than rethinking the whole coughing mechanism. Personally I think we should just live with it.

gilbertohasnofb:
@wlbragg One idea: is there a property that can be used to know if the nose cone is spinning or not?

gilbertohasnofb:
> is there a property that can be used to know if the nose cone is spinning or not?

Derp, that's called rpm. I am pushing a commit right now, please pull before proceeding. I also added this condition to the repair button which was also flashing when the engine was coughing.

wlbragg:
>the centre point in between the two cowl plugs (i.e. the point just behind the spinner)

-1.72701
0.00127
-0.0519

>the centre point of the front wheel

-1.11492
0.0
-1.18892

gilbertohasnofb:
Thanks!

PR #1176, opened 2018-08-12
Slight modifications and safeguards to exiting amphibious tutorials

wlbragg:
@gilbertohasnofb forgive me but there were too many small issues for me to let it go. The modifications to the landing and takeoff tutorials are as follows. There is also a couple related to the gear panel toggle switch.

Gear panel dim toggle switch needed a default value in -set.xml or it glitch at night when first used.
It also needed to be added to saved aircraft-data as well as nasal save switch state.

The landing and takeoff tutorial check point for distance to make the 180 was really buggy if your line was off course at all from the original. It was using only lat or lon and that wasn't perfectly perpendicular so it was buggy. I added distance to either mooring target or runway target so it is pretty much foolproof now.
Also there was some timing issues with the check gear advisory and that is much better now and makes sense as it comes on about the time you need to make changes to your gear.
For consistency I also added the Shift+[g] and [g] to the gear raising messages.

I added a night tutorial, it is the two day tutorials combined into one with a land runway touch and go in the middle. It is a really fun challenge. It also adds the nighttime dimension and shows off all the lighting, interior and exterior. It is really cool, you see at various points lots of AI taxiing around and taking off.

I think that is all.

Do you want me to also create an issue for this? I really think it may pass without any modifications.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -304,6 +304,7 @@ var switches_save_state = func {
         setprop("/controls/lighting/dome-white-norm", 1.0);
         setprop("/controls/lighting/dome-norm", 0.0);
         setprop("/controls/lighting/gps-norm", 0.0);
+        setprop("/controls/lighting/gearled", 0);
         setprop("/controls/gear/water-rudder", 0);
         setprop("/controls/gear/water-rudder-down", 0);
         setprop("/sim/model/c172p/brake-parking", 1);
--- a/Tutorials/amphibious-landing.xml
+++ b/Tutorials/amphibious-landing.xml
@@ -215,7 +215,7 @@ Pay attention to the heading bug on landing it will approximate the mooring posi
             <pitch-offset-deg>-12</pitch-offset-deg>
             <roll-offset-deg>0.0</roll-offset-deg>
             <x-offset-m>-0.21</x-offset-m>
-            <y-offset-m>0.19</y-offset-m>
+            <y-offset-m>0.23</y-offset-m>
             <field-of-view>73.6</field-of-view>
         </view>
     </step>
@@ -263,7 +263,7 @@ Pay attention to the heading bug on landing it will approximate the mooring posi
             </condition>
         </error>
         <error>
-            <message>Your gear must be down to take off.</message>
+            <message>Your gear must be down to take off Shift+[g].</message>
             <condition>
                 <equals>
                     <property>/controls/gear/gear-down</property>
@@ -345,7 +345,7 @@ Pay attention to the heading bug on landing it will approximate the mooring posi
             </condition>
         </error>
         <error>
-            <message>Raise your landing gear now.</message>
+            <message>Raise your landing gear now [g].</message>
             <condition>
                 <equals>
                     <property>/controls/gear/gear-down</property>
@@ -459,10 +459,16 @@ Pay attention to the heading bug on landing it will approximate the mooring posi
 
         <exit>
             <condition>
-                <greater-than>
-                    <property>/position/longitude-deg</property>
-                    <value>-157.9550502</value>
-                </greater-than>
+                <or>
+                    <greater-than>
+                        <property>/position/longitude-deg</property>
+                        <value>-157.9550502</value>
+                    </greater-than>
+                    <greater-than>
+                        <property>/sim/tutorials/targets/mooring</property>
+                        <value>5240</value>
+                    <greater-than>
+                </or>
             </condition>
         </exit>
     </step>
@@ -530,6 +536,10 @@ Pay attention to the heading bug on landing it will approximate the mooring posi
                         <property>/controls/flight/flaps</property>
                         <value>0</value>
                     </greater-than>
+                    <less-than>
+                       <property>/velocities/groundspeed-kt</property>
+                       <value>70</value>
+                    </less-than>
                 </and>
             </condition>
         </exit>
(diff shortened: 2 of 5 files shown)


gilbertohasnofb:
@wlbragg no worries and no problems, but if you won't mind it I will check this tomorrow. I am also trying to fix some stuff with the tutorials (though thankfully we are dealing with different files).

wlbragg:
So much for "passing without any modifications". I missed 4 closing tags, all three tutorials were breaking the launcher because of it. It's fixed now and I am testing all three tutorials, trying to break them.

gilbertohasnofb:
@wlbragg I had a quick look and I just have a couple of comments:

* on all tutorials, I would set the master off and the aircraft to cold (all switches off) so that it's up to the pilot to turn on the aircraft when commanded. It feels weird to have the master on but the engine off.
* in the night tutorial you could recommend the pilot to use the flashlight with the shortcut `f` before anything. That will help them start the engine.
* by the way, is it realistic at all to take off or land in the water at night? You can't see a thing and I thought that this would not be allowed. Runways have night illumination, the sea doesn't. I remember that once Thorsten had implemented a moonlight feature but I don't seem to see in the menu. If that's still available, you could crank that property a bit. edit: I've read that RL pilots only dare this in specific situations but that it is _very_ dangerous. In Alaska some bush pilots do it, but apparently only in the very right conditions (e.g. moonlight)
* I had another issue with the night tutorial. I was waiting and waiting for the 180 command and it never came and I had a controlled flight into terrain:

fgfs-screen-002

Note how the heading and altitude seem to be correct. This is the final result (changed the time to daytime for the shot, obviously):

fgfs-screen-004

wlbragg:
>on all tutorials, I would set the master off and the aircraft to cold (all switches off) so that it's up to the pilot to turn on the aircraft when commanded. It feels weird to have the master on but the engine off.

Even the one lined up on the runway? Don't autostart?

>by the way, is it realistic at all to take off or land in the water at night?
...
>I've read that RL pilots only dare this in specific situations but that it is very dangerous. In Alaska some bush pilots do it, but apparently only in the very right conditions (e.g. moonlight)

Did you just answer your own question?

>Is it legal to land a seaplane at night on the water?               
Yes
While many part 135 operators are specifically prohibited from landing a seaplane at night on a body of water lit specifically for that purpose, Part 91 operators are not specially prohibited from landing at night on an unlit body of water.
SPA must stress that landing on a body of water at night should only be done with very specific knowledge of a body of water and the obstructions that are present including boat and other traffic that may be present. This is an advanced technique and should be approached with great caution if a pilot chooses to do so.
There are a few facilities in the world that do accommodate night seaplane landings such as the landing lane on Lake Hood in Anchorage, Alaska that is specifically lit to facilitate night seaplane landings.

Not only is is legal but if you do a search about it is sound like it is not that uncommon, yet it does sound in some situations risky and requires a very strong knowledge of the body of water your going to land on.

One thing I got out of this discussion is that it would be nice to have a lot of lights along the shore line there at PHNL. Maybe for the following release I could cast some light on the water from the lights in the area like is done on the Shuttle night landing airport. I think that would be really cool!

>I had another issue with the night tutorial. I was waiting and waiting for the 180 command and it never came and I had a controlled flight into terrain:

When did this happen, was it with the very latest commits? I ask because that is what I specifically addressed with the following addition.
~~~
           <condition>
                <less-than>
                    <property>/sim/tutorials/targets/mooring/distance-m</property>
                    <value>5650</value>
                </less-than>
            </condition>
~~~
If this is the only condition required to release to the next step (the turn) then it is a fault of something else. I haven't had what you had happen since I added this condition and I have tested it pretty thoroughly, trying to make it break by being of course and such.

gilbertohasnofb:
> Even the one lined up on the runway? Don't autostart?

I noticed that on the night water take off, the master was definitely set to on when the tutorial started. And I meant that you could check if this was the case with the other two, I am not sure by heart whether that was the case. And I don't understand what you mean by 'Don't autostart?'.

> Did you just answer your own question?

I did, hence my `edit:` 😉 

>  Maybe for the following release I could cast some light on the water from the lights in the area like is done on the Shuttle night landing airport. I think that would be really cool!

Yeah, that would be cool!

> When did this happen, was it with the very latest commits?

Yeah, I am sure that it was with the latest commits. Maybe I just got unlucky, these conditionals can be a bit tricky. Maybe I was too low below 900 ft for a long time. I think we possibly just have to live with what we have now, as it seems to work most of the time.

wlbragg:
>Yeah, I am sure that it was with the latest commits. Maybe I just got unlucky, these conditionals can be a bit tricky. Maybe I was too low below 900 ft for a long time. I think we possibly just have to live with what we have now, as it seems to work most of the time.

On some of these that could be buggy, I made the conditional to go to the next step a standalone condition. In this case, looking at the message in your screen shot, you were at a conditional that only required the distance from the mooring or runway point to be met, nothing else. So it is something in the tutorial system that is buggy.

>the master was definitely set to on when the tutorial started.

I understand this and agree with you.

>And I don't understand what you mean by 'Don't autostart?'.

What I meant is in the daytime "landing" tutorial you start on the land runway with the engine already started. Do you think I should also start it cold and dark but still on the runway?
Because you refer to "all" the tutorials.

wlbragg:
>Would it be too much work to add it to the daytime tutorial too? 

No
 
Is there a reason you don't want to add it to that tutorial?

Not really, I had in my mind that you could take that to the extreme and add it to every step that if you fail to follow the instructions you could conceivably hit the hillside. Then it didn't seam reasonable anymore so I opted to only use it in that critical step in the nighttime version when you couldn't see what was going on.

In trying to decide how detailed and how failsafe to make any of these tutorials. We could add all kinds of tests to eliminate all possible deviations from the steps required to complete the tutorial. we don't do that. 
If you think I should add it to the climb step in daytime, it's not any problem to do so.

gilbertohasnofb:
Yeah, I agree that there is a limit of how much control over the conditions you want. But if you can do apply that to the daytime tutorial and we can merge this, they are all looking extremely good!

wlbragg:
@gilbertohasnofb this should be done then.

PR #1177, opened 2018-08-12
Checklist and tutorials

gilbertohasnofb:
@wlbragg this PR fixes all the stuff that was left from @legoboyvdlp's last PR. Please go through both preflight and securing tutorials. Then also go through both preflight and securing checklists. Should be ready to merge.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -3880,9 +3880,12 @@
                             <property>/fdm/jsbsim/gear/unit[23]/WOW</property>
                             <property>/fdm/jsbsim/hydro/active-norm</property>
                         </or>
-                        <not>
-                            <property>/sim/current-view/internal</property>
-                        </not>
+                        <or>
+                            <not>
+                                <property>/sim/current-view/internal</property>
+                            </not>
+                            <property>/nasal/tutorial/loaded</property>
+                        </or>
                     </and>
                 </condition>
                 <command>nasal</command>
@@ -3941,9 +3944,12 @@
                             <property>/fdm/jsbsim/gear/unit[23]/WOW</property>
                             <property>/fdm/jsbsim/hydro/active-norm</property>
                         </or>
-                        <not>
-                            <property>/sim/current-view/internal</property>
-                        </not>
+                        <or>
+                            <not>
+                                <property>/sim/current-view/internal</property>
+                            </not>
+                            <property>/nasal/tutorial/loaded</property>
+                        </or>
                     </and>
                 </condition>
                 <command>set-tooltip</command>
@@ -10730,15 +10736,15 @@
             <repeatable>false</repeatable>
             <binding>
                 <condition>
-                    <or>
-                        <and>
-                            <property>/sim/model/c172p/securing/pitot-cover-removable</property>
+                    <and>
+                        <property>/sim/model/c172p/securing/pitot-cover-removable</property>
+                        <or>
                             <not>
                                 <property>/sim/current-view/internal</property>
                             </not>
-                        </and>
-                        <property>/nasal/tutorial/loaded</property>
-                    </or>
+                            <property>/nasal/tutorial/loaded</property>
+                        </or>
+                    </and>
                 </condition>
                 <command>property-assign</command>
                 <property alias="/params/securing/pitot-cover"/>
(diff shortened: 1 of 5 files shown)


wlbragg:
I'll do this tomorrow.

gilbertohasnofb:
@wlbragg Ok, I solved the merge conflicts from lego's PR, this is again ready to be merged

wlbragg:
Will get to this later this afternoon

wlbragg:
Doing through this now. So far...

Preflight checklist

Water contamination check seems to still point to the courtesy lights?
Aileron, elevator, rudder and oil checks should all have green and yellow completion text indicators as we can perform those checks same as other items that are listed that way that have checks.
Would it be standard to check the exterior lights actually work with them turned on during the inspection?

Securing checklist
All looked OK

Preflight tutorial

Same as the checklist, the fuel sample location needs to be moved and the fuel check pops the menu up automatically, shouldn't you have to click on the fuel port location to open the menu?
Aileron, elevator, rudder should all have to be moved before the tutorial moves on.

Securing checklist
All looked OK

gilbertohasnofb:
@wlbragg I need a fix from you. Our hotspot for checking the fuel is a bit too unforgiving, and it's also only in the front part of the wing. I would suggest we make a large hotspot under the whole inner part of the wing. My suggestion is in blue:

image

Could you please adjust this for both wings? You can do it in the other PR in which you handling the 3D stuff.

gilbertohasnofb:
Ok, so I fixed everything and I also reordered the checklist and the preflight tutorial as to match the POH. I think it's now good to go :smile: 

And just one more thing, could you please make a small test for me, unrelated to our aircraft? Load FG and enable advanced weather and apply some weather theme. Now load a tutorial such as the preflight: do you also have now basic weather? If so then I think there is a bug in the weather system and we should report it.

wlbragg:
@gilbertohasnofb 
We use
~~~
<set>
            <property>/environment/weather-scenario</property>
            <value>Fair weather</value>
        </set>
~~~
in the tutorials, maybe that switches it. Although the GUI weather dialog doesn't show basic, it is still switched to detailed in my case because I was on detailed to start. The only indicator to me that it is basic is those high pillow clouds in rows, are those only in basic weather?

legoboyvdlp:
The flat 2D clouds are only basic weather.

wlbragg:
Can you tell the difference, if any, in the weather system used between these images?
cld1
cld2

gilbertohasnofb:
@wlbragg These two are advanced weather. But on my PC I get basic weather when I load the tutorial even if I already had AW running before.

> . Although the GUI weather dialog doesn't show basic, it is still switched to detailed in my case because I was on detailed to start. The only indicator to me that it is basic is those high pillow clouds in rows, are those only in basic weather?

That GUI is useless, you have to manually load AW or set it in your launcher. That option just tells you what will happen once you hit Apply, not what you currently have. @onox and I tried arguing about it in the past but frankly it's just hitting our heads against the wall. But your screenshots show quite clearly that you have AW loaded in the tutorials, I just don't understand how. @legoboyvdlp could you try to load them and see what you see?

gilbertohasnofb:
@wlbragg regardless of weather, this is ready to be merged

wlbragg:
>you have to manually load AW or set it in your launcher.

Yes, I have mine set in the launcher to AW. What is weird is I too have seen instances where it appeared to have loaded BW when I switched to a tutorial. That is if the only time you see those "rows" of high level clouds is BW.

gilbertohasnofb:
> Yes, I have mine set in the launcher to AW. What is weird is I too have seen instances where it appeared to have loaded BW when I switched to a tutorial. That is if the only time you see those "rows" of high level clouds is BW.

That's exactly my case, I have it set in the default launcher and I have those horrible rows of high level clouds when I load a tutorial such as this preflight one. I just don't get why we need to have two weather systems with one being factually worse than the other in every single level...

wlbragg:
I think maybe when we have some extra time we need to see if it is only a certain tutorial or a few that do this or if it is truly random.

gilbertohasnofb:
Yeah it's not the most important thing right now but it's something that has always bugged me. Later if we have time we can see if we can reproduce this. Otherwise I am super late with all those splash screens, I really did not have the chance to touch them yet... hopefully tomorrow

legoboyvdlp:
@gilbertohasnofb what tutorial did you see particularly that reverts to basic weather?

I'll test it tonight.

gilbertohasnofb:
It was the preflight tutorial. The weather looked quite bad to me every time I loaded it, it was clearly not AW with patterns of high clouds typical of BW. I will check later if any other tutorial has the same issue on my machine.

PR #1179, opened 2018-08-13
Adjust VSI and RPM UV-maps

### LINKED ISSUE #1178: Better UV-map for tachometer and VSI
Since we moved the gauges a bit backwards, both the VSI and the tachometer are having their textures too close to the borders, see:

image

image

@wlbragg would it be possible to slightly adjust the UV-map for those two gauges? Just proportionally resize the UV-map down keeping the correct central point (where the needles are pivoted) so that we don't need to change any xml files. We are talking about a very small resizing, ok? Just so those digits do not get cut.

wlbragg:
Fixes #1178


gilbertohasnofb:
@wlbragg Much better, thank you so much!

PR #1182, opened 2018-08-15
Miscellaneous issues

### LINKED ISSUE #724: Preflighting the C172 with outside view
So I took this for a test flight and preflight today. A few notes:
- The fuel tank quantity and sampling dialogs are neat. However the quantity could be also shown in liters in addition to gallons and pounds.
- Fuel sump clickpoint is a bit hard to find, maybe combine all fuel related stuff to one dialog and tie it to a big clickspot on the fuel tank itself? So that it shows quantity and lets you take the sample as well.
-  Is there a way to disable clicspots (and tooltips) from within the cockpit when you are in an outside view? There is a lot of stuff that is clickable and shows tooltips in the cockpit that gets in the way when you try to do a preflight check in an outside view. Same with outside clickspots when you are in the cockpi. I often accidentally clicked the engine oil clickspot when I used the GNS196.

### LINKED ISSUE #1180: Throttle lever does not move all the way in
This is the full throttle position with the new lever:

image

Throttle lever does not move all the way in. At full throttle the black part should be touching the back side, just like the mixture.

wlbragg:
Replaces misc-issues PR due to messy commit history.

Fixes #1180
Fixes #724


wlbragg:
This PR still does not contain the last know bug fixes needed which are
The propeller not rotating with the S keystroke.
Cowl plugs not disappearing with cranking only.
S before autostart displaying "Use primer ...".

gilbertohasnofb:
> S before autostart displaying "Use primer ...".

What bug is that?

> Cowl plugs not disappearing with cranking only.

That would need the `<logic>` implementation I mentioned. Do you want to take care of that yourself?

> The propeller not rotating with the S keystroke.

That's a big one and I think this was introduced recently.

wlbragg:
>Cowl plugs not disappearing with cranking only.
>>That would need the <logic> implementation I mentioned. Do you want to take care of that yourself?

Yes, I'll do it because it also relates to the one below.

>The propeller not rotating with the S keystroke.
>>That's a big one and I think this was introduced recently.

I'll do this one and the logic implementation above. This was introduced recently. It is because there is no code to remove the visibility of the cowlplug version of the prop when the engine cranks. It will work out correctly when we add the logic for the cranking.



>S before autostart displaying "Use primer ...".
>>What bug is that?

Start the sim and immediately try to use S to crank.

wlbragg:
@gilbertohasnofb I think I got it. I wasn't able to use a logic statement to do this. I ended up adding a switch in the c172p.nas "/engines/active-engine/cranking" listener like we do for the other objects that get set to invisible when the engine cranks. Because of all the propeller, spinner and propeller.fast object logic and animations and the damage logic, it was a bit bigger problem than first expected. But I think I got it all sorted out. Please test carefully. I had to move around some of the propeller animation in order to group them together to make it easier for future developers to read.

dany93:
Tested:
- "Use primer" is no longer displayed with "S" at sim start (with battery OFF),
- the cowl plugs and strap do disappear, the propeller and cone do spin at each starter activation.

If I forgot nothing, I'd say it is OK.

wlbragg:
@dany93  thanks for testing. I forgot to mention the cause of the "Use primer" with "S", it was happening when the batt was off which is why if felt so out of place. The logic that controlled when that prompt was displayed didn't include a conditional for the batt being off, so I just added it.

gilbertohasnofb:
@wlbragg I am just back home and I will start testing this. Is there anything left to be done still?

gilbertohasnofb:
@wlbragg This PR looks really good, I think we can close the other one, it seems everything is here. I think this is nearly ready to be merged, I only found two small issues:

* The baggage door hotspot is not properly posiitoned:

image
image

* the yellow chocks are floating above the ground a bit. I forgot to check the wooden ones but I suspect they would also be:

image

All else looks good. The primer message is displayed only with BAT on, you can't use the starter without BAT, the cowl plug logic is perfect. I just wishes we could interpolate the propeller to the correct place, but I understand that was not possible and it is not a big deal. But would look so much prettier :smile:

gilbertohasnofb:
> but for now I'm afraid it is all we get.

Totally understandable. Once we merge this I will open an issue for that so that we can think about it in the future. Maybe a solution would be to not use the rpm for the spinner rotation but rather use the rpm to generate a new property `propeller-angle` (modded at 360) which is then used for the animation and would give us the precise position at any given point.

gilbertohasnofb:
@wlbragg I've noticed another small bug: the propeller disk is having a very slow procession movement. To reproduce, fly level and then go to the external view and look at the aircraft from the front, something like this:

image

Now press a until the simulation is running at 8x. Now look at the disk: notice how it 'wiggles' a bit.

gilbertohasnofb:
@wlbragg and we still have that issue with the beacon light being inside of a 'square', would you have any ideas how to solve it?

wlbragg:
@gilbertohasnofb I adjusted the chocks and wooden chocks, adjusted the baggage door hot spot and adjusted the beacon light as much as I could. It's a little better I think. It is good as I can get it.
As far as the fastprop wiggling, I don't think I can do anything about that. You can't tell at normal speed and I really wouldn't know how to go about making that minute of an adjustment.

gilbertohasnofb:
@wlbragg Yeah, I understand your point about the propeller, not much of a big deal. I just suspected that the pivot point for its rotation animation was correctly centred and that would have been the problem. But agreed that's not the highest priority in the world. 

Is there anything else we need to do before merging this one?

wlbragg:
@gilbertohasnofb I think were done here.

gilbertohasnofb:
@wlbragg the baggage door has not been fixed. Perhaps you have forgotten to adjusted it from other angles, right now it is matching only the back camera, but the other angles are still not matching. This is the back view:

image

This is from other angles:

image
image
image

wlbragg:
@gilbertohasnofb something weird is going on here. The first time I tried to adjust I wasn't showing the same offsets you did. I double checked it in the sim and it was lined up in my version.
Now I just booted it up and it is wrong. I don't get it. I will try it one more time.

Remind me again, in this case why cant we use the entire interior and exterior door as the hotspot for this animation code block? Wouldn't that choice be locilized only to the code block it is used in?

gilbertohasnofb:
> Remind me again, in this case why cant we use the entire interior and exterior door as the hotspot for this animation code block? Wouldn't that choice be locilized only to the code block it is used in?

We don't want to pollute the Ctrl+C yellow contours. When the door is closed, only the handle should open it. But it's confusing as hell if the whole door changes your mouse when hoovering and the yellow contours shows the whole door, that feels like a bug. Also, this way, I can control what tool tip to show, one for opening, one for closing. We are almost there, would you mind trying to match it just a bit better?

gilbertohasnofb:
Just tested it, it's perfect now. Thanks a lot!

PR #1189, opened 2018-08-23
Create a startup and saved state system

### LINKED ISSUE #789: Create startup states and save mission state

wlbragg:
Fixes #789 
Fixes #932 

~~This is done and ready to test. I am happy with it for now.~~
There is possibly some work still to be done on the state system launcher and c++ side. When and if that come to fruition I want to revisit this.
In case anyone is wondering why I closed the other PR for this and started a new one. I thought it best to start from the latest master as this was a pretty old base.


wlbragg:
Change my mind once again. I have some changes to make to this tonight. It should be ready by tomorrow if all goes well.

wlbragg:
@gilbertohasnofb  @onox I am really happy with both the startup state system and save/resume system. I think you will find them both fairly robust and really useful.

To test the startup state system simply go through the Summary State selector listings and try them one at a time. Pay attention to the description and follow what it says. After you test them all please feel free to critique.
Be aware, I think after this next release we need to have a brainstorming session to decide what all startup states we really want and if this is working the way we want it to with the existing default systems already implemented internally, or if we need to rethink the "Save switch setting on exit".
If you don't use the Summary State selector then the Location - Runway, Parking works like it always has with the exception that if you select Runway, and tick On approach you will automatically be put into an approach state with the aircraft engine running, flaps down, brake off at whatever distance you selected in the On approach input field. That is something else we can discuss later on, exactly what state we want the aircraft in for this "automatic "on approach" state. Such as flaps, no flaps, etc. Or if we want to also offer one or more custom "on approach" states selectable from the Summary State selector. We can set up as many different configurations as we decide we need for any conceivable state, not just on approach but others as well.

The more standard states I already set up are...

Cruise
Parked, cold and dark
Parked, secured
Ready for Take-off
Ready for Take-off, lights on

I also included a few exotic states that are more of a combination of variant, location and aircraft state. These really suggest what could be done, limited only by our imagination.
Those are...

Alaska Mountain Top, bush tires
Columbia Glacier, skis
Lake Tahoe, floats
Kachemak Bay, amphibious

 There was much discussion on the dev list about the intent of this system and how the c182s requested certain accommodations. I can't say I agree with everything they did and everything that was suggested as how it should be done. But, for the most part I ended up doing things similarly. If you compare their nasal state manager with ours you immediately see ours is infinitely easier to read and extend. It is simple by comparison. I didn't see a need to complicate it. I also am still hoping that some of the feedback I provided James will sink in and maybe eventually allow us to move the property settings out of the nasal state manager and into the overlay file. It would make it so much more modular and it make much more sense to have all the settings for one state together in one file VS having many state settings commingled together in one huge file. I did a piss poor job communicating the concept. I still think it could be achieved, but I don't know that, as I don't know the capabilities or order of how and when you can read and then set properties assigned in the overlay file. For documentation purposes, suffice it to say, as  it stands now, persistent aircraft data loads and sets properties after the state overlay. As far as I can tell that is the only thing keeping it from being achievable.

I am extremely happy with the Save and Restore feature. It is a stand alone choice in the GUI and is capable of saving up to 5 different save and restore points. I tested two slots. What I did was start the aircraft at an airport location that also had a mooring attached to it. You can use PHNL. After I used the "Go Now" button to go to the water mooring location I moored the aircraft and manually set the switches to a cold and dark state. Then I saved that session. Then I restarted the sim using the Summary State of Alaska Mountain Top and after it loaded I shut the engine off. It starts engine running and I don't allow you to save a session with the engine running. We could extend it to allow for that, but for now I didn't want to complicate things and I think we need to discuss capabilities we want. Then I saved that session in slot 2. Then I both switch back and forth using the Save/Restore feature during that session and also exited the sim, started it back up and tried the save/restore again a few times. It all worked flawlessly. I'm really happy with it and we could make it even better if we waned to.
Right now it doesn't keep track of thinks like weather settings or time, but i suppose you could extend it to allow for such things if we wanted to.

Anyway, Please test it and if it meets your approval as it is, I would really like to get this in this release. I'm sure we could spend some time to polish both the available states and the save/restore feature, but they are already really nice new features for this aircraft.

wlbragg:
I'm still discussing this on the dev list and think I am narrowing down my misunderstandings and concerns. If I am understanding this correctly now, then to get the most effective use out of this we need to redo our -set file persistent data choices. We can't use the overlay as intended and have persistent data overwrite the state overlay. I don't want to do any of that until after this next release cycle is over. I'm still OK with implementing this as is for now as it is working fine and not touching much of the other code we have. In other words I feel it is safe and a good place to start in identifying what will need to be changed eventually.

wlbragg:
While I am at it, I should restate that the c182s implementation is closer to what we need to do than i realized. They are using the nasal part of the state manager section to set all their state related items though and I don't think you should have to do that. But they may be doing it because of cached data we have no control over.

onox:
@wlbragg I don't have the Qt launcher + I'm quite busy in RL these weeks, so can't really test it.

wlbragg:
@onox, I am pulling this one down shortly. The "Startup state" portion is just too buggy and also is going to change drastically after we discuss some issues and it gets re-factored. So I am going to put this hold until after this release.

PR #1191, opened 2018-08-23
Correct Astro Tech Digital Clock name in aircraft dialog

### LINKED ISSUE #1190: Digital Clock: wrong name in gui
Should be _Astro_ Tech, not _Aero_ Tech!

legoboyvdlp:
Fixes #1190

### DIFF
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -394,7 +394,7 @@
             
             <checkbox>
                 <halign>left</halign>
-                <label>Show Aero Tech LC2 Digital Clock</label>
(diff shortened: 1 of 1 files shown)


legoboyvdlp:
Since its so minor, I'll just go ahead and merge. I probably could just have commited to master for something like this but thought it better not.

gilbertohasnofb:
> Since its so minor, I'll just go ahead and merge. 

That's all right in this case, but do not get used to this 😉 

> I probably could just have commited to master for something like this but thought it better not.

You should always create PR, even for very minor things. So please never commit directly to the master. This keeps our history cleaner and more understandable for 'future generations'

PR #1193, opened 2018-08-29
Save and resume option and misc bug fixes

### LINKED ISSUE #789: Create startup states and save mission state

### LINKED ISSUE #906: Improvements to the tutorials and checklists
Work on several improvements to the tutorials and checklists:

- all the securing elements can be forced even if "securing aircraft" is false, which is a bit messy (e.g. if your "securing aircraft" is false, you can still add a pitot tube cover via the checklist but then can't remove it by clicking on it. similar stuff for tie-downs, chocks, etc). So better is to deactivated the [>] buttons if that property is false. For instance, this can be accomplished by using a `<condition>` block for that, e.g.:

```
            <binding>
                <property>sim/model/c172p/securing/pitot-cover-visible</property>
                <value type="bool">true</value>
            </binding>
```

becomes:

```
            <binding>
                <condition>
                    <property>/sim/model/c172p/securing/allow-securing-aircraft</property>
                </condition>
                <property>sim/model/c172p/securing/pitot-cover-visible</property>
                <value type="bool">true</value>
            </binding>
```

(It's a pity that `<enable>` does not seem to work with a checklist item, as it not only disables a button but as well make it look disabled.)

- Also, in the tutorial "Checklist: Securing Airplane", you can for some reason add a pitot tube cover even if the securing aircraft is `false`, which is not possible outside the tutorial. This seems potentially buggy so forcing that option to `true` is really important in all these

- I tried loading the tutorial "Securing Aircraft" and it didn't take me to Barcelona, is that intended? I think all tutorials should take place in Barcelona, except for the "Checklist: <something>" types.

- We now have static objects, including ladders, which can be activated in the Ground Equipment dialog. This ladder can be climber by the walker, making it easier to click on the fuel tank caps to add/remove fuel. This could be added to the pre-flight check tutorial.

wlbragg:
Fixes #906
Fixes #789

This is not really attached to any open issues I don't think. But it does correct a few bugs, a couple of them major bugs, and it cleans up some old commented out code.

The following is what is being done in this PR...

1) Save and resume option in GUI, up to 5 saved sessions. Saves position, variant selections and all interior switch and lighting selections. It basically is taking a snapshot of the current state of the aircraft.
2) Fixes a newly discovered bug in the mooring relocation code.
3) Reorganizes a couple saved data properties.
4) Adds a version check per warning in the log.
5) removes a few blocks and lines of old, unused and commented out code.
6) Corrects a glaring bug in the newly added "Variant" logic .
7) Adds a very simple automatic state logic to the QT launcher, thus making the QT Location selections work correctly. See the description of what it does below.
8) Removed the necessity of having to check "Allow securing aircraft".


onox:
(Removed the milestone because it's a PR, not an issue)

wlbragg:
@legoboyvdlp @onox this is finished and ready to merge after one more final testing of what I did in the last commit.

Support for...
adjusting cockpit and exterior lighting to match conditions
mixture at altitudes
altimeter calibration
heading offset
and carb heat.

I also removed some superfluous properties that do not exist out of autostart and that I copied into the state support. I don't know where they originally came from.
setprop("/controls/engines/engine[1]/primer", 3);
setprop("/controls/engines/engine[1]/primer-lever", 0);
There is no such properties, both engines use engine[0] primer and primer-lever.

wlbragg:
OK, it's really done this time .I wasn't satisfied with the approach configuration you start out with, these aren't perfect either, but starting the the aircraft the way we have to leaves it impossible to get a perfect approach configuration.
Most importantly I fixed the "nose five bug".

PR #1195, opened 2018-09-01
Add integral lighting to panel instruments for default renderer

### LINKED ISSUE #1192: Add non-ALS lightmaps
Add non-ALS lightmaps for non-ALS users doing night flights.

The current lightmaps require ALS. The aircraft cannot do night flights presently without ALS. User "Megaf" has been complaining about this on Discord.

It is possible to add non-ALS lightmaps which require only the Model shader using model-combined-deferred as seen on the IDG-A32X - would it be possible to add these, enabled only when ALS is off?

wlbragg:
Fixes #1192 

@onox 
This is the bare minimum so far. Shader set to lowest setting does not work, you have to have ALS off.
Do we also want this as a fallback for the shader setting?

I can try to integrate some ambient radiant light later on this evening too if we wan't. Without it it looks a bit dark as you can't see anything but the gauges.

### DIFF
--- a/Models/Effects/interior/c172p-interior-radiance.eff
+++ b/Models/Effects/interior/c172p-interior-radiance.eff
@@ -20,10 +20,10 @@
         <light-radius type="float">13</light-radius>
         <irradiance-map-type type="int">2</irradiance-map-type>
         <irradiance-map-strength type="float">.2</irradiance-map-strength>
-        <residual-ambience-r type="float"><use>/sim/model/c172p/lighting/rgb-ra-r-factor</use></residual-ambience-r>
+        <!--residual-ambience-r type="float"><use>/sim/model/c172p/lighting/rgb-ra-r-factor</use></residual-ambience-r>
         <residual-ambience-g type="float"><use>/sim/model/c172p/lighting/rgb-ra-g-factor</use></residual-ambience-g>
         <residual-ambience-b type="float"><use>/sim/model/c172p/lighting/rgb-ra-b-factor</use></residual-ambience-b>
         <ra-irradiance-map-type type="int">0</ra-irradiance-map-type>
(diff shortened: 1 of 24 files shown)


wlbragg:
I noticed while I was doing this the magnetic compass still had some integral lighting in the ALS light map version, but I thought we had decided only the rpm and oil were supposed to have it. If so then that was a mistake that needs to be corrected in the current release PR. Not a hugh bug, in fact there would be some that might like it. There was some discrepancy as to what all was supposed to have both post and integral lighting. There was mention that rpm, oil, egt and mag compass all were supposed to have both. But we also found live video of only rpm and oil having both.

Megaf:
Testing right now.

Megaf:
Yeah, this works well. It looks like the real thing but with all instrument lights on.
image

If this is wrong. Can someone please provide evidence of so? Like a book or something?

legoboyvdlp:
Yes, I can quote the POH - except I don't have it right here.

It would be better to use a material animation imo. But if everyone else thinks it is fine, that's fine - so long as the behaviour is still correct under ALS ;)

legoboyvdlp:
To be fair, we have a feature freeze tomorrow, so this is definitely acceptable - we can work on this more next release :wink:

onox:
What I've seen in the screenshot is suffcient. I will merge it in the morning.

wlbragg:
Oops, sorry about that last description. It was supposed to contain the words...
Correction to hobbs default renderer material animation
It's late and I am done with this. It's work for now.

Megaf:
Instrument lighting dial is changing the brightness of the "flood light". Is that expected?

legoboyvdlp:
@Megaf The Panel Light knob adjusts the flood light and Radio Light knob adjusts the integral / avionics lights.

PR #1197, opened 2018-09-03
Ambient lighting correction

wlbragg:
This adds the ambient lighting from post and radio stack lights back into the interior panel lighting scheme. It was erroneously removed due to incorrect calculations of the two involved  lighting schemes. It needs to be in this release.


onox:
@wlbragg Can you cherry-pick this commit to the release branch?

```sh
git checkout release/2018.3
git cherry-pick ee434918d0c72a857fdba9314f2ba90952847861
```

PR #1198, opened 2018-09-04
Complete default renderer lighting scheme (Rembrandt compatible)

wlbragg:
This finishes the default renderer fall back. It mimics the same lighting scheme as the ALS interior lighting only using material animations.


wlbragg:
@Megaf please check this out and provide some feedback if you get a chance.

legoboyvdlp:
@wlbragg is it possible to make the whole instrument object emit light - not just the needles or white bits and therefore have the correct behaviour with no integral light on the main instruments while preserving visibility? This looks really good regardless, ice work :)

wlbragg:
>Would I be asking too much to extend the light animation to the instruments so it behaves like the ALS one? As you can see, instruments are not visible with POST,

I'm a bit confused by this. Maybe this will explain.
I basically mimicked the ALS version best I could except for the following...

ALS version gives you the entire panel with post lighting (instrument dial) and only the oil and rpm with integral (radio dial).
VS
Default version gives you the entire panel with post lighting (instrument dial) and nothing with integral (radio dial). 

That was because of the complexity required to add those two instruments to integral via another set of filters to add it in that particular condition. It would have required several more filters and there are more than I care for already so I assumed that feature could be lived without. That "feature" being the ability to fly as in the one video where we see a night flight with only the oil and rpm integral lighting and no other lighting.

Default post lighting mimics ALS post lighting
Default overhead white mimics ALS overhead white
Default flood red mimics ALS flood red (except it is cockpit wide VS a focused spot). 

>instruments are not visible with POST

So this I don't understand, they are visible with post if you have the switch and the panel dial set correctly.
To get post lighting you have to have the overhead red flood switch set to post or both and also have the instrument dial turned up.

@Megaf
Of the pictures you posted the first two in order are

1) flood red with post lighting = overhead flood red switch set to both and instrument light dial turned up

2)  flood red with no post lighting = overhead flood red switch set to flood only and instrument light dial turned up (thus no post lights)

So unless I am misunderstanding your issue, I think I did address everything with the above integral rpm and oil exception.

>with no integral light on the main instruments while preserving visibility

@legoboyvdlp Yes, kind of. That is what is being done for the overhead lighting, even the integral lighting is doing that but more extreme. All I could do to mimic the ALS post lighting is to do what in the ALS version were calling integral. It could be softened? You can see for yourself if that would work by turning down the post lights (instrument light dial) to a point your talking about, where it doesn't look so bright.
Keep in mind everything your seeing in the default rendering scheme is material animation of varying degrees of intensity and colors.
The default rendering scheme lighting of the gauges, were now calling "post", does actually include the faces, but because of the material settings of the object itself, they don't light up well.

wlbragg:
>The radio stack should probably be brightened,

I think they are the same as in the ALS renderer. It does rely on the radio knob for it's brightness, same as ALS version. Now the Rembrandt renderer is darker and you also can't control intensity nor does it have the "day mode". The c172p lighting under Rembrandt version is a completely different animal and I think there is much room for improvement for anyone that wants to work on it. I don't, because I can't run Rembrandt without issues and I am not willing to live without ALS effects, atmosphere effects alone are reason enough for me. While vivid and pretty, the sunset scenes are totally unrealistic in Rembrandt. So I prefer to put my time elsewhere. I feel the basic renderer does need to be supported whenever possible, at least for now.

>and perhaps the integral lighting on the engine instruments could go on the RADIO knob? 

The reason I didn't do that was because we don't see any RL aircraft with integral lighting on the entire instrument panel. So I am trying to mimic the "post" lighting and the post lighting is on the instrument dial.

I have inspected all the commits in the default renderer changes and I cannot find anything that would account for the change to that lever animation, nothing! I even did a "find in file" search for the FlapSwitch object and the only place it is adjusted is in flaps.xml and that didn't change. I am totally stumped as to what is happening here.
Is anyone else seeing this in this branch?
bug1

dany93:
@wlbragg wrote
> change to that lever animation.....
> Is anyone else seeing this in this branch?

Yes, I'm seeing it too.
At he risk of writing things which are obvious for you....
It looks like the animation of `FlapSwitch` is too large (the zero is good).

In 1b989dfd you **added** these lines in flaps.xml
```
    <animation>
        <type>material</type>
        <object-name>FlapSwitch</object-name>
        <object-name>FlapPos</object-name>
        <object-name>rect</object-name>
        <condition>
            <not>
                <property>/sim/rendering/shaders/skydome</property>
            </not>
        </condition>
        <emission>
            <red-prop>/sim/model/c172p/lighting/rgb-ra-r-factor</red-prop>
            <green-prop>/sim/model/c172p/lighting/rgb-ra-g-factor</green-prop>
            <blue-prop>/sim/model/c172p/lighting/rgb-ra-b-factor</blue-prop>
        </emission>
    </animation>

```
In the parent one, cb7a388f, this bug does not exist.

I don't understand this code well, it seems only for the lighting. I guess this the reason for your perplexity.

wlbragg:
@dany93 thank you for verifying it is the material animation that is causing it. I refused to entertain that notion even though it was looking like it was the only possible cause. 
Why, is another question. It is the same material animation that is now in every object in the cabin. Why does it affect the FlapSwitch and only the flap switch. It looks to me like it is actually doubling the movement amount. I verified the numbers in the translate function and they are correct.
This is what I call a "true" bug. I have had only one other "true" bug, many, many years ago in a Lotus spreadsheet function that went haywire for no discernible reason. I'll never forget that bug. I guess this is very good argument against letting a computer control ones fate.

wlbragg:
FYI: Yes, all the code does is apply some lighting to the objects if not in ALS. The values are assembles in System/als-lights.xml

wlbragg:
Well the solution was to move the material lighting section to the end of the animations. I'll bring this up on the dev list just because. But at least it is solved here.

gilbertohasnofb:
@wlbragg Looks wonderful, thanks for taking your time to implement this. I really dislike that we have three rendering systems but it's really nice that our aircraft will work well on all three of them :smile: :+1:

PR #1201, opened 2018-09-07
setting initial frequencies

gilbertohasnofb:
Currently, the frequencies are set to `0` in the `-set.xml` file. I've changed them to the minimum value each can have:

`118.0` for COMMs
`108.0` for NAVs
`200` for ADF

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -727,14 +727,14 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <volume type="double">0.7</volume>
             <frequencies>
                 <dial-1-khz type="int">0</dial-1-khz>
-                <dial-100-khz type="int">0</dial-100-khz>
+                <dial-100-khz type="int">2</dial-100-khz>
             </frequencies>
         </adf>
         <comm n="0">
             <power-btn type="bool">1</power-btn>
             <frequencies>
                 <dial-khz type="int">0</dial-khz>
-                <dial-mhz type="int">0</dial-mhz>
+                <dial-mhz type="int">118</dial-mhz>
             </frequencies>
             <volume-selected type="double">0.7</volume-selected>
         </comm>
@@ -742,7 +742,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <power-btn type="bool">1</power-btn>
             <frequencies>
                 <dial-khz type="int">0</dial-khz>
-                <dial-mhz type="int">0</dial-mhz>
+                <dial-mhz type="int">118</dial-mhz>
             </frequencies>
             <volume-selected type="double">0.7</volume-selected>
         </comm>
@@ -750,7 +750,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <volume type="double">0.7</volume>
             <frequencies>
                 <dial-khz type="int">0</dial-khz>
-                <dial-mhz type="int">0</dial-mhz>
+                <dial-mhz type="int">108</dial-mhz>
             </frequencies>
             <ident-audible type="bool">false</ident-audible>
         </nav>
@@ -758,7 +758,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <volume type="double">0.7</volume>
             <frequencies>
                 <dial-khz type="int">0</dial-khz>
-                <dial-mhz type="int">0</dial-mhz>
+                <dial-mhz type="int">108</dial-mhz>
             </frequencies>
             <ident-audible type="bool">false</ident-audible>
         </nav>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -727,14 +727,18 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <volume type="double">0.7</volume>
             <frequencies>
                 <dial-1-khz type="int">0</dial-1-khz>
-                <dial-100-khz type="int">2</dial-100-khz>
+                <dial-100-khz type="int">0</dial-100-khz>
+                <standby-mhz type="double">200.0</standby-mhz>
+                <selected-mhz type="double">200.0</selected-mhz>
             </frequencies>
         </adf>
         <comm n="0">
             <power-btn type="bool">1</power-btn>
             <frequencies>
                 <dial-khz type="int">0</dial-khz>
-                <dial-mhz type="int">118</dial-mhz>
+                <dial-mhz type="int">0</dial-mhz>
+                <standby-mhz type="double">118.0</standby-mhz>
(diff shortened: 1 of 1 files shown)


wlbragg:
@gilbertohasnofb I think what is happening is the com/nav radio is being populated at load time by 
/instrumentation/comm[1]/frequencies/selected-mhz
/instrumentation/comm[1]/frequencies/standby-mhz
This is what the fgfs main UI for "Radio" is using and our radio appears to match that on startup after an aircraft-data wipe.
So I am guessing that on a clean install maybe those two properties are not set and thus it defaults to what is in our -set
<dial-khz type="int">0</dial-khz>
<dial-mhz type="int">118</dial-mhz>
maybe we want to add
standby-mhz
standby-khz
to our set file with the the same "default" numbers?

To test we need someone to do a new install after making these changes.
@wkitty42 you up to the task?

wkitty42:
On 09/07/2018 02:38 PM, wlbragg wrote:
> To test we need someone to do a new install after making these changes. 
> @wkitty42 you up to the task?

new install? do you mean just delete the aircraft-data file? sure... what do i 
need to look for first before doing that? i'm guessing it would be the same 
after deleting the xml file and reloading?

wlbragg:
>new install? 

Unfortunately I mean a completely new fgfs install.  I know you seem to recompile a lot an was hoping maybe you could test this as a completely new user installing for the first time.
We're looking to see what the nav and com frequencies show after a fresh install under this branch. We're thinking it will now show
dial-khz type=0
dial-mhz=118
and until or unless we change it
standby-mhz = 0
standby-khz = 0
Whereas before this PR it would show
dial-khz type=0
dial-mhz=0
standby-mhz = 0
standby-khz = 0

gilbertohasnofb:
@wlbragg This PR is actually wrong... those `dial` properties set the position of the knobs, not the frequencies themselves. They should range from 0 to 360 (or 1440 in the case of the 25K knob in the KX165). I think we need to set those back to `0` and use `selected-mhz` and `standby-mhz` for all radios. I will push a new commit now.

gilbertohasnofb:
Done, this should have solved it. But I don't want to do a reinstall to test this because my FG install is already _very_ fragile and often breaks completely. :cry:

gilbertohasnofb:
@wlbragg after this gets merged, what do we need to do in order to get these last fixes into FGData in time for the release? Are you familiar with the process?

wkitty42:
On 09/07/2018 03:21 PM, wlbragg wrote:
>     new install?
> 
> Unfortunately I mean a completely new fgfs install. I know you seem to recompile 
> a lot


yep, pretty much any time there's a new commit to any of the three or four 
binary projects that make up the FG project... PLIB, OSG, SG, and FG...


> an was hoping maybe you could test this as a completely new user installing
> for the first time.

oh... that should still be easy enough... there's at least two cache files, 
maybe three, that would be deleted which would simulate a new install...

the one in aircraft-data to ensure no old craft settings get in the way...
the autosave_XXXX_Y.xml where XXXX is the year and Y is the release version... 
the navdata_XXXX_Y.cache file is the only other one i can think of...


> We're looking to see what the nav and com frequencies show after a fresh install 
> under this branch. We're thinking it will now show
> dial-khz type=0
> dial-mhz=118
 > and until or unless we change it
 > standby-mhz = 0
 > standby-khz = 0


ok, so the above is after the changes...


> Whereas before this PR it would show
> dial-khz type=0
> dial-mhz=0
> standby-mhz = 0
> standby-khz = 0

ok... i'll take a look and see what i can see...


FWIW: here's what i find right now in my existing installation... the last time 
i was in the c172p was 2018 Jul 4... i don't recall which airport, though... 
possibly PHNL or KRDU but those are/were my most common ones...


$ printf "\n\"hz\" in autosave_2018_3.xml\n"; egrep -e "hz" 
~/.fgfs/autosave_2018_3.xml; printf "\n\"hz\" in 
aircraft-data/c172p.xml\n";egrep -e "hz" ~/.fgfs/aircraft-data/c172p.xml

"hz" in autosave_2018_3.xml
     <frame-rate-throttle-hz type="double">0</frame-rate-throttle-hz>

"hz" in aircraft-data/c172p.xml
         <source 
type="string">/instrumentation/nav[0]/frequencies/selected-mhz</source>
         <selected-mhz type="double">114.8</selected-mhz>
         <selected-mhz type="double">123.9</selected-mhz>
         <standby-mhz type="double">118.3</standby-mhz>
         <dial-khz type="int">0</dial-khz>
         <dial-mhz type="int">0</dial-mhz>
         <selected-mhz type="double">114.8</selected-mhz>
         <standby-mhz type="double">113.9</standby-mhz>
         <dial-khz type="int">0</dial-khz>
         <dial-mhz type="int">0</dial-mhz>
         <selected-mhz type="double">127.9</selected-mhz>
         <standby-mhz type="double">120.9</standby-mhz>
         <dial-khz type="int">0</dial-khz>
         <dial-mhz type="int">0</dial-mhz>
         <selected-mhz type="double">116.1</selected-mhz>
         <standby-mhz type="double">117.7</standby-mhz>
         <dial-khz type="int">0</dial-khz>
         <dial-mhz type="int">0</dial-mhz>
         <dial-1-khz type="int">0</dial-1-khz>
         <dial-100-khz type="int">0</dial-100-khz>
         <standby-khz type="int">242</standby-khz>
         <selected-khz type="int">367</selected-khz>


so this would indicate that really the only file i would need to look at would 
be the saved settings file in aircraft-data...

wkitty42:
On 09/07/2018 03:23 PM, wlbragg wrote:
> It can't be tested with an aircraft-data wipe because the sim itself I think is 
> setting these values to the last known freq the user set outside the 
> aircraft-data cache.


not from what i can see grepping the sim's autosave_XXXX_Y.xml file... the only 
"hz" in there is the frame rate throttle value...

gilbertohasnofb:
@wkitty42 maybe it is being saved without a label 'hz'? The fact is that I tried deleting my aicraft-data folder and the frequencies did not reset neither to zero nor to my newly set defaults above but to some other values which I believe are from the previous session.

wkitty42:
On 09/08/2018 09:26 AM, Gilberto Agostinho wrote:
> @wkitty42 <https://github.com/wkitty42> maybe it is being saved without a label 
> 'hz'?


i suppose that is possible... my grep was looking for the properties with 
[k|m]hz in them...


> The fact is that I tried deleting my aicraft-data folder and the frequencies
> did not reset neither to zero nor to my newly set defaults above but to some
> other values which I believe are from the previous session.

that would seem to indicate they are in the sim's autosave_XXXX_Y.xml file, 
then... if not there, then the only other place i'm aware of would be in the 
sim's startup code...

PR #1211, opened 2018-09-27
Limit visor rotations

### LINKED ISSUE #1210: Visors rotate beyond window limit
I don't know when this happened or why but the visors now rotate too far forward. They can rotate through the windshield. I know they didn't used to be able to do this and they were tested by @gilbertohasnofb and myself. Now suddenly they rotate beyond rational limits.

fgfs-screen-026

wlbragg:
Fixes #1210 

Visor rotations are allowing the visors to rotate out of bounds and through solid objects. This PR corrects that issue.

I have some confusion as to why this is happening and also the numbers don't seem accurate to me.

The rotation visually looks like it should be close to 180, it was at 140 but now needs to be at 90.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -3202,8 +3202,8 @@
                 <command>property-adjust</command>
                 <property>controls/visorleft/position-deg</property>
                 <factor>5</factor>
-                <min>0</min>
-                <max>140</max>
+                <min>10</min>
+                <max>90</max>
                 <wrap>false</wrap>
             </binding>
         </action>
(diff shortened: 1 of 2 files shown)


wlbragg:
This is ready to be merged after testing and verification by another member of our team.

onox:
The code seems to work, but I have still some remarks about it:

- `/controls/visorrodleft/position-deg` and `/controls/visorrodright/position-deg` default value is nil.

- You should group the properties. So put it under `/controls/visors/left-rod/position-deg` and `/controls/visors/left/position-deg`, etc. I prefer to put it under `/aircraft/controls/`. That will better indicate that it is aircraft-specific.

- The drag indicator should be vertical instead of horizontal (the rods can be left horizontal).

> Is something like this still supposed to be able to get committed to this new release?

I think 2018.3 was chosen to be a 'extra stable' release, so we can backport the fix to the release branch.

wlbragg:
Due to the time restraints of this release and me personally not knowing when it will be too late to get this fix backported into fgdata, after thorough testing by myself and making as many of the changes as @onox suggested as practical, I am merging this to master and then will cherry pick it back to the 2018.3 release branch and notify the dev list it is ready to go.

PR #1216, opened 2018-10-09
Add altitude-agl-m signal to flight recorder

### LINKED ISSUE #1212: Ground shadows just a few metres below the aircraft in flight in replay
In replay modus the ground shadow ist jsut a few metres below the aircraft.
 Instead it should be on the ground.
You can see it here in the video: [https://www.youtube.com/watch?v=LfCcw6t04k0](https://www.youtube.com/watch?v=LfCcw6t04k0) at 0:21 and the following seconds.

c172p from fgdata, pulled yesterday, matching FlightGear bin (218.4.0)

wlbragg:
Fixes #1212 

I'm doing this in the blind as I am unable to record tapes right now for some reason. It crashes my fg session when I hit save. I have tried a few different directories to no avail

If someone can test this and see if this ail all that is needed I would appreciate it.

### DIFF
--- a/Systems/flight-recorder/components/effects.xml
+++ b/Systems/flight-recorder/components/effects.xml
@@ -85,4 +85,10 @@
         <property type="string">/fdm/jsbsim/hydro/spray-wake-speed-kt-actual</property>
     </signal>
 
+    <!-- 3d Shadow -->
+    <signal>
+        <type>float</type>
+        <property type="string">/position/altitude-agl-m</property>
+    </signal>
+
 </PropertyList>


legoboyvdlp:
Unfortunately not:

image


Perhaps the 2D shadow could be hidden above approx 75 feet, which is when it starts to become noticeable?

Or else, I didn't test it properly. 😕

wlbragg:
Hum, OK I think we have two issues here.
If I remember correctly...
The shadow height AGL is computed using altitude-agl-m. Now that property wasn't being recorded and was needed to make the shadow behave during replay. So that should have fixed one issue of the shadow not being positioned at all.
But there is also a trick the shadow shader code that is also adjusting the shadow AGL at aircraft AGL's > 75'. I think that is the issue here now.
~~~
 if (dist < 75)
    {

         pos.z = -0.9 * alt_agl + 0.05 * vertex.z;
         //pos.z  = 0.05 * (vertex.z + gear_clearance);

         pos.xy -= deltaxy;

    }
    else
    {
         //The code below to shrink the shadow while keeping it 75 m. away has some issues that need to be fixed. Making the shadow shrink at x^2 rate partly to cover up this problem until it can be solved . . .
         //The problem is that the aircraft isn't flat/parallel to the ground any more, but it appears to be at an angle to the ground.
         //The various shrinkages perhaps mess with the angles somehow, meaning that when the animations apply the roll & pitch corrections they just don't quite work as they should
         pos.z = (-0.9*75)*alt_agl/dist + 0.05 * vertex.z ; //if the shadow is more than about 75 meters from the aircraft it disappears so we are going to just keep it right at 75 m. & make it smaller to simulate greater distance.
         //(-0.9*75) is the same factor for altitude we were using above when dist=75.  *alt_agl/dist keeps it at the right height proportionally to simulate the location at a further distance, while actually just keeping it at 75 m. distance.

         //pos.z = 0.05 * vertex.z ; //if the shadow is more than about 75 meters from the aircraft it disappears so we are going to just keep it right at 75 m. & make it smaller to simulate greater distance.



         //shrink the size FIRST, THEN move where it needs to be.  If you shrink later you're also shrinking the deltaxy distance moved, which doesn't work well
         pos.xy = 75 / dist * pos.xy;  //shrinking the size of the shadow to simulate further distance.  Should be linear shrinkage but doing it ^2 for now to help ucover up the issues in code above.

         pos.xy -= 75 * deltaxy/dist; // Similarly to above, * deltaxy/dist; keeps it at the right XY position proportionally to simulate the location at a further distance, while actually just keeping it at 75 m. distance.

    }
~~~

legoboyvdlp:
The shadow is that small at 75 feet that I think just hiding it would be fine.

wlbragg:
I'm not sure what you mean. So it quits displaying a shadow in the sim and also then does not record it?
I'd rather fix the shader code to do this right. I don't really remember why we did this in the shader? Was it because it was sinking below the ground the higher you got?

legoboyvdlp:
If you can't fix it in the shader for whatever reason, not displaying it at all if altitude is greater than 75 whether in replay or in-sim is surely an option?

wkitty42:
On 10/11/2018 01:06 PM, wlbragg wrote:
> I'm not sure what you mean. So it quits displaying a shadow in the sim and
> also then does not record it?

i would keep on recording the property because you might have some point where 
you are within 75m of ground level... seeing it live and in reply would be nice...

being able to tilt/distort the shadow to display "flat" on that ground mesh is a 
different problem... especially when it goes over peaks and valleys...

wlbragg:
@legoboyvdlp what are we seeing in your last screen capture, does is show a shadow at all? Is that the shadow in the red circle?

shadow

wlbragg:
Sorry missed the "Comment" button.

legoboyvdlp:
Yep, that's the shadow.

wlbragg:
@legoboyvdlp 

>If you can't fix it in the shader for whatever reason, not displaying it at all if altitude is greater than 75 whether in replay or in-sim is surely an option?

That is what was happening by default because the angle computation in the shader, it was disappearing at 75 AGL.
I think the opinion was  to make it work above 75 AGL.

I'm more of the opinion to just not show the shadow in the replay.

I don't know of a simple answer for this. I guess if left as is then at least you get the "correct" shadow in replay from an interior view or near interior view but not from a ground view.

What happens in the sim to the shadow when viewing from ground?

legoboyvdlp:
This may actually be a problem with the shadow itself:

image

Can anyone else see this? It seems like the shadow lifts off the ground as you climb - even when **not**
in replay.

wlbragg:
Yes, it is by design. The higher the AGL the hire the shadow lifts off the ground. That was to allow for the shadow to not flicker when over rough ground or cast below a hillside. The higher the aircraft the lees likelihood that the shadow would disappear or flicker.
It was really appreciated from a pilot perspective, but the downside was seeing it from a observer point of view.
~~~
<filter>
        <type>gain</type>
        <!--gain>0.3048</gain-->
        <gain>0.3500</gain>
        <input>
            <property>/position/altitude-agl-ft</property>
        </input>
        <output>
            <property>/position/altitude-agl-m</property>
        </output>
    </filter>
~~~

gilbertohasnofb:
@wlbragg what's the status of this PR?

legoboyvdlp:
I think this is fine... what I thought was a bug wlbragg tells me is intended behaviour?

wlbragg:
@gilbertohasnofb what @legoboyvdlp said is correct. It is a consequence of cheating the shadow clipping that happened after you exceeded a certain AGL. As you gain in altitude, after x agl the shadow no longer translates lower as you rise, it shrinks instead. I think maybe we should limit having the shadow visible from the exterior view only if it is below the AGL where it switches from translation to scale?

gilbertohasnofb:
Great, we have no opened PRs left! Thanks to all of you for all the hard work

PR #1223, opened 2018-10-24
Use auto-mixture in autostart function

### LINKED ISSUE #1206: Add auto mixture to the standard autostart in c172p.nas
I created an fcs_function outputting the property /fdm/jsbsim/engine/auto-mixture in c172p-engine.xml to use in the "on approach" state to insure (best we can) that the aircraft will start when in this state, regardless of elevation.

It is simple to use...
~~~
var auto_mixture = getprop("/fdm/jsbsim/engine/auto-mixture");
setprop("/controls/engines/current-engine/mixture", auto_mixture);
~~~

I totally forgot to do this when I created it but think it is a good idea. Unless of course we don't want to let autostart work in higher altitudes. :fearful:

wlbragg:
Fixes #1206 

This was an easy fix as we had the auto-mixture function working already.

Please test and merge when ready.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -19,7 +19,10 @@ var autostart = func (msg=1) {
     # Setting levers and switches for startup
     setprop("/controls/switches/magnetos", 3);
     setprop("/controls/engines/current-engine/throttle", 0.2);
-    setprop("/controls/engines/current-engine/mixture", 0.95);
+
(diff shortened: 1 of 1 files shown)


legoboyvdlp:
I'll test this one in the morning. Bogota at 8000 feet should work nicely? :smiley:

legoboyvdlp:
It works great! I'll go ahead and merge this.

PR #1229, opened 2018-11-07
Mixture set to 1 at sea level in autostart

### LINKED ISSUE #791: use airfield elevation to calculate optimal mixture for autostart
Currently our autostart function does not take into consideration the elevation of the airport. This means that while trying to start a plane in a high altitude airfield, autostart will always use full mixture and will be cranking the engine forever. Maybe we could try to take the altitude in consideration, something like using full mixture up to a certain altitude, then using some formula to calculate (or guess) the optimal mixture for starting the engine

Inspired by this post here: https://forum.flightgear.org/viewtopic.php?f=25&t=29996

wlbragg:
Fixes #791

Simple change to autostart mixture at sea level to 1 VS .95 for full power.

This should be ready to merge after review.

### DIFF
--- a/Systems/c172p-engine.xml
+++ b/Systems/c172p-engine.xml
@@ -206,7 +206,7 @@
         </switch>
 
         <fcs_function name="auto-engine-mixture">
-            <function> <!-- Mixture -- RICH (above 3000 feet, LEAN to obtain maximum RPM) -->
+            <function> <!-- Mixture rich above 3000 feet, LEAN to obtain maximum RPM -->
                 <sum>
                     <value>0</value>
                     <table>
--- a/Systems/c172p-engine.xml
+++ b/Systems/c172p-engine.xml
@@ -206,13 +206,13 @@
         </switch>
 
         <fcs_function name="auto-engine-mixture">
-            <function> <!-- mixture to 1 at sea level in autostart, state autostarts and tutorials, default mixture setting is .95 -->
+            <function> <!-- Mixture -- RICH (above 3000 feet, LEAN to obtain maximum RPM) -->
                 <sum>
                     <value>0</value>
                     <table>
                         <independentVar>/position/altitude-ft</independentVar>
                         <tableData>
-                            0       1.00
+                            3000    1.00
                             13000   0.75
                         </tableData>
                     </table>
--- a/Systems/c172p-engine.xml
+++ b/Systems/c172p-engine.xml
@@ -206,13 +206,13 @@
         </switch>
 
         <fcs_function name="auto-engine-mixture">
-            <function>
-                <sum>
+            <function> <!-- mixture to 1 at sea level in autostart, state autostarts and tutorials, default mixture setting is .95 -->
+                <sum
                     <value>0</value>
                     <table>
                         <independentVar>/position/altitude-ft</independentVar>
                         <tableData>
-                            0       0.95
+                            0       1.00
                             13000   0.75
                         </tableData>
                     </table>
--- a/Systems/c172p-engine.xml
+++ b/Systems/c172p-engine.xml
@@ -207,7 +207,7 @@
 
         <fcs_function name="auto-engine-mixture">
             <function> <!-- mixture to 1 at sea level in autostart, state autostarts and tutorials, default mixture setting is .95 -->
-                <sum
+                <sum>
                     <value>0</value>
                     <table>
                         <independentVar>/position/altitude-ft</independentVar>


Systems/c172p-engine.xml
```diff
@@ -206,13 +206,13 @@
         </switch>
 
         <fcs_function name="auto-engine-mixture">
-            <function>
-                <sum>
+            <function> <!-- mixture to 1 at sea level in autostart, state autostarts and tutorials, default mixture setting is .95 -->
+                <sum
```
dany93: Please close the tag.
`<sum>`
Otherwise, it works. Autostart tested up to about 9000 ft (in flight).

wlbragg: Wow, good catch. Don't know how I manage to do that.

Systems/c172p-engine.xml
```diff
@@ -207,7 +207,7 @@
 
         <fcs_function name="auto-engine-mixture">
             <function> <!-- mixture to 1 at sea level in autostart, state autostarts and tutorials, default mixture setting is .95 -->
-                <sum
+                <sum>
```
dany93: Sorry, one more thing (unless I don't understand).
The default setting for mixture in FG is 1 (not 0.95). I think that 0.95 came from onox in Autostart. See commit 3662ef5d, c172p.nas, line 14. Possibly to allow autostart at a slightly higher altitude?

wlbragg: OK, so you think we also need to set the default at 1 VS .95.

wlbragg: @dany93 I just tested this and it is starting with a mixture of 0 by default. The autostart is adjusting the mixture according to the table and at my elevation here in KS it is setting it to .93. SO i think this is finished and working as we wish, I think I spoke out of place and assumed the default setting at startup was .95. I think your correct that was solely a function of the autostart in an attempt to raise the startup ceiling limit.

onox: Try testing at KLXV (9900 feet) and VNLK (9300 ft). See https://github.com/c172p-team/c172p/issues/791#issuecomment-234784355

onox: We have a table now which controls the mixture to use during autostart? (I'm a bit busy with other projects and RL stuff)

wlbragg: It's already been tested at the ceiling limits and works fine.
I went ahead and tested it at KLXV and that works as well.

wlbragg: >The mixture is set at 0 at least in c172p-set.xml (lines 602 and 793). 

Is this what we want? Or should we set or allow it to default to 1? At 0 the knob is pulled all the way out. I always thought that looked odd to start up the simulation with that knob pulled out instead of pushed in. Unless it says different in POH?

dany93: @wlbragg it can make sense. In the POH-180hp-552SP (at least) it says
(p. 4-17, or PDF 102/384)

> SECURING AIRPLANE
[....]
> 4. Mixture -- IDLE CUTOFF (pulled full out)
> 

Hence a likely initial state.

onox: Looks history (added in first commit): https://github.com/c172p-team/c172p/blame/7a2d70c69b3b7b65337cb1b08d1edc9db4bccaee/c172p-set.xml#L474

Haven't tested whether things will break if you remove it.

Systems/c172p-engine.xml
```diff
@@ -206,13 +206,13 @@
...
         <fcs_function name="auto-engine-mixture">
-            <function>
+            <function> <!-- mixture to 1 at sea level in autostart, state autostarts and tutorials, default mixture setting is .95 -->
                 <sum>
                     <value>0</value>
                     <table>
                         <independentVar>/position/altitude-ft</independentVar>
                         <tableData>
-                            0       0.95
+                            0       1.00
```
onox: @wlbragg Doesn't this already starts lowering the mixture above 0 ft? (linear interpolation between 0 and 13_000 ft). Perhaps insert a row `x 1.00` where x is an altitude where 100 % can still be safely used.

wlbragg: Makes sense, I'll run some tests and see where it should start decreasing.

dany93: I also had this thought.

The POH-180hp/552SP says:
(p. 4-16, PDF 100/384)

> ENROUTE CLIMB
> 3. Mixture -- RICH (above 3000 feet, LEAN to obtain maximum RPM).

wlbragg: Thanks @dany93 that'll give us some kind of a hint or target to shoot for.

wlbragg:
Testing confirmed "rich" to 3000' showed stable max RPM. At 5000' lean slightly produced max RPM. The POH proved accurate.

This is done, ready to merge I think.

wlbragg:
Pair of dashes in comment caused a critical JSBSim error. Glad I tested it after I made the change to the comment.

This is done, ready to merge I think.

legoboyvdlp:
Tested, everything looks good. Merged.

PR #1233, opened 2018-11-07
FDM, pitching moment at zero alpha

### LINKED ISSUE #1232: FDM, pitching moment at zero alpha
FDM, change in cmo: pitching moment at zero alpha.
Alpha dependence of the pitching moment factor = cmo + (cm x alpha).

- Currently, cmo = 0.1 (constant).

_(With no flaps)_ 
It gives a climbing airspeed of 70 to 80 kt with no action on the yoke and no elevator trim correction after takeoff, when the airspeed increases (with pitch up and down oscillations). My impression is that the pitch up tendency is slightly too high.
Cruise (2400 RPM, 110 - 115 kt) necessitates an elevator trim correction of about about 0.2 to stabilize.

- This change:
cmo depends on qbar-psf (the dynamic pressure, depending on airspeed) as per [Megginson's tables](http://www.megginson.com/Aviation/roskam-coefficients.html#sec.data).

With this change, one needs to slightly pull the yoke to rotate and takeoff. After that, the aircraft climbs at about 85 to 95 kt kt with no action on the yoke and no elevator trim correction. Climbing at 80 kt is obtained by slightly pulling the yoke. Loosening the yoke gives an acceleration. More reassuring and, I think, closer to a real behavior (although this behavior is a choice which can be made by the elevator trim at takeoff).
For cruising (about 110 kt straight and level flight) an elevator trim correction of a bit less than 0.1 is enough.

dany93:
Fixes #1232

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -1672,7 +1672,13 @@
                     <property>aero/qbar-psf</property>
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/cbarw-ft</property>
-                    <value>0.1</value>
+                    <table>
+                        <independentVar lookup="row">aero/qbar-psf</independentVar>
+                        <tableData>
+                            13.6  0.09
+                            21.2  0.04
+                        </tableData>
+                    </table>
                 </product>
             </function>
 


legoboyvdlp:
is this mergeable?

legoboyvdlp:
I can tonight.

dany93:
Thank you.

legoboyvdlp:
I didn't notice anything massively wrong although I obviously don't have any actual flight experience to comment on it. Everything looks fine.

I think its fine to merge it.

dany93:
Thank you for your cooperation.

legoboyvdlp:
I'll merge this later if there are no objections?

dany93:
Thank you, lego.

PR #1236, opened 2018-11-11
fixing flap click sound

### LINKED ISSUE #1234: Flap lever sound: should only sound when lever position changes
If you currently drag the flap lever upwards when flaps are retracted or downwards when they are fully extended you still get the lever clicking, even though the lever doesn't move at all?

The same occurs with the [ / ] keys.

gilbertohasnofb:
Fixes #1234

* flaps click sound do not play any longer when trying to extend/retract it more than possible at max or min positions.


Nasal/electrical.nas
```diff
@@ -500,7 +500,6 @@ var avionics_bus_1 = func() {
 var flapsDown = controls.flapsDown;
 controls.flapsDown = func(v) {
     flapsDown(v);
```
onox: It seems you can completely remove the overriding of `controls.flapsDown`. Without the `click` function it is pointless to override it.

gilbertohasnofb: done

onox:
@gilbertohasnofb It looks like `SourceSelectorKnob` to `ki266.xml` was added by @wlbragg in https://github.com/c172p-team/c172p/commit/3e514845f8314572e5614952eae745cd35dadf52

onox:
Have you checked the value of `instrumentation/dme/switch-position`?

wlbragg:
It's switching the knob 30 degrees at a time which is definitely not correct.

wlbragg:
Hum, actually 90 degrees total is not that far off. it is actually supposed to be 60 total, ie: 20 degree at a time.

wlbragg:
The center origin of the rotation is correct.

gilbertohasnofb:
I once had pushed a fix for the volume knobs behaving strangely (turning many rotations instead of the expected 3/4 of a turn or so). Today I noticed the volume knobs were buggy but to my surprise my 'fix' was still there. I reverted that and they started behaving normally again. I really don't know if something changed in the core FG code or elsewhere. That perhaps could also be happening to the DME switch? I remember it working properly and I couldn't track any changes to the rotation factors in the history of our repo. @wlbragg thanks for taking a look at it.

gilbertohasnofb:
@onox concerning the other sounds, it seems to me that the only way to fix them not sounding at max/min positions would be to use the `in-transit` mode, but that does not work for many of them. E.g. we want the magnetos to 'click' at every key turn, we want three position switches to also click in between the 'transit', etc. So I guess there is not much I can do on our side short of asking the devs to perhaps address this issue.

legoboyvdlp:
Can this be merged soon?

gilbertohasnofb:
@legoboyvdlp not yet, there is still that DME bug that needs fixing. @wlbragg do you think you could take a look at it when you have a chance?

wlbragg:
@legoboyvdlp I'll look at this later tonight and see if I can correct the rotation issue. As I read this the only issue is ...
>the DME mode selector animation is totally screwed up and I can't figure out why. Position N1 is fine, but HOLD and N2 is completely wrong:

gilbertohasnofb:
@wlbragg there is also the DME code duplication issue:

> In that case the files should be deleted (and make sure they're not imported in some other XML file)

wlbragg:
Yes, I'll delete that now as well so no one else get confused by it.

gilbertohasnofb:
@wlbragg When you have a chance could you take a look at the DME animation?

wlbragg:
Looking now, I confirm it is messed up. Will push fix shortly.

wlbragg:
Something is really messed up with this. The rotation is correct. I'm trying a different method and it isn't working correctly either.

wlbragg:
Bizarre, I don't know how many of you know about this bug. It has to do with a knob animation being before a material animation, breaks the knob animation. Same thing happened to the flaps. Anyway, that was the issue. I'm gong to push the fix now.

wlbragg:
I guess @legoboyvdlp or @onox can merge this if @gilbertohasnofb is OK with it.

legoboyvdlp:
Sure. Looks great, thanks! Will wait for @gilbertohasnofb s confirmation

gilbertohasnofb:
Everything seems to be working fine now, thanks a lot for the fix, @wlbragg 

> Bizarre, I don't know how many of you know about this bug. It has to do with a knob animation being before a material animation, breaks the knob animation. Same thing happened to the flaps. Anyway, that was the issue. I'm gong to push the fix now.

That's a weird bug. Perhaps it would be worth to mention it in the dev-list so that it might get fixed one day or at least so that our solution gets documented there.

PR #1240, opened 2018-11-27
Animate front castor wheel rotations around pivot point

### LINKED ISSUE #1239: Front wheels of amphibious have no caster animation
The front wheels should rotate or caster freely around their pivot point when turning the aircraft on the ground.

wlbragg:
Fixes #1239 

This PR also corrects a couple texture mapping issues.
1) Wheel axles
2) Interior rudder release mechanism


wlbragg:
I believe this is ready to merge after testing.

gilbertohasnofb:
@wlbragg textures look really good and the animation looks amazing, but just one small thing: when I zoomed into the wheels and I started rolling, I had the impression the wheel was rolling to slowly in comparison to the airplane speed which makes it look like it's sliding.

wlbragg:
I thought the same thing, I'll speed it up.

dany93:
I've tested the animations.

The wheel rotation (castering) works well in usual conditions, but the `wheel-speed-fps` condition gives an angular "jump" if we start from stopped (after braking). However, I don't see what to do. Without this condition, the wheels pivot close to 90° at stop. Due to the JSBSim's way of calculation.
0.1 fps condition is not enough to avoid this.
Maybe that a 0.5 fps condition would be enough?

I have a small issue (tell me if it needs an issue number)
My Aircraft Options:
- Engine power option,
- Landing gear options,
are no longer saved between sessions. Even with "Save state of aircraft between sessions" checked.

(the others are, at least these that I usually have "ON")

With an older version of the aircraft the options are saved well, which should mean that it is not due to my c172p cache.

wlbragg:
@dany93 
> have a small issue (tell me if it needs an issue number)

Let me check this out before creating a new issue.

>Maybe that a 0.5 fps condition would be enough?

I'll take a closer look at this as well. You can test 0.5 and let me know what you think. But you may have helped my identify what bad symptom I was seeing and I may have a fix for it.

dany93:
I have tested the 0.5 fps value and I think that it might be enough. It was just to ask for your opinion.

wlbragg:
>I have a small issue (tell me if it needs an issue number)
My Aircraft Options:
    Engine power option,
    Landing gear options,
    are no longer saved between sessions. Even with "Save state of aircraft between sessions" checked.

Right, this is the new behavior because we now can use the variants choices.
The "Save state of aircraft between sessions" was ever only for the cabin switch setting and not the variant selections such as gear and engine.

This is something we still need to discuss. What I think we need to do is to finish the "State System". Right now all there is is "Automatic" which only plays nicely with the QT launcher "Locations" tab and does nothing else. I left it that way because we need to discuss a few things before proceeding.
Eventually we should include other startup states like Cold and Dark, Secured, Cruise and the important one pertaining to this conversation would be "Last State" which would be everything you had going on the last time you shutdown, both switch settings and variant selections.

The idea would be if you select a variant, that is what you get.

If you select state = Automatic (Default State), you get the QT launcher choices taking precedence, ie:
Runway = auto-started and brake on at the runway threshold
Parked = not running at the location tab parking selected
On Approach = auto-started in the air configured for landing

Then there would be other states...
Cold and Dark = parked at QT location tab setting no brakes and unsecured
Secured = parked at QT location tab setting secured
Running =parked at QT Location tab  setting but with engines running and brake on
On Approach = (not sure explicit "on approach" is needed if the Automatic state and QT Launcher handle this state effectively.
Last State = Everything set as it was the last time you exited, including variant. I don't think this would overrule the variant selection though, but it might?

I'm still looking at the caster code. 0.5 is good enough but doesn't do anything for the from stopped to moving snap to 0. The only thing I can think of to deal with that is to use a JSBSim function to produce a filter value to use for the rotation combining /fdm/jsbsim/gear/unit[19]/wheel-speed-fps and our own property we can control. Because I don't think we can do anything with /fdm/jsbsim/gear/unit[19]/wheel-speed-fps, which means we are at its mercy and will have to live with the position of the gear at all points in time including raising it and the jump at the start of rolling.

I have another issue. That is what happens if you raise gear with the gear rotated, it raises it that way. Something tells me it should straighten out as it is raised?
Also I haven't checked this out on water yet to see if it behaves when on the water.

legoboyvdlp:
The tricycle nose gear has a centering mechanism when the gear is fully decompressed - not sure if this has a shock absorber or not? Given that the wheel is catering, the airflow would probably centralise it.

dany93:
@wlbragg wrote

> The "Save state of aircraft between sessions" was ever only for the cabin switch setting and not the variant selections such as gear and engine.

I know and I fully agree with this behavior. I just made an attempt with  "Save state of aircraft between sessions" (that I usually never use) just in case, because I had some doubt.

My (current) opinion about this:
- The previous behavior (choices by options in the sim, separately saved) was fully satisfying.
- The "Save state of aircraft between sessions" should only be for the cabin switch setting and not the variant selections such as gear and engine. However, that can be incompatible because everything is interdependent. In this case, see next item.
- Keeping a possibility of "gear and engine selection" saved (independently from the other settings) is important.
- The "Aircraft Options" feature gives a great flexibility, difficult to transfer with the variant choices. For example, you will also have to add the 180 hp with default gear (chosen by many users, I think), 180 hp with 26" Bush Tires, 180 hp with 36" Bush Tires.

dany93:
@gilbertohasnofb can we merge this PR? (I think so)

gilbertohasnofb:
It all looks good, thanks a lot!

PR #1249, opened 2018-12-24
Define unuseable fuel

legoboyvdlp:
This defines unuseable fuel. I will write to the JSBSIM git repository asking this actually do something - it is better that FlightGear handle this if possible, without us doing any hacks to the fuel system.

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -688,6 +688,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
                 <water-contamination type="double">0.0</water-contamination>
                 <sample-water-contamination type="double">0.0</sample-water-contamination>
                 <fuel-sample-taken type="bool">false</fuel-sample-taken>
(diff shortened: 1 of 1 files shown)


c172p-set.xml
```diff
@@ -688,6 +688,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
                 <water-contamination type="double">0.0</water-contamination>
                 <sample-water-contamination type="double">0.0</sample-water-contamination>
                 <fuel-sample-taken type="bool">false</fuel-sample-taken>
+				<unusable-gal_us type="double">1.5</unusable-gal_us>
```
onox: Tabs @legoboyvdlp :stuck_out_tongue_closed_eyes:

legoboyvdlp: 😀

legoboyvdlp:
@onox that's it :)

legoboyvdlp:
I think we can merge this, then merge #1254 just before release?

PR #1254, opened 2019-01-12
Account for the volume of unusable fuel.

bcoconni:
JSBSim now allows to set a volume of unusable fuel below which fuel can longer be drained from a tank.
This feature has been added following the request in issue JSBSim-Team/jsbsim#141 from @legoboyvdlp. This also allows the PR #1249 to become functional.

This PR updates the fuel system of the c172p to account for the volume of unusable fuel.

### DIFF
--- a/Systems/fuel.xml
+++ b/Systems/fuel.xml
@@ -44,11 +44,19 @@ Fuel system for 2 different engines at choice, JSBSim FDM
     <!-- ****************************************************** -->
     <channel name="To Tank2">
         <!-- from Tank 0 (to Collector Tank 2) -->
+        <fcs_function name="propulsion/tank[0]/contents-volume-gal">
+            <function>
+                <quotient>
+                    <p> propulsion/tank[0]/contents-lbs </p>
+                    <p> propulsion/tank[0]/density-lbs_per_gal </p>
+                </quotient>
+            </function>
+        </fcs_function>
         <switch name="fuel/from-tank0-to-tank2">
             <default value="0"/>
             <test logic="AND" value="0.1">
                 propulsion/tank[0]/priority EQ 1 
-                propulsion/tank[0]/contents-lbs GT 0
+                propulsion/tank[0]/contents-volume-gal GT propulsion/tank[0]/unusable-volume-gal
                 /consumables/fuel/tank[2]/level-lbs LT 0.05
                 propulsion/tank[2]/priority EQ 1
                 accelerations/Nz GE 0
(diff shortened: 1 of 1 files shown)


wlbragg:
I'm getting property not found
~~~
FGPropertyValue::GetValue() - does not exist
propulsion/tank[0]/density-lbs_per_gal
~~~

bcoconni:
> I'm getting property not found

Have you pulled the last revision of FlightGear `next` ?

wlbragg:
OK, I wasn't sure where the JSBSim changes went. So I have to rebuild FGFS.

@gilbertohasnofb @legoboyvdlp @onox 
How does this affect us as far as merging into master? 

I ask because if this is merged into master, then master will only work on systems that pull and build a new FGFS. Not a problem per se, but it is if testers and even us developers haven't built the latest FGFS. I know I don't usually build FGFS until shortly before or after a release.

wkitty42:
On 1/12/19 8:22 PM, Jonathan Redpath wrote:
> I think we want this to be in the 2019.1 release; if we merge this directly into 
> the 2019.1 branch maybe then rebase onto master once everyone has 2019.1 set up?

that's basically my thoughts, too... i don't know if 2018.4.0 will become 2019.1 
or not, though... probably since we're already into 2019...

legoboyvdlp:
No, we will have a 2018.3.2 this incoming week and then go directly to 2019.1 as far as I remember.

wkitty42:
yeah, i know about the patch release for 2018.3 upcoming... i was speaking of what 2018.4.0 will become... either way, i was agreeing with you :smile_cat:

onox:
> I think we want this to be in the 2019.1 release; if we merge this directly into the 2019.1 branch maybe then rebase onto master once everyone has 2019.1 set up?

The current workflow is that any branch like `release/2019.1` is a stable branch and is created when we release 2019.1 version of the c172p. So it's easier to just keep this PR open and merge it when we can expect most developers to have recompiled SG/FG.

bcoconni:
FlightGear 2019.1 is now a Release Candidate. Could it be an opportunity to make some progress on this topic ?

legoboyvdlp:
If we are going to make a release branch for 2019.1 I think we agreed to merge this immediately before cutting the branch as some of us are not using next / RCs?

In any case maybe it's time to do a release with various bug fixes?

wlbragg:
We need to find out if we can still get it in this included in 2019.1. I am going to be tied up the next several days working on something else so I am afraid I wont be of much help.
I am also really concerned about the broken parts of the work we have already done and is included in current stable release that is now broken. That is my priority at the moment. But I have no issues with including this into master regardless. Do you mind pinging the dev list and asking if we cut a new release branch if someone will put it in the 2019.1 release?

onox:
When is 2019.1 going to be released? I think people have been given enough time to install FG git of January 12th or later, so this PR can be merged in my opinion.

wlbragg:
@gilbertohasnofb you OK with merging this?

gilbertohasnofb:
@wlbragg Yes, definitely, this looks good.

@bcoconni Thanks for the PR!

onox:
Added milestone because we don't have a corresponding issue in this project.

PR #1256, opened 2019-02-16
FDM adjustment based on it0uchpods' feedback

legoboyvdlp:
Hi,
@dany93 @gilbertohasnofb @it0uchpods 

it0uchpods finally had time to revise the FDM according to his experience. This pull request is the initial work based on his comments and suggestions. It incorporates the following changes:

- P-factor adjusted to a value of 5: this is accurate for the Lycoming IO320 and calculated by measuring the rolling rate and adjusting the p-factor value to match
- Removal of aileron trim: this is artificially correcting for a value of p-factor that is much too high. Instead we should use a more accurate p-factor value.
- Tweaking of adverse yaw: this requires much testing with regards to stall and spin behaviour and could could be further tuned
- Fixing of a bug with roll due to rudder: it was reversed in that left rudder would actually cause right 
roll! Furthermore, crucially, rudder alone is now able to counter p-factor and propwash, instead of requiring aileron input as well. Previously, it was impossible to fly in a coordinated manner and also counteract both p-factor and propwash
- Propwash: and finally, propwash is now reduced which was previously much too large.

Given that this is a somewhat major change, I think this requires much testing, but I hope you'll appreciate that all these changes completely match it0uchpods' experience in the Cessna 172.

Finally, it0uchpods is currently away for a few days, so he might be delayed in responding.

⚠️ WIP DO NOT MERGE

### DIFF
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -584,8 +584,8 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
 
     <controls>
         <flight>
-            <aileron-trim type="double">0.022</aileron-trim>
-            <rudder-trim type="double">0.02</rudder-trim>
+            <aileron-trim type="double">0.0</aileron-trim> <!-- c172 has no such thing -->
+            <rudder-trim type="double">0.01</rudder-trim>
             <rudder-trim-knob type="double">0.0</rudder-trim-knob>
             <freeze-yoke type="bool">false</freeze-yoke>
             <aileron-cmd type="double">0.0</aileron-cmd>
@@ -1144,7 +1144,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             </settings>
             <aero>
                 <coefficient>
-                    <spiral-propwash-coeff type="double">0.180</spiral-propwash-coeff>
+                    <spiral-propwash-coeff type="double">0.120</spiral-propwash-coeff>
                 </coefficient>
             </aero>
             <fcs>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -1144,7 +1144,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             </settings>
             <aero>
                 <coefficient>
-                    <spiral-propwash-coeff type="double">0.25</spiral-propwash-coeff>
+                    <spiral-propwash-coeff type="double">0.180</spiral-propwash-coeff>
                 </coefficient>
             </aero>
             <fcs>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -1144,7 +1144,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             </settings>
             <aero>
                 <coefficient>
-                    <spiral-propwash-coeff type="double">0.120</spiral-propwash-coeff>
+                    <spiral-propwash-coeff type="double">0.175</spiral-propwash-coeff>
                 </coefficient>
             </aero>
             <fcs>
--- a/c172p-set.xml
+++ b/c172p-set.xml
@@ -1144,7 +1144,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             </settings>
             <aero>
                 <coefficient>
-                    <spiral-propwash-coeff type="double">0.175</spiral-propwash-coeff>
+                    <spiral-propwash-coeff type="double">0.185</spiral-propwash-coeff>
                 </coefficient>
             </aero>
             <fcs>
--- a/c172p.xml
+++ b/c172p.xml
@@ -588,7 +588,7 @@
                     <yaw> 0.0 </yaw>
                 </orient>
                 <sense> 1 </sense>
-                <p_factor> 59 </p_factor>
+                <p_factor> 5 </p_factor>
             </thruster>
         </engine>
 
@@ -617,7 +617,7 @@
                     <yaw> 0.0 </yaw>
                 </orient>
                 <sense> 1 </sense>
-                <p_factor> 60 </p_factor>
+                <p_factor> 6 </p_factor>
             </thruster>
         </engine>
 
@@ -1868,7 +1868,7 @@
                         <independentVar lookup="column">aero/beta-rad</independentVar>
                         <tableData>
                                     -0.35        0           0.35
-                            0       -0.0216      -0.0216     -0.0216
+                            0       -0.0166      -0.0166     -0.0166
                             0.070   -0.039       -0.0786     -0.039
                             0.094   -0.025       -0.0504     -0.025
                         </tableData>
--- a/c172p.xml
+++ b/c172p.xml
@@ -1626,7 +1626,7 @@
                     </table>
                 </product>
             </function>
-
+            
             <function name="aero/coefficient/Cldr">
                 <description>Roll_moment_due_to_rudder</description>
                 <product>
@@ -1634,7 +1634,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/rudder-pos-rad</property>
-                    <value>0.0163</value>
+                    <value>-0.30659</value>
                 </product>
             </function>
 
@@ -1863,7 +1863,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/aileron-pos-rad-avgd</property>
-					<value>0.45</value> <!-- Way to much adverse... -->
+                    <value>0.45</value> <!-- Way to much adverse... -->
                     <table>
                         <independentVar lookup="row">aero/alpha-wing-rad</independentVar>
                         <independentVar lookup="column">aero/beta-rad</independentVar>
--- a/c172p.xml
+++ b/c172p.xml
@@ -1634,7 +1634,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/rudder-pos-rad</property>
-                    <value>-0.30659</value>
+                    <value>-0.20752</value>
                 </product>
             </function>
 
--- a/c172p.xml
+++ b/c172p.xml
@@ -617,7 +617,7 @@
                     <yaw> 0.0 </yaw>
                 </orient>
                 <sense> 1 </sense>
-                <p_factor> 6 </p_factor>
+                <p_factor> 5 </p_factor>
             </thruster>
         </engine>
 
@@ -1634,7 +1634,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/rudder-pos-rad</property>
-                    <value>0.0147</value>
+                    <value>0.0163</value>
                 </product>
             </function>
 
@@ -1863,6 +1863,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/aileron-pos-rad-avgd</property>
+					<value>0.45</value> <!-- Way to much adverse... -->
                     <table>
                         <independentVar lookup="row">aero/alpha-wing-rad</independentVar>
                         <independentVar lookup="column">aero/beta-rad</independentVar>
@@ -1890,7 +1891,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/rudder-pos-rad</property>
-                    <value>-0.0645</value>
+                    <value>-0.0505</value>
                 </product>
             </function>
         </axis>
--- a/c172p.xml
+++ b/c172p.xml
@@ -1634,7 +1634,7 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/rudder-pos-rad</property>
-                    <value>-0.20752</value>
+                    <value>-0.1365</value>
                 </product>
             </function>
 
@@ -1863,13 +1863,13 @@
                     <property>metrics/Sw-sqft</property>
                     <property>metrics/bw-ft</property>
                     <property>fcs/aileron-pos-rad-avgd</property>
-                    <value>0.45</value> <!-- Way to much adverse... -->
+                    <value>0.75</value> <!-- Way to much adverse... -->
                     <table>
                         <independentVar lookup="row">aero/alpha-wing-rad</independentVar>
                         <independentVar lookup="column">aero/beta-rad</independentVar>
                         <tableData>
                                     -0.35        0           0.35
-                            0       -0.0166      -0.0166     -0.0166
+                            0        0.0216      -0.0216     -0.0216
                             0.070   -0.039       -0.0786     -0.039
                             0.094   -0.025       -0.0504     -0.025
                         </tableData>


wlbragg:
I'm not qualified to test this as an experienced pilot of this aircraft, but I do like the changes I feel as a user of the features offered in this simulation.

The left pulling behavior still appears to be there enough that I 
The hydro portion appears to be much easier to negotiate, especially the right turn without "water" rudders.

I've had a long standing bug (or difficulties) in the right rudder where I have had instances where I loose almost all right rudder. From memory it did not appear to be a p-factor issue or the likes, it was a physical lack of rudder movement to the right. Although I think that it still could be connected to the p-factor or left pull tendency or trim and the way we have to simulate those actions (see: https://github.com/c172p-team/c172p/issues/1213). So far this has been somewhat alleviated. I need to continue to test this behavior throughly before I can make a definitive comment on this particular issue.

dany93:
At testing, pleasant at first view, except a too great rolling effect following an action on the rudder.
By elsewhere, too easy IMHO.
I disagree with the (near) cancellation of the P-Factor and adverse yaw effects.

A few important points to clear up.

@legoboyvdlp , @it0uchpods 
 

> * P-factor adjusted to a value of 5: this is accurate for the Lycoming IO320
>  and calculated by measuring the rolling rate and adjusting the p-factor value to match

I don't understand the relation:
- The P-factor effect is not on rolling rate, it is mainly on yaw moment.
- Rolling is due to the propeller moment. Or due to yaw rate, but as a secondary effect, only if there is a yaw rate. Also due to side-slipping, skidding, if there is one.
A value of 5 for the P-factor comes down to almost cancelling it. With this value, I hardly noticed any effect during my tests (or not at all, not sure).
I admit that the amplitude of this effect is guessed (although correct IMO, not excessive), but between 60 and 5 there is much room.

>     * Tweaking of adverse yaw

You divided it by a factor of 2.9. Testing it shows no adverse-yaw at all.

>   * Fixing of a bug with roll due to rudder: it was reversed in that left rudder would actually cause 
>   right roll! Furthermore, crucially, rudder alone is now able to counter p-factor and propwash, instead of requiring aileron input as well. Previously, it was impossible to fly in a coordinated manner and also counteract both p-factor and propwash

Firstly, I dont see how you could write "it was reversed in that left rudder would actually cause right roll!". This is wrong, an action on left rudder makes it rolling in the correct (left) direction. Such a bug wouldn't have survived for so many years.

I don't understand how and why "flying in a coordinated manner and counteract both p-factor and propwash" was impossible to you, I never have any difficulty to control the aircraft, to fly in a coordinated manner. About 0.1 to 0.2 amplitude for the rudder at climbing, full thrust, max 0.3 at very very low speed. A bit of aileron input at full thrust is to counteract the propeller moment. There is room left for stronger action, it just needs a normal coordination in the actions from the pilot.

More important :
It concerns flight aerodynamics and JSBSim.

As far as the rolling moment due to rudder deflection:
The rudder deflection normally (according to aerodynamics and in correct FDMs) gives a very weak rolling moment in the unexpected direction. So weak that I couldn't feel it.
The main rolling moment (in the expected direction) is due to the side-slipping (if there is one) and to the difference in airspeed between the two wings (during turning). Not a primary effect from rudder deflection

In JSBSim, the rolling moment coefficient due to rudder deflection angle is positive. Not negative as you set it, and way too large. By doing this, you compensate the unnecessary aileron effect by a inappropriate rolling moment due to rudder.

Unless this is due to me (mistakes with these signs are quickly done) . In this case, sorry and please fix me, this is a bad mistake and I would not like to go on with such a huge misunderstanding, above all on flight aerodynamics. However, Aeromatic and  every JSBSim aircraft in FG would be wrong, which is hard to admit.

Reminder:
https://github.com/c172p-team/c172p/issues/1188#issuecomment-415414384 (dany's references)
https://github.com/c172p-team/c172p/issues/947#issuecomment-360999442 (tigert)
Pilots :
Megginson (coefficient tables)
@tigert
@algefaen

Simulation :
@okroth
_________

dany93:
@wlbragg 

> I've had a long standing bug (or difficulties) in the right rudder where I have had instances where I loose almost all right rudder.(....) it was a physical lack of rudder movement to the right.

Weird.... I never had  this issue. Could it be that you had a bug?

> The amphibious taxing (...). I can now control the roll-out down the runway without brakes,....

Already with the master version, I could (can) control the amphibious in line with the air rudder alone (no brakes on the ground, no water rudders).  Just a larger action (0.8? memory) for a short time at very low speed, then 0.2 is enough.

it0uchpods:
I didn't test your version, Lego, after your final changes. I'm on vacation
now. So I can't know if it is right or not. My commit improved it a bit,
but I didn't have time to finish it


Sorry dany, but as someone who has also flown a real 172, this FDM doesn't
behave correct.

I don't know what you're pilots hardware is, but I measured my data based
on actual percent movement of pedal.

And yes, p factor doesn't affect yaw, that is silly. Propwash does. If p
factor did, it wouldn't be called p factor, it be r factor. p is roll rate.
It comes from the fact that the engine is turning the prop, which tried to
push the body in the opposite direction. However, when the aircraft rolls
left it also yaws left (obviously). Because the rudder in most planes
including the C172 and PA28 that I've flown many times has a big roll
moment, (in fact I have actually practiced flying with only engine power,
elevator trim, and rudder to control attitude, it's easily doable.) This
comes in handy when writing down ATIS, pulling up airport charts in flight,
etc. You can cancel 100% of the p factor roll and propwash yaw with about
10-15% rudder maximum, in a climb. When level in most situations, we
actually need little to no rudder at all. But in climbs, about 10-15% as I
said. At low rpms for descent, 1500 or so, you will actually need left
rudder as the plane will try to pull right. I suspect this is due to having
some actual right bias in rudder.

Kind Regards,
Josh

legoboyvdlp:
> The P-factor effect is not on rolling rate, it is mainly on yaw moment.
Correct: but it also causes rolling rate. It may be that jsbsim's implementation of p-factor is incorrect: does it 

?  A value of 5 for the P-factor comes down to almost cancelling it. With this value, I hardly noticed any effect during my tests (or not at all, not sure).

There certainly is an effect: I have to push right rudder or I will end up turning left. 


Finally regarding the signs of the rudder coefficient, well, when on the master branch I changed the coefficient to become a much larger value for testing (0.5 or something like that?) I had a roll in the opposite direction to the rudder deflection: maybe you could test that by increasing the coefficient yourself on the master branch and testing?

it0uchpods:
Roll due to rudder should definitely be positive.

Right rudder (positive) should cause Right roll (positive)

However, if you are still using the Aeromatic FCS system (which I believe you are, another thing that should be changed!) then right rudder is actually negative. So check your moment. If your roll_rudder moment is positive with right rudder (see prop browser) then it is ok.

J

PS: P factor isn't a fast turn.... its slow roll to the left.

legoboyvdlp:
yep: I saw that fcs problem with the nose wheel.

That is why the rudder coefficient should be reversed.

And I can confirm that with right rudder there is a positive roll_moment.

However, I also see that at cruise the roll moment is still too large: I will reduce it again

legoboyvdlp:
Furthermore, regarding Megginson's coefficient tables, I have high doubts about their accuracy, given that they were seemingly the product of a computer simulation in 1995 -- certainly not wind tunnel. I would not discount them but I would trust wind tunnel data more.

legoboyvdlp:
@dany93 I agree that adverse was now actually too low: I increased it, and also tuned the roll due to rudder again.

dany93:
> I don't know what you're pilots hardware is, but I measured my data based on actual percent movement of pedal. 

I measure my amplitudes on controls by the Internal properties like
`/controls/flight/aileron`
which are independent of my hardware in signs and indicated amplitude. The percent movement of pedal may be different, depending on 'power' in the xml joystick binding file.


> And yes, p factor doesn't affect yaw, that is silly. Propwash does. If p factor did, it wouldn't be called p factor, it be r factor. p is roll rate. It comes from the fact that the engine is turning the prop, which tried to push the body in the opposite direction.

It seems that you are confusing with the propeller (engine) torque.
P-Factor: (among others)
http://cessnachick.com/2016/06/01/left-turning-tendencies/
https://www.boldmethod.com/learn-to-fly/aerodynamics/4-left-turning-tendencies/

>  In a nose high attitude, the downward swinging blade of the propeller generates more lift (thrust) due to the higher angle of attack (of the blade, not necessarily the airplane). This means in that attitude, the right side of the propeller (downward swinging blade) generates more thrust than the left side (upward swinging blade) which causes the airplane to yaw to the left.

legoboyvdlp:
I didn't particularly notice any left yaw at all, but I certainly notice left roll so perhaps it is actually JSBSIM at fault?

dany93:
Based on my current understanding, JSBSim is not at fault.

Left roll at full thrust is due to the propeller torque. Correct in the principle. Also quantitatively correct when I made calculations based on engine power and RPM.
The aircraft reactions at this torque and at controls is another subject.
The absence of left yaw even at high AoA is seemingly due to the too low P-factor (changed from 60 or 59, to 5)

it0uchpods:
59 or 60 is a ridiculous number… it does not roll that fast by itself, sorry.

Kind Regards,

--
Joshua Davidson

------------------------------------------------------------------------------
CONFIDENTIALITY NOTICE: If you have received this email in error,
please immediately notify the sender by e-mail by replying to the message.  
This email transmission may contain confidential information.  This 
information is intended only for the use of the individual(s) or entity to 
whom it is intended even if addressed incorrectly.  Please delete it from 
your files if you are not the intended recipient.  Thank you for your 
compliance.

dany93:
As I already wrote, the 59 or 60 for the P-factor is a guess (not exactly a guess, rather based on other readings from David Culp if I remember well, and missing Hal V. Engel). A value which can obviously be discussed. Maybe trying at least 30, but not 5, which gives no effect.
Please read (again?) @tigert comment https://github.com/c172p-team/c172p/issues/947#issuecomment-360999442

But firstly, we cannot agree about the P-factor effect as long as we don't speak the same language (please don't laugh, sorry for my English :smile: ).

>  it does not roll that fast by itself, sorry

 You are still speaking of roll as the effect, I think yaw (cf my links above).

_(Anyway, have a good holidays)_

it0uchpods:
Hi dany,
That isn't true. in my warrior, I have a pfactor of 5 and it rolls with engine torque exactly how the real one does. That isn't the problem in the code. Cranking up p factor to counteract that poor latitudinal behavior isn't the solution. 

Kind Regards,
josh

dany93:
@it0uchpods 

> in my warrior, I have a pfactor of 5 and it rolls with engine torque exactly how the real one does.

Once more, roll is primarily due to the propeller torque. Also due to yaw rate, but only as a secondary effect. But true, the P-factor has a direct effect giving a yaw rate. Which means that everything is cross-coupled.
> ....cranking up p factor to counteract that poor latitudinal behavior isn't the solution.

Sorry, I don't understand. Which (what is) "poor latitudinal behavior"?

it0uchpods:
I mixed up the terms, whoops. Yes. But the same can be said about Yaw. The roll, however, it the noticeable effect because yaw causes roll. Hence why pushing the rudder kills the roll rate due to p factor.

I'll get back to you further when I'm back from vacation if I find the time.

Kind Regards,
Josh

onox:
Are @tigert or @algefaen able to comment whether this PR is correct? @wlbragg Perhaps the mailing list could be notified about the existence of this PR and see if there are any people there who can comment.

wlbragg:
>Perhaps the mailing list could be notified about the existence of this PR and see if there are any people there who can comment.

I have something typed and ready to go for the dev list, but prefer to wait until Josh has a chance to test these changes and comment.
If we all can come to a consensus here before involving others that would be ideal!

dany93:
@it0uchpods wrote
>  Previously, it was impossible to fly in a coordinated manner and also counteract both p-factor and propwash
And yes, p factor doesn't affect yaw, that is silly. Propwash does.
59 or 60 is a ridiculous number… it does not roll that fast by itself, sorry.


I've made tests and measurements, of the effect of P_factor and Spiral_propwash, on the needed correction by ailerons and rudder.

**P_factor, Spiral_propwash : effect on rudder et ailerons controls**
 
c172p master, initial P_fact = 60, Spiral_propwash = - 0.05 x 0.25
 
Tests with 180 hp from current master, Load = Default (pilote only), No flaps
Climbing 60 kt (low airspeed, on purpose), full thrust, rudder et ailerons coordinated to keep slip-ball centered.
 
Aileron and rudder values from /controls/flight/ Properties


P_factor | Spiral-prop-coeff | aileron | rudder
-- | -- | -- | --
60 | 0.25 | 0.05 to 0.10 | 0.15 to 0.20
5 | 0.25 | 0.07 to 0.10 | 0.06 to 0.12
60 | 1.00E-05 | 0.07 to 0.10 | 0.15 to 0.17
0.01 | 1.00E-05 | 0.08 to 0.09 | 0.06 to 0.07

P_factor = 60, Spiral-propwash-coef = 0.25 give:
==> **No noticeable effect on aileron control**. Roll (in straight and coordinated flight) is mostly due to propeller torque.
==> **Effect on rudder : about + 0.1**, less than + 0.15. **Main effect is on rudder, controllable**.

Which doesn't mean that the P_factor and Spiral_propwash must not be decreased a bit (@wlbragg :smile: , although that's not my feeling). But the initial values are not "silly", not to be as much decreased as @it0uchpods does for the P_factor. If @tigert or @algefaen tell me to decrease them, I do it with confidence.

it0uchpods:
What do I know, I just did a bunch of tests on the real aircraft over
several flights and marked my findings... lol

I got the terms mixed up, that was my bad.

Now I can tell you for sure we never need aileron to keep it level. All we
need is rudder input. So if you won't take my actual aircraft experience
seriously, at least fix this.

legoboyvdlp:
I agree with it0uchpods.

First of all:

The cruicial bug is certainly that roll response to p-factor is incorrect. **Whatever value of p-factor or propwash is chosen, this needs corrected**


We can go on and on for hours on this -- but let me just remind you that these values for p-factor and propwash are specifically tuned to match the real aircraft, and I think you should take that into account (and wait for feedback from @tigert and @algefaen ) before rejecting this.

gilbertohasnofb:
@it0uchpods perhaps you should reconsider the way you are writing some of your opinions here. I do not think it is very diplomatic to write 'lol', 'silly' or 'ridiculous' in relation to other people's opinions. By doing so, you are coming across as immature and your opinion becomes more likely to be ignored. Changes to the FDM are a big deal and also a huge can of worms, with people from all sides complaining with us at every single change. This is partially why some of us here are very reluctant to merge FDM changes without having a very deep conversation about what those causes. This is also why we try to get as many RL pilots to join the discussion, since a lot of these can be subjective and having more people with real life experience always give us a more solid idea of the best path to follow.  But such conversation is not possible if we descend into immature one-upping.

it0uchpods:
Well, everytime I spoke here about this rather serious issue, I was spoken
down to and told I was wrong, and there is nothing incorrect despite my
extensive testing and comparison to the real craft. Hard to remain serious
after that.

Well. I'll be on my way since my feedback is clearly unvalued.

Blue skies...
Josh

gilbertohasnofb:
> Well, everytime I spoke here about this rather serious issue, I was spoken down to and told I was wrong, and there is nothing incorrect despite my extensive testing and comparison to the real craft. Hard to remain serious after that.

Again, this is a non-diplomatic approach. If you react to people criticizing you or disagreeing with you by 'loling' them or calling them 'silly' or their opinions 'ridiculous', then you bring the discussion to a lower level and, in all likelihood, will not be making friends. I need to emphasize that I have not tested this PR and I have no real life experience in a 172, so what I am writing here is solely concerning the communication I've been reading so far and not concerning your work whatsoever. We are all very grateful to contributions to this repo, Josh, and it would be a pity if you decide to cut the conversation short because of the discussion. I am sure that no one here is intending to speak down to you, but taking offence at people's reluctance to accept your ideas won't bring the matter much further. I would then ask you to reconsider the approach to the communication here.

dany93:
Pilots feedback is appreciated (more: precious) but you were unclear. I don't wish to negatively argue, I would prefer that we agree on the explanation. That's the best way, and that's the only one to be positive, instructive. But, as you understand, I'm not convinced by your arguments. Not that I don't want, I can't.
If my position was only to refuse any change, I would already have  stopped this discussion.
So, please take my words as being without animosity

If if you got the terms mixed up, can you rewrite them as as they should have been?

Your experience is that you never need aileron to keep it level (but think that you don't have a spring as in a joystick, that can be misleading). Although I guess you mean you don't need to exert any force.
But in that case, how do you think that the propeller torque is countered? I don't believe that the rudder alone is efficient to hold the aircraft wings level. I will explain this later (I'm drawing a sketch). In straight flight, the rudder is just able to keep coordination. And if the flight is coordinated, can you explain me how is the propeller torque countered?

it0uchpods:
Well, I apologize for those unprofessional words, but as you can imagine, I've got a bit fed up and my "word choice" suffered a bit. However, I am not sure if I agree with everything.

If I add a lift coefficient that is 5x or 6x too high, causing far too much lift, don't you think that is a "silly" number? Maybe my way of speaking is not inline with yours.

@dany93 
Here is how I did it. On ground, I found the centered position of the yoke. I had someone on the outside let me know when both ailerons were exactly flush with the wing. I noted the yoke position. In the c172 I flew, it was actually almost dead straight, but there was a very slight amount of play in it. Once I had that, in flight, I did a test, and set the aileron to what not only was my "noted" amount of yoke turned, but could just about see that the ailerons were level on the wings. I was able to easily stop it from rolling left with the rudder, only about 10%-15% max of rudder, depending on airspeed, pitch, etc. (level flight needs less rudder counteract than climbing). I even moved the aileron a bit left, and by pushing the rudder a little more, it stopped rolling. (and the slip ball was still about in the middle). In fact, you can use the rudder to control roll, while writing down ATIS, or turning chart pages etc, something I've done  before, and ailerons stay neutral. If I drop power in a descent? No yoke roll change needed, just get off right rudder, and depending on how much power reduced, you may even need some left rudder! (this may be due to the slight right rudder trim, or something related to prop, I don't exactly know the cause of this motion, that c172 has no in-cockpit rudder trim so I was not able to test it). 

That is my experience of the real thing.

Kind Regards,
Josh

dany93:
@legoboyvdlp wrote
 > The cruicial bug is certainly that roll response to p-factor is incorrect. **Whatever value of p-factor or propwash is chosen, this needs corrected**
> (......)
> (and wait for feedback from @tigert and @algefaen ) before rejecting this.

If you think that there is a bug in roll response to P-Factor (I don't), did you report it to the JSBSim developers?
IMO, if this is in the c172, your current fix is not correct (I will explain).

Be reassured. My intention is not to reject with no discussion (or that's too late! :smile:  ). I've always been attentive at @tigert and @algefaen feedbacks and advices.

However, when I am in a domain where I can, I don't admit something when I think there is a defect or that I don't understand.

it0uchpods:
I don't think it is wrong either. I just think your number is too high. In your data above, this one looks the most correct to me:
pf    sprial    aileron           rudder
5     0.25      0.07 to 0.10   0.06 to 0.12

However. If you could make the aileron less, and the rudder a little more (10-15% is my testing) then I think it matches quite nice (like I did in the pa28)

Kind Regards,
Josh

it0uchpods:
Remember that a rudder can control roll if you wanted too (or your roll control cables snapped or something). Roll due to rudder comes to mind here.

dany93:
@it0uchpods 
Thank you for your explanation, I'm convinced that your intention was to observe this seriously.
> In fact, you can use the rudder to control roll, (...), something I've done before, and ailerons stay neutral

Yes, I've done this too. My opinion is that even small actions on the rudder give small side-slippings and small turns which, as secondary effects, give a roll (by dihedral  effect and airspeed difference between both wings).

legoboyvdlp:
@dany93 
testing a quick aeromatic job results in this:

`<function name="aero/moment/Roll_rudder">
       <description>Roll moment due to rudder</description>
       <product>
           <property>metrics/bw-ft</property>
           <property>fcs/rudder-pos-rad</property>
           <value> 0.0100 </value>
       </product>
    </function>`


If positive rudder (what FlightGear sees) is to the right then it should cause positive roll (to the right) -- ie coefficient should be positive


Thus, I think it is that the sign was accidentally reversed?

it0uchpods:
Careful!
JSB's default fcs/rudder-cmd-norm and trim are backwards! So negative is actually right, wheras with /controls/flight/rudder negative is left. Aeromatic uses the JSB ones.

Because it's confusion, I usually switch em to the /controls/flight ones.

Kind Regards,
Josh

tonghuix:
Have a try for *Slow Flight Maneuver*,this video's really helpful https://www.youtube.com/watch?v=9Wek1TISeZc&t=536s

I think by testing Slow Flight Maneuver, you will feel what should be done here.

wlbragg:
Thank you guys for working this out.
@it0uchpods I have told you on several occasions I really value your input here as you are a currently experienced RL pilot of this aircraft. Your willingness to defend your input and knowledge is why your still here talking and I personally appreciate it.
It can be difficult to work as a group here when your idea is challenged, everyone here has had that happen many times. But in the end we always seem to get to a good solution by hashing it out.
Right now we have the best of both worlds, RL experience and theoretical understanding. If you and @dany93 can merge the two, well, we'll have a really good FDM!
Thank you!

dany93:
@it0uchpods 
> If positive rudder (what FlightGear sees) is to the right then it should cause positive roll (to the right)

That's an important point on which I don't agree.

During this last period of time, I was doing a sketch.
c172-roll-due-to-rudder


Reminder, (for other colleagues...) JSBSim Body frame is:
X to the front, Y right, Z downward

Aircraft seen from behind.
The rudder is deflected to the right (not very visible, but please believe me or imagine).
The incoming airspeed, by hitting the right side of the rudder exerts a force **to the left**.
The roll moment of this force  relative to the X axis is **negative** (roll to the left).
In fact, this moment can only be **very weak**. The lever arm is very small compared with the aileron arm. IMO, it cannot give a noticeable roll.

it0uchpods:
Hi dany:
I believe it depends on how you simulate the roll effect that comes from stepping on rudder.

As I see it:
You either do it with rudder
Or simulate both wings so that you can get the rolling moment due to differential lift causes by yaw.

I was talking for the first one, which is how most people end up simulating it. Yes, it's less realistic, but I think it has to do until the second option is simulated.

You are correct in that when speaking literally, however

Bye,
Josh

it0uchpods:
Hi,
I am also wondering if roll_beta would be better here. I make some test.

Bye,
Josh

dany93:
I see with pleasure that our point of views got closer.

Mine about simulation is that the rudder mainly does the yaw (for what it's done).
After that, the dihedral angle and the differential lift cause the roll. JSBSim does this quite well.

In fact, the roll moment mainly depends on the side slip and differential airspeed on both wings. And this happens independently from the instantaneous rudder deflection, even when the rudder is centered. Hence the necessity to uncouple rudder deflection and roll moment (at first approximation).
Rudder : mainly for yaw, roll is a secondary effect,
Ailerons: mainly for roll, yaw is a secondary effect (adverse yaw).
IMO, JSBSim is quite good in its usual form.

it0uchpods:
Hi dany,
I did not notice that this was able in JSB. I never saw such things? Are they done internally, or need in aerodynamics?

Bye,
Josh

gilbertohasnofb:
> I see with pleasure that our point of views got closer.

Thanks to you both, I really appreciate how constructive these last messages are.

dany93:
I'm a bit surprised, not sure to understand your question well. Sorry if I respond beside.

JSBSim is the frame, the FDM file contains the tools to make the calculations of the forces and moments, which themselves will give accelerations. Apart from a minority which are internally calculated, like P-Factor, engine torque...., you need good aerodynamics knowledge to set the coefficients which link the effect of airspeed (forces) to movements.

Everything is like rudder for yaw and ailerons for roll, with coefficients.

1 - Let's start by yaw, primary rudder effect.
Its main secondary effect is induced roll (by rolling moment due to differential lift)
```
            <function name="aero/coefficient/Clr">
                <description>Roll_moment_due_to_yaw_rate</description>
                <product>
                    <property>metrics/bw-ft</property>
                    <property>aero/bi2vel</property>
                    <property>velocities/r-aero-rad_sec</property>
                    <table> (or value)
```
`aero/bi2vel` is the important factor. `bi2vel = b.w/2`V, where V is the airspeed.
If omega is the yaw rate, the velocity difference between both wings is
`V1^2 - V2^2 = 2.V.omega.bw`
Which enables to calculate the difference in lift between wings, hence the rolling moment. A calculation shows that it is proportional to` bi2vel.omega`, that is
```
                    <property>aero/bi2vel</property>
                    <property>velocities/r-aero-rad_sec</property>
```
2 - Roll moment due to side slipping
Physical cause: Dihedral effect, which stabilizes the aircraft when it skids or slips.
```
            <function name="aero/coefficient/Clb">
                <description>Roll_moment_due_to_beta</description>
                <product>
                    <property>metrics/bw-ft</property>
                    <property>aero/beta-rad</property>
                    <value>-0.092</value>
```
Which writes that the rolling moment is proportional to the beta (slip) angle and the aerodynamic pressure.

Difficult to give more details for the calculation here, all the more that I'm not sure to respond your question.

it0uchpods:
Hi dany,
Yes, we are pretty much saying the same thing now :)

So, I'll have a play around with the coeffs and see if I can get it to behavior as my notes.

Bye,
Josh

wlbragg:
https://www.youtube.com/watch?v=aqggWcd-uMg

Why is our aircraft turning left instead of right?

legoboyvdlp:
Not quite sure, but I would imagine any right-turn is an overcorrection to the p-factor and propwash?

I assume you are talking about the takeoff roll?

dany93:
My guess is that there is some cross wind from the left (see at landing too, where we can assume that there is no P-factor, no propwash effects).

My question would rather be: Why is _this_ aircraft rolling right instead of left?

Difficult for me to understand what they say, but at 9:36 he says "more right rudder". It looks like a last decrabbing with crosswind from left.

PR #1261, opened 2019-02-28
FDM: change P-factor (from 60 to 15)

### LINKED ISSUE #1260: Too much turn-to-the left tendency at takeoff and climbing.
This issue is connected to this one by @legoboyvdlp  and @it0uchpods . https://github.com/c172p-team/c172p/pull/1256#issue-253643544
I hope they can join it.
I have preferred opening a new issue starting back from the master version. This will avoid polluting the git history with features like reversing the proposed wrong roll-moment-due-to-rudder code and less important ones that are still to be discussed.
Also, I would prefer that we change not too many things at a time, that we work more analytically.

We have a newcomer in the french forum, newcomer in FG. He also noticed that the aircraft is pulling too much to the left, which confirms @legoboyvdlp  and @it0uchpods opinions. I invited him to join us in this discussion because he has a 20 years experience in C172 piloting, which means that his feedback should be valuable. All the more that @tigert and @algefaen no longer respond. He still has to discover FG a bit and he is waiting for rudder pedals to be delivered.

~I will prepare a pull request as a starting point.~
Done PR #1261 (change P-factor from 60 to 15).

:warning: :warning: :warning: DO NOT MERGE YET! :warning: :warning: :warning:

dany93:
Fixes #1260

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -588,7 +588,7 @@
                     <yaw> 0.0 </yaw>
                 </orient>
                 <sense> 1 </sense>
-                <p_factor> 59 </p_factor>
+                <p_factor> 15 </p_factor>
             </thruster>
         </engine>
 
@@ -617,7 +617,7 @@
                     <yaw> 0.0 </yaw>
                 </orient>
                 <sense> 1 </sense>
-                <p_factor> 60 </p_factor>
+                <p_factor> 15 </p_factor>
             </thruster>
         </engine>
 


tonghuix:
I tested this one and #1261 for lifting off and climbing, the left tendency not such strong as master. I have to say the P-factor maybe correct now, but I think other factors that contribute to left-turning tendency are still too strong. The evidence is when I level off and cruise flight, still  bank (not yaw) to the left, so I suppose is the spiral wash or/and torque still too strong ? 

As for #1261, I think its P-factor is too weak.

`P.S. I am still learning to get my pilot licence, and not have such much experience, so I ask my instructor to tell me the real feeling. (He did not test fgfs all the time).`

wlbragg:
@dany93  where are we at with this PR and issues?

dany93:
@tonghuix wrote

> I think other factors that contribute to left-turning tendency are still too strong. The evidence is when I level off and cruise flight, still bank (not yaw) to the left, so I suppose is the spiral wash or/and torque still too strong ? Or we should add a fixed rudder trim?

There's no such evidence. It obviously depends on RPM and load. Endless issue.

With the pilot alone, it is close to be balanced near 2200 RPM. Also near 2500 RPM.
And try with the copilot...

dany93:
@wlbragg wrote
> where are we at with this PR and issues?

I'd like this PR (P-factor --> 15) to be merged. It goes in the direction of a common request.

Anyway, re-adjusting this later is very easy.

tonghuix:
@dany93 As I experience build up and ask more instructors, I would say you are right at this point.

However, manufacturers would minimum propwash by mount engine offset longitudinal axis a bit or higher dihedral on left wing than right.

wlbragg:
>There's no such evidence. It obviously depends on RPM and load. Endless issue.

>I'd like this PR (P-factor --> 15) to be merged. It goes in the direction of a common request.

Agreed, well go this direction one more time and call it good until or unless we get some other hard data.

@it0uchpods if you happen to be watching this thread, what control system do you use for FlightGear, mouse, keyboard, flight stick, rudder peddels?

dany93:
Thank you, @wlbragg

legoboyvdlp:
>  unless we get some other hard data.

I would agree with everything you said there - only to comment on one thing: we already do have one such data point - that it is possible to counter both left rolling moment and left yawing moment by applying right rudder. If the FDM could be tuned in such a manner as that can be done in our cessna then I think that would probably be the last major inconsistency resolved.

legoboyvdlp:
I tested this and the behaviour looks a lot better - although not having flown myself I naturally cannot judge that correctly...

However, just making an observation based on the above data, I noticed that it is now possible to counter the left rolling and yawing using rudder - however quite often I found that to counter left rolling I had to apply more right rudder than was needed to counter left yawing, meaning I found myself in uncoordinated flight. Nevertheless, the ball deflection was not more than 1/2 of a diameter (half of the slip-skid ball still within the central region), so I think that is acceptable, especially now with rudder-trim :)

dany93:
The main use of the rudder is for keeping coordinated flight.
Rolling control is done by the ailerons.
 _(of course)_

During short sequences (e.g. climbing), the rudder must be used to keep the slip-ball centered.

During cruise, my observation in that the ailerons are the most efficient to keep leveled wing and route. If the slip-ball is correct, these weak aileron settings have a very weak effect on coordinated flight.
Thus, rather than having an action on rudder trim, I think having an action on aileron trim would be the best for our comfort in simulation. Keeping in mind that the real C172 does not have this.

I have an aileron trim setting on my joystick for this. Very comfortable. Cheating, but we are in simulation with joysticks with return springs (which are sometimes too stiff). Also, very very useful for accurate tests and adjustments.
IMO, an action on the rudder trim is much less needed (although I have it too).
IRL, this tendency exists, but usually softer close to the neutral position and the friction in the controls (cables, axes) makes it slightly less noticeable. Anyway, even IRL, a well-balanced aircraft is much more pleasant.

it0uchpods:
Dany is right, when levelled off cruising we need little to no rudder, and little to no aileron. In climbs. We need rudder.

I'll test it on my machine later and let you all know.

I personally use a Logitech 3D pro, and a pair of pedals I made myself out of some racing pedals. But my opinions are calibrated using percent deflection, and not linear.

Kind Regards,
Josh

algefaen:
I was out flying a C206 (so slightly bigger and not directly comparable, but same principles) last weekend. Once again I was reminded just how much I have to push the right rudder, especially during just after take-off (low airspeed, high AOA) - it really takes a lot of force. But I also realized that I'm not actually pushing it that far in - maybe 5 cm - it's mostly heavy because of the force pushing back through the rudder pedals. Which of course we'll never see because we don't have force feedback.

I think we should require a small amount of right rudder to stay on center line (propwash) during the take-off roll and then a bit more right rudder during the climb (added p-factor). Which seems to be what we've concluded on here. 👍

dany93:
Thank you once more @algefaen for these additional clarifications.

Apart from the fact that we can not have force feedback (which implies for us a subjective feeling in replacement) what is your opinion about our current settings (Spiral propwash, P-factor)? 

They are very easy to change in the code, the issue is the balance, compromise, subjectivity...
I can propose in a test branch, if you can test.
Or I can tell you how to change the values by yourself. 
The P-factor has to be changed in the FDM file, which normally needs a simulator start for each change.
The spiral propwash can be changed in flight, by manually setting a value in the Internal Properties.

Thank you for your cooperation.

PR #1272, opened 2019-04-30
Added fallback model support

Zaretto:
The fallback model (in fgdata) for the C172 is index 1; this allows use of the low detail option and avoids blue/yellow glider when the model is loading (over MP).

Also fixed an invalid XML definition; the xml tag refers to the xml version not the file version and must be 1.0

### DIFF
--- a/c172p-float-set.xml
+++ b/c172p-float-set.xml
@@ -1,4 +1,4 @@
-<?xml version="1.1"?>
+<?xml version="1.0"?>
 <!--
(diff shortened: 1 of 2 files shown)


legoboyvdlp:
Thank you - much appreciated!

wlbragg:
Yes, thank you. Sometimes the small stuff is allowed to slip through the cracks. It is very much appreciated when someone takes the time to fix it.

PR #1277, opened 2019-05-19
Load GPS only once and translate instead

### LINKED ISSUE #1255: Load GPS-Model only once
<s>It is a fact that loading the GPS-Model twice for lighting effect simplicity is costing us load time and overhead. So we need to load it only once and dynamically reposition it to the alternate "Float" variant location when called for.</s>
I had it that way originally and had trouble getting the interior shadow positioned correctly but understand it a bit better and should be able to get it close enough.

wlbragg:
Fixes #1255 

This is finished, We load the GPS model once and use a translate animation to move it if the amphibious bushkit is selected.

The interior shadow when in amphibious mode is slightly off, but not enough to warrant major effort required to fix it, if it is even possible.

If @legoboyvdlp or @gilbertohasnofb can merge this after testing I will cherry pick it to the release branch.

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -164,31 +164,48 @@
             <z-m> 0.08948</z-m>
             <heading-deg>-19.0</heading-deg>
         </offsets>
-        <condition>
-            <not>
-                <equals>
-                   <property alias="/params/bushkit"/>
-                   <value>4</value>
-                </equals>
-            </not>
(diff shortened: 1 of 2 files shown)


legoboyvdlp:
:+1:

legoboyvdlp:
Translation works perfectly here. Thanks! Sorry for the delay 😞

PR #1280, opened 2019-06-08
Add 8.33 radio

### LINKED ISSUE #1275: Add 8.33 khz support
The KX165A radio has support for 8.33 khz.
Add support for 8.33, in order to allow our aircraft to be used in Europe.
Documentation:
http://antena.fe.uni-lj.si/literaturgithub/Razno/Avionika/Kx155/Kx155a.pdf

While we are at it, we might as well simulate all the other functions of the KX165A.

@wlbragg I wonder if you could edit the 3D of the instrument to closer match the KX165A?
https://vimeo.com/177075004

Basically, the most urgent thing is to allow the radio to show three digits. This is required if we are to simulate 8.33.

The second thing would be to remove the separator between the COMM and NAV segments - the instrument should just have one LCD display.

image

image

legoboyvdlp:
Fixes #1275 

Replaces the KX165 with the KX165A radio which supports 8.33 frequencies.


legoboyvdlp:
I hope that was the right milestone?

PR #1283, opened 2019-06-19
Fix xy-plane and missing texture error messages

### LINKED ISSUE #1282: Invalid OSGText axis alignment causing error
Hi,
Julian Smith has fixed the "ignoring unknown axis-alignment'yx-plane" error, as well as some path errors in textures (which I have never seen). I'll be commiting this shortly. Should this go to the release branch also?

Please test and merge.

legoboyvdlp:
Fixes #1282 

Tested - ready to merge

### DIFF
--- a/Models/Effects/exterior/bumpspec-pontoon.eff
+++ b/Models/Effects/exterior/bumpspec-pontoon.eff
@@ -8,7 +8,7 @@
     <grain-texture-enabled type="int">1</grain-texture-enabled>
     <grain-magnification type="float"><use>/environment/aircraft-effects/dirt-fuselage</use></grain-magnification>
     <texture n="7">
-        <image>Models/Effects/exterior/dirt.png</image>
+        <image>Aircraft/c172p/Models/Effects/exterior/dirt.png</image>
         <type>2d</type>
         <filter>linear-mipmap-linear</filter>
         <wrap-s>repeat</wrap-s>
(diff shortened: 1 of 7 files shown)


gilbertohasnofb:
@wlbragg are you able to test and merge this?

legoboyvdlp:
Should this go in the release branch? I think it should if possible, but maybe its too late?

wlbragg:
If it's what I think it is then I don't think it is anything important, It's been reporting this for awhile now, I believe. But if Stuart can get it in then I have no objections.

wlbragg:
@legoboyvdlp

> I don't think it is anything important

I just looked at which labels these were and at least some of them I know were ones I added for new animated models we have introduced to the c172. Apparently the default was all that was needed in those objects because I tested that at the time and the labels were correctly positioned and orientated, so with the wrong definition I guess it took the default? So with that said, it is probably not critical. Also, if I recall the grain effect texture path currently isn't even used yet, it was put in there for future activation, so it shouldn't be critical either.
I am stating all this as I see Stuart had updated fgdata already with the newest picks to our release branch. 
So my question is, should you ask Stuart first if he will get it into fgdata and then pick it to our release, or go ahead and pick it to our release and then ask him to merge one more bug fix?

legoboyvdlp:
Thanks for merging!

Well, it may cause error reports if there are warning messages about the bad plane, so probably best to fix it for the release?

I will ask Stuart if he can get it in now.

wlbragg:
@gilbertohasnofb @legoboyvdlp 

Stuart added this manually to fgdata so I went ahead and cherry picked it to the 2018.1 release branch.

PR #1291, opened 2019-08-01
Amphibious gear switch not to trigger repair timer if pontoon damaged

### LINKED ISSUE #1290: Amphibious gear switch should not triger repair timer while  pontoon damaged
I missed this in my rush to get this in 2019.1. This fixes the behavior but there is still an issue that the amphibious gear trigger a repair when they are switched when not damaged. So if say a wing is damaged and you raised or retract the gear, it gets repaired. I can't find the trigger here. I looked at the "repairing" flag but don't see that triggering a repair, only a delay on all damages and tiedown support.
Bottom line, it's not critical and we can look at it later, I simply want to get the major bug fix in 2019.1 and this finishes that part of it.

wlbragg:
Fixes #1290 

I need to merge this and back port to our 2019.1 release immediately as I am holding up 2019.1 at remote.
I found this after one last test, after I merged the main issue https://github.com/c172p-team/c172p/pull/1289
Total apologies for doing this out of our normal procedures.

### DIFF
--- a/Nasal/physics.nas
+++ b/Nasal/physics.nas
@@ -195,8 +195,10 @@ setlistener("/fdm/jsbsim/wing-damage/right-wing", func (n) {
         killengine();
 }, 0, 0);
 
-setlistener("/controls/gear/gear-down-command", func (n) {
-    setprop("/fdm/jsbsim/damage/repairing", 1);
-    bushkit_changed_timer.restart(bushkit_change_timeout);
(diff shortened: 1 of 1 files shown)


wlbragg:
I'm not going to delete this and https://github.com/c172p-team/c172p/pull/1289 right now, anyone that wished can review and then cleanup.

PR #1298, opened 2019-08-13
Fix pitch oscillations at rest, Brakes ON, with pure crosswind

### LINKED ISSUE #1297: Strong pitch oscillations at rest, Brakes "ON", with pure crosswind
Connected to:
- this remark by @algefean https://github.com/c172p-team/c172p/issues/1130#issuecomment-399561812
- this one, https://github.com/c172p-team/c172p/issues/1130#issuecomment-403204924
 and 
- this issue (description, investigation and circumventing solution) in JSBSim https://github.com/JSBSim-Team/jsbsim/issues/89

Very rare, we were the first ones to notice it. It happens with a pure crosswind (less than +/- 1 deg, about), aircraft at rest, brakes "ON". Every JSBSim aircraft is potentially affected, but our C172p is more sensitive due to its implemented gear damage possibility.
All the more unlikely that this occurence is almost impossible in flight.

The conclusion was that it was due to an imperfection and a divergence in a JSBSim calculation.

After careful investigation, @bcoconni found no true solution, but he proposed a circumventing which works. However, this change is not in the JSBSim core code, it has to be applied to every affected aircraft. Fortunately, the change is very easy to add in.

dany93:
Fixes #1297

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -1485,9 +1485,13 @@
             </function>
 
             <function name="aero/coefficient/CLadot">
+                <!-- 'qbar-psf' has been replaced by 'qbarUW-psf'.
+                See JSBSim issue #89 Pitch oscillations at rest, Brakes "ON", with pure crosswind
+                https://github.com/JSBSim-Team/jsbsim/issues/89 -->
                 <description>Lift_due_to_alpha_rate</description>
                 <product>
-                    <property>aero/qbar-psf</property>
+                    <!-- <property>aero/qbar-psf</property> -->
+                    <property>aero/qbarUW-psf</property>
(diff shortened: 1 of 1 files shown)


wlbragg:
I haven't applied this to the Cub yet but I verified this fixes the issue in the c172p so I am merging it now.

dany93:
Thank you, @wlbragg

wlbragg:
@dany93
Sorry it took me so long to test and merge. I just applied the fix to the Cub as well, works perfectly. Nice fix, thank you!

dany93:
And thank you too @bcoconni for the code investigation and the explanations..

PR #1311, opened 2020-01-12
Issue-1309

### LINKED ISSUE #1309: Add screeching and skidding sounds to tires
I added skidding logic and visual effects to the tires awhile back. It would be nice to also add the sound effects. If one of our many audio artists could come up with a screeching on pavement and skidding on dirt sound, I would be more than happy to do the coding involved to apply it to the aircraft.

The current logic handles skidding of each individual left and right tire. So you potentially have one or two sound paths.

It isn't necessarily simple to add in the sound effect at this point, because there is now a lot of terrain logic that needs to be taken into account. On property needed will be the same property cuing the visual smoke. There will be at least one other that accounts for the ground surface.

I don't know if it would be possible to use a pitch change tied to ground hardness VS different audio files for different surfaces?

wlbragg:
Fixes #1309 

@gilbertohasnofb This is ready to test and merge if your satisfied.

Thanks again for the help.


gilbertohasnofb:
I'll test this as soon as I can if nobody beats me to it, I need to head off right now and I will be out the whole day.

gilbertohasnofb:
> Do we need to adjust the volumes of these three competing sounds, tire, rumble, engine?

@wlbragg I really like the sound levels, everything seems to be working fine. On gravel surfaces, the rolling sound is really easy to perceive, and the screech is super clear. On hard surfaces, the rolling sound is much more subtle but it is still there (it disappears when the tires leave the ground). I think this is excellent and ready to be merged. Are you missing anything or can I go ahead?

wlbragg:
@gilbertohasnofb If your happy with with sound levels we're good to go, you can merge it. Thank you for the sounds.

gilbertohasnofb:
@wlbragg thanks for implementing this too! I find the result really immersive

PR #1317, opened 2020-03-04
Starting with battery and avionic-master on for tutorial

### LINKED ISSUE #1258: Radio tutorial has no power to radio upon startup
See:https://forum.flightgear.org/viewtopic.php?f=37&t=35246&p=342952#p341392

We need to add a section at the beginning of the tutorial on how to get power to the radio stack.

ltoenning:
Fixes #1258 

As this tutorial didn't take long time I guess just using battery is enough. Otherwise we may can also start the engine.

### DIFF
--- a/Tutorials/radios.xml
+++ b/Tutorials/radios.xml
@@ -37,6 +37,18 @@ weather, the altimeter setting and the runway in use.
             <value>1</value>
         </set>
         <set>
+            <property>/controls/switches/master-bat</property>
+            <value>false</value>
+        </set>
+        <set>
+            <property>/sim/model/hide-yoke</property>
+            <value>false</value>
+        </set>
+        <set>
+            <property>/controls/switches/master-avionics</property>
+            <value>false</value>
+        </set>
+        <set>
             <property>/controls/flight/flaps</property>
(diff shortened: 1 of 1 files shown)


gilbertohasnofb:
@ltoenning I think it's important to start the tutorial with the battery and avionics off and have the players switch them on. These tutorials are particularly important for new players, and this teaches them how the radio operates. So could you consider implementing that?

ltoenning:
Sure. I just thought that normally the user would start with the first tutorials ("Preflight", "Engine Start" etc.) so they already knew how to power the aircraft up.

But I agree if someone just jumps into the "Radio Stack Tutorial" it might be confusing.

gilbertohasnofb:
Great, many thanks! I think some users might do things out of order, others might take some time remembering things, so it might be nice to include this in this tutorial. Definitely not a super deal breaker, but nice nevertheless. Thanks for implementing it!

ltoenning:
Ok will do tonight.

ltoenning:
Update pushed.

gilbertohasnofb:
@ltoenning Excellent work, thank you so much!

PR #1323, opened 2020-05-08
Turn magnetos on when using ignition key

### LINKED ISSUE #1322: Ignition switch fails when using the xml animation "Start"
I'm not sure what changed or why but the issue is if you use the interior view keyed ignition to start the aircraft it fails if you mouse over the "start" hot spot without first clicking on mag 1 and 2 .I don't think this used to be this way and I don't know that we changed any of the code in Models/Interior/Panel/Instruments/magneto-switch/magneto-switch.xml to cause this behavior. 
Auto start still works and if you first set up the aircraft systems for startup appropriately then the "s" key start still works.
~~~
    <!-- Magneto switch -->
    <animation>
        <type>knob</type>
        <object-name>key</object-name>
        <object-name>start</object-name>
        <object-name>magsw</object-name>
        <visible>true</visible>
        <action>
            <binding>
                <command>property-adjust</command>
                <property>/controls/switches/magnetos</property>
                <factor>1</factor>
                <min>0</min>
                <max>3</max>
            </binding>
        </action>
        <increase>
            <binding>
                <command>nasal</command>
                <script>c172p.click("magneto-forward")</script>
            </binding>
        </increase>
        <decrease>
            <binding>
                <command>nasal</command>
                <script>c172p.click("magneto-back")</script>
            </binding>
        </decrease>
        <hovered>
            <binding>
                <command>set-tooltip</command>
                <tooltip-id>magneto-switch</tooltip-id>
                <label>Magnetos: %s</label>
                <property>controls/switches/magnetos</property>
                <mapping>nasal</mapping>
                <script>
                    var m = arg[0];
                    if (m == 1) return 'RIGHT';
                    if (m == 2) return 'LEFT';
                    if (m == 3) return 'BOTH';
                    return 'OFF';
                </script>
            </binding>
        </hovered>
    </animation>

    <!-- Starter -->
    <animation>
        <type>pick</type>
        <object-name>click-S</object-name>
        <visible>false</visible>
        <action>
            <name>starter</name>
            <button>0</button>
            <binding>
                <property>/controls/switches/starter</property>
                <value>1</value>
            </binding>
            <binding>
                <command>nasal</command>
                <script>c172p.click("magneto-forward")</script>
            </binding>
            <mod-up>
                <binding>
                    <property>/controls/switches/starter</property>
                    <value>0</value>
                </binding>
                <binding>
                    <command>nasal</command>
                    <script>c172p.click("magneto-back")</script>
                </binding>
            </mod-up>
        </action>
        <hovered>
            <binding>
                <command>set-tooltip</command>
                <tooltip-id>starter-switch</tooltip-id>
                <label>Engine Starter</label>
            </binding>
        </hovered>
    </animation>
~~~
Or it may be a change in fgdata/Nasal/controls.nas

wlbragg:
Fixes #1322 

@gilbertohasnofb I think this is good.

### DIFF
--- a/Models/Interior/Panel/Instruments/magneto-switch/magneto-switch.xml
+++ b/Models/Interior/Panel/Instruments/magneto-switch/magneto-switch.xml
@@ -99,6 +99,11 @@
                 <value>1</value>
             </binding>
             <binding>
+                <command>property-assign</command>
+                <property>/controls/switches/magnetos</property>
+                <value>3</value>
+            </binding>
+            <binding>
                 <command>nasal</command>
                 <script>c172p.click("magneto-forward")</script>
             </binding>


gilbertohasnofb:
@wlbragg this seems to have fixed the issue, good catch and thanks for the fix!

PR #1328, opened 2020-08-02
Issue 1327

### LINKED ISSUE #1327: CG of wing tanks too far outboard
in the c172p.xml file, the CG position of tank 0 and 1 seem to be wrong. The y-position shall be 112" outboard, whereas in reality the tanks are, according to the parts catalogue, positioned between WS38.38 and WS65.125. That would result in a y-position near 51.8 inches.
On top the x position is given with 56" aft of firewall, but the POH indicates 48" on page 6-9.
The capacity is given with 185lbs eq. to 30.6gallons. Is this correct? 
Although of minor influence, one should position the float chambers tank 3 and 4 near the engine, about x=-10, y=0, z=26.

dany93:
Fixes #1327


wlbragg:
@gilbertohasnofb @legoboyvdlp @dany93 
BUMP
Anyone want to tackle testing this so we can close it up?
It looks like a lot in my list, but it's not really that much.
I am wanting someone familiar with the contamination routine to check that, but I think I got it right configuring it the way I did. :)

gilbertohasnofb:
@wlbragg could you please guide me on what needs testing exactly? I've been quite busy recently and I have not followed the conversations that were happening here in detail, my apologies.

wlbragg:
See https://github.com/c172p-team/c172p/issues/1327#issuecomment-691266507

dany93:
I started to do part of the tests after your https://github.com/c172p-team/c172p/issues/1327#issuecomment-691266507 list, but it makes a lot of states and everything is not clear.

> 4. Fuel tanks configured correctly at startup, dependent upon tanks selected with save state on
> 5. Fuel tanks configured correctly at startup, dependent upon tanks selected with save state off

With save state off, the tank content is (I guess) the default one but the selected tank at start up is the last as at exit: the active tank is saved and retrieved.
With save state on, the tank effective content is saved (if different from default). This for default or integral tank. Of course, the selected default or integral tank is saved and retrieved.

The engine number (160 or 180 hp) is not saved, but it may not be the current subject.

> 6. Fuel tanks configured correctly upon autostart, dependent upon tanks selected

"autostart" does not change the tank configuration nor the effective tank content.
Do you mean "Start with engine running?
save state off: selected tank kept (saved), effective content not saved (correct, at least for the content).
save state on: selected tank kept (saved), effective content saved (all correct).

Not sure I have tested every possible combinations because save state "ON", "OFF" needs many simulator startups, which is tedious.

> 7. Test fuel cap on wing logic (click on fuel caps on wings), dependent upon tanks selected

Display in accordance with the effective contents: correct.

> 8. Check cockpit fuel tank selection knob works correctly, dependent upon tanks selected

Both, Left, Right, Off: in accordance with Fuel and Payload checked boxes.
But on the fuel selector, 24, 24 and 48 gal are written, which is not correct.

I don't know for the rest, I let people who implemented these features test them.

wlbragg:
>With save state off, the tank content is (I guess) the default one but the selected tank at start up is the last as at exit: the active tank is saved and retrieved, which I don't know if it is expected.

Yes, I think it was decided to make the type of tanks selected at end of session persistent at startup always. Only the amount of fuel in the tank should be the "default" startup amount if "save state" is NOT active. That number should be approx. 25% of capacity of the selected tank type. If the "save state" is active then both the tank type and the amount of fuel left at the last sessions exit will be what you start with at the next session. I just verified this is the behavior I am seeing. So I think this is as expected and as designed.

>Do you mean "Start with engine running"?

No, I meant "autostart", I actually didn't even think about start with engine running. But even if we don't end up adopting @stuartbuchanan controller changes, we should merge the startup state work he did. That will eliminate the start with engine running as there will effectively be states for that. I can test that when the time comes. So I think we're good in this aspect.

>The engine number (160 or 180 hp) is not saved, but it may not be the current subject.

No it is not the current subject, and I noticed it isn't saved with the "save state" active either.
I think it is designed to be tied to the "variant" selected. That is one of the issues of having something like "Save State of the Aircraft" and being able to select variants. Variants should always override the saved state, but then what use is having the saved state option? Something we can discuss some other time I guess.

>Both, Left, Right, Off: in accordance with Fuel and Payload checked boxes.
But on the fuel selector, 24, 24 and 48 gal are written, which is not correct.

Ah, good catch, @gilbertohasnofb would you have time to make the correct textures for the new tank sizes. We need textures for both...
21.5 left - 21.5 right and 43 both
and
34 left - 34 right and 68 both

This is what we have now
fuel

gilbertohasnofb:
> Ah, good catch, @gilbertohasnofb would you have time to make the correct textures for the new tank sizes. We need textures > for both...
> 21.5 left - 21.5 right and 43 both
> and
> 34 left - 34 right and 68 both
> 
> This is what we have now

@wlbragg sorry, I missed this message. Sure, I will do that and upload it shortly. Can you confirm to me that the 24/48 texture is not going to be used any longer?

gilbertohasnofb:
@wlbragg done and pushed. New textures are in `/Models/Interior/fuel-selector` instead of just `/Models`. They are named according to the fuel tank size. Let me know if this works.

wlbragg:
> Can you confirm to me that the 24/48 texture is not going to be used any longer?

I'm not sure what your talking about, let me look at what you gave me and then maybe it will make sense.

wlbragg:
@dany93  @gilbertohasnofb  @legoboyvdlp 
This is complete and ready for final review and merge.

gilbertohasnofb:
> I combined the two textures so I could use texttranslate for the animation.

Looks good, I updated the Gimp files in the other repo too.

> This is complete and ready for final review and merge.

I'm afraid I won't have too much free time in the next days to check this. @dany93 would you be able to do a final test and merge this please?

wlbragg:
@dany93  Sorry for the mix up. I made an assumption as to where that new texture belonged and got it wrong. It belongs with the main model textures and not an independent directory as there is no additional xml associated with it.
It should be good now.

>As nobody else brought remarks, I will merge it when the texture issue is solved, but it will be with that I was able to test. Agreed?

If no one else remarks with testing input to this within 24 hours as far as I am concerned, go ahead.

dany93:
Good for the fuel selector textures.
Merged.

gilbertohasnofb:
Thanks a lot for this work!

PR #1329, opened 2020-08-06
Repair aircraft at the start of each tutorial

legoboyvdlp:
Fixes the ticket just opened on sourceforge.

### DIFF
--- a/Tutorials/altimeter.xml
+++ b/Tutorials/altimeter.xml
@@ -13,6 +13,18 @@ The source of this altimeter setting varies. At high altitudes, a standard press
 
 This tutorial will teach you how to set the altimeter based on the ATIS (Automatic Terminal Information Service) for Hilo Intl Airport
     </description>
+	
+    <nasal>
+        <script>
+            c172p.repair_damage();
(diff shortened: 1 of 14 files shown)


wlbragg:
@legoboyvdlp does this fix the stuck brakes?

legoboyvdlp:
I didn't see stuck brakes?

This is about the problem where a user left the radios on too long and drained the battery entirely during a tutorial. This now does a full repair at the start of each tutorial.

wlbragg:
Ah, ok let me see what effect if any this has on the other issue with the parking brake. Thanks for fixing this issue. I'll merge as soon as I test it.

wlbragg:
Nice, this actually fixed the brake issue for tutorial relocations. The only brake issue now is when doing and Airport relocation. Thanks @legoboyvdlp

onox:
@legoboyvdlp I see a lot of repeated code. You could put it all in one function (something like `c172p.repair_aircraft_for_tutorial()`) so you only have to modify 1 location if you ever need to modify something.

PR #1331, opened 2020-08-15
limiting AI offset knob

### LINKED ISSUE #1237: AI knob rotates infinitely
AI knob rotates infinitely (in both directions)

image
image
image
image
image
image

gilbertohasnofb:
Closes #1237

### DIFF
--- a/Models/Interior/Panel/Instruments/AI/AI.xml
+++ b/Models/Interior/Panel/Instruments/AI/AI.xml
@@ -139,7 +139,7 @@
             <binding>
                 <command>property-adjust</command>
                 <property>instrumentation/attitude-indicator/horizon-offset-deg</property>
-                <factor type="double">0.5</factor>
+                <factor type="double">0.2</factor>
                 <min>-7.0</min>
                 <max>2.0</max>
             </binding>
--- a/Models/Interior/Panel/Instruments/AI/AI.xml
+++ b/Models/Interior/Panel/Instruments/AI/AI.xml
@@ -140,6 +140,8 @@
                 <command>property-adjust</command>
                 <property>instrumentation/attitude-indicator/horizon-offset-deg</property>
                 <factor type="double">0.5</factor>
+                <min>-7.0</min>
+                <max>2.0</max>
             </binding>
         </action>
         <property>instrumentation/attitude-indicator/horizon-offset-deg</property>
@@ -158,7 +160,7 @@
             <binding>
                 <command>set-tooltip</command>
                 <tooltip-id>ai-horizon-offset</tooltip-id>
-                <label>Horizon Offset: %2d</label>
+                <label>Horizon Offset: %.1f</label>
                 <property>instrumentation/attitude-indicator/horizon-offset-deg</property>
             </binding>
         </hovered>


gilbertohasnofb:
@wlbragg Ok, so I found where the max and min limits of the AI offset knob were missing. I could not find any data online about it, and from the photographs of the instrument the range seems to be quite limited, and thus I had to estimate some reasonable values (so please test and let me know). 

Also, it really bothers me that the value of 0.0 does not correspond exactly to where the horizon line is exactly in the centre, it is around -2.5. Is this a problem with the model? We could add a filter to fix this but that is not the best solution I think. Otherwise, this is ready to go.

dany93:
@gilbertohasnofb wrote
> Also, it really bothers me that the value of 0.0 does not correspond exactly to where the horizon line is exactly in the centre, it is around -2.5. Is this a problem with the model?

_(just guesses; I've never used this instrument)_

I guess your value -2.5 is for the horizon-offset-deg on the ground (where the aircraft has a 4.6 deg pitch-deg). II think it is not a significant indication..

However, in straight and horizontal flight, the artificial horizon is close to the center.
Particularly, around 105 kt (2250 RPM) with the 160 hp, where `/instrumentation/attitude-indicator/indicated-pitch-deg` is close to zero deg. There is a small shift, but only close to 0.5 deg.

The AP can help for these tests.

I observe:
/instrumentation/attitude-indicator:
- indicated-pitch-deg,
- horizon-offset-deg

/orientation:
- pitch-deg

gilbertohasnofb:
@dany93 Thanks Dany, that's good to know. I think we can leave these arbitrary values as they are then, as I don't think they are representative of degrees or anything like that. I will probably make the knob a bit finer then since these variations in flight are so small.

gilbertohasnofb:
@dany93 @wlbragg I pushed a smaller modification, the range is the same as before but I used a finer control for the offset. It feels (and looks) much better. If you like it, this is ready to be merged I think.

wlbragg:
OK, thanks. I have some real life distraction that are going to keep me busy for awhile so it may take a few days before I have a chance to review this, and anything else for that matter. But I'll look at it soon and merge if @legoboyvdlp doesn't beat me to it.

wlbragg:
@gilbertohasnofb I looked at this and it looks good, thanks.

dany93:
@gilbertohasnofb wrote
> I think we can leave these arbitrary values as they are then, as I don't think they are representative of degrees or anything like that

After some more thinking, I'm wondering if the offset button really has a "notch" to sense "a zero" or if the pilot sets the artificial horizon for a given pitch angle (more usually in horizontal flight at a given velocity, but possibly for a climb) and uses it as a temporary reference to keep his pitch angle.

As the instrument has neither offset graduation, nor pitch angle graduation (although we can read them in FG properties) the difference between the aircraft pitch angle and the offset needed to compensate it might be of no importance.

gilbertohasnofb:
@dany93 Yes, I can imagine this having a notch at a reference zero in real life, but we cannot feel this in the sim. So setting the default position (currently -2.5) to zero could help with this. But I think this should be done model-wise, a filter to offset the property by +2.5 would feel quite hacky to me. @wlbragg any opinions on this?

dany93:
I'm not sure to understand well what you are meaning.
The default position of the instrument horizon (blue) at start? Currently, it indicates a pitch up at rest on the ground. Which is true considering the physical aircraft. It is centered with `horizon-offset-deg` = -2.5. Would you like it to be centered with the instrument at start with `horizon-offset-deg` = 0?

Personally, I don't think it is a good thing. Currently, it is set such as it is close to centered in cruise flight with zero pitch angle. Which is logic IMO. Setting it as you say would make it indicate a pitch down at zero pitch angle (average cruise). Which would need a (re-) setting by the pilot at cruise.

Otherwise, I don't think that a filter or changing in the 3D model is needed to get what you wish. I think that a mere offset of the animation would do it.

NB: Quantitatively, it's difficult to speak "angle values" as long as we do not know the relation between the horizon deviation on the instrument and the real pitch angle (variation) of the aircraft. However, this can be left as it is for now.

wlbragg:
I almost asked the question if we were OK with the starting value of -2.5 before I merged but I read in a prior post the same information @dany93  provided about it being close to 0 when in cruse (level flight).
>Currently, it is set such as it is close to centered in cruise flight with zero pitch angle. Which is logic IMO.

Even though a total guess om my part, it make logical sense the vehicle pitch offset when on the ground could cause this -2.5. You would want the AI horizon to match during level flight and the -2.5 offset at startup achieves that.

gilbertohasnofb:
Apologies, I had misunderstood what @dany93 had initially written. I think the key is this:

> In straight and horizontal flight, the artificial horizon is close to the center.

In this case, there is nothing else needed to be done, particularly due to this:

> Currently, it is set such as it is close to centered in cruise flight with zero pitch angle. Which is logic IMO. 

So it's all good. @wlbragg thanks for merging.

Issue #1350, opened 2021-01-25
Typo in Nasal/engine.nas oil service hours?


D-ECHO:
Hi there,
while trying to adapt your engine simulation to the Cessna 150L, I noticed what seems to be a typo/bug to me:
https://github.com/c172p-team/c172p/blob/master/Nasal/engine.nas, l.141ff.
```
service_hours = getprop("/engines/active-engine/oil-service-hours");
service_hours_increase_qph = 0.00020 * math.pow(service_hours, 2);
service_hours_increase_qph = std.min(1.5, service_hours_increase); # limit increase to 1.5 (at which point you really should think of changing it)
service_hours_increase_qps = service_hours_increase_qph / 3600;
```
In this segment, servie_hours_increase is limited to 1.5 using std.min(), but service_hours_increase is only once set 0 (l.99), never touched again later. I think what this is meant to say is service_hours_increase_qph both in std.min() as well as in the declaration/initialization in line 99.

Regards

dany93:
Hi @D-ECHO 

This part has been done by Jonathan Redpath.
@legoboyvdlp do you read this?

legoboyvdlp:
Yes that seems correct, if someone wants to commit that fix?

D-ECHO:
See PR #1351 @legoboyvdlp @dany93

dany93:
@D-ECHO, sorry, I don't understand the code, don't know what to check in the code and how to check the functioning in the sim.
@legoboyvdlp, can you confirm that it is correct and merge it please?

PR #1351, opened 2021-01-26
Fix typo in engine.nas concerning oil service hours increase

D-ECHO:
Fixes issue #1350


legoboyvdlp:
Verified ok; as the service hours increase, oil consumption increases. Thanks!

dany93:
Thank you @D-ECHO and @legoboyvdlp .

PR #1354, opened 2021-03-01
Fix iCCP Color Profile

wlbragg:
@legoboyvdlp 
This is a simple image swap from Scott at mailing list. In case you didn't see it he updated these four (4) images for iCCP Color Profile errors in the logs.

I think it is safe to go ahead and get this merged into master.


legoboyvdlp:
Looks good here!

PR #1367, opened 2021-06-27
Issue 1358

### LINKED ISSUE #1358: Parking brake setting for joystick doesnt work.
Setting a joystick button for parking brake, will not switch the parking brake.

wlbragg:
Fixes #1358 

@Megaf please test this and se if this fixes your issue without any intervention of the code or JS bindings

@dany93 maybe you could check this out as well and verify it fixes the issue as you know it.

@legoboyvdlp after it is tested by @Megaf and @dany93 would you take a look at it and make sure I didn't miss any unforeseen consequences of the logic change? You could then merge it.


dany93:
> maybe you could check this out as well and verify it fixes the issue as you know it.

I don't know what to check, The Parking Brake always worked for me (using the keyboard shift-B or by clicking on the lever).
It stills works correctly by using the keyboard shift-B or by clicking on the lever (checked by Internal properties and brake effect).

In the console: (5 such lines)
`[ALRT]:nasal      ERROR: Cannot add listener to tied property /controls[0]/gear[0]/brake-parking[0]`

But there is a new issue: **normal brakes have no braking effect**.
Despite properties correctly changing in controls/gear/brake-left and brake-right to 1 and 0. Squeaking is heard too.

I still have FG 2020.3.0 but I don't think this is an issue for this.

I'm surprised to see that getting such a basic effect is so complicated (parking brake in FG + model animations...).

dany93:
Everything works.
Thanks @wlbragg,
And @legoboyvdlp for the normal braking fix.

wlbragg:
Hum, @legoboyvdlp why did that break the normal braking when I didn't touch anything to do with the left or right brake (/controls/gear/brake-left or /controls/gear/brake-right) other than add the fcs function to catch the parking brake flag?

legoboyvdlp:
@wlbragg Specifically because you were outputing to the left and right brakes but the only input was from the parking brake; not from the normal brake.

PR #1369, opened 2021-06-28
Issue 1344

### LINKED ISSUE #1302: Upgrade KAP-140 to new version
@WoodSTokk brought it to my attention, we have a newly refactored KAP-140 that is fully functional. We should consider replacing the existing KAP 140 with this newer version.

wlbragg:
Fixes #1302


wlbragg:
@legoboyvdlp @dany93 and anyone else that wants to help, This is read to "thoroughly" test. I will also spend time testing this.

legoboyvdlp:
OK. What's the order of merging. Is this linked to the FG1000 at all?

wlbragg:
This would be the next iissue to merge. No it's not tied to the fg1000 at all.

legoboyvdlp:
So I don't know if I'm using it wrong; but when you engage the autopilot isn't the wing leveller supposed to hold the wings level. I engaged it in the climb out, and the servos did not seem to take over ... the plane just went into a left roll when I centered the controls.

wkitty42:
On 6/29/21 12:12 PM, legoboyvdlp wrote:
> So I don't know if I'm using it wrong; but when you engage the autopilot isn't 
> the wing leveller supposed to hold the wings level.

yes, it is supposed to... that's how/why i always use it when i fly the c172p...

> I engaged it in the climb out, and the servos did not seem to take over ...
> the plane just went into a left roll when I centered the controls.
i haven't tried in a (long) while... which version of the craft are you flying?

wlbragg:
It seems to be working on mine. As soon soon as I turn on the AP  I loose control of the elevator and the ailerons. The first time I did this today when testing, I was on the ground and noticed that the elevator trim was maxed out at.999. Had that happened in the air I would have surely stalled and I assume it would appear to just roll out to the left or right depending on my exact aircraft orientation. @legoboyvdlp we may have to get help testing from those intimately familiar with the KAP140.

wlbragg:
@legoboyvdlp hold off for now with the testing until I find out why @reiszner @hbeni are wanting to make so many code changes to the KAP. I thought it was ready to go with maybe a small change here or there. But the there are massive changes being proposed an I don't know why that is.

hbeni:
> I was on the ground and noticed that the elevator trim was maxed out at.999.

That is probably expected, and part of the preflight check:
When you are on the ground, wanting to use the AP, you need to verify its function on the ground.
When you activate the AP and change the VS setting, that should make the trim wheel go. As the logic will never reach a satisfied VS-rate, the trim will go until you either disconnect the AP, or reset the VS setting to `0`.

> But the there are massive changes being proposed an I don't know why that is.

You now do :) (see other thread)

wlbragg:
@hbeni  @reiszner

>When you activate the AP and change the VS setting, that should make the trim wheel go. As the logic will never reach a satisfied VS-rate, the trim will go until you either disconnect the AP, or reset the VS setting to 0

That is exactly what it is doing. So how do you successfully use the VS option? You can't set it to +100 and expect it to keep the aircraft climbing at +100 VSI? That was how it used to work.

I set the BARO to the altitude indicated and got airborne, then hit the vs +100 and it trims to the limit instead of +100 ASI. Are you telling me your supposed to turn it of when your VSI indicates +100 ASI?

That makes no sense, if I am on the ground and I want to preset the VS to +100 then it starts the trim wheel turning and you have no way to know when to stop it because the ASI indicator is not functioning on the ground.

wlbragg:
Hum, well it worked this time once I was airborne I was able to successfully set it and it followed the settings. You have to be spot on with your power settings or you can quickly get into trouble.

wlbragg:
OK, one more thing, the digital indicators on the autopilot are not tied to our panel lighting. The buttons are but not the digits or text. It used to be tied using the same property, Did that change? Are you using a different property for the digits?

Maybe referring to this...
>I had now a look into the material animation and there are <property> and <factor-prop>.
I know we can use <property alias="…"/>, but is this also possible with <factor-prop alias="…"/> ?

and this...
>Hm. Maybe just make a custom property in the /settings/ subtree of the autopilot, so aircraft devs can wire that, instead?

hbeni:
Will do some test flights with this branch and report back!
Thanks for merging :) I'm glad that this effort was going so fast (however it ate my evening, haha) :)

hbeni:
Is it normal that the radio lights knob control the RPM and Oil lighting?
grafik

hbeni:
**Tested OK!** :ok: :ok_hand:
Looks done for me :) The lights are okay, too :) :tada: 

----
**Details**

Tested Vertical modes:
(btw i didn't had the stalling issues you mentioned - but you have to be stable+trimmed (as IRL!) before activating the AP)
- VS mode
- Alt intercept and ALT hold
  - ALT adjustment: increments (button push)
  - ALT adjustment: increments (button hold)
- GS following

Tested Lateral modes:
- ROL
- ROL, intercept VOR (all-angle-intercept)
  - shallow angle
  - steep angle
- HDG
- HDG intercept VOR (45° intercept)
- NAV
- REV
- APR

Tested special stuff:
- AP Disconnect (`SHIFT+D`)
- AP CWS (`d')
- AP G-force disconnect

wlbragg:
@hbeni
>Is it normal that the radio lights knob control the RPM and Oil lighting?

Yeah, it's from the POH. I guess they were the most important gauges and are the only gauges that had integral lighting (material animations). The rest of the lights only had post lighting (lightmaps). But before we had the lightmaps everything was just using integral to simulate both, the two with integral and the rest with post lighting. In the shader off model mode, or  low spec rendering, it's all integral.

>but you have to be stable+trimmed (as IRL!) before activating the AP)

You know, that make total sense. It can be deadly if your not or you use it wrong. I wonder if you get the importance of that from the POH or manual. @legoboyvdlp I can verify this is the exact scenario I caused it to fail with, by trying to use it at takeoff and again trying to use it in the air before I was in a "trimmed and stable" mode.

Thank you so much for all your help. After seeing your cleanup of my attempt to implement this, I saw the mistakes I made and I probably would have found it, but it would have cost me more than a night I'm sure, ha!

Not to mention all the work prior in the c182s and @reiszner for the original conversion. We've been wanting this upgrade for a long time.

So once again thank you all for the fantastic effort.

hbeni:
It was fun to work with you all on this!
And the new KAP already reached three planes, afaik!

legoboyvdlp:
OK... I don't know much about it but I can't see anything obviously wrong, and it works as expected for me, a novice with the KAP140. Happy to merge!

Octal450:
> It was fun to work with you all on this!
> And the new KAP already reached three planes, afaik!

Likewise, a real pleasure it was :) Looking forward to future collaborations.

Kind Regards,
Josh

wlbragg:
Thanks @legoboyvdlp 
Next , see: https://github.com/c172p-team/c172p/pull/1366

reiszner:
Also from my side, it was a pleasure to work with you all.
I think i had said it allready but i cant say it enough times.
Big thanks to Josh (Octal450) for the excellent control loops and
Benedikt (hbeni) to nitpick me on the smallest contradictions about the logic.
This update is also a big step toward realism for the default aircraft of FGFS.

PR #1379, opened 2021-07-07
Allow default action in viewfrom in mouse manager

### LINKED ISSUE #1373: Mouse scroll to change zoom doesn't work anymore.
As the title says. You can no longer change FOV with the mouse wheel.

wlbragg:
Fixes #1373

Adds back in the default mouse action when in a viewfrom view mode.

@Megaf 
Please pull and test this.

@legoboyvdlp after @Megaf tests and confirms this works, would you mind testing what you can as far as normal views you use and if satisfied, merge and cherry pick again to 2020.3.
Thanks

### DIFF
--- a/Systems/views.xml
+++ b/Systems/views.xml
@@ -45,10 +45,6 @@
                     <value>lookat</value>
                 </equals>
                 <not-equals>
-                    <property>/sim/current-view/type</property>
-                    <value>lookfrom</value>
-                </not-equals>
-                <not-equals>
                     <property>/sim/current-view/name</property>
                     <value>Tower View</value>
                 </not-equals>
--- a/c172p-mice.xml
+++ b/c172p-mice.xml
@@ -43,6 +43,19 @@
                         gui.popupTip(sprintf("%d meter", abs(getprop("/sim/current-view/z-offset-m"))));
                     </script>
                 </binding>
+               <binding n="2">
+                    <condition>
+                        <and>
+                            <not>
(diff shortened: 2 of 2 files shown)


Megaf:
@wlbragg This fixes my reported issue yes. Now it is possible to control FOV/Zoom/Distance with right click + scroll.

BUT! It seems like "distance" is in use for all or at least most of the camera views, including tower views, should we open another issue for that?

wlbragg:
>BUT! It seems like "distance" is in use for all or at least most of the camera views, including tower views, should we open another issue for that?

Oops hit a wrong button.
This is subjective and in my opinion requires some discussion. This view manager has been there for a long time with no complaints. About the time we change it then we will get complaints for that. As far as I know your the only one who brought up an issue with the manager and that issue is corrected now, without causing adverse side effects to the view manager. I'm not going to tell you not to make an issue, but I doubt I will be involved with it unless I am convinced of its merits.

wlbragg:
@legoboyvdlp this is ready for a last test and if satisfied, merge and cherry pick again to 2020.3.

legoboyvdlp:
@Megaf I don't know if I'm not seeing it or what but I'm not seeing that distance is in use for zoom in views. It still changes FOV for me (x / X keys). Which, is a good thing. Can you confirm?

I can't test this @wlbragg unfortunately as I have a touchpad; not a mouse. OK to just merge it?

wlbragg:
Yeah, go ahead. I'm confident in what it does.

PR #1385, opened 2021-10-25
Add multi-key support for FG1000

PlayeRom:
As we have at https://wiki.flightgear.org/FG1000#Multikey_Support, the FG1000 should support multi-keys.


wlbragg:
Thank you PlayeRom!

PR #1392, opened 2021-11-18
Remove airport-id from tutorials (Issue #1387)

### LINKED ISSUE #1387: Taxing tutorial not starting in the correct location
dloeckx said...

Not sure if this is the right place to comment... It took me quite some searching, but I found out that my plane isn't positioned at the location indicated in the xml at the start of the taxiing tutorial, yet - as mentioned above - on the runway. Using Flightgear 2020.3.11 on a Macbook Pro (M1 chip), yet I had the same issue on an older (early 2015) Macbook Air a few months ago. Happy to help debugging!

dloeckx:
Fixes #1387


wlbragg:
Thank you @dloeckx

PR #1395, opened 2021-12-06
Fuel selector objects pointing at wrong effect file

### LINKED ISSUE #1394: Fuel selector pointing to wrong effect file in default variant

wlbragg:
Fixes #1394 

@legoboyvdlp if you don't mind, when you get a chance please review, merge and close. It's minor but I can't believe we never noticed.

### DIFF
--- a/Models/Effects/interior/lm-interior-fuel-selector.eff
+++ b/Models/Effects/interior/lm-interior-fuel-selector.eff
@@ -1,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<PropertyList>
+    <name>lm-interior-fuel-selector</name>
+    <inherits-from>Aircraft/c172p/Models/Effects/interior/c172p-interior</inherits-from>
+    <parameters>
+        <texture n="3">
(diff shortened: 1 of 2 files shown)


legoboyvdlp:
Wait, if you re-added lm-interior-fuel-selector.eff but changed the object to inherit from <inherits-from>Aircraft/c172p/Models/Effects/interior/lm-fuel</inherits-from>, how does that work? Am I missing something?

wlbragg:
Yeah, I first pointed it to what I thought was supposed to be the correct effect file. Then I realized the issue wasn't that it was  pointing to the wrong effect file at all. It was that the effect file it was pointing to was missing. So in the second commit I reversed the first commit and added the missing effect file.
Let me know if that doesn't make sense to you.
So, first commit was wrong, second commit reversed the first commit and added the missing effect file.

legoboyvdlp:
OK - thanks - I'll try and review this tomorrow morning

wlbragg:
Thank you @legoboyvdlp

PR #1415, opened 2023-05-04
Fix hpa displayed values at KAP140 on c172p

jimishol:
Fixing rather a typo so as the right value of hpa to be displayed when flying a c172p of release 2020.3


wlbragg:
@jimishol 
Hi, thanks for catching this. I see the typo of hpa-digit3 -> hpa-digit4 and hpa-digit4 -> hpa-digit5, but I don't understand the reverse ordering of the digits. Why do we need to reverse the decimal order? Did you actually look in the Blender file and see the logic behind this? I'm not saying your wrong, just hoping you could explain it to me as I am not seeing it? In fact, even though I totally see the typo in the code, my kap140 hpa appears to be functioning correctly when I dial it? I'm totally not understanding something here.

PR #1416, opened 2023-05-18
FDM: decrease pitch up tendency at full thrust.

### LINKED ISSUE #1414: Pitch instability
Per https://sourceforge.net/p/flightgear/codetickets/2803/

>The current c172 seems to have pitch instability (tends to hunt up and down, even when properly trimmed). This may be a contibuting factor to takeoff stalls. Movement of cg forward from 39.06
to 37.06 ameliorated the problem.
Also elevator response seemed abrupt. Changing flight_control name="FCS: c172
from <time>0.15</time>[code]to <time>0.30</time>
reduced the abruptness.
Kindly review.
Thanks

@dany93 would you mind reviewing this if you have the time?

dany93:
By AERORP, CoG, propeller height. Slow down elevator kinematics.

Fixes #1414


dany93:
Thank you @wlbragg

PR #1420, opened 2023-07-08
Revert "Material animation and effects definitions come after knob an…

wlbragg:
…imations"


wlbragg:
I messed up, I didn't refresh my browser and had not seen your merge. I got some very strange results when responding and creating a post. I got confused and reverted the wrong thing.

PR #1425, opened 2023-07-18
Track engine hours per engine

### LINKED ISSUE #1424: Hobbs calculates both engine hours combined
The Hobbs meter is calculating both engines as one engine. The engine change-outs are meant to be different aircraft and engine hours should be calculated separately.

wlbragg:
Fixes #1424

@dany93 please test.

The only other thing I would like to verify is if we can delete the unused filter in Systems\instruments.xml. Starting at line 189.
~~~
<filter>
        <name>Engine Total Hobbs Hours</name>
        <type>gain</type>
        <input>
            <expression>
                <div>
                    <sum>
                        <property>/sim/time/hobbs/engine[0]</property>
                        <property>/sim/time/hobbs/engine[1]</property>
                    </sum>
                    <value>3600.0</value>
                </div>
            </expression>
        </input>
        <output>
            <property>/instrumentation/clock/hobbs-meter-hours</property>
        </output>
    </filter>
~~~
If is is indeed not being used anywhere I want to delete it before merging.

@wkitty42 maybe you can help verify it is not being used anywhere?

### DIFF
--- a/Nasal/engine.nas
+++ b/Nasal/engine.nas
@@ -37,8 +37,18 @@ var update_hobbs_meter = func {
     # This uses minutes, for testing
     #hobbs = hobbs / 60.0;
     # in hours
-    hobbs = (hobbs_160hp + hobbs_180hp) / 3600.0;
+    #hobbs = (hobbs_160hp + hobbs_180hp) / 3600.0;
     # tenths of hour
+
+    var hobbs = 0;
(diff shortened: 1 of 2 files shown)


wlbragg:
@dany93 
This issue is ready to merge after review. There is no reference to the removed filter in any of the code. So I felt it was safe to remove.

PR #1440, opened 2023-08-12
Refactor tiedown logic

### LINKED ISSUE #1383: Tiedowns make the plane jump around
Hi there,
we found an issue with the tiedowns at the c182s (which we ported over initially from the c172p).
Under some circumstances the plane starts to oscillate wildly. We found out that this is linked to FDM framedrop which will mess up the PID that calculates the forces.

For the c182s we switched the force calculation to use a simple `<table>`.
Additionally we refactored the tiedown-breakage code from nasal to XML.
See the changes here: https://github.com/HHS81/c182s/pull/447.

I think they are quick and straightforward to port back here, only the lat two commits seem to apply here:
- https://github.com/HHS81/c182s/pull/447/commits/7dc7e9c16c60c70b7a80129db3653f0e0183a13e
- https://github.com/HHS81/c182s/pull/447/commits/395ed94a8bdc0c4205d5d3e61f3d71a45da9fa56

wlbragg:
Fixes #1383

### DIFF
--- a/Systems/c172p-ground-effects.xml
+++ b/Systems/c172p-ground-effects.xml
@@ -1942,79 +1942,69 @@
             <input>/sim/model/c172p/tiedowns/left/length</input>
             <input>-/sim/model/c172p/tiedowns/left/ref-length</input>
         </summer>
-
+        <fcs_function name="fcs/tiedown-left-magnitude">
+            <function>
+                <product>
+                    <property>/sim/model/c172p/securing/tiedownL-visible</property>
+                    <table>
+                        <independentVar>fcs/tiedown-left-mag-error</independentVar>
+                        <tableData>
+                           -1.0   -50.0
+                           -0.1     0.0
+                            0.0     0.0
+                            0.1     0.0
+                            1.0   7000.0
+                        </tableData>
+                    </table>
+                </product>
+            </function>
+            <output>external_reactions/tiedown-left/magnitude</output>
+        </fcs_function>
         <summer name="fcs/tiedown-right-mag-error">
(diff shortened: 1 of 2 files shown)


wlbragg:
@dany93 

This is straightforward from c182s team. It seems to tone down the intensity and frequency of the tiedowns causing the plane to jump. It also adds a check for strain on the tiedowns and disengages them if excessive force is reached.

Ready to merge

dany93:
Merged.
However, blindly done because I couldn't see the aircraft "jumps" with tiedowns, even in the previous c172p version.

wlbragg:
That's ok, It was hard for me to get it to "jump". I cranked the wind up and maybe got something similar. The new code from the c182 allows for the ties to break under strain so it should eliminate any chance of violent or strange stretching of the tiedowns.

wkitty42:
On 8/12/23 11:41 AM, wlbragg wrote:
> That's ok, It was hard for me to get it to "jump". I cranked the wind up and 
> maybe got something similar. The new code from the c182 allows for the ties to 
> break under strain so it should eliminate any chance of violent or strange 
> stretching of the tiedowns.

maybe i/we could test like i used to do when tiedowns (and lightning) were first 
added?

sit at KSFO, select the stormiest weather scenario from the list, and hold on... 
when i did this back then, i was flipped up in the air and upside down on top of 
the passenger terminal where i was parked... this is also when the "repair" menu 
appeared because it was so painful to have to restart the sim all over again to 
recover :)

wlbragg:
Thanks, that is more or less what I did but just wind and at a random airport. The logic that allows for the ties to break keeps any of that from happening now.

Issue #1398, opened 2022-03-09
Add a winter kit (cowl grill), expand icing effects


DaSovietPotato:
Today I was flying around in the rocky mountains and I faced a comical problem, the plane engine is unable to keep a decent temperature. Normally this wouldn‘t be a problem within the simulation, however my engine died on landing and I could not get it started to clear the runway until I used the external heat. Normally in real life, cowl blocks would be used to reduce the amount of air flowing through the engines to allow them to stay warm. I think that It would be a good option to add Cowl Blocks somewhere in the aircraft options menu.

wlbragg:
Hi @DaSovietPotato 

I think the description of what you explain here may be a bit off. When you say "until I used the external heat", are you talking about the carb heat control? If so this is the correct solution. On a long approach in icing conditions it is recommended to use the carb heat. It prevents the carb from icing.
The c172p simulation only simulates that condition. I don't remember off hand the exact logic of the code that simulates this.

My questions are...

In Real Life...

Does warming the engine in this manner (using airflow restrictions/cowl blocks) keep the carb from icing.
Why is this method used VS the carb heater?

It seems a bit risky to me to use such a method to control and issue that the carb heater was designed to prevent.

Are you a real-life pilot with experience in this issue? Can you enlighten us?

DaSovietPotato:
Hello, when I was referring to external heat I was referring to the external heat in the ground services tab. As far as flying was concerned, it was not humid enough for carburetor icing to occur but simply cold. The use of cowl blocks is in no way intended to replace carb heating as they accomplish related but completely different tasks. The use of cowl blocks is normally used to ensure that the engine is able to stay within the operating range on extremely cold days, they are commonly used on air cooled engines in places like Canada. A similar device exists for the 2CV to allow for the heater to work. The use of cowl plugs would indirectly help the carb heater to work more efficiently because of the higher engine temperature. I am not a pilot, however I have experience working in cold temperatures where similar measures to cowl blocks are usually taken. I hope that this answers some of your questions about the matter.

wlbragg:
Hi @DaSovietPotato 

Are they used while flying ore strictly on the ground?

I'm not very familiar with the ground services tab, but I will take a look at it. I was not aware that the ground services were temperature dependent.

I don't think it would be difficult to add functionality to account for extra heat because of air flow restrictions.

DaSovietPotato:
[...]

PR #1445, opened 2023-08-12
Change cowl plug condition logic

### LINKED ISSUE #1444: Cowl plugs are non-operable
The cowl plugs option in the GUI isn't working because the conditions for using it are not met. There are two conditions
engines/active-engine/running = false
engines/active-engine/rpm = 0

Even under cold and dark aircraft conditions the rpm is reading positive. It is actually winding up slowly starting somewhere around 0.00273...... 

So to repair the bug, engines/active-engine/rpm = 0 needs to be changed to engines/active-engine/rpm< 1

wlbragg:
Fixes #1444 

@dany93 

This is a bug fix that I will cherry pick back to the new 2020.4.1 release. I plan to investigate a little more in issue #1398
But for a simple bug fix for the 2020.4.1 release, this is adequate.

HOLD - don't merge yet.

### DIFF
--- a/gui/dialogs/aircraft-dialog.xml
+++ b/gui/dialogs/aircraft-dialog.xml
@@ -241,10 +241,10 @@
                         <not>
                             <property>/engines/active-engine/running</property>
                         </not>
-                        <equals>
+                        <less-than>
                             <property>/engines/active-engine/rpm</property>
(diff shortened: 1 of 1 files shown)


wlbragg:
Handeling in #1398

PR #1448, opened 2023-08-24
Clarify alt static port configuration, change response time of interior static port

### LINKED ISSUE #1447: KAP140 uses same static port as the exterior port, configure static ports
@hbeni  @HHS81  @dany93 

Does the KAP140 have its own internal sensors? If so then it should not be using static[0]/pressure-inhg[0] as it appears to be. It should use it's own isolated port.
Or the aircraft instruments and the icing system should use static[1]/pressure-inhg[1].

It appears we are adding a "mythical" interior static port static[1]/pressure-inhg[0] to the system when in reality that should be the port the AP is using.

Otherwise the AP is also being iced during an icing condition and we never switch it off the bad port as it is hard tied to static[0]/pressure-inhg[0]

We use /systems/static-pressure-inhg-selected for instruments but as far as I can tell we are still using static[0]/pressure-inhg[0] for the KAP140.

Also  "Alternate source is from cabin, see POH p. 7-42" does anyone know where this reference came from and exactly what it is referring to? Is it meant to be the AP or something else?

See
~~~
    <static>
        <name>static</name>
        <number>0</number>
        <tau>0.1</tau>
        <!-- Side-slip angle influence only. Type 2 will also give AoA
             influence. See Docs/README.systems.
        -->
        <type>1</type>
        <error-factor>0.2</error-factor>
    </static>

    <static n="1"> <!-- Alternate source is from cabin, see POH p. 7-42 -->
        <name>static</name>
        <number>1</number>
        <tau>0.1</tau>
        <type>0</type>
        <error-factor>0.35</error-factor>
    </static>
~~~

wlbragg:
Fixes #1447

@dany93 this is a documentation clarification change, it shouldn't affect anything performance wise.

I'm trying to test the value with and without the error for type=0. I am getting different results but don't know if it is related to the error value o something else.

### DIFF
--- a/Systems/systems.xml
+++ b/Systems/systems.xml
@@ -42,15 +42,15 @@
         <!-- Side-slip angle influence only. Type 2 will also give AoA
              influence. See Docs/README.systems.
         -->
-        <type>1</type>
-        <error-factor>0.2</error-factor>
+        <type>1</type> <!-- Type 1 implies exterior port location influences -->
+        <error-factor>0.2</error-factor> <!-- location prone to introduction of error -->
     </static>
     <static n="1"> <!-- Alternate source is from cabin, see POH p. 7-42 -->
         <name>static</name>
         <number>1</number>
         <tau>0.1</tau>
-        <type>0</type>
-        <error-factor>0.35</error-factor>
+        <type>0</type> <!-- Type 0 implies no exterior port influences -->
+        <error-factor>0.0</error-factor> <!-- Venturi effect dynamically applied in Systems/static-sources.xml -->
     </static>
 
     <vacuum>
--- a/Systems/systems.xml
+++ b/Systems/systems.xml
@@ -48,7 +48,7 @@
     <static n="1"> <!-- Alternate source is from cabin, see POH p. 7-42 -->
         <name>static</name>
         <number>1</number>
-        <tau>0.1</tau>
+        <tau>0.3</tau> <!-- Slower response when interior port is used -->
         <type>0</type> <!-- Type 0 implies no exterior port influences -->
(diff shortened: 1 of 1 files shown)


hbeni:
>  I am getting different results

What exactly is different for you?

wlbragg:
The tau change from .1 to .3 mention in the [c182 issue](https://github.com/HHS81/c182s/issues/505#issuecomment-1692152192) is also a good idea and we should adopt it. I'll do that here as well.

@hbeni the pressure was different each time I loaded it using "Fair Weather" with and without error = 0. I'm still testing and it may be that it will change regardless of changing that value. Hard to quantify.

wlbragg:
I continued testing and it didn't change with type = 0 between error = 0 and error = .35 I was sitting with the sim loaded for a little bit the first time I tested. My pressure becomes dynamic (to about the 10th position numerically) when sitting for more than a minute or so, even at cold and dark. It may have just been the time the sim was loaded before I checked. When loading and checking immediately, they were the same.

wlbragg:
>Therefore, if you want to see the effect of the measurement failure, "tau" should be 0.1 or smaller.

I don't understand?

hbeni:
> > The output property /systems/static[n]/pressure-inhg is filtered. Therefore, if you want to see
> > the effect of the measurement failure, "tau" should be 0.1 or smaller.
> 
> I don't understand?

The final value is filtered trough a simple time filter, which is configured by paramter "tau".
tau is basicly the rate of change per second (so tau=1 is a lag of about a second or so)

wlbragg:
Ok, thanks.

wlbragg:
@dany93 ok, finished here merge at your convenience.

PR #1453, opened 2023-09-03
Glass rain effect improvements

### LINKED ISSUE #1452: Improve glass rain effects
Back window rain now flows the correct direction. The side windows are a little tamer.

wlbragg:
Fixes #1452 

This is not quite finished. Anyone concerned, please test the rain effect on all the windows and critique. It is certainly better on the back window better, the front window is unchanged, but I think I want to slow down the speed of the animation of the front window, it's too intense. The side windows required a major refactoring in order to give each side their own unique effect. That was necessary as the Y vector have to be reversed for each side or you are limited on any animations. There are still quirks with it though. I need to stop for awhile and come back to it. It was perfect during testing, but when actually applied in the sim, the right side didn't behave as it did in the tests. It was an exact duplication of vector except the reversed Y vector and performed correctly while testing with hard values. As soon as I interpolated in the sim it got a little less perfect.

But you'all can see for yourself.

What I think needs to still be done is...
Front window high IAS animation slowed down
front window, non-moving aircraft needs to still get animated drops falling on the window. It never did, I need to adjust that.
The left and right windows need to also have better animated drops falling on the window when the aircraft is not in motion. I think it is ok when moving. It is impossible to get consistent rain streaks moving from front to back on the side windows.


wlbragg:
@dany93 

I'm finished with this. The pain to gain ratio to get it any better is not worth it. I think this good.

dany93:
Received. I hold off.

wlbragg:
@dany93 I really don't think I am going to do any more. I feel this is quite a bit better than before and I am happy with it.
It is about the best I think I am going to get. There are little quirks that always pop up if we try to get too fancy. Like the interpolation values you get between going from one direction to another. Or little stutters when the aircraft logic is crunching some event and the rpm is the condition in control of the effect. So with that in mind, I have the basics that were of concern covered as best I could.
I don't think we will get any other resolution with the shader. Because we have a solution available by glass positioning and then repositioning, I'm ready to put this to bed.

They key point are as follows.

1) Front windshield lass has rain drops forming and slowly animating down to gravity until prop spins up. Then it animates towards the back. As airspeed increases so to does the speed of the animation, slightly.

2) Front windshield side glass has rain drops forming and slowly animating down to gravity until prop spins up. Then it animates towards the back. As airspeed increases so to does the speed of the animation, slightly.

3) Side window glass covered or mostly covered by the wings has as limited drops as I could program. No moving animation until airspeed exceeds 20 kts, then a slow backwards flow of drops.

4) Back window rain drops slowly move down when the aircraft is stationary and no prop. Back window rain moves down to the back when in motion or from the prop. From the prop, maybe a questionable effect, would the air from the prop reach the back with enough force to animate the rain accumulations? Airspeed gives a slight increase in animation speed.

So unless someone makes a very good argument for any changes or improvements, please merge at your convenience.

wlbragg:
My head is spinning. I'm stuck on the fact that you cannot see the rain on glass effect from the outside, even with double sided glass. The exterior glass interferes for some reason? If you move the interior glass to the exterior and leave it double sided, it appears you can still see the effect and you can still see the reflection effect on the exterior glass which is now positioned as the inside glass. It has always bugged me we cannot see the rain effect from the outside view. Is it worth moving the glass positions and all that this entails?

wlbragg:
@dany93 if you get a chance, check out [this branch](https://github.com/c172p-team/c172p/branches). It is with the interior and exterior glass swapped so you can see the rain when looking at the aircraft in external view. All the effects still work, but the edges of the front windshield and the side glass on the windshield I think is a little more pronounced. What do you think, is it better to be able to see the rain on the glass when in exterior view at maybe a little bit of expense of loss of quality in the shape of the windshield?

dany93:
@wlbragg 
GlassSwapped branch.

From the interior, I don't notice a difference in the contrast (which already existed) between the greenish front and the edges of the windshield.
From the exterior, yes. The frontier is much more pronounced. Instead of smoothed in the initial version.

But the rain effect is very much improved, very impressive. From the interior the rain drops go down or back. They are seen and nice with their movement from the exterior too.

Even if imperfect, I think that the effect is that nice and impressive that it is worth the loss of quality (contrast between the front windshield and the sides seen from the exterior).

wkitty42:
On 9/15/23 8:02 PM, wlbragg wrote:
> @dany93 <https://github.com/dany93> if you get a chance, check out this branch 
> <https://github.com/c172p-team/c172p/branches>. It is with the interior and 
> exterior glass swapped so you can see the rain when looking at the aircraft in 
> external view. All the effects still work, but the edges of the front windshield 
> and the side glass on the windshield I think is a little more pronounced. What 
> do you think, is it better to be able to see the rain on the glass when in 
> exterior view at maybe a little bit of expense of loss of quality in the shape 
> of the windshield?

please forgive my ignorance... i gotta ask why we cannot have the rain effect on 
both sides of the glass? it *does not* have to be active and visible on both 
sides *at the same time*, though... only when we are on that side of the glass 
should we see it... so in all interior views, we would see the effect on the 
interior glass like normal... if we switch to an exterior view, we would see the 
effect on the exterior of the glass... i assume it would switch off if we zoom 
out to the point where we can't see it if we tried...

i don't know if there's a difference in "selector" or "trigger", though... like 
would the effect switch on based simply on selected view or would it be better 
to switch on/off based on actual camera position? in other words, if we select 
an external view we should see the effect on the exterior glass... now it move 
the camera inside the craft or building, should we see the effect on the 
interior glass? it would be cool if we could and it would seem, in my limited 
knowledge, that that would be based on camera position... and somehow knowledge 
of what is around our position at the time...

wlbragg:
>please forgive my ignorance

>i don't know if there's a difference in "selector" or "trigger", though... like
would the effect switch on based simply on selected view or would it be better
to switch on/off based on actual camera position?

Yeah, that is not a thing in this case. This is strictly, do we make the glass with the rain effect 2 sided and make it the outside layer so you can see the rain from exterior view, but lose a little of the smoothed effect of the exterior reflection of the other layer of glass because it is now the inner layer of glass. Turning these on and off would make no difference even if we could. You still have to have both effects active at the same time to see both effects.

What I don't understand is why does the reflective glass layer interfere with the rain glass layer visual but the rain effect visual doesn't "totally" interfere with the reflective layer? I'll ask on the dev list and see if there is an answer for this.

wlbragg:
@dany93 I pushed the swapped glass to this attached PR. I saved the original glass configuration in the blend file in case we ever decide to go back. I guess we hold of on merging a bit longer for a few more questions and answers.

dany93:
@wlbragg 
Do you know, do you have an idea about where this greenish color comes from? Why the front windshield, the back window and not the side glass?

wlbragg:
@dany93 
It's the exterior reflection effect. I did some more tests to determine a couple things.
I thought maybe it was because when I reordered which glass layer was the first layer to the outside of the aircraft, I recombined all the previous single pieces that were the interior glass with the rain effect into the new interior glass that now holds the exterior reflection, I thought maybe it was because they were just individual glass that I "joined" in Blender. That that somehow is different than a contiguous piece of glass and that is what was messing with the side of the front windshield. So last night I actually took the original exterior glass and sized it down to be the interior glass and used it as the exterior reflective glass. Same characteristics, still that really flat sharp cut of from front facing to side facing. So just now I ran through a series of tests reordering glass layers, turning them on and off to see what they were doing. I think the bottom line is, if the interior glass with the rain effect is placed outside the reflective glass which is actually supposed to be the exterior glass, then the rain affected glass gets its own amount of reflectivity, not from the designed effect for exterior reflections, but from just being in the scene as a transparent. It is definitely interfering with the inside glass which has the exterior reflection effect applied. The effects just aren't designed to handle any of what we are asking them to do. Sucks, but I don't have the skill to change the behavior in the shader and those that might I don't think are interested.

I did figure out on the current pushed version to this attached PR that the side portion of the front windshield that has the rain effect applied was messed up in a really weird way. It was like the normals on the side were no longer working at all? I just pushed a new version with that fixed. If you pan around the exterior in bright sun, you can see that the reflection works better at some angles. Again it's the rain layer causing interference. This is the best I can do with rain visible on the outside.

So I guess our decision is whether to use this or leave it alone and not have rain visible from the outside?

I weigh that question with the following in mind.

The interior visual didn't change as far as I can tell. The new rain effects are decent either way.

From the exterior view, what is worse, the really noticeable lack of raindrops on certain glass surfaces or the reflections and any noticeable interference as you pan around the aircraft. How much do you really notice the undesirable effects caused by allowing visible rain on the glass in the exterior view?

Which issue is most noticeable?

wlbragg:
https://github.com/c172p-team/c172p/assets/11069680/5431aa96-be4f-4a5e-8b49-8046cb8a51cd

wlbragg:
@dany93 instead of swapping these glass layers, how about just hiding the reflective layer when it is raining? In the exterior view, I think the rain on the glass is more important when it is raining than the reflection. In anything other than rain I think the reflection is really important. Does this solve both cases?

wlbragg:
@dany93 after sitting on this for a few days, what do you think, is this last iteration plausible? Just hide the reflective glass that is blocking the rain effect glass when it is raining? That way when it is not raining we always still get the rich reflection. I think it is a better option than losing the nice reflection just to get rain visual from the exterior view. I don't think in overcast, raining conditions you would get much reflection anyway. The important visual in that situation is the rain.

dany93:
@wlbragg 
Sorry, but I don't understand the issue with the reflection + rain.
d2108b77 "Fix normals os side glass on front winshield" improved the side glass reflection (smoothed frontiers). Fine, but this is not the rain issue.
I'm confused because, with d2108b77, I could see the rain drops from the exterior, with the reflection. From the interior too, of course.
c172-d2108b77

wlbragg:
>The only fault I notice is a disappearance of reflections on certain particular orientations (minority) when I scan around the aircraft. Is this your problem?

Yes, and I changed my mind. After looking at it one more time I am in agreement with you. It is fine. The minor loss in smoothness and the "disappearance of reflections on certain particular orientations (minority)" is minimal. It look good enough. I'm finished and you can merge at your convenience. Thanks for all the help with this particular issue.

dany93:
Merged.
You are welcome. I have no skills in these animations but I understand that you had a very hard time at implementing this. Thumb up!

PR #1462, opened 2023-11-14
Magnetic Compass enhancements

hbeni:
Magnetic Compass enhanced
- simulated pivot mount
- can get stuck when outside the pivots pitch limits
- avionics have electrical influence on the bearing (compass correction is the same as it was before with all avionics on; by coincidence its the same)

(Ported over from the c182s)

### DIFF
--- a/Models/Interior/Panel/Instruments/mag-compass/mag-compass-fg1000.xml
+++ b/Models/Interior/Panel/Instruments/mag-compass/mag-compass-fg1000.xml
@@ -76,6 +76,38 @@
     <animation>
         <type>rotate</type>
         <object-name>Ring</object-name>
+        <property>instrumentation/magnetic-compass/roll-deg-final</property>
+        <factor>0.5</factor>
+        <axis>
+            <x> 1.0 </x>
+            <y> 0.0 </y>
+            <z> 0.0 </z>
+        </axis>
+        <center>
+            <x-m>-0.3769</x-m>
+            <y-m>0.0000</y-m>
+            <z-m>0.2405</z-m>
+        </center>
+    </animation>
+    <animation>
+        <type>rotate</type>
+        <object-name>Ring</object-name>
+        <property>instrumentation/magnetic-compass/pitch-deg-final</property>
+        <factor>0.5</factor>
+        <axis>
+            <x>  0.0 </x>
+            <y> -1.0 </y>
+            <z>  0.0 </z>
+        </axis>
+        <center>
+            <x-m>-0.3769</x-m>
+            <y-m>0.0000</y-m>
+            <z-m>0.2405</z-m>
+        </center>
+    </animation>
+    <animation>
+        <type>rotate</type>
+        <object-name>Ring</object-name>
         <property>instrumentation/magnetic-compass/indicated-heading-deg-final</property>
         <factor>-1</factor>
         <axis>
--- a/Models/Interior/Panel/Instruments/mag-compass/mag-compass-fg1000.xml
+++ b/Models/Interior/Panel/Instruments/mag-compass/mag-compass-fg1000.xml
@@ -76,7 +76,7 @@
     <animation>
         <type>rotate</type>
         <object-name>Ring</object-name>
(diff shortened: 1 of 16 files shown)


wlbragg:
Wow. thanks @hbeni you've been busy.  I take a look at everything you've done and get it merged when I get a free moment.

hbeni:
Cool,
maybe the animation and stuck-conditions need revision, but i think the current values are reasonable.
And these are easily to be changed just by parameters.
Currently enjoying the changes at Grenchen/GA-Wednesday and it looks good to me :)

hbeni:
> Wow. thanks @hbeni you've been busy. I take a look at everything you've done and get it merged when I get a free moment.

When you try it, notice the following things:
- When switching avionics on, the compass will move off course for some degrees.
- When pitching up/pitching down, the compass disc will pitch/bank accordingly to "keep it straight to the earth surface
(actually the magnetic field lines, but we assume perfect calibration to the location where you are currently flying; thus perfect calibration factory sided to the location)
- When pitching up/down or banking off-limits the compass dic will hit the casing and be unable to move.
Once out of that condition (currently set to +-9° pitch and +-25° bank), the compass will be free and rotate to the magnetic field again
- In cruise flight, the compass is expected to move freely; in climb and descend quickly hit limits

hbeni:
Also, you really should consider adding the Compass correction card to the cockpit somewhere (or the compass itfself); the values are the same as for the c182 (is i took the initial deviation table from the 172 :) )
https://github.com/HHS81/c182s/blob/master/Models/Instruments/CompassCorrectionCard.xml

wlbragg:
So is the compass deviation card something a pilot simply refers to so they know how much offset to apply to the compass readout? The compass has no calibration dial to set it to read accurately?

wlbragg:
I also read the mailing list thread related to this. I guess we are not going to do this in c++ for now with
>config/has-pivot and config/max-pivot-angle to select this behaviour on / off

wlbragg:
I'm a bit confused, I saw the compass deviation card in the c182. You don't have a magnetic compass like the 172 does (on top of the dash) I see how that applies to that compass. If you want to fly 90 then you set coarse to 92. But what about the compass on the panel, that has a calibration knob, yet looking at the card, it is different deviations for different headings, so one calibration change wouldn't apply to all?

wlbragg:
@hbeni

I'm a bit confused, I saw the compass deviation card in the c182. You don't have a magnetic compass like the 172 does (on top of the dash) I see how that applies to that compass. If you want to fly 90 then you set coarse to 92. But what about the compass on the panel, that has a calibration knob, yet looking at the card, it is different deviations for different headings, so one calibration change wouldn't apply to all?

How do you use that card in the c182?

hbeni:
> So is the compass deviation card something a pilot simply refers to so they know how much offset to apply to the compass readout? The compass has no calibration dial to set it to read accurately?

No, it is a simple magnetic compass.
This gets calibrated from a technican (by adjudting the compendator magnets), and the remaining error is noted on the deviation card.
The Card shows the error that can not be corrected away.

hbeni:
> I also read the mailing list thread related to this. I guess we are not going to do this in c++ for now with
> 
> > config/has-pivot and config/max-pivot-angle to select this behaviour on / off

I also don‘t think it will be added shortly. Hence this addition.

hbeni:
> @hbeni
> 
> I'm a bit confused, I saw the compass deviation card in the c182. You don't have a magnetic compass like the 172 does (on top of the dash) I see how that applies to that compass. If you want to fly 90 then you set coarse to 92. But what about the compass on the panel, that has a calibration knob, yet looking at the card, it is different deviations for different headings, so one calibration change wouldn't apply to all?
> 
> How do you use that card in the c182?

Are you reffering to the Directuonal gyro?
That must be calibrated against the magnetic conpass.
Every plane needs a simple magnetic compass by regulation.
The 172p has it mounted to the top of the panel and the 182 has it mounted to the windshield.

wlbragg:
OK, thanks for the explanation. Didn't see the 182 mag compass.

PR #1465, opened 2023-11-16
Add magnetic deviation reference card

### LINKED ISSUE #1464: Add Magnetic Deviation Card
This issue adds a magnetic deviation card for pilot reference in the pocket where the POH and control lock is.

wlbragg:
Fixes #1464 

Add magnetic deviation reference card, adjust control lock card in pocket.

@dany93 Ready to merge after review. Thanks for the suggestion @hbeni


hbeni:
Cool idea!

However I personally think it is ok for simulation purposes, the FAA actually requires this information near the compass:
https://aviation-regulations.com/27.1547

I know for the C182 it is not actually „near“ there, but for the panel placard position I got footage…
I also know of compass mounts.
Maybe its an idea to make the card smaller and put it below the compass?

wlbragg:
@dany93 or @hbeni I added a deviation card graphic to the mag compass. So now you can either click on the deviation card graphic on the compass or on the deviation card in the POH holder below left of the pilot seat to bring up the card so you can read the offset values. Please merge PR at your convenience after testing. This PR is finished.

hbeni:
@wlbragg Just checked, looks good to me :)
The solution with the card beeing clickable is a nice Idea!

This should not hinder merging, just some food for tough and curiosity:
Honestly, I would just remove the card from the bag. Never seen such a card not glued to the panel or at the compass.
Another question: Why is the card at the compass so low-res? Other planes have it in hires, so you can zoom in and see it clearly.

wlbragg:
>Why is the card at the compass so low-res? Other planes have it in hires, so you can zoom in and see it clearly.

I can try to scale it to fit the compass. But I don't think it's possible to fit in that space legibly.

I agree, about the card in the bag, there is really no need for it now. I will remove it. I mainly don't like deleting work I already did. :)

.

dany93:
With the clickable card (good idea, yes,) I dont think many people will try zooming to look at the one on the compass...

dany93:
Question: are the values on the card in accordance with the deviations of the compass?
(or rather, in simulation: are the deviations of the animated compass in FG in accordance with the values on the card?

hbeni:
> Question: are the values on the card in accordance with the deviations of the compass?

Yes, it is ;)

wlbragg:
OK, I'm done, we have both. You can zoom in and look at the card on the compass, resolution was ok. I was thinking of a pick that zooms in for you, but it gets too complicated with all the different views. So I also made the card graphics on the compass clickable, it opens the larger card for viewing so no zoom is required. Clicking on on either card or graphic on the compass will close the displayed card if it is visible. Everything is contained in the panel/instruments/mag-compass folder. One new property to control the visual sim/model/c172p/cockpit/view-deviation-card.

@hbeni thanks for all the suggestions. @dany93 feel free to merge if you are good with everything.

hbeni:
I really like it :)

PR #1480, opened 2023-12-19
Add Towbar.

hbeni:
Planes ground handling dialog now gives access to the towbar item. When attached, it can be grabbed with the mouse (hold button) and then left/right movement will adjust direction and back/forwad controls force. Once the mouse (=force) is released, the plane will slowly stop.

Works with all tires as well as skis.

(This also fixes a hidden syntax bug in the cranking-listener)

----
Ported from C182: https://github.com/HHS81/c182s/pull/549


Nasal/c172p.nas
```diff
@@ -704,7 +704,8 @@ setlistener("/sim/signals/fdm-initialized", func {
 
     setlistener("/engines/active-engine/cranking", func (node) {
         setprop("/engines/active-engine/external-heat/enabled", 0);
-        setprop("/sim/externalheater/enable", 0)
```
hbeni, line 707: By accident I found the missing statement terminator here,
so i decided to just fix it too

hbeni:
That was Quick 😅

PR #1481, opened 2023-12-20
Improved Towbar model from MariuszXC

hbeni:
@MariuszXC provided a better model for the towbar :)

grafik


SP-NTX:
Shouldn't the towbar be a little longer? Once I was putting one aircraft out of hangar with towbar and it was pretty long, for sure not right under the propeller, or is it just a perspective

hbeni:
There are different Models out there.
Mariusz told me, he did use the measurements from the shop item

MariuszXC:
I checked several models for Cessnas (at aircraftspruce (dot) com ) and the one I modeled has specs:
Collapsible Length: 22.5"
Overall Length: 39.5" 

There are two other models:

for Cessna twin 300 and 400 series, with specs:
Collapsible Length: 35"
Overall Length: 57" 

and 'Cessna 182-XL Bogi Bar', with specs:
extends to 55" and closes down to 35"

So I can extend the model to 55" (1.4 m) if you want, or add a second version, so either can be used depending on user preferences.

hbeni:
We have one that works, and its just visuals, AND one existing in real life - so I think we could accept it as is.

MariuszXC:
Yes, but the one I did was sized somewhere between the two (I lengthened the short one after you suggested it is too short). Now we have two models with exact dimension of real life parts. It may be important for some.

hbeni:
Added that and some minor other fixes

wlbragg:
Thanks everyone, hbeni for porting it over to the 172.

MariuszXC:
:+1:

PR #1486, opened 2024-01-17
Towbar: Release force instantly instead of "slow" decay

hbeni:
This helps with a small bug with quick clicks not "gripping". Handling does not make a difference.


hbeni:
Tested, ready for merge IMHO

PR #1490, opened 2024-01-29
Update gyro instruments to current "next" enhancements

hbeni:
**Update gyro instruments to current "next" enhancements**

This I made "opt-in", so the pilot needs to activate this from the aircraft options menu.
  If this is not activated at startup, it will adjust the instruments config to not exhibit
  errors. The instruments will work always good, not wander and not tumble; also
  they will spinup very fast (few seconds, like previously).

Gyros will have more realistic behaviour:
* HI, AI and TC: the gyro spinup will take longer (about 2-3 minutes).
* HI:
  - can be caged now, and when adjusting the disc, it will realistically cage while moving. Cage can be toggled by shift-click on the knob. The knob appears pushed and the tooltip will also report a hint to the user.
  - can tumble now with high G loads
  - models errors (drift/precession due to latitude (transport wander, high yaw-forces))
  - has latitude nut simulation (factory setting to be set in aircraft options)

Settings are stored between sessions.
Autostart will spin up all gyros instantly (as if they have already run for some time) and HI is aligned (while removing all erros/drift from the instrument).

:warning:  Needs a change from FGFS 2020.4/next, so not compatible to 2020.3/stable!

-----
**Refs:**
* Related SF Ticket: https://sourceforge.net/p/flightgear/codetickets/2839/
* Mailing list discussion: https://sourceforge.net/p/flightgear/mailman/flightgear-devel/thread/6e66f08e3fb3f02a9d5c557cfd7a2910%40hallinger.org/#msg53245268
* C182s sister ticket: https://github.com/HHS81/c182s/pull/527
* FGFS Wiki Newsletter Jan/2024: https://wiki.flightgear.org/FlightGear_Newsletter_January_2024#Gyro_based_instruments_enhancements

### DIFF
--- a/Models/Interior/Panel/Instruments/hi/hi.xml
+++ b/Models/Interior/Panel/Instruments/hi/hi.xml
@@ -95,10 +95,10 @@
     </animation>
 
     <!-- Heading Offset -->
-    <animation>
+<!--    <animation>
         <type>knob</type>
         <object-name>OBS-Knob</object-name>
-        <property>instrumentation/heading-indicator/offset-deg</property>
+        <property>instrumentation/heading-indicator/align-deg</property>
         <factor>12</factor>
         <axis>
             <x>-1</x>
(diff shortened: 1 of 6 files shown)


hbeni:
Already tested it, and it looks good.

However, some more flying/testing could not harm.

hbeni:
> Already tested it, and it looks good.

Also verified in spin/stall.
Looks good overall.
Did about 15 landings/takeoffs and pattern work and some flight maneuvers, and the HI/DG was indicating okay after that (minor errors, but expected due to steep turns).

hbeni:
We should put merging to hold for the time beeing - the code in fgfs/next makes the gyro swap weirdly.

hbeni:
> We should put merging to hold for the time beeing - the code in fgfs/next makes the gyro swap weirdly.

Seems to be resolved.

hbeni:
Just did another test flight.
Gyro works as intended, both with realistic and unrealistic setting.

If you ask me, it's ready.

PR #1492, opened 2024-02-12
Adjust mixture setting to be able to go over-rich

hbeni:
See:
 - Forums post from Davod Megginson: https://forum.flightgear.org/viewtopic.php?f=49&t=42055
 - JSBSim discussion: https://github.com/JSBSim-Team/jsbsim/issues/1037

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -1163,6 +1163,23 @@
                 </function>
             </fcs_function>
 
+            <!-- Mixture unit
+                 Used for normal operation.
+                 Mixture lever 0.0->1.0 setting is translated to jsbsim internal values.
+            -->
+            <fcs_function name="fcs/mixture-unit-cmd">
+                <function>
+                    <product>
+                    <value>1.27</value> <!-- 1.65 / 1.3 -->
+                    <sqrt>
+                        <property>/controls/engines/current-engine/mixture</property>
+                    </sqrt>
+                    </product>
+                </function>
+                <output>fcs/mixture-cmd-norm[0]</output>
+                <output>fcs/mixture-cmd-norm[1]</output>
+            </fcs_function>
+
             <pure_gain name="fcs/mixture-primer[0]">
                 <input>fcs/mixture-cmd-norm[0]</input>
                 <gain>fcs/mixture-primer-cmd</gain>


hbeni:
Marked as draft, as this is not yet finished. Not sure, the resulting effect is realistic.

hbeni:
From the [Forums](https://forum.flightgear.org/viewtopic.php?p=419031#p419031):

> You should need to lean to get max power at sea level — that's what I think confused whoever wrote that JSBSim code, who just assumed, wrongly but understandably, that you'd want the most-efficient fuel-air mixture when the lever/knob is at its maximum setting.
> 
> Aircraft piston engines have long been set up so that full-rich mixture is inefficient — you're burning more fuel than you should for the power you're producing, or, to put it another way, you're producing less power than you should for the amount of air being sucked into the engine. It's most obvious with a fixed-pitch propeller at full throttle, because you can see it on the tachometer. As you start to lean, the RPM will initially increase, then it will start to fall again as you keep pulling the mixture back.
> 
> That must seem confusing to a physicist, but there's a sound engineering reason behind it. Above about 75% power, you need to operate the engine inefficiently (excessively rich fuel/air mixture) to keep internal pressures from getting high enough to cause predetonation, where the fuel/air mixture ignites before the spark, throws off the engine timing, and can eventually tear the engine apart (or at least cause tens of thousands of dollars in premature wear).
> 
> So nearly all aircraft piston engines over the past century are designed in such a way that if the pilot simply pushes the throttle and mixture all the way in, they will get that inefficient mixture that keeps the pressure below limits when the engine is producing > 75% power for takeoff and initial climb (the exception would be the very few aircraft piston engines that have FADEC or some other kind of automated mixture control). I have hear that many Lycomings and Continentals also have an "enrichment" circuit that adds even more to that when the throttle is right at the stop, but I don't know about that.
> 
> There is no instrumentation for internal pressure, but CHT moves in lockstep with it, which is why many engine manuals use CHT as a limit (it's not the heat itself, though, but the pressure that's the main danger).
> 
> Once you're above 3,000 ft density altitude, a normally-aspirated piston engine can't produce a lot more than 75% power anyway, and peak CHT/pressure is a little lean of that point, which is why they recommend leaning for max RPM at that DA or higher.
> 
> D

hbeni:
So lets let this settle a bit.
either this gets fixed in JSBsim, which would be preferable according to David; or we have the aircraft-sided fix right here.

hbeni:
JSBsim closed again as wontfix.
On the FGFS ML I did ask for clarification if this should be fixed at the jsbsim/fgfs integration layer (Bertrand proposed a change in the [jsbsim Ticket](https://github.com/JSBSim-Team/jsbsim/issues/1037#issuecomment-1967919602)).
If that too gets rejected, we can apply this fix here at the aircraft level.

wlbragg:
I'm following the jsbsim thread on this. So should we still hold off on this?

hbeni:
As Long as it is not sorted out…
Or merge and revert later.

hbeni:
Related to
- https://github.com/JSBSim-Team/jsbsim/issues/96
- https://github.com/c172p-team/c172p/issues/1128

wlbragg:
@hbeni I'd like to get this cleared out. It looks like this is the fix for now, do you agree?

hbeni:
This is a valid aircraft sided fix in my opinion.
It needs to be revised, when the jsb-team decided to adjust the engine sim code.

So, I would merge for now.

wlbragg:
Thank you!

PR #1496, opened 2024-03-01
Issue 1495 elec sys bugs

### LINKED ISSUE #1495: Electrical system bugs
Hi,
investigating POH 7-25 reveals some bugs:
- [x] Turn coordinator is currently linked to avionics switch. But according to POH it is powered by the primary electrical bus (via the turn coord braker), and thus should be powered when the master switch is on. (this was probably overlooked in #79)
- [x] Also, the elctrical.nas currently supplies voltage also to `outputs/DG` but the DG/HI is vaccum driven. This prop looks obsolete and probably should be cleaned up.
- [x] Turn coordinator flag is currently linked to the `spin` of the gyro. But that couldn't be measured directly and afaik the flag just displays if the equipment is powered or not.
(see also: https://pilotinstitute.com/turn-coordinator/ _"It is important to note that the flags only indicate a loss of power/vacuum pressure; they do not check the validity of the gyroscope. In other words, if the gyroscope itself were to fail, the instrument would indicate incorrectly and no red flag would be present."_)
- [x] Currently, attaching the GPU will energize the electrical bus regardless of Master-switch state. Supplement 6 deals with the GPU, and the GPU power source should only provide energy when the master (BAT-part) switch is ON (also in accordance with POH 7-26: "the right half of the switch, labelled BAT, controls all electrical power to the airplane.").
- [x] POH 7-35 describes the Avionics cooling fan, which should be modelled (breaker is the strobe light one), and it should be audible when the fan is working. Currently we already have a sound when electrical power is on, and we should reinvestigate if this should already be that fan (if so, we need to link the sound to the fan; if not, then we need an additional sound implemented).

hbeni:
resolves #1495

### DIFF
--- a/Models/Interior/Panel/Instruments/tc/tc.xml
+++ b/Models/Interior/Panel/Instruments/tc/tc.xml
@@ -98,7 +98,7 @@
     <animation>
         <type>translate</type>
         <object-name>Flag</object-name>
-        <property>instrumentation/turn-indicator/spin</property>
+        <property>systems/electrical/outputs/turn-coordinator</property>
         <factor>0.0040</factor>
         <axis>
             <x>0</x>
--- a/Nasal/electrical-fg1000.nas
+++ b/Nasal/electrical-fg1000.nas
@@ -654,17 +654,6 @@ var avionics_bus_1 = func() {
         setprop("/systems/electrical/outputs/avionics-fan[0]", 0);
     }
 
-     ##############????????#############
-    # Turn Coordinator and directional gyro Power
-    if ( getprop("/controls/circuit-breakers/turn-coordinator") ) {
-        setprop("/systems/electrical/outputs/turn-coordinator", bus_volts);
-        setprop("/systems/electrical/outputs/DG", bus_volts);
-        load += 14 * bus_volts;
-    } else {
-        setprop("/systems/electrical/outputs/turn-coordinator", 0.0);
-        setprop("/systems/electrical/outputs/DG", 0.0);
-    }
-
     # register avn1 voltage
     avn1_volts = bus_volts;
 
--- a/Nasal/electrical-fg1000.nas
+++ b/Nasal/electrical-fg1000.nas
@@ -646,6 +646,14 @@ var avionics_bus_1 = func() {
       setprop("/systems/electrical/outputs/is", 0.0);
     }
 
+    # Avionics Fan Power
+    if ( bus_volts > 12 and getprop("/controls/circuit-breakers/pfd-avn") ) {
+        setprop("/systems/electrical/outputs/avionics-fan[0]", bus_volts);
+        load += bus_volts / 24;
+    } else {
+        setprop("/systems/electrical/outputs/avionics-fan[0]", 0);
+    }
+
      ##############????????#############
     # Turn Coordinator and directional gyro Power
     if ( getprop("/controls/circuit-breakers/turn-coordinator") ) {
@@ -695,11 +703,9 @@ var avionics_bus_2 = func() {
     # Nav 2 Power and Avionics Fan Power
     if ( getprop("/controls/circuit-breakers/nav2") ) {
         setprop("/systems/electrical/outputs/nav[1]", bus_volts);
-        setprop("/systems/electrical/outputs/avionics-fan", bus_volts);
         load += 5 * bus_volts;
     } else {
         setprop("/systems/electrical/outputs/nav[1]", 0.0);
-        setprop("/systems/electrical/outputs/avionics-fan", 0.0);
     }
 
     # Com 2 Power
@@ -736,6 +742,14 @@ var avionics_bus_2 = func() {
     } else {
       setprop("/systems/electrical/outputs/dme", 0.0);
       setprop("/systems/electrical/outputs/adf", 0.0);
+    }
+
+    # Avionics Fan Power
+    if ( bus_volts > 12 and getprop("/controls/circuit-breakers/mfd") ) {
+        setprop("/systems/electrical/outputs/avionics-fan[1]", bus_volts);
+        load += bus_volts / 24;
+    } else {
+        setprop("/systems/electrical/outputs/avionics-fan[1]", 0);
     }
 
     # register avn2 voltage
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -439,14 +439,21 @@ var electrical_bus_1 = func() {
         setprop("/systems/electrical/outputs/strobe-norm", 0.0);
     }
 
-    # Turn Coordinator and directional gyro Power
-    if (getprop("/controls/circuit-breakers/turn-coordinator") and getprop("/controls/switches/master-avionics")) {
+    # Directional gyro Power
+    # DG/HI is vacuum powered. This looks obsolete here and could probably be removed.
+    #if (getprop("/controls/circuit-breakers/turn-coordinator") and getprop("/controls/switches/master-avionics")) {
+    #    setprop("/systems/electrical/outputs/DG", bus_volts);
+    #    load += 14 * bus_volts;
+    #} else {
+    #    setprop("/systems/electrical/outputs/DG", 0.0);
+    #}
+
+    # Turn Coordinator Power
+    if (getprop("/controls/circuit-breakers/turn-coordinator") ) {
         setprop("/systems/electrical/outputs/turn-coordinator", bus_volts);
-        setprop("/systems/electrical/outputs/DG", bus_volts);
         load += 14 * bus_volts;
     } else {
         setprop("/systems/electrical/outputs/turn-coordinator", 0.0);
-        setprop("/systems/electrical/outputs/DG", 0.0);
     }
 
     # Gear Select Power
--- a/Nasal/electrical.nas
+++ b/Nasal/electrical.nas
@@ -271,7 +271,7 @@ var update_virtual_bus = func (dt) {
         bus_volts = alternator_volts;
         power_source = "alternator";
     }
-    if ( external_volts > bus_volts ) {
+    if ( external_volts > bus_volts and master_bat) {
         bus_volts = external_volts;
         power_source = "external";
     }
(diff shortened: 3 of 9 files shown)


hbeni:
Ready for review

hbeni:
For the Archives; regardind the GPU (ground power unit) I had a mail discussion with David Megginson to clarify, wether the GPU supplies power to the bus with the master-bat switch in the OFF position (like it is modelled now).

I believe that to be wrong, and found evidence.

> So, I did dig into the matter more and found this discussion:
> https://aviation.stackexchange.com/questions/72425/how-does-the-battery-master-switch-activate-the-battery-contactor-in-cessna-152
> 
> It means that this is really depending on plane, like you said.
> 
> C182S
> For the C182s, I looked closer to the schematics (POH 7-30, attached).
> There, we have a dedicated "external power relay", that is linked to the master switch. It also is linked to the battery.
> The POH stats that the "master BAT switch controlls all electrical power to the plane", and also "The primary buses are on anytime the master switch is turned on, and are not affected by starter or external power usage. The avionics buses are on when the master switch and avionics master switch are in the ON position.". For the case that the battery is dead,and therefore can not provide current via the master switch to ground, so the contactor closes, the external power does close the battery contactor too. But since the external power relay is placed before the contactor, and only closes when the BAT switch is ON, this also is linked to that switch. The result is also, that loading a flat battery using the external power unit is also only done when the BAT switch is ON. It's not enough to just connect the GPU.
> So for the C182s I would really see our toughts as "confirmed by evidence" wouldn't you say?
> 
> For the C172 the schematics is not as clear. It looks like the GPU is powering directly bypassing the switch,
> but the POH also has the same wording as the C182S POH regarding "controlls all power to the plane".
> It does not have the "external power relay" but a "reverse polarity contactor" (assumingly doing the same, but the schematics do not show wiring to the master switch).
> Like with the 182, the POH says "The primary bus is on anytime the master switch is turned on, and is not affected by starter or external power usage". To me, that tells me its working like with the C182, namely, the BAT switch needs to be in the ON position to allow power from the GPU to reach the primary bus.
> 
> Thanks again for your help :)

See also https://github.com/HHS81/c182s/issues/565

wlbragg:
@hbeni 

Thank you for digging into this as much as you have already. I have seen schematics both ways. Below is a schematic I have used for some time and it shows the batt switch is not inline for the GPU port.

I know this isn't even close to anything definitive, just another data point, but Chat GPT stated the following which was an interesting thought.

>No, the master battery switch on a Cessna 172P does not control the flow of current from a ground power unit (GPU) to the rest of the aircraft's electrical circuitry.
The master battery switch is primarily used to connect or disconnect the aircraft's electrical system from the main battery. When the master battery switch is turned on, it connects the aircraft's electrical system to the battery, allowing it to power various systems and components. When turned off, it isolates the battery from the electrical system, preventing it from supplying power.
The GPU current to the aircraft's electrical circuitry is typically controlled by a separate switch or mechanism. This allows the ground crew or pilot to connect or disconnect the external power source without affecting the aircraft's internal electrical system. The specific method for controlling GPU power may vary depending on the aircraft's configuration and equipment.

So I am concerned about this part.

The other piece, the avionics switch acting as a circuit breaker, is that from the 182 POH or both?

hbeni:
> I know this isn't even close to anything definitive, just another data point, but Chat GPT stated the following which was an interesting thought.

I would not trust ChatGPT further than I can throw the word AI. I also asked it alot of questions for fun when doing recherche for this - and it was not helpful at all, very vague etc.
I also had a chat with David Megginson about the matter.

This is the POH I was referencing. Elec system is described from page 7-24 (PDF page 156).
https://www.cpaviation.com/images/downloads/Cessna%20172P.pdf

It shows this schematic: 
grafik

> I also had a chat with David Megginson about the matter.

We concluded, that it is sensible to assume the BAT switch also disconnects the GPU, because
- safety reasons
- wording is the same as the C182s POH. There we have definite proof from the schematics for the behaviour

grafik
grafik

The GPU is described in supplement 6 (PDF  page 253ff).
grafik
And from the Schematics and the master switch description we know, that the battery contactor is only on when master switch is on.
So, in conclusion, there can be no power through the contactor. And the part with the GPU and master switch is the same as with the C182s POH. For me this is enough of evidence, but If this should not be changed, its easy.
Just ignore (or revert) commit 1ef6a2d12042c2174f66 for the merge. I isolated the change for this reason.

-----

> The other piece, the avionics switch acting as a circuit breaker, is that from the 182 POH or both?

No, the 182 has dedicated breakers for the avionics busses. Its just the 172p; from the 172P (POH 7-26):
grafik

wlbragg:
The way I read the following...

>Which will close the battery contactor when external power is applied with the master switch turned on.

Meaning only if the batt switch is turned on will the battery contactor be closed to the batt if GPU power is supplied, Meaning you have to isolate the batt circuit from the GPU circuit if the batt circuit is active when you apply GPU power. You don't want both. The contactors are there to isolate the various circuits from one another. If the master batt is off, then you don't need the battery contactor to close when you apply GPU power.

I am reading this totally different than your interpretation?

The battery contactor is taking power from either the batt or the GPU and allowing it on the the ammeter and primary buss, not both.

Look at the circuit and imagine the  battery contactor as an OR switch to decide which circuit to allow the ammeter and primary bus to have access to.

Does this make any more sense? I think the schematic supports my conclusion. The one I provided supports it even more as it is clear that it is isolated from the bat and if you look at the top of the primary bus it says "from alternator bus". So the primary bus has one circuit from the GPU "OR" one from the alternator bus.

wlbragg:
It is still unclear to me exactly what the battery contactor is for and how it works in this instance?

wlbragg:
Thanks for clearing up the avionics switch portion, that is crystal clear.

hbeni:
I put the conclusion first, then the details.

Conclusion
------------------

- I think, this is a difference between the C172 and the C182s.
- The C172 service manual I found, is not specific for the "p" variant.
  - it very probable states that the GPU connects to the bus, "bypassing" the master switch
- We also found evidence that the master switch will switch on the battery contactor, adding the battery to the system.
  - in case the battery is flat, one can switch BAT on, in order to add the battery to the circuit - the GPU then will load the battery. To do so, the GPU wiring can close the contactor, too.

- [ ] So, I need to refine the code a bit, so that the GPU can load the battery with master switch on, but the GPU providing general power to the busses.


Details
------------------
> It is still unclear to me exactly what the battery contactor is for and how it works in this instance?

See: https://www.batterydesign.net/introduction-to-contactors/
The contactor is an elecrically operated switch. The BAT switch is connected to ground at one end. The other end is connected to the contactors switch line. Therefore, if you close that circuit by placing the BAT switch in the ON position, you make a path between the battery and ground via the BAT switch. That (low power) current activates the contactor, which will then engage and close the actual circuit; in this case between the battery and the rest of the electrical system.
You do it this way, because otherwise all the power would need to go via the bat switch, which is clearly a dangerous thing and also rises the needed specs for the switch. With the contactor, the switch-circuit can be a very small power.

> The way I read the following...
>
>>   Which will close the battery contactor when external power is applied with the master switch turned on.

If you read further, you get the explaination for that sentence. The GPU has a special fused wiring, that will _also_ (in additon to the design laid out in the first sentence of my comment here) close the contactor, but only, when BAT is ON - not when it's off. It is for the case when the battery is flat: when the BAT has not anymore enough power to close the contactor when you switch BAT=ON, the contactor would remain closed. The alternate GPU circuit therefore can provide power to close the contactor.
At the very least, this is to connect the flat battery to the electrical system, so the GPU can load it.

> Does this make any more sense? I think the schematic supports my conclusion. The one [schematics] I provided

Do you remember if its for an C172P? Because what I have found, that this behavior is really plane/model specific. So unless you can clearly say "this is a c172p we are looking at" I would not model something after it.

But anyway; I found a (C172 servicing manual)[https://www.manualslib.com/manual/1240815/Cessna-172-Skyhawk-Series.html?page=197#manual] which has a detailed diagram for the battery's wiring: https://static-data2.manualslib.com//pdf6/125/12409/1240815-cessna/images/172_skyhawk_series_197_bg.jpg
There you can also clearly see a "ground service receptable contactor (19)".
Also that manual has a good description of the system regarding our question.
grafik
grafik

hbeni:
- [ ] The Ammeter also needs a little revision:
grafik

wlbragg:
Excellent, thank you for taking even more time on this.

>Do you remember if its for an C172P? Because what I have found, that this behavior is really plane/model specific. So unless you can clearly say "this is a c172p we are looking at" I would not model something after it.

I agree, I am not confident enough without finding the source where I got it originally. I'm going to make sure to save some of the new information provided here.

16-29, 16-30 and Note:,
I like the note telling you how to jury rig the GPU power to act as the batt VS using the supplied GPU relay circuitry.

This all is really good information and I appreciate your educating me on it. This looks great, Are you all signed off on it then, ready to merge?

wlbragg:
Good luck, it is a different electrical system. I'm not sure it is even close to correct and may be somewhat messy. So if you look at it, look at it from the perspective of having to possibly redo it or parts of it. How different is the 182t from the 182s?

hbeni:
> How different is the 182t from the 182s?

They share most of the main circuits. What's new is a standby battery and its wiring.

hbeni:
Searching for a POH I did only find a 172S with G1000 POH.
Are there any 172P's with G1000 where we can peek at its electrical system?
But I would expect the wiring to be similar.

https://wayman.edu/files/C172S-G1000-POH.pdf (7-47 / PDF p. 249 ff) looks _very familiar_ to the 182T.
- Figure 7-7 (sheet 1/3) is the same as with the 182s (note the master switch here really switches _all power_, ie no GPU with Master-BAT=OFF).
- Figure 7-7 (sheet 2/3)  is the same, except for  AVNCS1 Bus, where with the 172 there is no "Storm scope" equipped
- Figure 7-7 (sheet 3/3)  is the same

So I would figure we can just use the 182Ts electrical system nasal file?

hbeni:
OK; quickly looked at the fg1000 electrical system, and it looks not bad. Even the standby battery is modelled.

- [ ] Whats missing would be the avionics fan;
- [ ] 
I'll add that and then we are good to merge imho.

hbeni:
done, you can test and merge :partying_face:

wlbragg:
Outstanding, thanks for cleaning it up. I'm glad to have other eyes on the fg1000 electrical and that it passed muster. This is good, thank you.

PR #1497, opened 2024-03-10
172sp elec

hbeni:
Digged a little further into the C172SP FG1000 electrical system and fond two issues:

- [x] The STBY BAT TEST switch
  - [x] changed drag direction to a more natural vertical setting
  - [x] spring loaded TST position properly (more simpler) implemented trough xml
  - [x] test lamp should indicate battery state. Its not an ammeter (like it is wrongly implemented now)
- [x] PFD wiring corrected re AVN and ESS busses
  It was solely calculated from the ESS bus code; ie. if the ESS breaker was pulled, the PFD went dark, even when there was power from the AVN bus.


wlbragg:
Looks good, thanks!

PR #1498, opened 2024-03-12
Add ALT mode to Throttle/Mixture levers and trim wheels

hbeni:
Pressing ALT and moving the levers (mousewheel/dragging) gives finer control now.
This simulates throttles friction lock and the mixtures vernier feature.

### DIFF
--- a/Models/Interior/Panel/c172p-panel/c172p.xml
+++ b/Models/Interior/Panel/c172p-panel/c172p.xml
@@ -2181,9 +2181,23 @@
         </center>
         <action>
             <binding>
+                <condition>
+                    <not><property>/devices/status/keyboard/alt</property></not>
+                </condition>
                 <command>property-adjust</command>
                 <property>controls/flight/elevator-trim</property>
                 <factor>0.001</factor>
+                <min>-1</min>
+                <max>1</max>
+                <wrap>0</wrap>
+            </binding>
+            <binding>
+                <condition>
+                    <property>/devices/status/keyboard/alt</property>
+                </condition>
+                <command>property-adjust</command>
+                <property>controls/flight/elevator-trim</property>
+                <factor>0.0001</factor>
                 <min>-1</min>
                 <max>1</max>
                 <wrap>0</wrap>
@@ -2234,9 +2248,23 @@
         </axis>
         <action>
             <binding>
+                <condition>
+                    <not><property>/devices/status/keyboard/alt</property></not>
+                </condition>
                 <command>property-adjust</command>
                 <property>controls/flight/rudder-trim</property>
                 <factor>0.002</factor>
+                <min>-.18</min>
+                <max>.18</max>
+                <wrap>false</wrap>
+            </binding>
+            <binding>
+                <condition>
+                    <property>/devices/status/keyboard/alt</property>
+                </condition>
+                <command>property-adjust</command>
+                <property>controls/flight/rudder-trim</property>
+                <factor>0.001</factor>
                 <min>-.18</min>
                 <max>.18</max>
                 <wrap>false</wrap>
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -2710,11 +2710,26 @@
         <visible>true</visible>
         <drag-direction>vertical</drag-direction>
         <!-- no translation, we handle that seperately -->
+        <shift-repeat>10</shift-repeat>
         <action>
             <binding>
+                <condition>
+                    <not><property>/devices/status/keyboard/alt</property></not>
+                </condition>
                 <command>property-adjust</command>
                 <property>controls/engines/current-engine/throttle</property>
                 <factor>0.02</factor>
+                <min>0</min>
+                <max>1</max>
+                <wrap>0</wrap>
+            </binding>
+            <binding>
+                <condition>
+                    <property>/devices/status/keyboard/alt</property>
+                </condition>
+                <command>property-adjust</command>
+                <property>controls/engines/current-engine/throttle</property>
+                <factor>0.01</factor>
                 <min>0</min>
                 <max>1</max>
                 <wrap>0</wrap>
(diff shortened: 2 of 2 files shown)


hbeni:
I have seen, the default movement step is 2%.
The C182 uses 5% which may be a little better?

If you like, I can adjust it to match.

hbeni:
> So without "ALT" it will be a faster travel if you change it to 5, but with "ALT" you can get a slow more precise adjustment?

Currently it is so:
- shift usage = 10% steps
- Just use: 2% steps
- alt usage: 1% steps (introduced by this PR)

I propose to change „Just use: 2% steps“ to 5% increments.


> If this is so, yes, go ahead.

okay 👍

hbeni:
What I wonder tough: where do we document the different use modes of the levers, so pilots will see them?

wkitty42:
On 3/12/24 6:58 PM, Hbeni wrote:
> What I wonder tough: where do we document the different use modes of the levers, 
> so pilots will see them?

in the aircraft help is where i would do it...

hbeni:
Hm, or maybe directly at the tooltip?
Or just the wiki?

I tried adding it at the help gui, but that looked not so good.

wkitty42:
On 3/12/24 7:16 PM, Hbeni wrote:
> Hm, or maybe directly at the tooltip?
> Or just the wiki?
> 
> I tried adding it at the help gui, but that looked not so good.

yeah, i don't know about the formatting for the craft specific help but i 
generally check there first and rarely look anywhere else...

a tool tip would be fine if it can also look good and not be too much... 
especially if the tip is trying to note all three of the options...

of course the wiki should also be updated, too ;)

hbeni:
For the 182 I added a Dicumentation.md file years ago, to collect feature descriptions etc.

hbeni:
No, but we have an "about the plane" dialog, featuring buttons to open links:
grafik

There, a link to the github markdown file could be added, because that will open the file directly parsed.
But: It needs internet access. That is a bad thing.
Maybe we should add some markdown parser widget to display markdown in canvas dialogs; that way we could make this easy and generic.

wkitty42:
> maybe use leading dashes on each of the "using [...]" lines so it looks like an unordered list would help, too...

would look like this...
c172p-proposed-craft-help-dialog-with-unordered-list-look-Screenshot_20240313_120252

wkitty42:
On 3/13/24 12:43 PM, wlbragg wrote:
> After looking at the available GUI help categories, I think it belongs in Common 
> Aircraft Keys.

idk... i see Common Aircraft Keys as being common to all craft whereas this is a 
feature we're adding to the c172p so it is specific to the c172p like it is 
specific to Hbeni's craft... that's why i think the c172-help.xml is the proper 
place IMHO...

i will also note that i have not dug around to find out if there are other 
property types available that might fit the formatting better... i whipped that 
up using the existing line format because that's what i saw and could mock 
something up with real fast... if there's a list property, that could work, too...

wlbragg:
>idk... i see Common Aircraft Keys as being common to all craft whereas this is a
feature we're adding to the c172p so it is specific to the c172p like it is
specific to Hbeni's craft...

But that is one of the first places someone will look to quickly get the key bindings. Also this whole concept of "Common Aircraft Keys", that was my point. When does it become common, one, two, 10 or more aircraft? The throttle and mixture are common to almost every aircraft. These are just modifiers to those common controls. Again. I think we need a core change and remove the word Common. If we are to keep "Common",  then we need another heading such as "Custom Keys". I don't think it is necessary or proper to limit what goes in the quickest help popup for an individual aircraft to the word "common" I think it is just legacy wording that is past time for a change.

wkitty42:
i didn't realize the seat position keys were in the Common Keys menu... i thought they were in the "?" menu specific to each craft... but, i do see what you are saying about throttle and mixture being common... i've always seen that menu as common to all craft with feature or special stuff being in the "?" menu for each craft...

i don't know at what point something would be know as "common"... we face this same question in an old-school network where "things in common practice" are what is documented... in a network with thousands of systems that can connect to each other, at will, and exchange messages and files, "common handshakes", "common transfer protocols", and similar are used... at what point they were """counted""" and passed some "boundary" to become "common" i don't know...

anyway, i'm offering a POV with thoughts and ideas... you all are the implementers asking for input... i offer my input freely and leave it to you all to consider it and do with it what you will :+1:

wlbragg:
>i thought they were in the "?" menu specific to each craft...

Ah, that is the menu I was referring to. I guess i got confused. I see no problem putting it in there, in the same place as the seat position keys.

The "Common Aircraft Keys" is swollen and has grown. Is that custom to each aircraft or is it stored in fgdata? If it is in fgdata, then I don't think it should go there.

wkitty42:
yeah, looking briefly at the menubar code, i see nasal is used to display `/sim/help` and `/sim/help/common`... the only text line in `/sim/help/common` is the title line... the "?" menu comes from `/sim/help` and i do see that my testing modification (above) is loaded in there... i haven't gone digging far enough to figure out how and when the text is loaded from the xml file... i would have to dig a lot deeper to figure out where the "common" help is loaded from... looking at the xml files in fgdata's "gui" directory doesn't show me anything obvious in the filenames indicating that file...

wlbragg:
Revisiting this, 

> i don't know how you added it but i was thinking of something like this...

I actually like this. It is easy to read, even better with the bullets.

hbeni:
Just force-pushed the branch.
- Its now based on the current master (commit ccfa4e0318779dd49)
- Added a third commit to add an ALT mode to the trim wheels (in accordance with the [latest FGAddon commit fc55e00a458](https://sourceforge.net/p/flightgear/fgdata/ci/fc55e00a4585db191de057dae5dcb8416f64eb54/)/[ticket-2880](https://sourceforge.net/p/flightgear/codetickets/2880/) and C182)

I suggest we might merge that now and resolve the help menu issue later?

hbeni:
I would say so - if you are okay with that

wlbragg:
Thank you!

PR #1499, opened 2024-03-13
KR87 ADF: prevent overflow, implement countdown timer and directly setting the active FRQ in FLT/ELT mode

### LINKED ISSUE #1279: Improve the KR87
@ pfefferCode reported these potential bugs in the adf radio KR 87:

1. The rotating knobs seem to rotate the anticlockwise (wrong direction), clicking on it.

2. ~The standby frequency is not displayed after a click on "FRQ"-Button (always 200 is shown).~ user says issue disappeared

3. If on the right display the time is shown, the knob should directly change the active frequency, see: <https://www.bendixking.com/en/~/media/bendixking/files/006-18110-0000_5-kr-87-pilot-guide.pdf> page 44: "While FLT or ET is displayed the “in use” frequency on the left side of the window may be changed, by using the frequency select knobs, without any effect on the stored standby frequency or the other modes. This feature is especially useful when searching for stations with unknown frequencies."

4. If on the right display the standby freq is shown, it does not change, when clicking on the knob. Nevertheless, the standby freq is changed as a click on the "FREQ"-Buttons transfers it (the chnged freq) to the active freq.

5. ~I could not start the clock, it always shows "2:00". Should that already work?~ user says issue disappeared

6. Long pressing the "SET/RST"-Button should change to count-down mode (see page 43 in https://www.bendixking.com/en/~/media/bendixking/files/006-18110-0000_5-kr-87-pilot-guide.pdf ). Is that implemented?

7. Turning the inner knob should not overflow to the hundrets of kHz, see:

> With “FRQ” annunciated, the standby frequency is selected using the frequency select knobs which may be rotated either clockwise or counterclockwise. Pull the small inner knob out to tune 1’s. Push the smaller inner knob in to tune 10’s. The outer knob tunes the 100’s and the 1000’s up to 1799<
    Source: Pilot's Guide on KR 87, p. 42.

8. Improve the 3D model and texturing and confirm that the functionality matches the real KR87

hbeni:
Fix #1279 

- [x] Fix open things of #1279
- [x] Implement count-down mode of ET timer
  (by this occasion, I rewrote the internal timer implementation to use the much friendlier `maketimer()` interface)

### DIFF
--- a/Models/Interior/Panel/Instruments/kr87-adf/kr87.xml
+++ b/Models/Interior/Panel/Instruments/kr87-adf/kr87.xml
@@ -70,10 +70,31 @@
         <type>select</type>
         <object-name>indicator.ET</object-name>
         <condition>
-            <equals>
-                <property>instrumentation/adf[0]/display-mode</property>
-                <value type="int">2</value>
-            </equals>
+            <or>
+                <and>
+                    <not><property>instrumentation/adf[0]/set-mode</property></not>
+                    <equals>
+                        <property>instrumentation/adf[0]/display-mode</property>
+                        <value type="int">2</value>
+                    </equals>
+                </and>
+
+                <and>
+                    <property>instrumentation/adf[0]/set-mode</property>
+                    <less-than>
+                        <expression>
+                            <mod>
+                                <product>
+                                    <property>sim/time/elapsed-sec</property>
+                                    <value>10</value>
+                                </product>
+                                <value>10</value>
+                            </mod>
+                        </expression>
+                        <value>4</value>
+                    </less-than>
+                </and>
+            </or>
         </condition>
     </animation>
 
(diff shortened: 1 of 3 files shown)


hbeni:
The new count-down mode is operated in this way:
- Press SET button for more than two seconds (until the ET blinks)
- Dial in the desired time with the two knobs
- Press the SET button again to start the timer. it will of course also count down if the unit is switched to FT or FRQ modes.
- Once the alarm goes off, the unit switches to ET mode and shows flashing digits.
  - The unit can optionally have an aural alarm. I decided that this is not installed (found no sample); however, if you like to add it, just hook a suitable sound to the `/instrumentation/adf/enroute-timer/alarm-sound` property.
  - The flashing alarm goes off automatically after 15 seconds; or if manually confirmed by the SET button.
  - The optional aural alarm goes off automatrically after ~ 1 second.

hbeni:
One last thing I'm unsure about:
Does the FT timer show FT really just in `MM:SS` like it is modelled now?
I would assume that it will show it in `HH:MM`... Could not find something in the manuals, however.

PR #1505, opened 2024-04-20
Multiplayer engine sound

hbeni:
Adds engine sound to remote instances in multiplayer, so other c172 can be heard for example when taxiing

### DIFF
--- a/Models/c172p.xml
+++ b/Models/c172p.xml
@@ -34,6 +34,9 @@
         </sim>
     </multiplay>
 
+    <!-- Multiplayer Sound -->
+    <sound><path>Aircraft/c172p/c172-sound-multiplayer.xml</path></sound>
+
     <model>
         <name>Damage</name>
         <path>Effects/damage/damage.xml</path>
--- a/c172-sound-multiplayer.xml
+++ b/c172-sound-multiplayer.xml
@@ -1,0 +1,97 @@
+<?xml version="1.0"?>
+<!-- Multiplayer sound configuration -->
+<!-- Loaded for remote models -->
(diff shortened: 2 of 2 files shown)


hbeni:
:information_source: Note: This was developed against 2020.3.19 and verified OK.

Current next as of now has an issue preventing all remote sounds to be played. This is however nothing to be fixed at the aircraft side.

wlbragg:
@hbeni does the failure of 2020.4 sound have anything to do with the "type"?

Also, a while back we discussed, and I am pretty sure there were changes made to the fgcommand sound system. Addons use this sound system. I was having difficulty getting the addon sound to work. It ended up being the lack of defining the "queue" parameter.

see:
~~~
var sound_alert = {path: "[addon=org.flightgear.addons.missiongenerator]Sounds", file: "alarm.wav", volume:1.0, queue:"instant"};

fgcommand("play-audio-sample", props.Node.new(sound_alert));
~~~

I don't know that this would have anything to do with the FX sound system or even if the MP is using the FX sound system or if it is something else such as "mp"?

Anyway, I just wanted to bring it up in case it pertains to the issue.

hbeni:
> I don't know that this would have anything to do with the FX sound system or even if the MP is using the FX sound system or if it is something else such as "mp"?

I think it uses the "ordinary" fx sound system. And it should work like for all other "normal" sounds, at least per the wiki. And in 2020.3 it does! No fgcommands involved.

hbeni:
Opened a SF ticket for investigation.

It's your call if you want to merge... (maybe backport to 2020.3.19?)

hbeni:
C182 sister ticket: https://github.com/HHS81/c182s/issues/587

wlbragg:
Thank you!

PR #1508, opened 2024-06-05
Fix issue 1507 / ADF power switch shutting off DME

### LINKED ISSUE #1507: DME is fused by ADF
DME gets out if the ADFs power button is turned off.
They are protected by the same circuit breaker, however I think nonetheless they should be switching off separately when operating the individual power switches.
The DME pwr-switch works, but the issue is, that the ADFs power switch also switches off the DME, which probably is a bug.

hbeni:
Fix bug with ADF power switch shutting off also DME

Fix #1507


wlbragg:
Good catch, thank you!

PR #1509, opened 2024-06-17
fix yoke normals

xcvb85:
The normals of the yokes are partly flipped which leads to rendering issues (see https://sourceforge.net/p/flightgear/mailman/flightgear-devel/thread/BCBE54EF-49FD-4EDB-98EC-D0C58A665EBB%40albanarts.com/#msg58784722).


wlbragg:
Good catch. Thank you for fixing this.

PR #1518, opened 2024-09-09
Switched compass pitch/roll to damped c++ values

### LINKED ISSUE #1517: Magnetic Compass: Simplify
Simplify magnetic compass code to use c++ based pitch/roll values for the disc

See:
- https://sourceforge.net/p/flightgear/codetickets/2904/
- https://github.com/HHS81/c182s/pull/601

(:warning: this is exclusive to 2020.4 only, not to be backported)

hbeni:
Fix #1517

### DIFF
--- a/Models/Interior/Panel/Instruments/mag-compass/mag-compass-fg1000.xml
+++ b/Models/Interior/Panel/Instruments/mag-compass/mag-compass-fg1000.xml
@@ -82,10 +82,10 @@
             <clip>
                 <clipMin>-25</clipMin>
                 <clipMax>25</clipMax>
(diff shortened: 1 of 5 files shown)


wlbragg:
@hbeni can you please explain this. Is this something that was changed in c++ so we don't need the filter for effect anymore or what?

hbeni:
Yes, the current fgfs next code contains the impementation in c++ space, so the aircraft sided filters can be retired.

(However, the filters for "getting stuck" needs to stay in place)

wlbragg:
Thank you!

PR #1531, opened 2025-01-28
Issue 1527 kma20

### LINKED ISSUE #1527: Add FGCom integration (KMA20TSO)
The C182s has a nice feature which makes integrating fgcom more realistic (via audio panel).
With that enabled, you can just use SPACE as your FGCom transmit keybind, and which COM is used is selected b the insim/cockpit audio panel.
It is implemented optionally, so you can choose what mode you want to operate in.

It would be cool to have that option in the 172 too and more deepl simulate the KMA20TSO audio panel.

Image

----
Ref:
- C182s PR: https://github.com/HHS81/c182s/pull/327
- KMA20TSO manual: https://www.wpaviation.com/pdfs/king_kma20_pilot_guide.pdf

hbeni:
Tested OK, should be fine to merge

Fix #1527

### DIFF
--- a/Models/Interior/Panel/Instruments/kma20/kma20.xml
+++ b/Models/Interior/Panel/Instruments/kma20/kma20.xml
@@ -41,6 +41,7 @@
         <mkr>instrumentation/kma20/mkr</mkr>
         <sens>instrumentation/kma20/sens</sens>
         <knob>instrumentation/kma20/knob</knob>
+        <operable>instrumentation/kma20/operable</operable>
     </params>
 
     <path>kma20.ac</path>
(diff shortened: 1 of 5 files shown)


wlbragg:
I'll take a look soon and get this merged. Thanks!

hbeni:
Thank you.

One thing that came to my mind this night:
The listeners are currently configured to fire if the listened-to prop was written to (copied previous behavior).
While that works and also doesn‘t put much extra load (update function is called once a second or so), this might still be optimized by adding a second `0` to the `setlistener()` calls in the init function.
That way they only execute if the listened value _changes_, which should be enough (needs testing of course).

hbeni:
put on hold, I want to add some things.

hbeni:
OK, Wayne, you can have a look now.

wlbragg:
Very nice, thank you!

PR #1474, opened 2023-11-28
Cockpit flight controls react to trim settings

dany93:
> From Murmur79 (c172p git repository)

Fixes #1471

PR #1532, opened 2025-02-06
Cockpit flight controls react to trim settings

### LINKED ISSUE #1471: Cockpit flight controls not reacting to trim settings.
If trim settings are modified, the external model correctly shows the control surfaces changing their positions with centered flight controls.

Cockpit flight controls, though, will not react to trim changes, and the yoke and pedals will stay centered, even if the control surfaces are trimmed out-of-center. This is wrong for the C172, since it has direct mechanical link (cables) between flight controls and control surfaces.

The cause of this, is that the three variables:

```
/sim/model/c172p/cockpit/yoke-elevator
/sim/model/c172p/cockpit/yoke-aileron
/sim/model/c172p/cockpit/pedals-rudder
```

are output from:

```
<channel name="Pitch">
<channel name="Roll">
<channel name="Yaw">
```

whereas they should be output from:

```
<summer name="Pitch Trim Sum">
<summer name="Roll Trim Sum">
<summer name="Yaw Trim Sum">
```
which take into account the trim input.

I tried the modification and it works as it should.

hbeni:
Currently no aileron trim feedback to yoke. The C172 has an asymetrically mounted wing in order to compensate for roll-forces in cruise. This is simulated by the aileron-trim prop. Aileron _input_ (yoke) should be neutral, when the input device (joystick) is neutral. Also, don't confuse with rudder-trim (which there is an option for).

For disuccsion see: https://github.com/c172p-team/c172p/issues/1471
Fix #1471

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -817,6 +817,20 @@
                     <min>-1</min>
                     <max>1</max>
                 </clipto>
+            </summer>
+            <summer name="Yoke Roll Sum">
+                <!--
+                Currently no aileron trim feedback to yoke. The C172 has an asymetrically mounted wing in order to
+                compensate for roll-forces in cruise. This is simulated by the aileron-trim prop.
+                Aileron _input_ (yoke) should be neutral, when the input device (joystick) is neutral.
+                Also, don't confuse with rudder-trim (which there is an option for).
+                For disuccsion see: https://github.com/c172p-team/c172p/issues/1471
+                -->
+                <input>fcs/aileron-cmd-norm-filtered</input>
+                <clipto>
+                    <min>-1</min>
+                    <max>1</max>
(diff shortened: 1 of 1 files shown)


hbeni:
This replaces PR #1474

hbeni:
OK, looks good.
- Aileron trim is active but not reflected on the yoke
- Elevator trim is active and reflected on the yoke
- KAP140 can handle it
- Flight characteristics look unchanged

Ready for merge @dany93 @wlbragg ; and the other PR can be closed

wlbragg:
Thank you everyone!

hbeni:
So this went out the window, I guess; at least I couldn't find it in the `git log` of `master`....
Can this be remerged?

PR #1533, opened 2025-02-06
Modeling STICK-FORCE-PER-'g' and the effect of dynamic pressure on flight controls

hbeni:
Adds the option to adjust needed control forces (on joystick) by dynamic pressure

### DIFF
--- a/c172p-main.xml
+++ b/c172p-main.xml
@@ -412,6 +412,7 @@ http://forum.flightgear.org/viewtopic.php?f=4&t=25157
             <path>/sim/current-view/user/default-field-of-view-deg</path>
             <path>/sim/model/c172p/ruddertrim-visible</path>
             <path>/fdm/jsbsim/fuel/tank</path>
+            <path>/fdm/jsbsim/fcs/stick-force-per-g</path>
         </aircraft-data>
 
         <current-view>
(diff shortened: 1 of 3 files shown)


hbeni:
Replaces PR #1472 

I squashed the commits together to make it more easily distinguishable.
Also I deliberately left out Murmur79's latest change to the tables (see commit 6278a60ea2627984d742e45edbc7a7361db66174) because I barely couldn't take off with that numbers and elevator nearly had no control authority.

wlbragg:
@hbeni Thank you for clearing this up and getting it off the table!

wlbragg:
@hbeni after more testing, this elevator limiting filter is not able to be turned off. Also, it is very strong. I don't know if that was intended or not. My impression is it needs to be strong (limiting that elevator displacement) at the first half of the range of yoke travel but then of you continue to move the yoke and thus the elevator, it will still cause the aircraft to pitch as normal. I don't know if I described that correctly. I should have tested this more thoroughly before I merged, that is on me. Can you assist me in  whatever is needed to correct this. I was going o revert the merge but I request your guidance at to the procedure that should be used to do that.

wlbragg:
>Also I deliberately left out Murmur79's latest change to the tables (see commit https://github.com/c172p-team/c172p/commit/6278a60ea2627984d742e45edbc7a7361db66174) because I barely couldn't take off with that numbers and elevator nearly had no control authority.

I actually felt the same with the table as it is now.

wlbragg:
Do we just revert this and work from there? I accidentally started to revert the wrong merge and had to close that one, thus my hesitation to do this one.

hbeni:
Is it possible to reset master before the merge?
I can investigate on my branch, I still have it.
If we are done, then we can remerge easily.

PR #1535, opened 2025-02-08
Modeling STICK-FORCE-PER-'g' and the effect of dynamic pressure on flight controls.

hbeni:
From Murmur79

Added switch to make it optional to the user (BeniH) User can select/deselect from aircraft menu

(Second try, onto current master)


hbeni:
Hi @wlbragg I think I just overlooked to add the fix for the table. It was not in my other branch.

I decided to pull the current master (c413474512045b3) and prepare another PR for your convinience.
With that branch I just had a test flight and I think all is well.
I can enable/disable, and it has a notable effect.

To see it:
- take off
- CTR+U a bit, gain speed to abut 100 knots
- go to outside view
- push your joystick forward, and hold
- parallely activate the feature

You should see a notable difference looking at the elevator. With the feature the deflection is less and consequently you need more force on your physical stick to move it to the extremes.

You can toggle the feature on/off as desired in quick succession to make it more visible.

wlbragg:
Much better, thank you!

PR #1537, opened 2025-02-09
c172p should not overwrite fgdata's controls.throttleAxis()

### LINKED ISSUE #1524: c172p should not overwrite fgdata's controls.throttleAxis()
See https://sourceforge.net/p/flightgear/codetickets/2793/ and specifically the comment by Richard.

https://github.com/c172p-team/c172p/blob/0538dfb166c1f33b5394358cbcfdda3a484c57c4/Nasal/engine.nas#L490-L497

wlbragg:
Fixes #1524

### DIFF
--- a/Nasal/avionics.nas
+++ b/Nasal/avionics.nas
@@ -145,7 +145,9 @@ aircraft.data.add(
     "/controls/gear/brake-parking",
     "/controls/flight/flaps",
     "/controls/flight/elevator-trim",
-    "/controls/engines/current-engine/throttle",
+    #"/controls/engines/current-engine/throttle",
+	"/controls/engines/engine[0]/throttle",
+	"/controls/engines/engine[1]/throttle",
     "/controls/engines/current-engine/mixture",
     "/controls/engines/engine[0]/primer-lever",
 );
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -14,7 +14,9 @@ var autostart = func (msg=1) {
 
     # Setting levers and switches for startup
     setprop("/controls/switches/magnetos", 3);
-    setprop("/controls/engines/current-engine/throttle", 0.2);
+    #setprop("/controls/engines/current-engine/throttle", 0.2);
+	setprop("/controls/engines/engine[0]/throttle", 0.2);
+	setprop("/controls/engines/engine[1]/throttle", 0.2);
 
     var auto_mixture = getprop("/fdm/jsbsim/engine/auto-mixture");
     setprop("/controls/engines/current-engine/mixture", auto_mixture);
@@ -319,7 +321,9 @@ var switches_save_state = func {
         setprop("/controls/engines/engine[0]/primer", 0);
         setprop("/controls/engines/engine[0]/primer-lever", 0);
         setprop("/controls/engines/engine[0]/use-primer", 0);
-        setprop("/controls/engines/current-engine/throttle", 0.0);
+        #setprop("/controls/engines/current-engine/throttle", 0.0);
+		setprop("/controls/engines/engine[0]/throttle", 0.0);
+		setprop("/controls/engines/engine[1]/throttle", 0.0);
         setprop("/controls/engines/current-engine/mixture", 0.0);
         #setprop("/controls/circuit-breakers/aircond", 1);
         setprop("/controls/circuit-breakers/autopilot", 1);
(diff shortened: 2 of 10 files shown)


wlbragg:
@hbeni if you have some time, would you test this (DO NOT MERGE).
What needs verifying is that the throttle controls for engine 1 and engine 2 work as they always have. I tested it and it works for my method of control which is the mouse. I will test with a Thrustmaster JS soon. I noticed I used my local branch that has the Cockpit flight control commit in it. I'll either ignore it if you end up merging https://github.com/c172p-team/c172p/pull/1536 or I will close this and resubmit if not. I have a mental reservation for this entire issue as it seems so complicated for such a simple fix. I don't totally understand all the throttle control paths, ie: jsbsim, flightgear, js, etc. I'm just not grasping the whole picture no matter how much I try.

I'm also going to have @stuartbuchanan take a look to see if this fixes his Multi-Throttle JS issue.

hbeni:
No worries - I don't have write access to this repo. I cant merge on my own ;)

But you are right. I cant control the throttle with my Joystick anymore! :stop_sign:

wlbragg:
>I cant control the throttle with my Joystick anymore!

Do you know if you're JS has a custom connection to the throttle or is it using a default?

hbeni:
> > I cant control the throttle with my Joystick anymore!
> 
> Do you know if you're JS has a custom connection to the throttle or is it using a default?

I made it public: https://forum.flightgear.org/viewtopic.php?p=423781#p423781
It uses default bindings:
```
<axis n="3">
    <desc type="string">Throttle</desc>
    <binding>
      <command type="string">nasal</command>
      <script type="string">controls.throttleAxis();</script>
    </binding>
  </axis>
```

wlbragg:
@hbeni 

Could you please try this again. Feel free to test as deeply as you have time for. I have verified the mouse and my Thrustmaster WCS work correctly. I don't know if a keyboard is supposed to work for throttle, I never looked to see if there are bindings because I am mostly on a laptop with no number pad.

I also need to ask Stuart to test again against his multi-throttle control hardware.

As far as I know I have accounted for everything except the tutorials, I still need to fix those.

I feel pretty confident in this fix.

wkitty42:
On 2/19/25 9:37 PM, wlbragg wrote:
> I don't know if a keyboard is supposed to work for throttle, I never looked
> to see if there are bindings because I am mostly on a laptop with no number
> pad.

FWIW: PgUp and PgDn are the default kbd throttle controls...

FWIW2: Amazon has add-on number pads from ~$40US down to less than $10US... some 
are bluetooth connected... most are USB connected... some are only the number 
pad while others are the entire editing keys (w/ inverted T arrows) and number 
pad section like off of a 104-key keyboard... some even have an actual 
calculator with display built in...

search "add on number pad for laptop"



-- 
        *Please keep mailing list traffic on the list unless*
        *a signed and pre-paid contract is in effect with us.*

wlbragg:
@wkitty42 
Do you know if this ever worked with keyboard (PgUp and PgDn)? Do you have a controler and does it still work with it?

wkitty42:
before i got my controller, i used only kbd... mouse control is too janky for me... i've tried once or twice with my controller but haven't flown in a while... i mainly use a joystick (w/ throttle slider) when i fly non-UFO craft... in all cases, the throttle responded to all throttle controls i used no matter what they were... PgUp/PgDn, slider, shoulder bumpers, mouse grab/slide, etc...

i would hope this doesn't affect that... [time passes] hummm... looking at my controller xml file, i see it uses this... 

```
  <button n="4">
    <desc type="string">Throttle decrease</desc>
    <repeatable type="bool">true</repeatable>
    <binding n="0">
      <command type="string">property-adjust</command>
      <property type="string">/controls/engines/engine[0]/throttle</property>
      <step type="double">-0.01</step>
    </binding>
  # plus 7 more of the above for 1-7 of engine[?]...

```
and the same for the Throttle increase button... whereas the slider for the T16000M uses 

```
 <axis>
  <desc>Throttle</desc>
  <number>
   <unix>3</unix>
   <mac>3</mac>
   <windows>2</windows>
  </number>
  <binding>
   <command>nasal</command>
   <script>controls.throttleAxis()</script>
  </binding>
 </axis>

```
the difference between the two is the slider on the joystick and two buttons for Throttle up/down on the controller...
unfortunately my FG is broken at the moment and i cannot get a good build for testing things...

PR #1544, opened 2025-02-26
Version 2024.1-git

wlbragg:
Start of new 2024.1.1 LTS development branch.

PR #1546, opened 2025-02-28
New release 2024.1.1-LTS

hbeni:
see #1544


hbeni:
User report says, the 2024.1 git version (which is master) dies not work with 2020.3.
So we should not merge this one here.

PR #1555, opened 2025-08-03
update pitch moment by flaps

ysopflying:
Based on a quasi-2D vortex lattice calculation, giving the expected pitch up moment when setting flaps on downwind.
Double-check welcome.


wlbragg:
@dany93 are you still around? If so please review and possibly advise. Anyone else want to comment on this, would be welcome! I'd like to give this a week and see if there are any responses. Also I will test myself. @ysopflying is there a target date for the 2024.1.2 release?

wkitty42:
On 8/3/25 1:57 PM, wlbragg wrote:
> *wlbragg* left a comment (c172p-team/c172p#1555) 
> <https://github.com/c172p-team/c172p/pull/1555#issuecomment-3148602488>
> 
> @ysopflying <https://github.com/ysopflying> is there a target date for 
> the 2024.1.2 release?

i think there's an 2024.1.2-rc5 now? in testing but getting close for release... 
maybe can get c172p merged for it or next RC and eventually GA?

dany93:
Hi @wlbragg,

Sorry, my wife is hospitalized.

And I even don't know if I could use my current (old) FG next version to test it, I'm still with Linux Mint 20. I cannot update FG next.

Sorry.

Anyway (see the issue from feb 2025, and another one that I don't remember when) I have in mind that people doesn't seem to agree on this behaviour for the C172. For me, I totally ignore the real one. Only a C172 pilot can say it. Calculations are not a absolute reference.

The only thing which is sure is that flaps down increase lift. The immediate effect is to be confirmed.

ysopflying:
@dany93 Thanks for weighting in. Hopefully your wife gets well soon! More important than this stuff here.

If inclined to do so you can try the modification on the simple FDM testbed https://github.com/ysopflying/c172r, which has the new cm_flap.

Here there are minimum 3 steps of validation, which can be taken. Presumably not all details on this platform due to bots. More over the week.

dany93:
> The change refines a bit by also adding an influence of alpha on cmdf. On downwind leg you have an alpha of 3-5deg, so with the change cmdf is positive=nose up instead of the previous nose down. This is the "initial" behavior of the aircraft just when the flaps are set.

Please help me, I don't understand well.
Looking at tour cmdf table.
Between 0 and 4 deg flaps, 0 to 10 deg alpha, cmdf is negative. Which means a nose down moment. Am I correct ?

At highest values for flaps and alpha, cmdf positive = nose up.

ysopflying:
Alphas are in the columns, so cmdf positive at all alphas greater than 6.

An evening of research later: Difficult to find consistent data for this matter, so little possibility to validate own work. Certainly more progress on the weekend.

In one pilot's report it is said, that 100kts clean requires same trim as 70kts flap 10 on the C172. If no better solution, I could replicate this.

Seeing past discussion on this I am sorry to bring this up again, but there are various indications, that it should be nose up. Maybe it is worth doing some stick time on the C172.

dany93:
@ysopflying wrote
> Alphas are in the columns, so cmdf positive at all alphas greater than 6.

Sorry, thank you. I didn't look closely enough at the rows and columns. After all this time without practicing JSBsim...

dany93:
@ysopflying wrote
> Actually there are different effects like the shift of lift vector backwards due to the location of the flaps at trailing edge, added drag above centre of gravity and most importantly change of downwash on the horizontal stabilizer.

Very interesting summary.

Drag, and particularly change of downwash on the horizontal stabilizer, would be in favor of pitch up moment with flaps.

ysopflying:
That's why in principle two approaches are being taken:
- validation by calculations
- replicating know operation points (which can exclude the lift factor)

Pilot reporting is a bit questionable, but the indications between the lines are clear.
Let's put it like this: Almost everybody is driving a car, but this does not make everybody an expert of four-wheel-dynamics.

Currently some success on the calculation part. It is quite a larger operation, and it might be, that it just ends with tuning to meet the operation points.

ysopflying:
Progress thanks to our friendly FGFS community we have both qualitative confirmation of a positive pitch moment flap increment and video evidence.

https://youtu.be/hnV4JIsr0a0

Watch the brighter knobs on the yoke shaft relative to frame to assess elevator deflection.

1:02:03 95kts straight, level and trimmed
1:02:45 flaps 10 set
1:02:47 nose goes up
1:02:49 yoke goes forward
1:03:09 abt. 80kts same yoke position as in 95kts clean
1:04:02 abt. 60kts, flaps30, same yoke position

So there are three operation points with same elevator deflection, allowing to find a new cm_df.
A possible candidate is:
```
<independentVar>fcs/flap-pos-deg</independentVar>
                        <tableData>
                             0.0000      0.0000
                            10.0000    0.001
                            20.0000    0.002
                            30.0000    0.003
                        </tableData>
```
                        
I'll refine those and change the PR accordingly.

dany93:
@ysopflying wrote
> 1:03:09 abt. 80kts same yoke position as in 95kts clean
1:04:02 abt. 60kts, flaps30, same yoke position

How can we we have information about pitch trim adjustment during the procedure?

ysopflying:
> @ysopflying wrote
> 
> > 1:03:09 abt. 80kts same yoke position as in 95kts clean
> > 1:04:02 abt. 60kts, flaps30, same yoke position
> 
> How can we we have information about pitch trim adjustment during the procedure?

We can ask. But actually the video is showing yoke position=elevator position, which is in practical terms independent of trim settings.

NASA windtunnel test of the C210 with a similar fowler-high wing layout had the following results:
Screenshot from 2025-08-21 23-31-53
So small positive in the relevant AoA-region.
Apparently the numbers in the current FDM originate from the JSBsim-example. Comparing with above and available information the are IMHO not correct.

There are some more ideas for verification, which will unfortunately take a dozen hours or so.

dany93:
@ysopflying wrote
>  But actually the video is showing yoke position=elevator position

You are right.
If I understand well, IRL, setting a trim value changes the elevator deflection, thus the yoke position, aircraft trimmed with no action on the yoke.

ysopflying:
Updated proposal (commit 4 of this PR)
```

					<tableData>
<!--                 delta pitch moment to reflect 95kts approach, 75kts flaps 10 and 60kts flaps 30 final with no trim change -->
							0.0000      0.0000
                            10.0000    -0.008
                            20.0000    -0.016
                            30.0000    -0.024					
					</tableData>

```
By CFD calculations it could be confirmed, that cm_df has low negative(!) numbers. The values obtained are not in good agreement with the pilot's reports, so current commit included manual adaptions.
In any case previous values have been too much nose down.

I'll check with our RL-pilot in our community for additional confirmation.

wlbragg:
I'm still trying to follow along here, does it look like there are changes need and agreed upon? I'm going to finish up the interpolation table for the automixtue and then we need to get a request in to update fgdata version if we want any chance in getting these changes into the next release. I'll ping the list now to see if there is still time.

ysopflying:
Pilot's feedback from our community received:

> Hi, I did a preliminary (full flap landing) test, and it feels much more like the real thing! The nose goes slightly up with each flap increase, and I need to slightly correct the trim for the right glide IAS. I think this is a step in the right direction! Thanks!

In this light I propose to take over the PR as commit 4 (Hopefully I did nothing wrong with github here).

Work on Cessna 1xx-FDMS will continue.

ysopflying:
Sorry for doing a PR, which was not fully matured.

Yes, all final changes in commit 4.
Thanks for your message on the mailing list.
(and having a look at the automixture will be very interesting)

wlbragg:
We're too late for 2024.1.2. 2024.1.3 is supposed to be soon to follow so we should have no issues getting into it. @ysopflying are you finished or do you want any more time before merging?

ysopflying:
Merging would be fine for me.
Vacation is around the corner and further improvements would require time.

As said it is not the end of research. Further updates possible, but in careful consideration not to break other items while improving one issue.

wlbragg:
Thanks for sticking with this through all the analysis.

PR #1564, opened 2025-09-29
Bug fixes

### LINKED ISSUE #1325: Repairing flag set to true during "Go to Airport" relocation
There is a bug where the flag /fdm/jsbsim/damage/repairing is set to true during a GUI instigated airport relocation.
What that does is set the property /controls/gear/brake-parking to true. This requires a cycling of the "Repair" function or bush gear change or manually setting the repairing flag to false.
The function that is responsible for repairing flag to be set to true is in Systems/ground-effects.xml
~~~
<logic>
        <name>Parking Brake And Chock</name>
        <input>
            <or>
                <!-- The actual parking brake controlled by the pilot -->
                <property>/sim/model/c172p/brake-parking</property>

                <property>/sim/model/c172p/securing/chock-visible</property>

                <!-- When changing the bushkit, hold position so that new
                     initial length of tiedowns can be safely computed.
                -->
                <property>/fdm/jsbsim/damage/repairing</property>
            </or>
        </input>
        <output>
            <property>/controls/gear/brake-parking</property>
        </output>
    </logic>
~~~
The cause may be linked to a singleshot timer in Nasal/physics.nas that is set upon bush gear change out. This function works fine on a normal startup but not on a reset such as happens with a Airport relocation.
~~~
var set_bushkit = func (bushkit) {
    setprop("/fdm/jsbsim/damage/repairing", 1);
    reset_all_damage();
    bushkit_changed_timer.restart(bushkit_change_timeout);
};

# This timer object is used to enable damage again a short time after
# changing to the last bush kit option.
var bushkit_changed_timer = maketimer(bushkit_change_timeout, func {
    setprop("/fdm/jsbsim/damage/repairing", 0);
    setprop("/fdm/jsbsim/damage/traversing", 0);
});
bushkit_changed_timer.singleShot = 1;

# Update the 3D model when changing bush kit
setlistener("/fdm/jsbsim/bushkit", func (n) {
    set_bushkit(n.getValue());
}, 1, 0);
~~~

Further investigation is needed to determine what is causing this repairing flag to be set and by what function for sure.

Test if Airport relocation triggered by tutorial or mooring relocation trigger the same issue.

### LINKED ISSUE #1552: Auto mixture not working at mid ASL range
The auto mixture needs higher resolution in the interpolation table to correctly configure the mixture at the midrange of its performance envelope (6500)

wlbragg:
Fixes #1325 
Fixes #1552 - see last post in issue.

@hbeni this is another important bug fix I need to get merged asap so I can cherry pick it back to Release 2024.1.2.

### DIFF
--- a/Nasal/c172p.nas
+++ b/Nasal/c172p.nas
@@ -12,6 +12,9 @@ var autostart = func (msg=1) {
     # Reset battery charge and circuit breakers
     electrical.reset_battery_and_circuit_breakers();
 
+    #Guarantee repairing flag is reset
+    setprop("/fdm/jsbsim/damage/repairing", 0);
+
     # Setting levers and switches for startup
     setprop("/controls/switches/magnetos", 3);
 	setprop("/controls/engines/engine[0]/throttle", 0.2);
@@ -491,6 +494,8 @@ var reset_system = func {
     props.globals.getNode("/fdm/jsbsim/pontoon-damage/right-pontoon", 0).setIntValue(0);
 
     setprop("/engines/active-engine/kill-engine", 0);
+
+    setprop("/fdm/jsbsim/damage/repairing", 0);
 
     set_fuel();
 }
--- a/Systems/c172p-engine.xml
+++ b/Systems/c172p-engine.xml
@@ -226,35 +226,35 @@
             <sum>
               <!-- Pressure Altitude -->
               <product>
-                <subtract>
+                <difference>
                   <value>29.92</value>
                   <property>/environment/pressure-inhg</property>
-                </subtract>
+                </difference>
                 <value>1000</value>
               </product>
               <property>/position/altitude-ft</property>
               <!-- Temperature Offset -->
               <product>
                 <value>120</value>
-                <subtract>
+                <difference>
                   <property>/environment/temperature-degc</property>
-                  <subtract>
+                  <difference>
                     <value>15</value>
                     <product>
                       <sum>
                         <product>
-                          <subtract>
+                          <difference>
                             <value>29.92</value>
                             <property>/environment/pressure-inhg</property>
-                          </subtract>
+                          </difference>
                           <value>1000</value>
                         </product>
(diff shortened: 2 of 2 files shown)


Nasal/c172p.nas
```diff
@@ -12,6 +12,9 @@ var autostart = func (msg=1) {
     # Reset battery charge and circuit breakers
     electrical.reset_battery_and_circuit_breakers();
 
+    #Guarantee repairing flag is reset
+    setprop("/fdm/jsbsim/damage/repairing", 0);
```
hbeni, line 16: This confuses me a little.
As far as I remember, there are listeners acting on this flag. So when autostart, and we want a guaranteed "repaired state", we should call the appropriate repair function at some point here, which needs to deal correctly with setting/resetting the flag.

Alternatively, If I understand correctly, we do _not_ want autostart to fully repair, just attempt an engine start from the current state the plane is in (which then may fail if for example the engine is broken).
In this case, the big question is: why do we want to set the repairing flag here? If the flag is "1", repairing is either still running (in that case we need to leave it alone) or the reset mechanism in the repair function is broken (in that case we need to leave it alone here and fix it at the correct place)

hbeni: OTOH I did not find a suitable repairing function with my quick grep just now.
It looks like the C182 is structured a little differently in this regard.

hbeni: Ah ok, just read your explaination. That makes sense now.

hbeni (approved):

hbeni:
Verified, it worked (on my 2024.1.1 build)

wlbragg:
Let me state this another way. 

>if it was molten, keep it molten - unless we repair

/fdm/jsbsim/damage/repairing does not necessarily mean we executed reset_all_damage()

The entire logic is this.. 
repair_damage = func {
    set_bushkit(getprop("/fdm/jsbsim/bushkit"));
}

PR #1577, opened 2025-11-27
Add battery capacity dependency on battery temperature

hbeni:
Adds temperature dependency for the battery capacity.

(Ported from C182)


hbeni:
I already did a test and it looked good on 2024.1.3

Issue #1589, opened 2026-05-30
Garmin196.nas has intermittent issue saving parameters


rumpole6:
During power-up of the instrument garmin196.  The script garmin196.nas produces runtime errors when saving parameters. Because it is intermittent, I don't have the exact message.

rumpole6:
I'd like to work on this issue, please assign it to me.

wlbragg:
@rumpole6 Absolutely, thank you for the help. I can review any work and merge it when ready. It's been long needed, I am aware of some of these error or warning, we've been getting them for a long time. It has never really been a priority, but it needs investigating and fixed. I don't know how deep you're planning on digging. But currently the 172 has its own local version.  I don't recall why we originally went that way. I am not opposed to using the fgdata version if we can insure we don't lose anything or have to worry about future changes we have no control over.

rumpole6:
Lets use this issue to fix the saving parameters bug. I will open a new issue to investigate whether switching to the generic version.

rumpole6:
@wlbragg  I notice that there is no .clang-format file for this repository.  Do we want to run clang-format anyway? what about pre-commit? The  process seems slightly different here.  Also, should the SPDX headers be added?

wlbragg:
No, I don't think we need change the existing workflow. Every repo I have been a part of has been around for a long time and we haven't felt the need. Can you give me any examples as to why we need any of the items you're asking about? I had to look them up as I never worked in any serious business environment that used these things. I'm for simple VS complicated. We don't have a lot of contribution going on and from what I read, I don't see the need, but I am far from an experienced user or moderator.
Typically you just make an issue (#1589), make a local branch that matches your issue, like Branch-#1589 (this is not mandatory, but helps keep everything in sync). When you have something ready for review, submit that branch in the form of PR (pull request), we like a different person than the originator to review and help test and then when ready, someone that has commit authority commits it. There is one other thing that helps, when you make the merge request, at the top of the comments in the request  you type the wording Fixes #1589 and it will automatically close the corresponding issue when merged.

PR #1591, opened 2026-06-16
Fixing issue # 1589

rumpole6:
Intermittently in nasal, if you are completely removing a branch node on the property tree and the node you are removing is already not there, nasal will fail silently causing other strange errors. This is if you use prop.remove(). The actual method to use for a branch node is prop.removeChild() which always works regardless of the status of the branch node.

On branch topic/#1589

### DIFF
--- a/Models/Interior/Panel/Instruments/garmin196/garmin196.nas
+++ b/Models/Interior/Panel/Instruments/garmin196/garmin196.nas
@@ -224,7 +224,10 @@ var load_parameters = func{
 setlistener("/sim/signals/fdm-initialized",load_parameters);
 
 var save_parameters = func{
-    props.globals.getNode("/instrumentation/garmin196/save").remove();
+
+    var save_branch = props.globals.getNode("/instruments/garmin196");
+    if (save_branch != nil )save_branch.removeChild("save");
+
 
     ##preparation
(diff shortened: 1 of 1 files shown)


wlbragg:
Can you give me a test example that causes the garmin error that I can test to see that this actually eliminates the error? In other words, do you know what triggers the error specifically? Does it have something to do with viewing or adding the Garmin and then removing it that triggers the error. Then after this patch, that no longer happens? I'm going to try to test that and force an error condition so I can see it this patch eliminates that error from ever happening. Or, are you just assuming this is the cause? It may very well be, I would just like to verify this by a controlled trigger.

wlbragg:
I'm not sure it it needs to be in the first comment of the PR or the first line in any comment of the PR or anywhere in any comment of the PR that you place Fixes #1589. So I am adding it here to see it it will close the Issue #1589 on merge. We have always added it as the first line in the PR's first comment.

wlbragg:
I'm trying to test this now, if I can prove this gets rid of that annoying intermittent error, it would be great if the FGDATA version got this patch, if it even works the same way and has the same code.

wlbragg:
I cannot trigger this error in the before version.

PR #1596, opened 2026-07-15
Proposal: Implement parked ground-friction helpers to mitigate tail-drop and yaw-sliding

### LINKED ISSUE #1595: Potential Bug: Parked aircraft exhibits tail-drop / pitch-up behavior in mild tailwinds (12+ knots)
## Description

When the Cessna 172P is parked and secured (engines off, parking brake set, or wheel chocks active), a mild tailwind of approximately 12 knots causes the aircraft to pitch upward, lifting the nose wheel off the ground.

While the flight model should realistically react to severe weather, a standard tricycle-gear aircraft like the C172P typically remains statically stable on its three wheels in a 12-knot breeze. I am opening this issue to discuss whether this points to an underlying flight dynamics model (FDM) edge case when handling reversed airflow while static, as I am trying to prevent the tail from dropping in the first place.

---

## Hypothesized Causes Under Investigation

I have been looking through `c172p.xml` to understand what forces might be causing this pitch-up behavior, and I have a couple of hypotheses:

### 1. `Cmo` (Pitching Moment at Zero Alpha) and Dynamic Pressure

The current implementation of `/aero/coefficient/Cmo` relies on dynamic pressure (`aero/qbar-psf`), which is calculated using the square of velocity ($V^2$). Because it is squared, `qbar` remains positive regardless of whether the aircraft experiences a headwind or a tailwind.

My hypothesis is that in a tailwind, the FDM reads this positive `qbar` and applies a positive (nose-up) pitching moment coefficient. In the real world, a tailwind pushing against the back of the airframe wouldn't automatically generate a standard flight-state nose-up moment like this.

### 2. Ground Effect (`kCLge`)

The `/aero/function/kCLge` multiplier is currently a 1D table looking only at height above ground (`aero/h_b-mac-ft`). I wonder if, in a tailwind, this multiplier is still actively amplifying lift coefficients, artificially boosting vertical forces on the wings from reversed airflow.

---

## What I Have Tried So Far (Experiments)

To test these hypotheses, I applied a few experimental modifications to a local copy of `c172p.xml`. While they changed the behavior, they did not fully resolve the issue:

### Experiment 1: Gating `Cmo` with the Parking Brake

I tried multiplying the `Cmo` calculation by a table referencing `fcs/parking-brake-sum` to nullify the zero-alpha pitching moment whenever the aircraft is secured on the ground:

```xml
<function name="aero/coefficient/Cmo">  
    <description>Pitching_moment_at_zero_alpha</description>  
    <product>  
        <property>metrics/cbarw-ft</property>  
        <table>  
            <independentVar lookup="row">aero/qbar-psf</independentVar>  
            <tableData>  
                13.6  0.09  
                21.2  0.04  
            </tableData>  
        </table>  
        <table>  
            <independentVar>fcs/parking-brake-sum</independentVar>  
            <tableData>  
                0.0  1.0  
                1.0  0.0  
            </tableData>  
        </table>  
    </product>  
</function>

```

*Result:* This made the behavior a little better, but it did not completely stop the tail-drop. This suggests that other pitching moment coefficients (perhaps `Cmalpha` or elevator trim states) are still generating forces that overcome the aircraft's weight distribution.

### Experiment 2: Restructuring `kCLge` to Account for Wind Direction

I tried converting `kCLge` into a 2D table using `velocities/u-aero-fps` (longitudinal aerodynamic velocity) to scale the ground effect multiplier down to `1.0` (neutral) if the wind is coming from behind (negative $u$):

```xml
<function name="aero/function/kCLge">
    <description>Change_in_lift_due_to_ground_effect_factoring_wind_direction</description>
    <table>
        <independentVar lookup="row">aero/h_b-mac-ft</independentVar>
        <independentVar lookup="column">velocities/u-aero-fps</independentVar>
        <tableData>
                  -20.0     -5.0      0.0      20.0
             0.0000 1.0000   1.0000   1.2030   1.2030
             0.1000 1.0000   1.0000   1.1270   1.1270
             ... [rest of row data] ...
             1.1000 1.0000   1.0000   1.0000   1.0000
        </tableData>
    </table>
</function>

```

*Result:* While mathematically cleaner for handling tailwinds during taxi/roll, this also didn't prevent the parked tail-drop issue on its own.

---

## Steps to Reproduce

1. Start FlightGear with the default C172P.
2. Select an environment wind speed of approximately **12 knots**.
3. Choose a runway aligned with the wind, but spawn/start on the **wrong end** (so the wind acts as a direct tailwind upon loading).
4. Ensure the aircraft is parked (engines off, parking brake set, or wheel chocks active).
5. From the external view, observe the nose gear lifting off the ground as the tail drops.

## Request for Input

Since my experimental patches only partially alleviated the problem, I suspect there are other interactions at play—potentially involving how JSBSim computes elevator trim or alpha-dependent moments when the airflow vector is reversed.

Has anyone else noticed this behavior, or is there a recommended best practice for stabilizing aircraft aerodynamics when tied down or parked in JSBSim?

jimishol:
## Description

This PR proposes a set of pitch and yaw counter-moment "helpers" in `c172p.xml` to address the legacy issue where the aircraft's tail drops or the fuselage slides on the ramp when parked in moderate-to-strong winds.

Because I want to respect the core flight model's integrity, **this is presented as a tentative, empirical bypass rather than a definitive engine fix**. I highly welcome and encourage FDM and JSBSim experts to review this approach to see if it is acceptable for the main branch.

---

## The Suspected Problem

At exactly zero ground velocity (`v = 0`), I suspect that JSBSim's Lagrange multiplier static friction solver experiences a numerical feedback loop or convergence instability when a strong tailwind unloads the nose gear:

1. **Aerodynamic Leverage:** A tailwind creates a nose-up pitching moment around the main landing gear (acting as a fulcrum).
2. **Normal Force Fluctuations:** This moment unloads the nose gear and shifts weight to the main gear. The rapid weight transfer may trigger tiny, dynamic oscillations in the main gear's mass-spring-damper system.
3. **Potential Solver Instability:** Because the maximum allowable static friction on the tires is calculated dynamically from the normal force (`Max = mu * Fn`), these rapid normal force fluctuations cause the limit bounds to shift rapidly frame-by-frame.
4. **The 50-Iteration Limit:** In `FGAccelerations::CalculateFrictionForces`, the projected Gauss-Seidel solver is hard-coded to a maximum of **50 iterations** to resolve the static friction constraints. 
5. **The Slip:** I suspect that under the rapid force-transfer oscillations of a tailwind, the solver's constraint matrix fails to converge within those 50 iterations. Once it "gives up" trying to find a stable static friction value, the constraint momentarily slips to zero holding force, causing the tail to drop or the plane to slide sideways.

---

## The Proposed Reaction

To mitigate these dynamic force-transfer oscillations before they can overload the solver, I propose adding two localized, parked-only counter-moments (Pitch and Yaw) inside the `<aerodynamics>` section of `c172p.xml`.

### Key Design Aspects:

* **Velocity-Only Transition (Decoupled from Brakes):** I originally tied these helpers to the parking brake state. However, due to physical inertia, releasing the brakes takes a transient moment to translate into actual ground speed (`vg-fps > 0`). Tying the helper to the brake meant it shut off instantly at release while ground speed was still mathematically `0.0`, immediately re-triggering the solver crash. I have decoupled the helpers from the brakes so they rely **strictly on ground velocity** to transition.
* **Seamless Velocity Fade:** The system is fully active at a standstill and fades out completely once ground speed (`velocities/vg-fps`) exceeds 10.0 fps (approx 6 knots). This bridges the critical time-delay gap during brake release and lets physical rolling momentum smoothly take over, ensuring zero interference with normal taxi, takeoff, or landing physics.
* **Tailwind Isolation:** The pitch helper is asymmetric. It is strictly active for tailwinds (`velocities/u-aero-fps < 0`) and completely inactive in headwinds, allowing natural headwind nose-squatting.
* **The 22-Knot Saturation Limit (37.1 fps):** I chose a 22-knot limit for linear interpolation. It provides a relatively steep slope for strong stability in typical 15–18 knot ramp winds (where the helper is approximately 68% to 82% active), but saturates past 22 knots. This is intended to ensure that in extreme, unrealistic winds, the wind can still naturally overpower the helper.

---

## Observed Empirical Results

During my preliminary testing on the ramp, the proposed changes showed very promising behavior:

* **Baseline (No Helper):** Under an 18-knot tailwind, releasing the brakes to taxi caused the tail to drop instantly before the aircraft could build any roll velocity.
* **With Proposed Helper (Brake Release & Taxi):** Under an 18-knot tailwind, I can now release the brakes and taxi out smoothly. I successfully loaded the aircraft (very slowly) back to an aft CG of **49.7 inches** (beyond the maximum flight limit of 47.3 inches) without any tail-drop at standstill or during the initial roll.
* **Dynamic Loading:** Loading weight rapidly still triggers realistic strut compression and spring oscillations, suggesting the physical dynamics of the landing gear remain intact.
* **Headwind Bypass:** In an 18-knot headwind, the helper correctly disabled itself. The natural aerodynamic nose-down pitching moment allowed me to load the CG even further aft to **51.13 inches** before tipping.

---

## Call for Expert Review

As a non-expert on the deep inner workings of the JSBSim solver, I would appreciate feedback on:

1. Is this localized helper method an acceptable way to bypass this zero-velocity ground contact instability?
2. Do the sign conventions and lever arm coordinates (`cg-z-in` vs `-15.3` and `cg-x-in` vs `57.3`) align with the project's standards for dynamic moment additions?

*Fixes #1595*

### DIFF
--- a/c172p.xml
+++ b/c172p.xml
@@ -2533,7 +2533,7 @@
                     <table>
                         <independentVar>velocities/u-aero-fps</independentVar>
                         <tableData>
-                           -200.0   1.0
+                            -60.0   0.0
                             -37.1   1.0
                               0.0   0.0
                             100.0   0.0
@@ -2568,7 +2568,7 @@
                     <table>
                         <independentVar>velocities/v-aero-fps</independentVar>
                         <tableData>
-                           -200.0   1.0
+                            -60.0   0.0
                             -37.1   1.0
                               0.0   0.0
                              37.1   1.0
--- a/c172p.xml
+++ b/c172p.xml
@@ -2569,10 +2569,10 @@
                         <independentVar>velocities/v-aero-fps</independentVar>
                         <tableData>
                             -60.0   0.0
-                            -37.1   1.0
+                            -25.3   1.0
                               0.0   0.0
-                             37.1   1.0
-                            200.0   1.0
+                             25.3   1.0
+                             60.0   0.0
                         </tableData>
                     </table>
 
--- a/c172p.xml
+++ b/c172p.xml
@@ -2525,7 +2525,7 @@
                         <independentVar>velocities/vg-fps</independentVar>
                         <tableData>
                             0.0  1.0
-                            2.0  0.0
+                            10.0  0.0
                         </tableData>
                     </table>
 
@@ -2570,7 +2570,7 @@
                         <independentVar>velocities/vg-fps</independentVar>
                         <tableData>
                             0.0  1.0
-                            2.0  0.0
+                            10.0  0.0
                         </tableData>
                     </table>
 
--- a/c172p.xml
+++ b/c172p.xml
@@ -2548,41 +2548,6 @@
             </function>
         </moment>
 
-        <moment name="parked-yaw-friction-compensation" frame="BODY">
-            <description>Prevents ground sliding and violent weather-vaning by neutralizing direct base aero yaw moments during parking and low-speed taxi</description>
-            <direction>
-                <x> 0.0 </x>
-                <y> 0.0 </y>
-                <z> 1.0 </z> 
-            </direction>
-            <function>
-                <product>
-                    <table>
-                        <independentVar>velocities/vg-fps</independentVar>
-                        <tableData>
-                            0.0  1.0
-                            35.0  0.0
-                        </tableData>
-                    </table>
-
-                    <table>
-                        <independentVar>velocities/v-aero-fps</independentVar>
-                        <tableData>
-                            -60.0   0.0
-                            -25.3   1.0
-                              0.0   0.0
-                             25.3   1.0
-                             60.0   0.0
-                        </tableData>
-                    </table>
-
-                    <product>
-                        <value> -1.0 </value>
-                        <property>moments/n-aero-lbsft</property>
-                    </product>
-                </product>
-            </function>
-        </moment>
     </external_reactions>
 
     <system file="bushkit"/>
--- a/c172p.xml
+++ b/c172p.xml
@@ -2513,17 +2513,17 @@
             </direction>
         </force>
 
-	<!--
-            TEMPORARY WORKAROUND FOR JSBSIM SOLVER SATURATION (JSBSim #1489 / c172p #1595 / PR #1596)
-            Compensates for tailwind tail-drop by neutralizing direct base aero pitch moments 
-            during parking and low-speed taxi.
-            
-            Root Cause: JSBSim's PGS ground friction solver saturates (hits max 50 iterations limit) 
-            at low ground speeds under tailwinds, causing premature pitch-up / tail-drop.
-            
-            REMOVE THIS BLOCK once JSBSim #1489 is fixed upstream and integrated into FlightGear.
+        <!--
+                TEMPORARY WORKAROUND FOR JSBSim SOLVER SATURATION (JSBSim #1489 / c172p #1595 / PR #1596)
+                Compensates for tailwind tail-drop by neutralizing direct base aero pitch moments 
+                during parking and low-speed taxi (relying on u-aero-fps to naturally handle forward motion).
+                
+                Root Cause: JSBSim's PGS ground friction solver saturates (hits max 50 iterations limit) 
+                at low ground speeds under tailwinds, causing premature pitch-up / tail-drop.
+                
+                REMOVE THIS BLOCK once JSBSim #1489 is fixed upstream and integrated into FlightGear.
         -->
-	<moment name="parked-pitch-friction-compensation" frame="BODY">
+        <moment name="parked-pitch-friction-compensation" frame="BODY">
             <description>Compensates for tailwind tail-drop by neutralizing direct base aero pitch moments during parking and low-speed taxi</description>
             <direction>
                 <x> 0.0 </x>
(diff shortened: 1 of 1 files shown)


hbeni:
First a big thank you to this detailed and well explained PR (and also issue ticket)!

I think this looks like a good workaround, but I lack the time to review the code in depth right now (first glimpse looks good)
But if it is testes well, I think it is a GO.

Ultimatively I think this Problem needs to be solved in JSBSim code, because this probably affects all jsbsim (light) aircraft?
Would be good if you open a ticket over there too.

jimishol:
Updated the PR with a refined fix:

- Isolated base aero moments: Removed the geometric drag-moment offsets (fbx-aero, fby-aero) from the pitch/yaw compensation blocks.
- Preserved real drag moments: Direct aerodynamic drag forces were always active, but now their natural pitching/yawing moments about the CG act on the airframe and gear naturally without being counteracted by synthetic moments.
- Direct targeting: Targeted strictly the unphysical reverse-flow base aero moment tables (m-aero, n-aero).
- Extended blending threshold: Increased the linear fade-out cutoff (vg-fps) from 10.0 to 35.0 fps (~20.7 kts) while keeping the original $1.0 \to 0.0$ linear ramp logic untouched to prevent leakage during slow taxi under strong tailwinds.

Testing confirms this prevents artificial tail-drop during low-speed ground taxi while leaving true aerodynamic drag moments and takeoff/landing physics completely intact.

jimishol:
Update: Cleaned up the implementation further. Removed the redundant vg-fps table entirely. Because u-aero-fps natively incorporates both wind speed and aircraft ground movement, a single high-resolution $V^2$ parabolic table on u-aero-fps handles both the peak tailwind protection and the smooth taxi rollout fade-out seamlessly. Code is cleaner and mathematically unified.

PR #1599, opened 2026-08-19
fix(kap140): add +0.5 rounding bias to hPa baro display to fix off-by-one truncation

jimishol:
### Summary of Changes
This PR fixes an off-by-one display issue on the KAP140 autopilot when toggling the altimeter setting to hPa display mode.

When setting the altimeter to a standard inHg equivalent reported by METAR (e.g., 29.88 inHg for Q1012), the internal hPa conversion yields fractional values (e.g., ~1011.85 hPa). 

Because the `textranslate` digit animations in `kap140.xml` / `kap140-fg1000.xml` perform implicit integer truncation (floor), setting 29.88 inHg displays as **1011 hPa** instead of rounding to the expected **1012 hPa** when toggling into hPa mode via long-press.

### Solution
- Added an expression with `<sum>` and `<value>0.5</value>` bias to the `baro-hpa` input in `kap140-proprules.xml`.
- This mirrors the existing logic used in the inHg branch (which scales by 1000 and adds `+5` bias for 0.01 resolution).
- Since `baro-hpa` operates on 1 hPa integer steps, adding `+0.5` shifts implicit floor truncation into proper half-up rounding to the nearest whole hPa integer.

### Testing
- Set BARO to 29.88 inHg in-sim and long-pressed to toggle hPa display mode.
- Confirmed the KAP140 now correctly displays **1012 hPa** instead of **1011 hPa**.


wlbragg:
@jimishol Thank you!

PR #1600, opened 2026-08-29
save.nas: Persist per-slot engine damage level across state saves and resumes

jimishol:
### Summary
This PR adds persistence for `/fdm/jsbsim/engine/damage-level` to `Nasal/save.nas` so that exact engine wear and tear is saved and restored on a per-slot basis.

### Problem
When saved state slots are restored, `/fdm/jsbsim/engine/damage-level` currently resets to `0`. As a result, any damage accumulated during a flight is lost upon resuming a slot, forcing every resumed session to start with a pristine engine and breaking persistent damage tracking across multiple flights.

### Verification / Test Workflow
1. Set the **Enable damage** option to `true`.
2. Manually set `/fdm/jsbsim/engine/damage-level` to `13` via the Property Tree.
3. Save the aircraft state to a slot and exit FlightGear.
4. Relaunch FlightGear: verify that `damage-level` correctly initializes to `13` via `<aircraft-data>`.
5. Resume the saved slot: verify that `damage-level` accurately preserves the slot's saved value (`13`) instead of resetting back to `0`.

### Changes
* **`save_state`**: Captures `/fdm/jsbsim/engine/damage-level` into `/save/damage-level`.
* **`resume_state`**: Zeroes `/fdm/jsbsim/engine/damage-level` immediately on resume start to prevent aircraft-data bleed-through, then restores the slot's saved value inside the delayed damage timer.


wlbragg:
@jimishol thank you for this. Obviously overlooked during save/resume development. I'll take a look soon and get it merged.

